PR #2, opened 2015-04-10
Scaffold basic admin users controller and templates

himal-mandalia:


EdwardAndress:
I realise this is purely a skeleton thing but...
Where can 'role' be set?  I don't see it in _form (could be useful)
Should we also include 'role' in the show template?

PR #3, opened 2015-04-13
Set submitted at date/time on claim submission

himal-mandalia:


EdwardAndress:
Himal - debugging line at advocate_claims_steps.rb:52 ?

PR #6, opened 2015-04-14
Api for cms

EdwardAndress:


app/controllers/api/advocates/claims_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      def update
+        if @claim.update(claim_params)
+          render json: @claim, message: 'Claim sucessfully updated'
+        else
+          render json: { errors: @claim.errors.full_messages }, status: :unprocessable_entity
+        end
+      end
+
+      def destroy
+        @claim.destroy
```
struds: you might want to call @claim.destroyed? to check if it was successful - I might be overly cautious ? it might throw an exception anyway if it fails.. not sure

app/controllers/case_workers/application_controller.rb
```diff
@@ -1,5 +1,5 @@
 class CaseWorkers::ApplicationController < ApplicationController
-  before_action :authenticate_case_worker!
+  #before_action :authenticate_case_worker!
```
struds: why is this commented out ? if you need to remove it remove it. else keep it and override authentication in the api controller. While we are here you might as well add basic auth with a single username/password jsut for the API (temporarily to allow testing)

app/views/api/advocates/claims/edit.jbuilder
```diff
@@ -0,0 +1 @@
+json.(@claim, :attributes)
```
struds: is there an argument we don't need jbuilder right now and can just return @claim.json in the controller?

struds:
I think there should either be:

be a separate story and PR that authenticates the user of the API given that we have view, update and delete functionality and we only really want users modifying their own data (even when testing in alpha)

OR

we should remove update and delete methods and only allow creation of claims and no functionality to view existing claims over the API

PR #29, opened 2015-04-22
to turn api off by commenting out relevant routes

EdwardAndress:
I've investigated a few feature switching gems and will look into them some more - but maybe this is OK for now?


EdwardAndress:
Actually... the API tests will now fail.  I'll sort that, then push again.

PR #32, opened 2015-04-23
Additional af1 fields

himal-mandalia:


EdwardAndress:
There are some merge conflicts here.

PR #34, opened 2015-04-27
Refactor User model roles into Advocate and

himal-mandalia:
CaseWorker models using multi-table inheritance.


app/models/user.rb
```diff
@@ -4,26 +4,5 @@ class User < ActiveRecord::Base
...
-  ROLES.each do |role|
-    define_method "#{role}?" do
-      is?(role)
-    end
-  end
-
-  def is?(role)
-    self.role == role.to_s
-  end
+  belongs_to :rolable, polymorphic: true
```
struds: is roleable the best naming convention ?

user.rolable.is_a?(CaseWorker)

user.role.is_a?(CaseWorker)

or maybe even better ?

user.persona.is_a?(CaseWorker)

As a user might have a persona of CaseWorker and then also have roles as a CaseWorker

app/models/case_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+  ROLES.each do |role|
+    define_method "#{role}?" do
+      is?(role)
+    end
+  end
+
+  def is?(role)
+    self.role == role.to_s
+  end
+end
```
struds, line 27: I think this is the right approach.. much better than the old user class.

struds, line 27: this might end up being moved to a concern or PORO for handling roles at the point you need to do the same thing in Advocate.

But for now it's fine here.

app/controllers/case_workers/claims_controller.rb
```diff
@@ -5,9 +5,9 @@ class CaseWorkers::ClaimsController < CaseWorkers::ApplicationController
   def index
     @claims = case tab
       when 'current'
-        current_user.claims_to_manage.submitted
+        current_user.rolable.claims.submitted
```
struds: can we do current_user.claims.submitted ? as both types of user have many claims?

app/models/user.rb
himal-mandalia: "Persona" certainly works, "role" should be avoided due to user.role.role, but that may change depending on how we implement authorisation/permissions.

app/controllers/case_workers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class CaseWorkers::ApplicationController < ApplicationController
   private
 
   def authenticate_case_worker!
-    unless user_signed_in? && current_user.case_worker?
+    unless user_signed_in? && current_user.rolable.is_a?(CaseWorker)
```
struds: I guess you might want to do current_user.rolable.case_worker?

although I think I do prefer current_user.persona.case_worker?

app/models/case_worker.rb
himal-mandalia, line 27: Had the same thought and was going to move that into a concern... but held off for now.

spec/controllers/application_controller_spec.rb
```diff
@@ -4,22 +4,36 @@
...
     end
 
+    # context 'given an admin case worker' do
+    #   before { sign_in admin.user }
+    #
+    #   it 'returns case workers root url ' do
+    #     expect(subject.after_sign_in_path_for(case_worker.user)).to eq(admin_root_url)
+    #   end
+    # end
+
```
struds, line 32: whys this commented out ?

app/controllers/case_workers/claims_controller.rb
himal-mandalia: The claims relationship is on Advocate/CaseWorker and not on the user model. The relationship for advocates to claims is one to many, many to many for case workers to claims.

Could delegate claims to persona by putting in user model:

delegate :claims, to: :persona

Then we could all current_user.claims

spec/controllers/application_controller_spec.rb
himal-mandalia, line 32: That was meant to be fixed before I pushed... oversight. Fixing now.

struds:
+1 looks good

PR #38, opened 2015-04-29
Documents in claim rake

EdwardAndress:


EdwardAndress:
Tests now depend on the seeds, so the build fails.  I'll sort this tomorrow.

PR #41, opened 2015-04-30
Import offences and add select2

himal-mandalia:


EdwardAndress:
I'm happy for this to be merged once the conflict in claims/_form.html.haml is resolved

PR #47, opened 2015-05-05
Remove lorem ispsum entries from seed data

jsugarman:
...ion

factory girl generated random strings were being seeded by the rake claims tasks.
These were removed and "real" offence data used for seeding claims demo data.


jsugarman:
do the changes to Gemfile.lock matter and need to be excluded?

PR #48, opened 2015-05-06
Build assets in the container at docker-build time

ashb:
We need the assets in the container so that we have a working manifest.json packaged with the app else the templates can't know what that hash of the files are.

To generate assets properly we need to put rails into production environment. The only thing I'm not too hapy with here is having to specify SECRET_KEY_BASE - or more specifically how do we keep up to date with any other new required environment variables

### DIFF
--- a/docker/rails/Dockerfile
+++ b/docker/rails/Dockerfile
@@ -9,4 +9,7 @@ WORKDIR /rails
 
 EXPOSE 80
 
+#SECRET_TOKEN set here because otherwise devise blows up during the precompile.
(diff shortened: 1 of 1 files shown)


struds:
yeah, its not really desirable to do this.

not sure of an alternative as I don't think I understand the scope of the problem well enough.

ashb:
@struds which bit in particular do you not like btw?

struds:
the SECRET_KEY_BASE= being checked into the core CBO codebase

I think you've spoken to todd about it tho

PR #50, opened 2015-05-07
Fix lower case MAAT reference in advocate dashboard

struds:
the label was generated from the database field hence (easily) missed by Himal when he did (I assume) a search and replace.

Tested in UI, can't see any more lowercase references in codebase of UI


tatyree:
If you can put that into locales and then rebase, then I'm happy to merge.  
Thx. 
T

PR #51, opened 2015-05-07
Fix maat number

struds:
so it's always upper case and referenced from one place in locales.

Also noticed other text strings not in locals but think its out of scope for this PR


tatyree:
Please rebase and fixup/squash all of into a single commit.  It's difficult to tell what the original intent was from the history.

PR #65, opened 2015-05-13
Refactor advocate dashboard claim groups into

himal-mandalia:
submitted, allocated, completed and draft


tatyree:
Given that we've decided to stay with the states as mapped on the whiteboard, do we still need the pending scope and associated specs here?

himal-mandalia:
No we don't! Forgot to remove those after the group conversation. Gone now.

PR #70, opened 2015-05-14
add random advocates with claims to the test chambers/firm

jsugarman:
demo data for displaying on the advocate admin dashboard

### DIFF
--- a/lib/tasks/claims.rake
+++ b/lib/tasks/claims.rake
@@ -8,11 +8,14 @@ namespace :claims do
   desc "Create demo claim data for all states, allocating to case work as required"
   task :demo_data => :environment do
 
+    ADVOCATE_COUNT = 4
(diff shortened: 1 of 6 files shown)


lib/tasks/claims.rake
```diff
@@ -8,11 +8,14 @@ namespace :claims do
   desc "Create demo claim data for all states, allocating to case work as required"
   task :demo_data => :environment do
 
+    ADVOCATE_COUNT = 4
+    CLAIMS_PER_ADVOCATE_PER_STATE = 3
+
     example_advocate = Advocate.find(1)
```
struds, line 14: should we check for failure ? if no user find fail gracefully or do something useful like create one ?

jsugarman, line 14: sure, i will work on it. was thinking of adding error handling

PR #80, opened 2015-05-15
Claim summary for the advocate dashboard

struds:
A summary screen for the advocate dashboard that displays the total
value of outstanding and authorised claims.

If you are viewing the dashboard as an admin the summary is calculated
for your chamber otherwise it shows a summary for your user.
- add model specs for ClaimSummary
- add ClaimSummary model and views
- add cukes


struds:
this is half of the story on trello - the bit missing is claims details. I decided to split it into 2 PR's to get something out early.

PR #88, opened 2015-05-18
fix broken cascading offence list and improve for editing claims

jsugarman:
some css changes had resulted in the failing of the cascading effect
of the offence class select list on the offence (category) offence list.
This has been fixed and an improvement applied to display the class of 
the selected offence category on the claim edit page.


jsugarman:
thanks himal. hows things going over there

PR #99, opened 2015-05-21
Ensure documents are only accessible to authorised users

tatyree:
Authenticate and authorise all document access.

```
Previously this had not been the case.  This changes the workflow to use
consistent CanCan roles based on a user having a role of Advocate or
Case Worker, or being a member of the same chamber as the originating
advocate.
```


tatyree:
I'm unable to squash this - things are getting pulled in from a number of merges that means that 30 minutes into a squash I'm still merging conflicts for things completely out of scope.  Accept or reject as-is and I need to do some research on squashing to see how we avoid this problem.

PR #100, opened 2015-05-21
Use paperclip.io_adapters to find path to pdf duplicate

EdwardAndress:
evidence 'view' link isn't shown on Heroku / AWS because it depends on document#has_pdf_duplicate, which was written using document#path, in error.  Changed to use Paperclip.io_adapters and this works locally.  All tests passing.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -51,7 +51,7 @@ class Document < ActiveRecord::Base
   end
 
   def path_to_pdf_duplicate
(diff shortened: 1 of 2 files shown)


EdwardAndress:
One test failed.  I'll sort it now.

tatyree:
As per our hipchat conversation, will merge now, but please refactor so that these are stored as separate paperclip attachments named 'converted_preview_document'.

PR #125, opened 2015-05-29
Set Document advocate_id

himal-mandalia:
Document advocate_id should be set when created as nested resource via claim


EdwardAndress:
Some integration tests fail after merging (and resolving conflicts) with develop.

PR #129, opened 2015-06-01
remove expanded case view from dashboard

fordie:


himal-mandalia:
This can be re-opened once the partials have been corrected.

PR #140, opened 2015-06-02
Rename "QC alone" advocate category to "qc"

himal-mandalia:


EdwardAndress:
At the moment items in 'advocate category' appear as snake case in block caps.

See line 34 of advocates/claims/_form.html.haml : 
34  = f.collection_select :advocate_category, Claim::ADVOCATE_CATEGORIES, :to_s, :upcase, { prompt: true }, { class: 'select2' }

PR #181, opened 2015-06-10
Consolidate advocates dashboard search with dropdown

himal-mandalia:
Replace multiple search fields on advocates dashboard with dropdown to constrain search,

### DIFF
--- a/app/controllers/advocates/claims_controller.rb
+++ b/app/controllers/advocates/claims_controller.rb
@@ -1,6 +1,7 @@
 class Advocates::ClaimsController < Advocates::ApplicationController
   respond_to :html
   before_action :set_claim, only: [:show, :edit, :summary, :update, :destroy]
+  before_action :set_search_options, only: [:index]
   before_action :set_context, only: [:index, :outstanding, :authorised ]
   before_action :set_financial_summary, only: [:index, :outstanding, :authorised]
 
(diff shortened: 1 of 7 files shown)


jsugarman:
Hey Himal,

Have pulled locally and tested this branch and the code below (i believe) causes the `All` search  to search for an advocate AND defendant _even if you are an advocate and cannot search for anything but Defendant or All_ - this causes the search to unexpectedly return fewer results or none for an `All` search than for a `Defendant` search. `All` also implements logic as `advocate name like x AND defendant name like x` but perhaps OR logic is required.

``` ruby
claims = claims.find_by_advocate_name(params[:search]) if ['All', 'Advocate'].include?(params[:search_field])
claims = claims.find_by_defendant_name(params[:search]) if ['All', 'Defendant'].include?(params[:search_field])
```

 Also, i note that by searching this way we restrict the flexibility of the search as a user cannot search by a combination of **different** advocate and defendant names. However, the Use Case may not require this??

app/controllers/advocates/claims_controller.rb
```diff
@@ -8,8 +9,8 @@ def landing; end
 
   def index
     claims = @context.claims.order(created_at: :desc)
-    claims = claims.find_by_advocate_name(params[:search_advocate]) if params[:search_advocate].present?
-    claims = claims.find_by_defendant_name(params[:search_defendant]) if params[:search_defendant].present?
+    claims = claims.find_by_advocate_name(params[:search]) if ['All', 'Advocate'].include?(params[:search_field])
```
jsugarman, line 12: see my main comment - this code (i believe) makes an `All` search (signed in as an advocate) filter by advocate and defendant name despite not having an 'Advocate' search field option

PR #203, opened 2015-06-16
Require an S3 bucket name to be set for Paperclip in production envs.

ashb:
We can sort of have a sane default for develop and test, but on
production we require this to be set as the app shouldn't run without
this set correctly


ashb:
Well that won't work - it evaluates the production config section even in development simply to render the YAML config file. Dang.

PR #212, opened 2015-06-16
select2 for expense and document types

himal-mandalia:
select2 dropdown for expense and document type dropdowns on claim form.


app/views/advocates/claims/_document_fields.html.haml
```diff
@@ -10,7 +10,7 @@
   = f.label :document
   - unless f.object.id.nil? || f.object.document_file_name.nil?
     .note.file-exists
-      File already uploaded: 
+      File already uploaded:
```
cfq, line 13: We shouldn't mix whitespace changes with functionality

PR #228, opened 2015-06-19
Multiple rep orders

fordie:


app/views/advocates/claims/_defendant_fields.html.haml
```diff
@@ -25,7 +25,10 @@
     #documents
       = f.fields_for :representation_orders do |repo_form|
         = render 'representation_order_fields', f: repo_form
+        .links
```
EdwardAndress, line 28: Would it be useful to have a 'remove representation order' link?

PR #229, opened 2015-06-19
Start internationalizing

EdwardAndress:


app/views/advocates/admin/advocates/_form.html.haml
```diff
@@ -1,27 +1,26 @@
...
-    
+    %h2= "#{pluralize(@advocate.errors.count, 'error')} " + t("admin.form.errors")
+    %ul
+      - @advocate.errors.full_messages.each do |msg|
+        %li= msg
+
   %fieldset
     = f.fields_for :user do |user_fields|
-      = user_fields.label :email
+      = user_fields.label t("admin.form.user_email")
```
himal-mandalia: This should be used as t('.email'), so the en.yml struture should be:

```
advocates:
    admin:
      advocates:
        form:
          email: Hi
```

PR #244, opened 2015-06-22
fix 'moving Add Date Attended Buttons'- bug where buttons would appea…

EdwardAndress:
…r in wrong place
- It would seem that the extra rows for attendance dates were being inserted in the wrong place, pushing the Add Date Attended buttons out of the table
- I moved the lines which render the _date_attended_fields partial to the end of the _fee_fields partial and this seems to have worked


EdwardAndress:
The second build failed due to a flickering test, which Himal will fix tomorrow.  After he's done that, I'll reopen this PR.

PR #245, opened 2015-06-22
sanitize regexp interpolation for name (to escape problem punctuation…

EdwardAndress:
… from Faker)


himal-mandalia:
Probably a good idea to fix this up in other specs too (see screenshot)
screen shot 2015-06-23 at 08 56 29

EdwardAndress:
Thanks Himal.

I've refactored those specs where a string is interpolated into a regexp.  But some of those listed in the screenshot are not interpolations (e.g. "expect(response.body).to match(/Dashboard.*Advocates/)") ... so they should be fine as they are... should they not?

himal-mandalia:
That's right Eddie - only the ones with interpolations matter, should've clarified that. Good work, that should take care of all our flickers for now.

PR #255, opened 2015-06-24
Add Granting Body to representation order

stephenrichards:
and move MAAT reference and representation order date from defendant to rep order


jsugarman:
Nice! pulled locally and sanity tested adding and viewing multiple rep orders and searching by MAAT reference. all good

PR #282, opened 2015-07-02
Modified Dockerfile to shorten build time

bennythejudge:
Merged existing Dockerfile with the upstream
  Made changes to use caching more effectively
  Upgraded ruby version in Gemfile to 2.2.2
  Removed non necessary statement


ashb:
The one thing I might suggst adding is some "comment guards" to make it clear which bit of the docker file is the 'webapp-onbuild' Docker file.

Something like:

```
FROM ministryofjustice/ruby:2

###
### Start of copy from ruby:2-webapp-on-build
###

# add official nodejs repo
RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
        echo 'deb https://deb.nodesource.com/node jessie main' > /etc/apt/sources.list.d/nodesource.list


…
…

###
### End of copy from ruby:2-webapp-on-build
###

WORKDIR $APPHOME

…
…
```

And a comment about 'weve added libreoffice to this list - otherwise as per stock' so that we can unpick this change later if we can find a nicer solution.

PR #307, opened 2015-07-09
add dummy case workers to demo data

jsugarman:
This enables tests for allocation tool to have numerous
case workers to allocate to. NOTE that this may be
replaced with a seed file of real case workers.


jsugarman:
hold on this. going to increase fee amount data to have demo data over threshold of 20k

PR #324, opened 2015-07-20
Api squashed and merged

jsugarman:
This is the initial work on production of an API
for the submission of draft claims. The intent is to provide
a means by which Case Management Systems used by many big
advocate chambers can submit claims to ADP programmatically,
thereby removing the need for rekeying of data.

This branch includes:
- use of Grape gem for quick API
- use of Swagger gem for automated documentation
- use of Swagger UI for a testing interface for the API
- a claim creation API endpoint
- a claim validation API endpoint
- error handling for the above endpoints
- rspecs for the above endpoints


Gemfile
```diff
@@ -70,8 +73,9 @@ group :test do
   gem 'simplecov-multi',            require: false
   gem 'timecop',                    '~> 0.7.4'
   gem 'webmock',                    '~> 1.21.0'
+  gem 'airborne'
```
struds: airbourne is probably in here by accident. .I think I was going to use it then backed out. this would have been used in the API tests

Gemfile.lock
```diff
@@ -28,6 +28,11 @@ GEM
       thread_safe (~> 0.1)
       tzinfo (~> 1.1)
     addressable (2.3.8)
+    airborne (0.1.16)
+      activesupport (>= 3.0.0)
+      rack-test (~> 0.6, >= 0.6.2)
+      rest-client (~> 1.7, >= 1.7.3)
+      rspec (~> 3.1, >= 3.1.0)
```
struds: ditto with airbourne

app/interfaces/api/v1/advocates/claim.rb
```diff
@@ -0,0 +1,105 @@
...
+          end
+        end
+
+      end
+
+      add_swagger_documentation hide_documentation_path: true
+      end
+    end
+  end
+end
```
struds, line 105: I think some of this logic can be moved into other classes at a later date, theres logic in the helpers around arguments and valid arguments which could be moved into a service object of the model.

there is still a small amount of work to be done on the claims api so we can have a go at refactoring at that point.. I think this class is just on the tipping point of too much logic hiding the function of the class (which is to describe the API)

struds:
I think the only things are:
- remove airbourne gem
  - include a description in the PR with an overview of whats been added (grape, grape swagger and swagger ui for the claims API)

and we can do some minor refactors in the next PR for the claims API. After those 2 changes are in it looks good to merge.

jsugarman:
many thanks stephen. been working on another branch but will tweak the api
branch asap as you suggest and then get stephen R to review and merge.

i certainly think my error handling code could do with a refactor but
getting it in for now sounds good to me.

PR #326, opened 2015-07-20
Separate fixed misc fee sections

jsugarman:
fixed fees only apply to fixed fees case/claim types
which are to be applied in a future ticket. This
lays the groundwork for that ticket.

includes:
- remove non basic fee logic (model scopes, partials)
- add separate scopes and partials for fixed/misc fees
- amend specs in line with changes
- add cukes


stephenrichards:
Let's look at this method with the same name as an association together and see what's going on.

PR #332, opened 2015-07-24
Claims api

struds:
Refactored claims API to include:
- default logging of all API actions via rack middleware
- added source field to claims api so that we can tell if the request came from the API or web


struds:
my attempt to squash seems to have picked up too many previous commits making this seem bigger than it is.. damn git is annoying sometimes.

I'm not going to have a chance to seriously look at this again to Wednesday so either someone else will have to fix, or just go for the merge

PR #343, opened 2015-07-28
Implementation of VatRate model and specs

stephenrichards:
This just implements the model, factory and specs for VatRate.  This model will be used in a later PR to calculate VAT and display as a separate line on the claim.

### DIFF
--- a/app/models/vat_rate.rb
+++ b/app/models/vat_rate.rb
@@ -1,0 +1,52 @@
+# == Schema Information
+#
+# Table name: vat_rates
+#
+#  id               :integer          not null, primary key
+#  rate_base_points :integer
+#  effective_date   :date
+#  created_at       :datetime
+#  updated_at       :datetime
+#
+
+
+
+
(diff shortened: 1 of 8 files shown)


app/models/vat_rate.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  class MissingVatRateError < RuntimeError;end
+
+  validate :effective_date, uniqueness: true
+
+  @@rates = nil
+
+  class << self
+
+    def instantiate_rates
```
EdwardAndress: How about 'assign_rates' ?  I find 'instantiate' a little misleading here.

app/models/vat_rate.rb
```diff
@@ -0,0 +1,55 @@
...
+      rate_for_date(date)
+    end
+
+ 
+    private
+
+    def rate_for_date(date)
+      result = nil
+      @@rates.each do | rec |
+        unless date < rec.effective_date
```
EdwardAndress, line 45: Is 'rec' short for 'record'?  Why not use the singular 'rate' ?

PR #347, opened 2015-07-29
add amount as required param to fee API

jsugarman:


app/interfaces/api/v1/advocates/fee.rb
```diff
@@ -21,6 +21,7 @@ class Fee < Grape::API
               requires :claim_id, type: Integer
               requires :fee_type_id, type: Integer
               requires :quantity, type: Integer
+              requires :amount, type: Float
```
EdwardAndress, line 23: :amount needs to be added into #args (should it's presence also be validated on the model?)

jsugarman, line 23: @EdwardAndress -  those changes now in for you to review

PR #383, opened 2015-08-04
Refactor API endpoints to use UUIDs

himal-mandalia:
API endpoints updated to use UUIDs for ID params (claims, defendants etc).


jsugarman:
please see merged version of this branch

PR #384, opened 2015-08-04
Add meta tag to make iOS safari not render numbers as telephone numbers

aliuk2012:
We can still convert telephone numbers to display and work as telephone numbers but we have to explicitly define them as such.
reference: https://developer.apple.com/library/safari/featuredarticles/iPhoneURLScheme_Reference/PhoneLinks/PhoneLinks.html

### DIFF
--- a/app/views/layouts/advocate.html.haml
+++ b/app/views/layouts/advocate.html.haml
@@ -1,5 +1,6 @@
 - @disable_logo_link  = true
 - content_for :head do
+  %meta{:name =>"format-detection", :content => "telephone=no"}
   = stylesheet_link_tag "advocates/application", media: "all"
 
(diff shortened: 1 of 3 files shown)


jsugarman:
hey,
reran the travis and it passed second time (flickering test maybe - worrying).  Regarldess i note that in http://html2haml.herokuapp.com/ that it adds a trailing forward slash to the end of the meta tag to ensure self closing tags - as mentioned here http://haml.info/docs/yardoc/file.REFERENCE.html.

does it need this? `%meta{:content => "telephone=no", :name => "format-detection"}/`

aliuk2012:
Ah, forgot about trailing '/'. Well spotted and thanks. New commit added to reflect the changes.

PR #388, opened 2015-08-05
add defendant name(s) to allocation tool

jsugarman:


jsugarman:
Alistair, this minor change to the allocation tool adds a column which squeezes up the others and makes it somewhat unattractive. can anything be done about from front-end

jsugarman:
ok, i have just made that tweak. i think it would be useful to add tickets for you in backlog to look at the the styling for this page (allocation tool) and another ticket i am working on which is widening the expense section of the new/edit form and could do with some love. can chat about it after they are in.

PR #391, opened 2015-08-06
Add dates attended to expenses merged

jsugarman:
Add dates attended to expenses

expenses also require a date as specified
in the AF1 form. The existing dates_attended
model has therefore been associated with the
expense model as it is with the fee model.

includes:
- migration to reference dates_attended from expense
- update model associations
- update model spec for dates_attended
- update claims:demo_data
- update views

could add in future branch:
- cukes to excercise adding dates attended to expenses (js related)
- need to modify API in line with this


app/models/date_attended.rb
```diff
@@ -12,9 +12,19 @@
 
 class DateAttended < ActiveRecord::Base
   belongs_to :fee
+  belongs_to :expense
```
himal-mandalia, line 15: I really think this ought to be a polymorphic association. Replacing both belongs to associations with:

belongs_to :date_attendable

app/models/date_attended.rb
```diff
@@ -12,9 +12,19 @@
...
 
   validates :date, presence: true
-  validates :fee, presence:  true
+  validates :fee, presence:  true, if: "expense.nil?"
+  validates :expense, presence:  true, if: "fee.nil?"
+  validate  :belongs_to_fee_or_expense
+
+  def belongs_to_fee_or_expense
```
himal-mandalia, line 22: Having a polymorphic relationship between fees, expenses and dates attended will obviate the need for this validation.

himal-mandalia:
@jsugarman I've left some comments. We should sit down and refactor this as a polymorphic relationship.

PR #404, opened 2015-08-13
Break out amount assessed into fees and expenses

stephenrichards:
- Refactor so that assessment record is on a separate table (the same table will be used for determinstiaon)
- clalculated the assessed total from assesed expenses + assessed fees


app/controllers/advocates/claims_controller.rb
```diff
@@ -40,6 +40,8 @@ def show
     @doc_types = DocType.all
     @messages = @claim.messages.most_recent_first
     @message = @claim.messages.build
+    @enable_assessment_input = false
+    @status_change_disabled = true
```
EdwardAndress: The outcome of lines 43 and 44 is that the advocate can neither change status nor input assessment (i.e. both are negating an ability), but one is set using 'true' and the other 'false'.

How about:  44 @enable_status_change = false ?

PR #409, opened 2015-08-13
Caseworker Clerk Home page top section

aliuk2012:
Just removes the breadcrumbs and tabs. A "Claims" page heading added and tests updated to pass.


features/caseworker_claims_list.feature
```diff
@@ -11,14 +11,14 @@ Feature: Caseworker claims list
...
 
-  Scenario: View completed claims
-    Given I am a signed in case worker
-      And There are fee schemes in place
-      And I have completed claims
-     When I visit my dashboard
-      And I click on the Completed Claims tab
-     Then I should see only my claims
-      And I should see the claims sorted by oldest first
+  #Scenario: View completed claims
```
jsugarman, line 15: can this scenario be re-written to be whatever you have now

jsugarman:
further to our discussion i think you you should comment out entire scenarios that are otherwise not exercising/testing any components and use a # TODO comment on each one for the time being so that, at a later stage, we can run "rake notes" to easily identify those scenarios.

aliuk2012:
I've added the TODO notes and refactored the tests as discussed. BTW, running rake notes looks like it doesn't look at feature or spec folder. if you run rake notes now you will see that none of my TODO's are listed.

PR #414, opened 2015-08-14
Changed to add logstasher to the application

bennythejudge:
The application will log in ELK-ready JSON format to the ELK clusters.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -49,4 +49,22 @@ Rails.application.configure do
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+  # logstasher
+  # Enable the logstasher logs for the current environment
+  config.logstasher.enabled = true
+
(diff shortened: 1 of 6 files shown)


config/environments/development.rb
```diff
@@ -49,4 +49,22 @@
...
+  config.logstasher.source = 'Advocate Defence Payments App development'
+
+  # This line is optional if you do not want to suppress app logs in your <environment>.log
+  config.logstasher.suppress_app_log = true
+
+  # This line is optional if you do not want to log the backtrace of exceptions
+  config.logstasher.backtrace = true
+
+  # Enable logging of controller params
+  config.logstasher.log_controller_parameters = false
```
ashb, line 66: Did you chose to make this false rather than scrub the parameters we are worried about?

ashb:
Given that the same config settings are duplicated across all the environments it would be cleaner if it was moved 'up' to config/application.rb?

Also on dev we almost certainly do want parameters enabled

PR #449, opened 2015-08-24
Rename account number to supplier number on Advocate and Chamber

stephenrichards:


spec/models/advocate_spec.rb
```diff
@@ -163,7 +163,7 @@
 
   describe '#name_and_number' do
     it 'should print last name, first name and account number' do
```
jsugarman: should print last name, first name and supplier number

spec/models/advocate_spec.rb
```diff
@@ -32,30 +32,30 @@
...
-      a = FactoryGirl.build :advocate, account_number: 'AC12'
+      a = FactoryGirl.build :advocate, supplier_number: 'AC12'
       expect(a).not_to be_valid
-      expect(a.errors[:account_number]).to eq( ['must be 5 alhpa-numeric characters'] )
+      expect(a.errors[:supplier_number]).to eq( ['must be 5 alhpa-numeric characters'] )
     end
 
     it 'should fail validation if not alpha-numeric' do
-      a = FactoryGirl.build :advocate, account_number: 'AC-123'
+      a = FactoryGirl.build :advocate, supplier_number: 'AC-123'
```
jsugarman: this might fail validation just because it is too long, rather than non-alpha numeric - "AC-23" better?!

db/seeds/advocates.rb
```diff
@@ -1,4 +1,4 @@
-chamber = Chamber.find_or_create_by!(name: 'Test chamber/firm', account_number: 'A1234567')
+chamber = Chamber.find_or_create_by!(name: 'Test chamber/firm', supplier_number: 'A1234567')
```
jsugarman, line 1: should this account no not break validation because it is too long? - scrap that this the chambers account/supplier no. which has no such validation (should it?)

jsugarman:
noticed a few minor issues that are leftovers of previous code (see individual comments) rather than anything you changed but am otherwise happy to merge in any event. 

Also, just to flag this up, i notice that advocates have an account/supplier no. AND a chamber id while chambers have an account/supplier no. too - are the numbers different or are we possibly storing redundant data for advocates? (it may be that not all advocates have a chambers or they are different numbers!?)

aliuk2012:
Fixed Merge conflict in my commit b37d396

PR #495, opened 2015-09-08
Further styling to the advocates home report

aliuk2012:


aliuk2012:
I'm busying working on the failing tests

PR #502, opened 2015-09-09
Ajax file uploads

himal-mandalia:
AJAX multi-file uploader on claim form. Drag and drop or select multiple files. Endpoints to create and destroy documents.


jsugarman:
Just a couple of points of usability bonus points:
- a warning that invalid files will not be saved
- clear image for tick image

PR #534, opened 2015-09-15
this js was no longer in use

EdwardAndress:


himal-mandalia:
This is still used in the file upload/attachment for messages. It's probably going soon.

EdwardAndress:
Ah, yes, I forgot about that fileupload!

PR closed.

PR #567, opened 2015-09-22
Claims History Styling

aliuk2012:
A new Claims History Presenter was created by @himal-mandalia and used to construct the markup for the claims history and messaging section. New styles were then applied to the claim history message and to the date separator.

The AJAX now updates the full list of history/messages for the user when they submit a new message.


jsugarman:
hi alistair, have pulled this locally and it looks good. i have not had much to do with this aspect so not sure what ticket it relates too. am happy to merge if you think @himal-mandalia does not need to look at it.

one thing i note is that when you attach a document you have no way of knowing that you have attached it until you hit send (i.e. maybe some visual indicator in the background of the message body field or next to it) - new ticket?

aliuk2012:
Hi Joel, 

Sorry I didn't realise you would have looked at this last night. This is the ticket https://trello.com/c/rqrUKCcc it relates to. I dont think @himal-mandalia needs to review it, I only mentioned him as he had created the ClaimHistoryPresenter which was merged into develop as a separate PR.

You are right about the visual indictor of what file is being uploaded.  I'll create another ticket for it because its a fault with the design and not sure where best to position it.

PR #571, opened 2015-09-23
Automatically mark messages read when show action loaded

himal-mandalia:


aliuk2012:
Hi @himal-mandalia, I see theres two new issues that code climate is complaining about. There is two Identical code with the mark_messages_read being used in both the caseworker and advocate.  Could we not make that somehow global? it reduces the the repo code climate from 2.9 to 2.8

himal-mandalia:
@aliuk2012 , yes we could move the method and after action into a controller concern then include it in both controllers. I was thinking about doing that this morning, then long meeting happened... Leave this PR open, I'll do a quick refactor tomorrow morning.

PR #610, opened 2015-09-30
Remove the caseworker allocation tabs and add to the primary nav

aliuk2012:


app/views/layouts/_primary_navigation.html.haml
```diff
@@ -11,7 +11,8 @@
     - elsif current_user.persona.is_a?(CaseWorker)
       - if current_user.persona.admin?
         = link_to 'Summary', case_workers_admin_root_path
-        = link_to "Allocation", case_workers_admin_allocations_path
+        = link_to "Allocation", case_workers_admin_allocations_path(params.except(:controller, :action, :filter, :search).merge(tab: 'unallocated'))
```
jsugarman, line 14: Do we want this link text to include the count of applicable claims (as the tabs did)?

app/views/layouts/_primary_navigation.html.haml
```diff
@@ -11,7 +11,8 @@
...
+        = link_to "Re-allocation", case_workers_admin_allocations_path(params.except(:controller, :action, :filter, :search).merge(tab: 'allocated'))
```
jsugarman, line 15: Do we want this link text to include the count of applicable claims (as the tabs did)?

aliuk2012:
@jsugarman, I was thinking the same but wondered if it would actually be of any use to the user. A count doesn't seem useful other than if its a count of any new claims that have come in since they last visited the Claims list. Lets check with the others this morning before merging.

PR #624, opened 2015-10-04
New claim - Top part - Basic case information

aliuk2012:
Better usage of the white space at the top of the page by having two options on a line.


himal-mandalia:
This looks great @aliuk2012, especially the form.

PR #640, opened 2015-10-07
Fix submodel validation error bubbling

jsugarman:
- use of validate_associated to raise a single claim (summary level) error for sub model errors
  on defendants, fees and expenses that were otherwise not preventing successful claim submission
- applied validation standards to expense model


jsugarman:
need to align with demo data

PR #663, opened 2015-10-09
omit these atts from versioning

EdwardAndress:
We do not need to version changes to :evidence_checklist_ids because these will not change after submission (evidence will be added via the messaging tool) and we do not need to version the claims during their creation (prior to submission).

We do not need to track :submitted_at because state change is already versioned.


EdwardAndress:
Versioning these items was causing useless items to be added in claim history

PR #692, opened 2015-10-15
Only allow permitted state changes in GUI

stephenrichards:
- In the case worker detail view, the state change drop down now only show states which the claim can be transitioned to, rather than all states
- the status partial has been separated into two separate partials - one for case workeres where the state can be changed, and the other for advocates and caseworkers where the state cannot be changed.


app/presenters/claim_presenter.rb
```diff
@@ -1,6 +1,19 @@
 class ClaimPresenter < BasePresenter
   presents :claim
 
+
+  # returns a hash of state as a symbol, and state as a human readable name suitable for use in drop down
+  #
+  def valid_transitions
+    states = claim.state_transitions.map(&:to_name) - [:archived_pending_delete]
```
jsugarman, line 8: do we need to remove submitted aswell? i pulled this branch and tried to set status to submitted and it blew up
screen shot 2015-10-15 at 13 05 29

PR #706, opened 2015-10-19
must save claim before validating total > 0

EdwardAndress:
- save claim before forcing validation to update claim total
- wrap this all in a transaction so that save is undone if claim is then invalid
- do not apply validation for total on claims which come via the API because, in this case, the claim is created and saved before any fees or expenses


EdwardAndress:
closed whilst I fix specs

PR #710, opened 2015-10-20
only advocates change state to :await_written_reasons or :redetermine

EdwardAndress:
Addressing this bug report from Alistair:
- Case workers are getting options to request written information and to apply for redetermination.
- Test case
- Create a claim and get it to the point where the advocate has requested redetermination.
- Login as a case worker admin and allocate that claim.
- Login as caseworker and edit that claim
- You should see the 4 status options (refused, rejected,part-auth, auth).
- Choose "refused"(i suspect the same applies for other status types too) and click the "Update" button.
- The page submits and everything shows it been updated to the selected status except the case worker now sees two status options (redetermination and apply for written reasons) that were only previously available to advocates.

It doesn't seem to be an issue in staging so its something we have introduced recently

### DIFF
--- a/app/presenters/claim_presenter.rb
+++ b/app/presenters/claim_presenter.rb
@@ -13,9 +13,12 @@ class ClaimPresenter < BasePresenter
   end
 
   def valid_transitions_for_detail_form
-    valid_transitions(include_submitted: false)
+    if claim.state == "allocated"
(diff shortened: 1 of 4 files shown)


EdwardAndress:
- I've also dealt with part of a closely related ticket here
- The caseworker should not be able to change status in response to a request for written reasons

PR #717, opened 2015-10-21
Modified rake task in run.sh to load demo data

bennythejudge:


docker/run.sh
```diff
@@ -72,8 +72,8 @@ reload)
     bundle exec rake db:create 
     echo "executing rake db:schema:load"
     bundle exec rake db:schema:load
-    echo "executing rake db:data:load"
-    bundle exec rake db:data:load
+    echo "executing rake claims:restore"
+    bundle exec rake claims:restore
     ;;
 reseed)
     # db:clear
```
stephenrichards, line 73: You're going to kill me Ben, but I think we're going to drop the idea of dumping data and restoring, and go back to the old idea of re-seeding the database each deploy.

bennythejudge, line 73: Absolutely not a problem don't worry. it's just a minimal change. Please let me know if I need to close this, i will leave it for the time being. The complication was Trevis outage..

bennythejudge:
closing down due to duplication and confusion

PR #718, opened 2015-10-21
Remove Scheme model

himal-mandalia:
Removes Scheme model and associations.


stephenrichards:
Looks OK to me, but there are conflicts

PR #723, opened 2015-10-22
Table row checkbox can be selected by clicking anywhere in the row

himal-mandalia:
Clicking table row on allocation page selects the checkbox in that row.

### DIFF
--- a/app/assets/javascripts/advocates/claims/table_row_click.js
+++ b/app/assets/javascripts/advocates/claims/table_row_click.js
@@ -3,21 +3,22 @@
 var adp = adp || {};
 
 adp.tableRowClick = {
-  $selector: null,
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/advocates/claims/table_row_click.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+var adp = adp || {};
+
+adp.tableRowClick = {
+  $selector: null,
+
+  init : function(table_id) {
+    if($('#' + table_id).length) {
+      adp.tableRowClick.$selector = '#' + table_id;
```
aliuk2012: Why does it need an id? Class would be better because what if you had two tables on the page  you would not be able to give them the same ID as id need to be unique.

PR #731, opened 2015-10-29
Zendesk integration for feedback/user satisfaction

himal-mandalia:
Relies on env variables being set:

ZENDESK_USERNAME
ZENDESK_TOKEN

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -6,16 +6,36 @@ class FeedbackController < ApplicationController
   end
 
   def create
-    redirect_to redirect_after_create_url, notice: 'Feedback submitted'
+    @feedback = Feedback.new(merged_feedback_params)
+    if @feedback.save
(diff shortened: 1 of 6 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -6,16 +6,36 @@ def new
...
   end
 
   private
 
-  def redirect_after_create_url
+  def merged_feedback_params
+    feedback_params.merge({
+      email: (current_user.email rescue 'anonymous'),
+      referrer: request.referer,
+      user_agent: request.user_agent
```
aliuk2012, line 23: I like how you have merged the params instead of throwing them out on the page only to be read back in. From a security point of view at least people cant tamper with it. I think Prison visits implementation of referrer and user agent is on the page : https://github.com/ministryofjustice/prison-visits/blob/master/app/views/feedbacks/_form.html.haml#L15-L17

PR #736, opened 2015-10-29
Status message bugs and copy

EdwardAndress:


app/views/shared/_message.html.haml
```diff
@@ -61,4 +61,6 @@
         .event
           %ul
             %li
-              = state_change.change
+              %strong
```
EdwardAndress: @aliuk2012 For some reason this is only rendering the updates in bold on the advocates side... any idea why that might be?

PR #775, opened 2015-11-12
Json importer api token

EdwardAndress:
- The API key is now taken from current_user.persona.chamber in the json_document_importers_controller
- I have removed validation against the creator having admin rights (because regular advocates should be able to create claims, in their CMS, and submit via the API).
- Some adjustments to the example docs and specs were required.


EdwardAndress:
I've placed this PR now in order to break up the work into more manageable chunks and will do some refactoring on Monday that should sort out the codeclimate issues.

PR #796, opened 2015-11-19
some fees can be claimed for dates before the start of trial

EdwardAndress:
- this relates to a common complaint, from advocates, that some fees will validly relate to dates that are before the start of trial


jsugarman:
hey @EdwardAndress if this has been agreed by BA then i would suggest you remove the translation for this error too, in error_messages.en.yml

PR #816, opened 2015-12-02
Add fee calculations

jsugarman:
This PR reintroduces simple fee calculation for all fee types.

It also applies conditional logic that prevents recalculation of individual fees amounts for claims that
already exist (for private-beta/gamma environment). This conditional logic permits rates of 0 if an amount already exists.


jsugarman:
@aliuk2012 to review a javascript issue

PR #822, opened 2015-12-03
Rename service

jsugarman:
This PR change the name of the application from the users point of view (only) from Advocate Defence Payments to "Claim for crown court defence work"

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Advocate Defence Payments
+###a.k.a Claim for crown court defence work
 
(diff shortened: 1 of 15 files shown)


aliuk2012:
Just had a look through the commits in light of what we talked about at the last Tech COP. Shouldn't the following commits (https://github.com/ministryofjustice/advocate-defence-payments/commit/505a5e097ace13a8fcdf035dc2a6ec76e23bab9d and https://github.com/ministryofjustice/advocate-defence-payments/commit/a144e05e97237eccd94b2d7c156c997140ceac50) be a separate PR or be mentioned in the PR description because it doesn't actually relate to "Rename Service"

himal-mandalia:
Could we put the service name in the settings.yml or somewhere, then we wouldn't need to update all the occurrences.

PR #824, opened 2015-12-05
Update Claim details basic information section 

aliuk2012:
1. Update the Claims details basic information section layout.
2. A whole new layout for defendants to show how representation orders on other data relates to each individual defendant.
3. Include all trial details data (Date of first trial, estimated trial length, actual trial length and date trial concluded) where the case type is a "Trial" or a "Retrial.

### DIFF
--- a/app/assets/stylesheets/claim-accordion.scss
+++ b/app/assets/stylesheets/claim-accordion.scss
@@ -29,18 +29,6 @@
     }
   }
 
-  ul.case-data {
-    overflow: hidden;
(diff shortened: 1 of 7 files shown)


himal-mandalia:
There will be nil errors on some fields when people open it as a draft, e.g. defendant date of birth (undefined method `strftime' for nil:NilClass).

You can catch this by doing:

= defendant.date_of_birth.strftime('%d/%m/%Y') rescue ''

I'd check other fields on draft claims for the same problem.

PR #829, opened 2015-12-08
"Discontinuance" case type should not require trial dates

himal-mandalia:
"Discontinuance" case type should not require trial dates. Hides trial date fields on claim form and turns off trial date validations.


jsugarman:
:+1: merging for you because its awesome and you're busy

PR #833, opened 2015-12-09
Fix validation so does not trigger before advocate selected

himal-mandalia:
Fix advocate/creator validation so it does not trigger before advocate is selected.


jsugarman:
:+1: cool

PR #836, opened 2015-12-11
Claim Detail - Claim/Assessed amounts

aliuk2012:
Now includes a break down of VAT for users to easily see how much VAT is added to the total if it is applied.


app/models/determination.rb
```diff
@@ -25,6 +26,14 @@ def calculate_total
     self.total = self.fees + self.expenses
   end
 
+  def calculate_vat
```
jsugarman, line 30: hey @aliuk2012, discussed this with @EdwardAndress but just to show you a local manual test where i zeroed the vat_amount for a determination and then viewed it in show view - i.e. how it would be on gamma. This probably needs conditional logic to not confuse users (or ask Product Management if they will accept this)
screen shot 2015-12-14 at 12 08 47

EdwardAndress, line 30: I suspect that if existing determinations have VAT included the total will
reflect that, but the VAT line in the breakdown will be blank.  So the two
totals will be the same.

I'll check this though...

aliuk2012:
Good spot, yes I dont think it should show a figure for Claimed VAT amount.  I suppose we could potentially add conditional login to completely hide the VAT and Total (Inc VAT) rows but I think for the moment they just need this broken down for now.

Maybe a refactor thing for another time?

EdwardAndress:
I think we should find a very simple solution for the time being that will
stop people from being confused or concerned about the breakdown for their
claim (and prevent us from having to deal with their questions).  We could
temporarily add in a bit of explanatory text, for example.  Or calculate
the VAT based on whether the claim has vat and the total amount assessed
(i.e. deduce it from the total amount in cases where VAT is included)

aliuk2012:
We could hide the (VAT and Total inc VAT rows in the assessment if the claim shoudl not have VAT replied

EdwardAndress:
It would be more a question of when the determination was made, since some
old determinations will include VAT but not have the breakdown so the VAT
row will be 0 despite VAT being paid

aliuk2012:
:confused: But shouldn't we be looking up the vat rate using the date the claim was submitted not when redetermination was made?

EdwardAndress:
there's a claim#vat_date for that

but showing / hiding the VAT line dependent on whether VAT is applied will
not solve the problem where there's not vat_amount for determinations that
pre-date the migration.  In that case, the line will be shown but will not
have an amount (it will be the default value; zero)... so we would need to
either add some explanatory text (easiest option) or add some logic in the
presenter which can deduce the VAt amount.

PR #840, opened 2015-12-14
Move feedback link

EdwardAndress:
- simple solution for feedback / bug report issues


aliuk2012:
Why is there two different sets of changes being made in this PR? 

Couldn't that single commit be broken out into its own PR? using 

```
git checkout -b help-text 7971f05
```

EdwardAndress:
@aliuk2012 - that was done in error; thanks for pointing it out (i.e. I was being a bit lazy)

PR #841, opened 2015-12-14
popluate daily attendance fees from actual trial length

jsugarman:
The number/quantity of daily attendance fees is calculable from the actual trial length. This JS populates the DAF quantities based on the user-specified trial length to:
a. avoid user error
b. simplify form completion.

### DIFF
--- a/features/advocate_new_claim.feature
+++ b/features/advocate_new_claim.feature
@@ -181,3 +181,24 @@ Feature: Advocate new claim
       And I submit to LAA
      Then There should not be any Initial Fees saved
       And There should be a Miscellaneous Fee Saved
+
+  @javascript @webmock_allow_localhost_connect
+  Scenario Outline: Daily attendance fees derived from actual trial length
+  Given I am a signed in advocate
+    And There are case types in place
+    And I am on the new claim page with Daily Attendance Fees in place
+   When I fill in the trial details
+    And I fill in actual trial length with <actual_trial_length>
+   Then The daily attendance fields should have quantities <daf_quantity>, <dah_quantity>, <daj_quantity>
+
+  Examples:
+    | actual_trial_length         | daf_quantity | dah_quantity | daj_quantity |
+    | 1                           | 0            | 0            | 0            |
+    | 2                           | 0            | 0            | 0            |
+    | 3                           | 1            | 0            | 0            |
+    | 40                          | 38           | 0            | 0            |
+    | 41                          | 38           | 1            | 0            |
+    | 50                          | 38           | 10           | 0            |
+    | 51                          | 38           | 10           | 1            |
+    | 60                          | 38           | 10           | 10           |
+    | 70                          | 38           | 10           | 20           |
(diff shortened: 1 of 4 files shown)


features/advocate_new_claim.feature
```diff
@@ -181,3 +181,24 @@ Feature: Advocate new claim
       And I submit to LAA
      Then There should not be any Initial Fees saved
       And There should be a Miscellaneous Fee Saved
+
+  @javascript @webmock_allow_localhost_connect
```
aliuk2012, line 185: I don't think you need @webmock_allow_localhost_connect, as far as I understand it this is just to allow ajax requests to be used in the tests and this feature doesn't require any ajax. I've removed it locally and tested and the tests still pass.

features/step_definitions/advocate_new_claim_steps.rb
```diff
@@ -411,6 +411,32 @@
...
+  create(:fee_type, :basic, description: 'Basic Fee', code: 'BAF')
+  create(:fee_type, :basic, description: 'Daily attendance fee (3 to 40)', code: 'DAF')
+  create(:fee_type, :basic, description: 'Daily attendance fee (41 to 50)', code: 'DAH')
+  create(:fee_type, :basic, description: 'Daily attendance fee (51+)', code: 'DAJ')
+  visit new_advocates_claim_path
+end
+
+When(/^I fill in the trial details$/) do
+  select2('Trial', from: 'claim_case_type_id')
+  fill_in 'claim_actual_trial_length', with: 1
```
aliuk2012: I dont think this is required because the actual trial length is set to 1 only to then be changed on this line https://github.com/ministryofjustice/advocate-defence-payments/pull/841/files#diff-0ed77b0f3d6ca4f188322cbe3a22a955R191 in the feature. Maybe this step could be renamed to something like 'I select the case type as a trial'  (on a side note perhaps this should really be a global/shared step to set the case type as I'm sure other features use/ would do something similar)

features/advocate_new_claim.feature
jsugarman, line 185: This breaks my test - with:

```
 Real HTTP connections are disabled. Unregistered request: GET http://127.0.0.1:58747/__identify__ with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent'=>'Ruby'}
```

i believe because the JS assets are not available - may be a setup issue??

features/step_definitions/advocate_new_claim_steps.rb
jsugarman: cool, you're right and, in fact, i had written a step that can be used previously - will apply

features/cracked_trial.feature
```diff
@@ -11,7 +11,7 @@ Feature: Cracked trial
     Given I am a signed in advocate
       And I am on the new claim page
      Then I should NOT see Cracked trial fields
-      And I select2 "Cracked trial" from "claim_case_type_id"
+      And I select2 a Case Type of "Cracked trial"
```
jsugarman, line 14: just tweaked these while i was at it.

PR #842, opened 2015-12-14
add help text for expense hours

EdwardAndress:
- add help text for expense quantity


config/locales/en.yml
```diff
@@ -398,7 +398,8 @@ en:
         add_fixed_fee: 'Add another Fixed fee'
         add_misc_fee: 'Add another miscellaneous fee'
         expenses: 'Expenses'
-        expenses_help_text: 'Please enter details for each separate date'
+        expenses_help_text: Enter details and relevant dates for each separate expense
+        hours_help_text: Hours should be in decimal format (eg 1.25 hours for 1 hour and 25 mins) and will be rounded to the nearest quarter (eg 1.40 becomes 1.50)
```
aliuk2012: I'm not sure about this copy sounds rather "techy". i.e what if someone thinks 15mins is 0.25 (i.e a quarter)? the last eg is confused because someone might read the 1.50 as being 1hr 30mins

EdwardAndress: Actually that's a mistake (and I thought I had removed it but this ended up
bouncing back and forth between Rachel and I so I lost track...); it should
read "(eg 1.25 hours for 1 hour and 15 mins)"

EdwardAndress: This is the copy she's seen and approved though.

PR #874, opened 2015-12-31
Advocate maintain user

aliuk2012:
This PR just needs the search functionality bit added in the backend

### DIFF
--- a/features/step_definitions/advocate_admin_steps.rb
+++ b/features/step_definitions/advocate_admin_steps.rb
@@ -1,14 +1,10 @@
 Then(/^I should see all advocates?$/) do
-
-  @advocates << @advocate
-
-  @advocates.sort!{ |a, b| a.user.last_name != b.user.last_name ? a.user.last_name <=> b.user.last_name : a.user.first_name <=> b.user.first_name}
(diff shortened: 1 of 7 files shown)


features/step_definitions/advocate_admin_steps.rb
```diff
@@ -0,0 +1,42 @@
+Then(/^I should see all advocates?$/) do
+
+  @advocates << @advocate
+
+  @advocates.sort!{ |a, b| a.user.last_name != b.user.last_name ? a.user.last_name <=> b.user.last_name : a.user.first_name <=> b.user.first_name}
```
aliuk2012: Hi @himal-mandalia 

Is there a way of sorting advocates by last/first name? I need to sort it so that the cucumber tests pass but its currently failing this test and throwing this error:

```
  undefined method `user' for nil:NilClass (NoMethodError)
```

I know you had to join to :user to enable the search functionality. (https://github.com/ministryofjustice/advocate-defence-payments/pull/874/files#diff-6ffebb356b75665e83802b5e4540f6b7R8). I'm assuming you would need to do something similar to this.

PR #884, opened 2016-01-05
Optional supplier number and vat registration fields for advocates admin

himal-mandalia:
Optional supplier number and vat registration fields for advocates admin depending on provider type, chamber or firm.


aliuk2012:
I've already done some of this.....https://github.com/ministryofjustice/advocate-defence-payments/pull/876

only the cukes were failing. if you merge this then I'll pull this into my existing pr and merge the changes that should still pass the cuke tests mentioned in here

PR #888, opened 2016-01-06
use claim_presenter for advocate name and claim state

EdwardAndress:
- there was already a method for humaized_state
- advocate_name reads better in the view

### DIFF
--- a/app/views/case_workers/claims/_claims.html.haml
+++ b/app/views/case_workers/claims/_claims.html.haml
@@ -28,14 +28,14 @@
           %td
             = link_to claim.case_number , case_workers_claim_path(claim),{:class => 'js-test-case-number-link'}
           %td
-            = claim.advocate.user.name
+            = claim.advocate_name
           %td
             = claim.defendant_names
           %td.js-test-total
             = claim.total_inc_vat
           - if params[:action] == 'archived'
             %td
-              = claim.state.humanize
+              = claim.humanized_state
           %td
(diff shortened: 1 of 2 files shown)


app/views/case_workers/claims/_claims.html.haml
```diff
@@ -28,14 +28,14 @@
...
-            = claim.advocate.user.name
+            = claim.advocate_name
           %td
             = claim.defendant_names
           %td.js-test-total
             = claim.total_inc_vat
           - if params[:action] == 'archived'
             %td
-              = claim.state.humanize
+              = claim.humanized_state
```
jsugarman, line 38: not sure this presenter refactor is worth while since humanize, is pretty ubiquitous as a kind of presenter method in itself - so its kind of a presenter using a presenter. It also adds a layer of abstraction without improving clarity, unless humanzed_state does more than `claim.state.humanize`(currently not)

app/views/case_workers/claims/_claims.html.haml
```diff
@@ -28,14 +28,14 @@
           %td
             = link_to claim.case_number , case_workers_claim_path(claim),{:class => 'js-test-case-number-link'}
           %td
-            = claim.advocate.user.name
+            = claim.advocate_name
```
jsugarman, line 31: The advocate_name can actually be:
`= claim.advocate.name`
as advocate delegates `name` to user anyway. In which case is there any benefit to advocate_name?

PR #898, opened 2016-01-07
Improve cracked/trial details styling

aliuk2012:
screen shot 2016-01-07 at 22 51 52
screen shot 2016-01-07 at 22 52 06

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -15,6 +15,7 @@ ActiveRecord::Schema.define(version: 20160107163538) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pgcrypto"
   enable_extension "uuid-ossp"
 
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -15,6 +15,7 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pgcrypto"
```
jsugarman, line 18: al, can you run a local db:reseed/reload as this extension should have been removed

aliuk2012, line 18: I cant seem to get rid of it. I manually removed the line and then I tried the following

rake db:drop
rake db:create
rake db:migrate (this one actual seems to insert that line into schema.rb for me)
rake db:reload
rake claims:demo_data

I did try run bundle install before all of it as well.

PR #901, opened 2016-01-10
Sortable claim table columns joels

jsugarman:


jsugarman:
see sortable-claim-table-columns-v2

PR #903, opened 2016-01-11
Case worker sortable claim lists

jsugarman:
Himal, this branch represents the case worker aspect of sortable claims only. 

It squashes your previous commits and mine. It is fit for merging but please note i have removed the ability to sort advocate claim lists while not removing some of the logic from their controller as i was going to work on the advocate side next but wanted to avoid stale branching issues.

Could review together as i have some questions re json importer stuff in here.


app/controllers/external_users/claims_controller.rb
```diff
@@ -11,6 +13,7 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
   before_action :load_advocates_in_provider, only: [:new, :edit, :create, :update]
   before_action :generate_form_id, only: [:new, :edit]
   before_action :initialize_submodel_counts
+  before_action :initialize_json_document_importer, only: [:index]
```
jsugarman, line 16: @himal-mandalia as per our conversation this line seems unrelated to the sorting feature for advocates. was it just a refactor?

app/controllers/external_users/claims_controller.rb
```diff
@@ -389,4 +396,16 @@ def present_errors
     @error_presenter = ErrorPresenter.new(@claim)
   end
 
+  def initialize_json_document_importer
```
jsugarman, line 398: @himal-mandalia This method, used in the callback above, is the first line of the index action as well. should i remove the line in the index action and rely on the callback?

PR #914, opened 2016-01-14
Certification page - radio buttons

aliuk2012:
Changes certification check boxes to radio buttons.
CertificationType model added and boolean attributes on Certification model removed.
Migration to move existing certifications to new certification type paradigm.


aliuk2012:
Still getting rspec/cuke failures. 

It seems to be the fact that certication types arent being populated and the error messages being returned.

screen shot 2016-01-18 at 07 39 28

db/migrate/20160119133748_seed_and_move_to_certification_types.rb
```diff
@@ -0,0 +1,27 @@
+class SeedAndMoveToCertificationTypes < ActiveRecord::Migration
+  def change
```
jsugarman, line 2: should this not be in an up method? due to irreversible data migration

jsugarman:
:+1: just a suggestion that perhaps  a seed file should be created for the certification types and then loaded in the migration and the cuke step that relies on them, unless i am missing something

PR #920, opened 2016-01-16
Accessible accordion

aliuk2012:
The Claim details Accordion will now support keyboard navigation.

Another major change in the way it works is that default all accordions will be closed except if the user is using a "Messages" link which will instead open just the messages panel for them when the page is loaded

### DIFF
--- a/app/assets/javascripts/accordion.js
+++ b/app/assets/javascripts/accordion.js
@@ -1,0 +1,720 @@
+//Based off the example given here http://oaa-accessibility.org/example/37/
+moj.Modules.Accordion = {
+  //Store our keycodes
+  keyCodes : {},
+
+  //Set the keycodes that this Accordion will use
+  setKeyCodes : function (){
+    // Define values for keycodes
+    this.keyCodes.tab        = 9;
+    this.keyCodes.enter      = 13;
+    this.keyCodes.esc        = 27;
+
+    this.keyCodes.space      = 32;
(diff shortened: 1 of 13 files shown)


aliuk2012:
I seem to have a flicker on this PR with the rspec. I'm not sure how changing the view/js would affect the rspecs that are failing.

himal-mandalia:
The accordion.js needs to be broken up and refactored with better management of scopes. Code Climate isn't happy.

app/assets/javascripts/accordion.js
```diff
@@ -0,0 +1,395 @@
...
+  //
+  // @return N/A
+  //
+  togglePanel : function ($tab) {
+
+    var $panel = this.$panel.find('#' + $tab.attr('aria-controls'));
+
+    if ($panel.attr('aria-hidden') == 'true') {
+      $panel.attr('aria-hidden', 'false');
+      $tab.find('img').attr('src', 'http://www.oaa-accessibility.org/media/examples/images/expanded.gif').attr('alt', 'expanded');
```
Demwunz: @aliuk2012  I don't think we should be hotlinking images

aliuk2012:
Well spotted @Demwunz, we dont actually need that and I forgot to take that out during the refactor.

aliuk2012:
@Demwunz if you happy with it could you merge it?

PR #949, opened 2016-01-20
Modify ppe npw fee entry

jsugarman:
adds feature whereby PPE and NPW fees accept a quantity and amount, NOT a quantity and rate. 

This is encapsulated by the notion of an uncalculated fee and a calculated fee - for which there is a calculated boolean attribute on fee type.

Also includes:
- refactoring and removal of unused code to improve code climate
- codeclimate config file to exclude api endpoints from test coverage, for now.


db/migrate/20160118180130_add_calculation_to_fee_types.rb
```diff
@@ -0,0 +1,13 @@
+class AddCalculationToFeeTypes < ActiveRecord::Migration
+  def up
+    add_column :fee_types, :calculated, :boolean, default: true
+
+    FeeType.where(code: ['PPE','NPW']) do |record|
+        record.update_column(calculated: false)
```
himal-mandalia, line 6: Two levels of indentation?

PR #953, opened 2016-01-22
Conditional offence validation

jsugarman:
presence of offence only needed for non-fixed fee cases

### DIFF
--- a/app/validators/claim_textfield_validator.rb
+++ b/app/validators/claim_textfield_validator.rb
@@ -67,7 +67,7 @@ end
 
 # must be present unless case type is breach of crown court order
 def validate_offence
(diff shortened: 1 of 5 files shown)


app/validators/claim_textfield_validator.rb
```diff
@@ -67,7 +67,7 @@ def validate_advocate_category
 
 # must be present unless case type is breach of crown court order
```
EdwardAndress, line 68: Is this comment (line 68, not the one I am making right now) still relevant?

jsugarman, line 68: :+1: nope its not. can i remove this in a subsequent branch - just to avoid the flickering tests which took about 2 hours to overcome this time round

PR #956, opened 2016-01-22
Display certification details

### LINKED ISSUE #937: CCCD: Caseworker needs to know how advocate came to be Instructed Advocate
Certification selection to be added to detail view of claim

jsugarman:
displays the advocate certification details on the claim detail page for both case workers and advocates
fixes #937

### DIFF
--- a/app/views/shared/_claim.html.haml
+++ b/app/views/shared/_claim.html.haml
@@ -1,9 +1,24 @@
 %section
+
+  .certification-info-row
+    .bold-xsmall{ style: 'margin-bottom: 5px;' }
+      = t('.certification_box_title')
+    %div.bold-xsmall
+      = t('.certified_by') + ':'
+      %span.xsmall
+        = claim.certification.certified_by
+    %div.bold-xsmall
+      = t('.certification_reason') + ':'
+      %span.xsmall
+        = claim.certification.certification_type.name
+
   - if claim.case_workers.any? && CaseWorker === current_user.persona && current_user.persona.admin?
-    .bold-xsmall
-      = t ('.assignee')
-    .xsmall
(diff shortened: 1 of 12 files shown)


app/views/shared/_claim.html.haml
```diff
@@ -1,9 +1,24 @@
 %section
+
+  .certification-info-row
+    .bold-xsmall{ style: 'margin-bottom: 5px;' }
+      = t('.certification_box_title')
+    %div.bold-xsmall
```
aliuk2012: This line could just be ".bold-xsmall". HAML by default would create a div for any element that isn't explicitly mentioned. eg line 3. it would create a div with a class of "certification-info-row"

app/views/shared/_claim.html.haml
```diff
@@ -1,9 +1,24 @@
...
+      = t('.certified_by') + ':'
+      %span.xsmall
+        = claim.certification.certified_by
+    %div.bold-xsmall
```
aliuk2012: Same as line 6 above

app/views/shared/_claim.html.haml
jsugarman: cheers dude, this had just been left in from me mucking about using `%div{style: 'whatever:init;'}` previously. incidentally i need to talk to you about some styling on this and the main form PPE and NPW sections. til monday

aliuk2012:
I think we should reduce the left padding so the text is all left aligned and inline. I think it looks a bit odd with it having extra left padding. I'll check with Maria.

At the moment it looks like this:
screen shot 2016-01-25 at 07 46 56

but I think it looks better like this:
screen shot 2016-01-25 at 07 46 35

PR #958, opened 2016-01-25
fix nilclass bug; breaking MI on gamma only

EdwardAndress:
- Mel wants all unallocated claims to have a state of 'submitted'
- also some small refactors for consistency

### DIFF
--- a/app/presenters/claim_csv_presenter.rb
+++ b/app/presenters/claim_csv_presenter.rb
@@ -11,7 +11,7 @@ class ClaimCsvPresenter < BasePresenter
   end
 
   def sorted_and_filtered_state_transitions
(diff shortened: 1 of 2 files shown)


app/presenters/claim_csv_presenter.rb
```diff
@@ -54,17 +54,18 @@ def submission_type
   end
 
   def submitted_at
-    @journey.select { |step| submitted_states.include?(step.to) }.first.created_at.to_s
+    submission_steps = @journey.select { |step| submitted_states.include?(step.to) }
+    submission_steps.present? ? submission_steps.first.created_at.to_s : 'n/a'
   end
 
   def allocated_at
```
EdwardAndress, line 61: this, and the completed_at method, was fine but I added a little refactor

PR #983, opened 2016-01-30
Fix flickering MI report presenter spec

jsugarman:
there  is a time related spec which may
occasionally flicker (see pic).
screen shot 2016-01-30 at 10 03 42


stephenrichards:
Looks good to me  - go ahead and merge (I don't have permissions to merge yet)

PR #984, opened 2016-02-01
fixes a validation that relies on trial lengths

jsugarman:
Daily attendance fee quantity validations set an upper bound
of the actual trial length. The trial length should now 
be the retrial length for retrials.


stephenrichards:
Looks fine to merge - I don't yet have permissions to merge though, so go ahead and merge it yourself.

PR #1013, opened 2016-02-08
Add litigator role

EdwardAndress:
- add litigator role to external users
- add agfs and lgfs roles to providers, in part, so that existing users (at chambers) do not suddenly get the option to add litigators


spec/models/external_user_spec.rb
```diff
@@ -188,7 +209,142 @@
 
   describe 'ROLES' do
     it 'should have "admin" and "advocate"' do
```
jsugarman, line 211: should have "admin", "advocate" and "litigator"

PR #1037, opened 2016-02-15
sort providers by name for superadmins

EdwardAndress:


db/schema.rb
```diff
@@ -15,6 +15,7 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pgcrypto"
```
himal-mandalia: Get rid of this

PR #1050, opened 2016-02-16
Perform page reload when redetermination or awaiting written reasons

himal-mandalia:
…triggeredPerform page reload when redetermination or awaiting written reasons triggered via messages on external users claim detail view.

### DIFF
--- a/features/step_definitions/claim_messages_steps.rb
+++ b/features/step_definitions/claim_messages_steps.rb
@@ -19,6 +19,8 @@ When(/^I leave a message$/) do
 end
 
 Then(/^I should see my message at the bottom of the message list$/) do
+  sleep 1
+
   within '#panel1' do
     message_body = all('.message-body').last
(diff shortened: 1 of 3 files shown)


features/step_definitions/claim_messages_steps.rb
```diff
@@ -19,6 +19,8 @@
 end
 
 Then(/^I should see my message at the bottom of the message list$/) do
+  sleep 1
```
aliuk2012: why is there a sleep in this step? should it not be wait_for_ajax or is it even needed if the page is actually refreshed?

I just think that this step could potentially cause flickers if the content is not available once it continues.

himal-mandalia: It has to wait for the AJAX to return the JS that reloads the page - but the actual page reload cannot be waited for, hence the sleep statement. Exploring ways of getting around it.

PR #1055, opened 2016-02-17
Filter sensitive params from logs

himal-mandalia:
Params to filter out of logs:

:email,
:first_name,
:last_name,
:date_of_birth,
:supplier_number,
:body

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,4 +1,16 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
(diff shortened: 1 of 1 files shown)


config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,16 @@
...
+  :email,
+  :first_name,
+  :last_name,
+  :date_of_birth,
+  :supplier_number,
+  :body,
+  :supplier_number,
+  :email,
+  :first_name,
+  :last_name
```
aliuk2012: Why is email, first_name,last_name and supplier_number repeated?

himal-mandalia: That'll be a careless copy and paste. Correcting.

PR #1061, opened 2016-02-19
Fix bug with Draft claims expenses

aliuk2012:
Fixes Zendesk ticket 13398.

To replicate bug:
Create a draft claim that has some data entered but leave the expense type blank.
Edit the claim and it should throw an error.

### DIFF
--- a/app/presenters/expense_presenter.rb
+++ b/app/presenters/expense_presenter.rb
@@ -5,7 +5,16 @@ class ExpensePresenter < BasePresenter
     expense.dates_attended.order(date: :asc).map(&:to_s).join(', ')
   end
 
- def amount
+  def amount
     h.number_to_currency(expense.amount)
- end
+  end
+
+  def name
+    if expense.expense_type.blank?
+      "Not selected"
+    else
+      expense.expense_type.name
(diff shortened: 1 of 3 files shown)


app/presenters/expense_presenter.rb
```diff
@@ -5,7 +5,16 @@ def dates_attended_delimited_string
     expense.dates_attended.order(date: :asc).map(&:to_s).join(', ')
   end
 
- def amount
+  def amount
```
himal-mandalia, line 8: Indentation weird

spec/presenters/expense_presenter_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      expense.expense_type_id = nil
+      expect(presenter.name).to eql('Not selected')
+    end
+
+    it 'outputs the type name is there is an expense type' do
+      expense.expense_type_id = expense_type.id
+      expect(presenter.name).to eql(expense.expense_type.name)
+    end
+  end
+
```
himal-mandalia, line 45: Get rid of trailing newlines

PR #1066, opened 2016-02-22
Implement STI on Fee and FeeType classes

stephenrichards:
Remove logic on Fee class to determine whether it is not a basic, fixed or miscellaneous fee
by examining the assiciationed Feetype and its associated Fee Category with three separate classes for
both Fee and FeeType.  This enables us to get rid of the FeeCategory altogether.


app/models/fee/fixed_fee.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+
+class Fee::FixedFee < Fee::BaseFee
+
+  belongs_to :fee_type, class_name: Fee::FixedFeeType
+
+  validates_with Fee::FixedFeeValidator
+
+  def is_fixed?
+    true
```
himal-mandalia, line 24: Could this not be put on the BaseFee as a define_method?

```
define_method "is_#{self.class.to_s.underscore.split('_').first}?" do
  true
end
```

This would add the #is_fixed? #is_misc? methods to inheriting fee subclasses. Just a thought.

PR #1082, opened 2016-02-24
Fix basic fee bugs

jsugarman:
Fix bugs that came in with Fee abstraction work 
- [x] basic fees all getting buttons to add dates attended erroneously
- [x] basic fee ordering in form is by description but should be by id

### DIFF
--- a/spec/models/fee/basic_fee_type_spec.rb
+++ b/spec/models/fee/basic_fee_type_spec.rb
@@ -2,12 +2,36 @@ require 'rails_helper'
 
 module Fee
   describe BasicFeeType do
+
     let(:fee_type)  { build :basic_fee_type }
+    DATES_ATTENDED_APPLICABLE_FEES = %w( BAF DAF DAH DAJ PCM SAF )
+    DATES_ATTENDED_NOT_APPLICABLE_FEES = %w( CAV NDR NOC PPE NPW )
 
     describe '#has_dates_attended?' do
-      it 'returns true' do
-        expect(fee_type.has_dates_attended?).to be true
+      context 'for fees that can have dates associated with them' do
+        DATES_ATTENDED_APPLICABLE_FEES.each do |code|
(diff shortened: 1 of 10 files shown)


spec/models/fee/fixed_fee_type_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    describe 'default scope' do
+      before do
+        create(:fixed_fee_type, description: 'Ppppp')
+        create(:fixed_fee_type, description: 'Xxxxx')
+        create(:fixed_fee_type, description: 'Sssss')
+      end
+
+      it 'should order by description ascending' do
+        # expect(Fee::FixedFeeType.all.map(&:id)).to eq [1,3,2]
```
stephenrichards: Better to remove the commented-out line

spec/models/fee/basic_fee_type_spec.rb
```diff
@@ -2,12 +2,37 @@
 
 module Fee
   describe BasicFeeType do
+
     let(:fee_type)  { build :basic_fee_type }
+    DATES_ATTENDED_APPLICABLE_FEES = %w( BAF DAF DAH DAJ PCM SAF )
+    DATES_ATTENDED_NOT_APPLICABLE_FEES = %w( CAV NDR NOC PPE NPW )
```
aliuk2012, line 8: Why are these hard coded? Isn't there a risk that these dont match what is being used in actual code?

jsugarman, line 8: It is deliberately separating the app logic from the test so as to raise an error in the event of a change to the app that is not accounted for in the test

spec/models/fee/fixed_fee_type_spec.rb
jsugarman: done

PR #1089, opened 2016-02-26
Rewrite Date format feature tests as model specs

stephenrichards:


db/migrate/20160126121119_add_requires_retrial_details_flag_to_case_types.rb
```diff
@@ -1,9 +1,8 @@
 class AddRequiresRetrialDetailsFlagToCaseTypes < ActiveRecord::Migration
   def up
     add_column :case_types, :requires_retrial_dates, :boolean, default: false
-
-    # migrate data
-    CaseType.find_by(name: 'Retrial').update_column(:requires_retrial_dates,true) rescue nil
+    retrial = CaseType.find_by_sql("SELECT * FROM case_types WHERE name = 'Retrial'").first
```
jsugarman, line 4: what is the purpose of this - its an old migration and does not appear to change the result in any event - unless find_by(name: 'etc') will cease to function with other changes in this PR.

spec/factories/fee_types.rb
```diff
@@ -16,7 +16,7 @@
 FactoryGirl.define do
```
jsugarman, line 16: this has been changed already (in develop,etc) to a simpler factory method that avoids any validation issues (based on code)

db/migrate/20160222163526_add_roles_to_case_type.rb
```diff
@@ -4,7 +4,8 @@ def change
 
     CaseType.reset_column_information
 
-    CaseType.all.each do |ct|
+    case_types = CaseType.find_by_sql("select * from case_types")
```
jsugarman, line 7: is there a difference between these two iteration methods?

spec/models/case_type_spec.rb
```diff
@@ -3,4 +3,39 @@
...
+      @child_2 = create :child_case_type, :s74
+    end
+
+    after(:all) do
+      CaseType.delete_all
+    end
+
+    describe '.parents' do
+      it 'does not return child records' do
+        expect(CaseType.top_levels).to eq([@parent_1, @parent_2])
```
jsugarman, line 20: should this not be describe '.top_levels' or '#top_levels' or perhaps the scope itself and this spec should be '#parents'

i.e. CaseType.parents
CaseType.parents.each { |parent| puts parent.children }

PR #1093, opened 2016-02-29
Add roles to fee types

EdwardAndress:
- adding roles to fee_types ahead of LGFS implementation
- MIGRATION required: will alter existing fee_types on gamma to include "agfs" role


himal-mandalia:
Should also include the shared examples Roles in the BaseFeeSpec:

it_behaves_like 'roles', Fee::BaseFee, Fee::BaseFee::ROLES

EdwardAndress:
- done, thanks @himal-mandalia

Issue #1099, opened 2016-02-29
[FE] Claim detail page: Basic claim information not displaying correctly


aliuk2012:
View any type of claim
<img width="988" alt="screen shot 2016-02-29 at 16 48 24" src="https://cloud.githubusercontent.com/assets/3441519/13401518/538e036e-df04-11e5-9016-854bb238ff64.png">

Issue #1103, opened 2016-03-01
[FE] wonky date attended fields and inline date attendedfields


jsugarman:
wonky on expenses
<img width="1064" alt="screen shot 2016-03-01 at 12 54 37" src="https://cloud.githubusercontent.com/assets/7016425/13427576/d8d674cc-dfac-11e5-9604-7de3a3c6fd14.png">

inline for misc/basic(and fixed?)
<img width="904" alt="screen shot 2016-03-01 at 12 55 31" src="https://cloud.githubusercontent.com/assets/7016425/13427593/fa86c590-dfac-11e5-98c8-f4e7f93510e0.png">

PR #1109, opened 2016-03-01
quick fix for #1107, #1103

### LINKED ISSUE #1107: [FE] wonky trial and retrial fields
screen shot 2016-03-01 at 14 40 19

Demwunz:
fixes #1107 #1103

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -27,3 +27,5 @@
 @import "moj/forms";
 @import "moj/typography";
 @import "moj/footer";
+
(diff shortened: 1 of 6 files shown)


Demwunz:
@jsugarman

aliuk2012:
I took another look at the issue about the number of claims fled stretching and wondered if this PR would inadvertently fix it turns out it breaks the field so users cant click ok the input field

select2 is placing a layer over the input field so users can't type

aliuk2012:
good news is it fixes #1099  being all one long column but this is on the claim detail page not the start a new claim page

PR #1110, opened 2016-03-01
Start tagging cucumber features and scenarios

aliuk2012:
First wave to improve cucumber tests and manage them bette
1. Added tags to feature tests 
2. Moved features into user specific folders
3. Add descriptive text to example tables.

These targets feature tests can be run using:
Cucumber --tag @admin
Cucumber --tags @caseworker, @advocate
Cucumber --tags @advocate, @admin

Please review to the README.md in the features folder.


features/api_documentation.feature
```diff
@@ -14,18 +17,21 @@ Scenario: Sign in page for NOT API sandbox
      When I click on the API Sign up and Documentation link
      Then I should be directed to the API landing page
 
+  @advocate
```
jsugarman: maybe we should have a @vendor or @api tag since advocates, per se, are not the target of this specific feature

features/api_documentation.feature
```diff
@@ -14,18 +17,21 @@ Scenario: Sign in page for NOT API sandbox
...
      Then I should see a link to the API documentation
 
+  @advocate
   Scenario: API documentation not available from dashboard
     Given I am not on the API sandbox
       And I am a signed in advocate
      When I visit the advocates dashboard
      Then I should not see a link to the API documentation
 
+  @advocate
```
jsugarman: as last comment on this one @vendor is more targetted

features/case_workers/view_claim_evidence.feature
```diff
@@ -0,0 +1,22 @@
+@advocate
```
jsugarman: following your suggested dir structure should these advocate features not be in the features/external_users/advocates directory

features/caseworker_claims_list.feature
```diff
@@ -40,7 +41,7 @@ Feature: Caseworker claims list
      Then I should only see <archive_number> claims
       And the claim count should show <archive_number>
 
-     Examples:
+     Examples: List of defendants and the number of live/archive claims they're involved in
```
jsugarman, line 44: :+1:  ..i never knew you could do this

jsugarman, line 44: awesome stuff @aliuk2012: am happy to merge as is but a few points I wanted to note/query for future iterations:
1. I like your suggested dir structure but note that not all advocate features are under the external_uses/advocates folder, as yet
2. similar for case workers
3. there is one feature for advocates situatuated under case workers dir - separate comment above

aliuk2012, line 44: I'm aware of points 1 and 2 but given the number of tests and time we have available I think updating the locations/tests in a piecemeal fashion is the only real option of getting it into a usable state.

Point 3 it should have actually been a case worker test not advocate. I've fixed it up as part of the latests commits 92a8376

PR #1133, opened 2016-03-07
Govuk elements

Demwunz:
:see_no_evil: integrated @robmckinnon gem for GDS assets.
:speak_no_evil: cleaned up some UI
:hear_no_evil: better place to start to update remainder of app.


Demwunz:
@himal-mandalia @jsugarman

PR #1134, opened 2016-03-07
Refactor Feedback and BugReport into single model and controller

himal-mandalia:
Refactor Feedback and BugReport into single Feedback model and controller. Removes duplication of code and tests.


app/models/feedback.rb
```diff
@@ -10,9 +10,22 @@ class Feedback
...
+    feedback: [:rating, :comment, :email],
+    bug_report: [:event, :outcome, :email]
+  }
+
+  FEEDBACK_TYPES.keys.each do |type|
+    define_method "#{type}?" do
+      is?(type)
+    end
+  end
+
```
stephenrichards: wouldn't this be better as two separate hard coded methods rather than one dynamic one?

def feedback?
 FEEDBACK_TYPES[:feedback].include?(@type)
end

def bug_report?
 FEEDBACK-TYPES[:bug_report].include?(@type)
end

PR #1136, opened 2016-03-07
Litigator dashboard first pass

jsugarman:
This PR gets demo data working for both advocate agfs claims and litigator lgfs claims and enables their viewing in the relevant index-like views. includes: 
- [X] split and modify demo data
- [X] allocation
- [X] reallocation
- [X] your claims
- [x] archive views
- [x] claim details views
- [X] add provider has_many relation to claims via creator (not external_user)
- [X] amend context for claims controller and financial summary
- [X] amend claims controller new and create to handle different types of claims
- [X] fix broken specs and cukes (rewrote financial summary by provider)
- [X] add base claim model specs
- [X] add simple claims list litigator journey cuke


app/views/external_users/claims/new.html.haml
```diff
@@ -1,3 +1,10 @@
-= render partial: 'layouts/header', locals: {page_heading: t('.heading')}
+%header.main-header{role: 'banner'}
```
aliuk2012: I dont think this should be in here because the layout/header partial will output similar markup so the page would have two main headings.

app/views/external_users/claims/new.html.haml
```diff
@@ -1,3 +1,10 @@
...
+
+  - if @claim.agfs?
+    - page_heading = t('.agfs_heading')
+  - elsif @claim.lgfs?
+    - page_heading = t('.lgfs_heading')
+
```
aliuk2012: Couldn't the above be put into a helper method like https://github.com/ministryofjustice/advocate-defence-payments/blob/develop/app/helpers/application_helper.rb#L101

Just so we can remove this type of logic in the view

app/views/external_users/claims/new.html.haml
jsugarman: think that came in from a merge conflict

app/views/external_users/claims/new.html.haml
jsugarman: yeah, that was a quick fix . done

PR #1153, opened 2016-03-10
Add Disbursements for LGFS claims.

zheileman:
New database tables created to store the different types of disbursements and to link them to the claims. The Calculations module has been updated to process disbursements net and vat amounts, and a new field to store the disbursements total has been added to the claims table.

New DisbursementGenerator to generate demo data for litigator claims.


db/seeds/disbursement_types.rb
```diff
@@ -0,0 +1,36 @@
+[
+  'Accident reconstruction report',
+  'Accounts',
+  'Computer experts',
+  'Consultant medical reports',
+  'Costs judge application fee',
+  'Costs judge preparation award',
+  'DNA testing',
+  'Engineer',
+  'Enquiry Agents',
```
stephenrichards: lower case first letter on all subsequent words

lib/demo_data/disbursement_generator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  class DisbursementGenerator
+
+    def initialize(claim)
+      @claim = claim
+    end
+
+    def generate!
+      DisbursementType.order('RANDOM()').first(rand(0..5)).each { |type| add_disbursement(type) }
+    end
```
stephenrichards, line 11: let's keep it 0..2

lib/demo_data/base_claim_generator.rb
```diff
@@ -104,6 +105,10 @@ def add_expenses(claim)
       ExpenseGenerator.new(claim).generate!
     end
 
+    def add_disbursements(claim)
```
jsugarman: would this not be better placed in the litigator_claim_generator class

PR #1217, opened 2016-04-01
Reinstate original behaviour of has_roles? method and ExernalUser on Litigator Claims

stephenrichards:
This PR addresses two related issues:
#1) The has_roles? method in the Roles concern

The has_roles? method was returning true if and only if the list of 
roles specified as a parameter exactly matched the roles on the receiver.

The correct behaviour is to return true if the receiver's roles include 
all the roles specified as parameters (i.e. the receiver may have more
roles than were specified as params). This PR re-instates that behaviour, 
and also allows roles to be specified either as symbols or strings.
#2) Litigator Claims have an external user

The concept of an external user was previously removed.  This PR replaces it, and 
ith it the logic that is used to determine what claims are visible to whom.  Thei is 
done by means of the ContextMapper which has been greatly simplified as a result.


features/step_definitions/shared/claims_list_steps.rb
```diff
@@ -2,7 +2,8 @@
 #       Steps that are applicable for Your Claims, Archived, Outstanding and Authorised claims list, at least
 
 Then(/^I should see all claims$/) do
+  save_and_open_page
```
jsugarman, line 5: this needs removing

PR #1238, opened 2016-04-11
Move parent-child relationship from case types to fixed fees

himal-mandalia:
Parent-child relationship on case types removed and moved to fixed fees.

### DIFF
--- a/app/models/case_type.rb
+++ b/app/models/case_type.rb
@@ -13,22 +13,15 @@
 #  requires_maat_reference :boolean          default(FALSE)
 #  requires_retrial_dates  :boolean          default(FALSE)
 #  roles                   :string
-#  parent_id               :integer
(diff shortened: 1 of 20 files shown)


app/models/case_type.rb
```diff
@@ -13,22 +13,15 @@
...
 class CaseType < ActiveRecord::Base
   ROLES = %w{ lgfs agfs }
   include Roles
-  
-  auto_strip_attributes :name, squish: true, nullify: true
-
-  has_many :children, class_name: CaseType, foreign_key: :parent_id
-  belongs_to :parent, class_name: CaseType, foreign_key: :parent_id
 
-  default_scope -> { order(parent_id: :desc, name: :asc) }
```
stephenrichards, line 29: We still need default scope of name ascending

PR #1264, opened 2016-04-18
Supplier number selector for litigator claims.

zheileman:


app/models/provider.rb
```diff
@@ -36,14 +36,14 @@ def ordered_by_last_name
     end
   end
 
+  has_many :supplier_numbers
```
himal-mandalia: Dependant  destroy?

app/validators/claim/litigator_claim_validator.rb
```diff
@@ -47,4 +47,9 @@ def validate_offence
   def validate_case_concluded_at
     validate_presence(:case_concluded_at, 'blank')
   end
+
+  def validate_supplier_number
+    validate_presence(:supplier_number, 'blank')
+    validate_pattern(:supplier_number, @record.supplier_number_regex, 'invalid') if @record.supplier_number
```
himal-mandalia: @record.supplier_number.present?

^ might be clearer

PR #1268, opened 2016-04-19
Implement InterimFeeTypePresenter

stephenrichards:
The InterimFeeType presenter is necessary to provide data attributes to the
front end so that when a fee type is selected from the drop down box, the
front end knows which elements to hide or show.

remove unneeded file


zheileman:
@stephenrichards: We don't need this anymore. We can remove this PR/branch. All what was done in this PR has been already done as part of PR merged in the last week for interim claims.

PR #1293, opened 2016-04-22
Add VAT to expenses for LGFS claims

himal-mandalia:
Adds VAT input to expenses for LGFS claims.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160419100425) do
+ActiveRecord::Schema.define(version: 20160421104849) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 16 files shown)


db/schema.rb
```diff
@@ -265,6 +265,7 @@
     t.integer  "mileage_rate_id"
     t.date     "date"
     t.integer  "hours"
+    t.float    "vat_amount",      default: 0.0
```
stephenrichards: This should be a decimal surely?

spec/factories/expenses.rb
```diff
@@ -19,6 +19,7 @@
 #  mileage_rate_id :integer
 #  date            :date
 #  hours           :integer
+#  vat_amount      :float            default(0.0)
```
stephenrichards: decimal surely?

spec/models/expense_spec.rb
```diff
@@ -19,6 +19,7 @@
 #  mileage_rate_id :integer
 #  date            :date
 #  hours           :integer
+#  vat_amount      :float            default(0.0)
```
stephenrichards: decimal

PR #1305, opened 2016-04-26
Setting allocation type set on submit

stephenrichards:
The allocation type is set whenever a TransferClaim transitions to submitted state.
This is acheived by a after_transmission hook which calls set_allocation_type on the claim.
This method is implemented as an empty method on BaseClaim, but on TransferClaim it
calls TransferBrain to work out the allocation type according tothe fields on the TransferDetail and
save it to the claim record.

### DIFF
--- a/app/models/claim/transfer_brain_data_item_collection.rb
+++ b/app/models/claim/transfer_brain_data_item_collection.rb
@@ -19,7 +19,6 @@ module Claim
     def initialize
       lines = load_file
       @collection = []
-      lines.shift
       lines.each { |line| @collection << TransferBrainDataItem.new(line) }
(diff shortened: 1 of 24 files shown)


app/models/claim/transfer_brain_data_item_collection.rb
```diff
@@ -19,7 +19,6 @@ class TransferBrainDataItemCollection
     def initialize
       lines = load_file
       @collection = []
```
doctorpod, line 21: `@collection` probably doesn't need to be an instance var - it's only referred to in `construct_collection_hash` - it could be passed in as a param here: `construct_collection_hash(collection)`

PR #1331, opened 2016-04-29
Handle pre-populated vat_amount

doctorpod:
PT: https://www.pivotaltracker.com/story/show/118554905

### DIFF
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -124,22 +124,7 @@ module Claim
 
     accepts_nested_attributes_for :basic_fees,        reject_if: :all_blank, allow_destroy: true
     accepts_nested_attributes_for :misc_fees,         reject_if: :all_blank, allow_destroy: true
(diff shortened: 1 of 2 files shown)


app/models/claim/base_claim.rb
```diff
@@ -124,7 +124,22 @@ class BaseClaim < ActiveRecord::Base
...
+          atts["expense_type_id"].blank? &&
+          atts["location"].blank? &&
+          atts["distance"].blank? &&
+          atts["reason_id"].blank? &&
+          atts["reason_text"].blank? &&
+          atts["hours"].blank? &&
+          atts["amount"].blank?
+      },
+      allow_destroy: true
     accepts_nested_attributes_for :disbursements,     reject_if: :all_blank, allow_destroy: true
```
stephenrichards, line 128: I think we need tests for this - I suspect that if we fill in a value in the vat amount and none of the other fields, the record will not be created.

PR #1367, opened 2016-05-06
Refactor ExternalUser#available_roles into service object

himal-mandalia:


slorek:
As discussed this would probably better as methods on the respective model classes.

PR #1370, opened 2016-05-06
Move #available_claim_types to service object

himal-mandalia:


slorek:
As discussed this would probably better as methods on the respective model classes.

PR #1372, opened 2016-05-09
Move Provider#regenerate_api_key into service object

himal-mandalia:


slorek:
Consider moving the `update_column` method call back to the model class. This is/should be an internal implementation detail.

PR #1400, opened 2016-05-12
add missing transfer claim validations

jsugarman:
case number and 0 transfer fee amount validations
were not included and functioning respectively


jsugarman:
JASMINE phantom.js install failure causing test suite not but report success

PR #1412, opened 2016-05-17
Some small changes to improve readability dealing with claim service results.

zheileman:
Also removing the coupling with Google Analytics, and returning the service result instead of the service once we run call()

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -102,24 +102,28 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
 
   def create
     if submitting_to_laa?
-      action_service = Claims::CreateClaim.call(@claim)
+      result = Claims::CreateClaim.call(@claim)
       tracking_args = %w(event claim submit started)
     else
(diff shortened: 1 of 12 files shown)


slorek:
Definitely better now. Not directly relevant to this PR but consider having the services' `call` method return the `result` object.

Also consider moving the instantiation of the `Claim` object (`@claim`) into the service in all cases.

app/controllers/external_users/claims_controller.rb
```diff
@@ -344,15 +342,15 @@ def render_action_with_resources(action)
...
 
-  def handle_submit_result(service)
+  def render_or_redirect(service)
     if service.result.success?
-      handle_submit_success(service)
+      render_or_redirect_success(service)
     else
-      handle_submit_error(service)
+      render_or_redirect_error(service)
     end
```
slorek: Consider the following:

``` ruby
return render_or_redirect_error(service) unless service.result.success?
render_or_redirect_success(service)
```

slorek:
Made some minor changes. Still quite a bit to do, but will be done in another PR to prevent changes becoming stale.

zheileman:
@slorek happy with the changes. Are we gonna merge this PR? If so, go ahead :thumbsup:

PR #1416, opened 2016-05-17
Update logstash source to reflect the true environment 

stephenrichards:
Currently, logs from all environments (production, staging, dev, demo, sandbox) were being written to logstash with a source of 'Advocate Defence Payments production', making them impossible to distinguish on kibana.

This PR substitutes production for the name of the environemnt (production, staging, etc). derivied from the ENV environment variable

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -101,7 +101,7 @@ Rails.application.configure do
   config.logstasher.enabled = true
 
   # This line is optional, it allows you to set a custom value for the @source field of the log event
-  config.logstasher.source = 'Advocate Defence Payments App production'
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -101,7 +101,7 @@
   config.logstasher.enabled = true
 
   # This line is optional, it allows you to set a custom value for the @source field of the log event
-  config.logstasher.source = 'Advocate Defence Payments App production'
+  config.logstasher.source = "Advocate Defence Payments App #{ENV['ENV']}"
```
slorek: not `RAILS_ENV`?

PR #1418, opened 2016-05-17
validate existence of transfer fee on transfer claim

jsugarman:
A user should not be able to submit a transfer claim without a transfer fee. This adds a validation to enforce this logic together with error message, translations, and error links.


jsugarman:
code climate complaining unnecessarily

PR #1435, opened 2016-05-20
Adding in side bar + minor layout fixes

cleargif:
- updated new and edit views to render the sidebar
- applied generic css classes
- updated spec

Totals not being added
https://www.pivotaltracker.com/story/show/119812465


cleargif:
@zheileman this is ready

PR #1445, opened 2016-05-21
improve js using literals/hashes

jsugarman:
Hey, refactored this JS we looked at previously, as you suggested, to use literals/hash.

### DIFF
--- a/app/assets/javascripts/modules/external_users/claims/TransferDetailFieldsDisplay.js
+++ b/app/assets/javascripts/modules/external_users/claims/TransferDetailFieldsDisplay.js
@@ -7,13 +7,32 @@ moj.Modules.TransferDetailFieldsDisplay = {
   caseConclusionSelect: '.js-case-conclusions-select',
   transferStageLabel: '.js-transfer-stage-label',
   transferDateLabel: '.js-transfer-date-label',
+  params: {},
 
   init: function() {
     if ($(this.tdWrapper).length > 0) {
+      this.cacheEls();
       this.$tdWrapper = $(this.tdWrapper);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/modules/external_users/claims/TransferDetailFieldsDisplay.js
```diff
@@ -42,17 +62,15 @@ moj.Modules.TransferDetailFieldsDisplay = {
...
   },
 
   constructParams: function() {
-    var params = '';
-    params = params + this.getParamVal('litigator_type',    this.$tdWrapper.find(this.litigatorTypeRadio + ':checked'));
-    params = params + this.getParamVal('elected_case',      this.$tdWrapper.find(this.electedCaseRadio + ':checked'));
-    params = params + this.getParamVal('transfer_stage_id', this.$tdWrapper.find(this.transferStageSelect).find('option:selected'));
+    var params =  this.getParamVal('litigator_type') +
+                  this.getParamVal('elected_case') +
+                  this.getParamVal('transfer_stage_id');
```
cleargif: I would change this to a loop,

``` javascript
var params;
$.each(this.params, function(key, val){
  params += this.getParamVal(key);
});
```

makes it a little more extendable - but you and I both know there are only ever going to these 3 fields. 

nice job on this too - the testing is going to be simpler after these refactors.

...just wrote this into the console - did not actually test it - but it looks right..

jsugarman: cool, will amend and repush

cleargif:
Nice job!

PR #1453, opened 2016-05-24
Repair Missing Docs on cloned claims

stephenrichards:
This PR provides the rake task clone:repair whcih reads a file containing details of claims
with missing documents and identifies which of those is a clone, and update the clone_source_id
field accordingly.

This enables the ClonedDocumentRepairer to be run on those claims to copy the documents over 
from the source claim.


app/rake_helpers/cloned_claim_repairer.rb
```diff
@@ -0,0 +1,51 @@
+# This class attempts to find the source claim of a cloned claim
+# and update the clone_source_id coloumn with the id of the original claim.
+
+class ClonedClaimRepairer
+
+  def initialize(claim_id)
+    @claim_id = claim_id
+    @claim = Claim::BaseClaim.find_by(id: claim_id)
+
+    @updated_claims = {}
```
jsugarman, line 10: unused?

PR #1456, opened 2016-05-24
Remove ap in test suite

stephenrichards:


jsugarman:
ap print removed in previous PR

PR #1465, opened 2016-05-25
change interim claim warrant fee to interim warrant

jsugarman:
Interim claims were using "final warrants" (Fee::WarrantFee of type Fee::WarrantFeeType with code XWAR) objects for their warrant fees. This has been changed to use Interim fees of description warrant (Fee:InterimFee of type Fee::InterimFeeType with code of IWARR).

This necessitated changing validations (and error messages) as well as presenters which were also refactored. summary and claim details views were also changed in line with the changes.

In addition "final warrants" (Fee::WarrantFee) have been amended to require an amount, with commensurate changes to validators and presenters and specs

factories and spec modified in line with changes too.

### DIFF
--- a/app/presenters/base_presenter.rb
+++ b/app/presenters/base_presenter.rb
@@ -14,8 +14,16 @@ class BasePresenter < SimpleDelegator
 
   private_class_method :presents
 
-  private
+  def format_date(date)
+    date.strftime(Settings.date_format) rescue nil
+  end
 
+  def date_format(options={})
+    options.assert_valid_keys(:include_time)
+    options[:include_time] ? Settings.date_time_format : Settings.date_format
+  end
+
+  private
 
   def h
     @view
--- a/app/presenters/fee/interim_fee_type_presenter.rb
+++ b/app/presenters/fee/interim_fee_type_presenter.rb
@@ -8,18 +8,21 @@ class Fee::InterimFeeTypePresenter < BasePresenter
         legal_aid_transfer:   fee_type.is_retrial_new_solicitor?,
         trial_concluded:      fee_type.is_retrial_new_solicitor?,
         retrial_dates:        fee_type.is_retrial_start?,
-        ppe:                  unless_disbursement_or_warrant?,
-        fee_total:            unless_disbursement_or_warrant?,
+        ppe:                  is_not_disbursement_or_warrant?,
+        fee_total:            is_not_disbursement?,
         warrant:              fee_type.is_interim_warrant?,
         disbursements:        is_not_warrant?
     }
   end
 
-
   private
 
-  def unless_disbursement_or_warrant?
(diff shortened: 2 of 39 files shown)


app/presenters/base_presenter.rb
```diff
@@ -14,8 +14,16 @@ def self.presents(name)
...
+  def format_date(date)
+    date.strftime(Settings.date_format) rescue nil
+  end
 
+  def date_format(options={})
+    options.assert_valid_keys(:include_time)
+    options[:include_time] ? Settings.date_time_format : Settings.date_format
+  end
+
+  private
```
stephenrichards, line 27: Wouldn't it be better to scrap format_date, and just use date_format everywhere?

app/presenters/fee/interim_fee_type_presenter.rb
```diff
@@ -8,18 +8,21 @@ def data_attributes
...
     }
   end
 
-
   private
 
-  def unless_disbursement_or_warrant?
-    !(fee_type.is_disbursement? || fee_type.is_interim_warrant?)
+  def is_not_disbursement_or_warrant?
+    is_not_disbursement? && is_not_warrant?
```
stephenrichards, line 21: I think it's great to have renamed this method, but I think to make it even clearer what it does, it could be renamed is_neither_disbursement_nor_warrant

app/validators/fee/interim_fee_validator.rb
```diff
@@ -42,11 +39,17 @@ def validate_disbursements
...
 
-  def validate_warrant
-    if @record.is_interim_warrant?
-      add_error(:warrant, 'blank') if @record.claim.warrant_fee.nil?
-    else
-      add_error(:warrant, 'present') if @record.claim.warrant_fee.present?
-    end
+  def validate_warrant_issued_date
+    return if !@record.is_interim_warrant?
+    validate_presence(:warrant_issued_date, 'blank')
```
stephenrichards, line 44: I think return unless @record.is_interim_warrant? is better here

app/validators/fee/interim_fee_validator.rb
```diff
@@ -42,11 +39,17 @@ def validate_disbursements
...
+    validate_not_before(Settings.earliest_permitted_date, :warrant_issued_date, 'check_not_too_far_in_past')
+    validate_not_after(Date.today, :warrant_issued_date, 'check_not_in_future')
+  end
+
+  def validate_warrant_executed_date
+    return if !@record.is_interim_warrant?
+    validate_not_before(@record.warrant_issued_date, :warrant_executed_date, 'warrant_executed_before_issued')
```
stephenrichards, line 51: and here

app/validators/fee/interim_fee_validator.rb
slorek, line 44: I think Rubocop will enforce this if set.

PR #1478, opened 2016-05-27
Ui changes to split the sections

cleargif:
We may need some copy here:
screen shot 2016-05-27 at 12 52 21

Look at me Ma.... I am doing ruby specs...


cleargif:
@jsugarman @zheileman

zheileman:
@cleargif I've pushed some small changes to improve the Code Climate issue. I've also removed `html_safe` as we are not using `hint_tag` anymore in the section_header so is redundant.

cleargif:
@zheileman thank you... it built now.. can you merge please..

PR #1481, opened 2016-05-30
Update api

jsugarman:
Ensures only AGFS fee types and expenses returned from API lookup data endpoints, reintroducing skipped specs. Plus some refactoring.

### DIFF
--- a/app/models/concerns/roles.rb
+++ b/app/models/concerns/roles.rb
@@ -2,16 +2,29 @@ module Roles
   extend ActiveSupport::Concern
 
   included do |klass|
+    klass.extend(ClassMethods)
     klass.serialize :roles, Array
     klass.before_validation :strip_empty_role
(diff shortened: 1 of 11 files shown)


app/models/concerns/roles.rb
```diff
@@ -2,17 +2,30 @@ module Roles
...
 
   included do |klass|
+    klass.extend(ClassMethods)
     klass.serialize :roles, Array
     klass.before_validation :strip_empty_role
     klass.validate :roles_valid
 
     klass::ROLES.each do |role|
-      klass.scope role.pluralize.to_sym, -> { klass.select { |m| m.roles.include?(role) } }
+      klass.scope role.pluralize.to_sym, -> { matching_role_query(["#{role}"]) }
```
stephenrichards, line 12: Nice!

PR #1486, opened 2016-06-01
Direct exceptions to Raven/Sentry

stephenrichards:


features/step_definitions/common_steps.rb
```diff
@@ -34,7 +34,7 @@
 
 When(/I click the claim '(.*?)'$/) do |case_number|
   @external_user_home_page.claim_for(case_number).case_number.click
-  sleep 0.5   # wait for page to load and populate select lists
+  sleep 2   # wait for page to load and populate select lists
```
slorek, line 37: Perhaps a `wait_until` condition would be more reliable?

app/controllers/application_controller.rb
```diff
@@ -18,12 +18,18 @@ class ApplicationController < ActionController::Base
...
       else
+        Raven.capture_exception(exception)
         redirect_to error_500_url
       end
     end
   end
 
 
+  def dummy_exception
+    Raise ArgumentError.new('Boom!')
```
slorek: Consider adding a more descriptive message (so people looking at Sentry know where it came from).

app/controllers/application_controller.rb
```diff
@@ -18,12 +18,18 @@ class ApplicationController < ActionController::Base
       if exception.is_a?(RuntimeError)
         raise exception
```
slorek: Not directly related, but what is the value in re-raising a `RuntimeError`?

PR #1487, opened 2016-06-01
Attempt to resolve flickering cukes

stephenrichards:


slorek:
Consider using `until` blocks to test and wait for the existence of required page elements in the DOM.

PR #1501, opened 2016-06-07
Fix exceptions when submitting a claim more than once

slorek:
Example: https://sentry.service.dsd.io/mojds/staging-adp/group/965/ - the problem was the `result` object was not returned.

I've refactored the unit tests for the affected service as well so that the individual methods and their return values are tested and with more meaningful contexts.

https://www.pivotaltracker.com/n/projects/1520095

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -150,7 +150,6 @@ Style/AlignArray:
 Style/AndOr:
   Exclude:
     - 'app/controllers/external_users/claim_types_controller.rb'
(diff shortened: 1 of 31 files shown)


slorek:
The CI failures are related to front-end Capybara tests; could someone confirm if these were broken by this change or if there's some other issue?

stephenrichards:
I.m going to take this over - we need to do it for the other claim creation / update services as well, and I'll see what's going on with the cukes.

PR #1503, opened 2016-06-07
Apply Rubocop newline rules for improved consistency and readability

slorek:
There were a lot of empty lines applied inconsistently across the project which makes it a little annoying to read. I've applied Rubocop's default rules to clean this up.

Dependency on https://github.com/ministryofjustice/advocate-defence-payments/pull/1502 as I've included the `spec/` directory in this.

Also fixed up a couple of new unrelated failures.


stephenrichards:
I think we need to park this for a bit - It's gong to be a nightmare merging it with a codebase that is changing so fast with 3 devs working on it, and Joel, Jesus and I had agreed on what standards we wanted to follow  - some of the spacing standards we feel are too strict.

PR #1511, opened 2016-06-08
Date validation for Interim claims

stephenrichards:
PT: https://www.pivotaltracker.com/story/show/121119527
This PR validates the the following dates are not in the future:
- effective_pcmh_date
- retrial_started_at
- trial_concluded_at

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -44,8 +44,8 @@ claim_csv_headers:
 expense_schema_version: 2
 
 # temporary feature toggle til interim fees ready
-allow_lgfs_interim_fees?: true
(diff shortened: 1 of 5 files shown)


config/settings.yml
```diff
@@ -44,7 +44,8 @@ claim_csv_headers:
 expense_schema_version: 2
 
 # temporary feature toggle til interim fees ready
-allow_lgfs_interim_fees?: false
+allow_lgfs_interim_fees?: true
```
zheileman: We don't want to allow these claims yet, are we?

PR #1512, opened 2016-06-08
Removing google analytics and replacing it with GTM. Some virtual pages implemented.

zheileman:
**PT#120839091**
**PT#120838641**

We are now using Google Tag Manager.
As per David request, all Google Analytics stuff have been removed.

All of the virtual pages in the spreadsheet have been implemented. More to come as separated PR.


zheileman:
@stephenrichards ready to merge if we are happy with it. More tracking can be added as part of separated PRs.

PR #1526, opened 2016-06-10
W3c fixes

cleargif:
- nesting issues
- wrong / extra attrs removed
- updated deprecated attr usage
- table header / cols fixes
- general fixes relating to semantic validation

### DIFF
--- a/app/views/shared/_summary.html.haml
+++ b/app/views/shared/_summary.html.haml
@@ -14,7 +14,8 @@
           %col
           %col
           %col
-          %col
+          - if claim.agfs?
+            %col
         %thead
           %tr
             %th{ scope: 'col'}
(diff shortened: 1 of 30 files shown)


app/views/shared/_summary.html.haml
```diff
@@ -71,7 +72,9 @@
             %col
             %col
             %col
-            %col
+            - if claim.lgfs?
```
zheileman, line 132: Do we really need this colgroup? As part of another PR for a separate issue, I removed the colgroup completely, as I don't see the point of it:

https://github.com/ministryofjustice/advocate-defence-payments/pull/1528/commits/259fa74218c4f49bd4033f384dd74196c6a2ef6c

If we need it, I can reintroduce it on my PR, as otherwise there will be conflicts when merging yours.

cleargif:
@zheileman we can likely remove it - we are not actually using it for its intended purpose.

I will create a separate ticket for it

PR #1538, opened 2016-06-15
updated min / max number on quantity

cleargif:


cleargif:
@zheileman little pr needs merging

PR #1539, opened 2016-06-16
Ensure we validate fees when importing a claim via JSON.

zheileman:
**PT#121584253**

Some fees are rigidly validated on the backend when created via web, but with the JSON importer, these validations were not triggered, generating a draft impossible to modify (i.e. the rate field is not visible/editable for PPE on the web but was populated in the JSON), rendering the draft unusable.

The issue has been addressed at the time of parsing the JSON file and generating the fees, raising an error when there are invalid fees.

**NOTE**: we don't really want to rigidly validate the whole claim and associated objects on import, only those that can render the draft impossible to edit, like in this case Fees.
Otherwise it can be very frustrating to use the JSON importer. Missing or invalid fields can be filled/fixed via web at a later stage by the user, that's the point of a Draft.

### DIFF
--- a/app/views/external_users/json_document_importers/create.js.erb
+++ b/app/views/external_users/json_document_importers/create.js.erb
@@ -1,4 +1,5 @@
 var $importerResults = $('.importer-results');
+var resetUploaderForm = function() { $('#importer').find('form').addClass('no-file-selected').find('.file-upload-name').empty(); };
 
 $importerResults.find('.js-import-failed').remove();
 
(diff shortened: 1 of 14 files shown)


app/views/external_users/json_document_importers/create.js.erb
```diff
@@ -73,7 +68,7 @@ $importerResults.find('.js-import-failed').remove();
       '</ul>' +
     '</div>'
   );
-  $('#importer').addClass('has-errors');
+  $('#importer').addClass('has-errors').find('form').addClass('no-file-selected').find('.file-upload-name').empty();
```
slorek: It looks like this form field reset function should be extracted to a function, as it's repeated three times in this change alone.

PR #1546, opened 2016-06-20
AdpFormBuilder#adp_text_field

stephenrichards:
This PR makes a method adp_text_field() avaialble on the AdpFormBuilder which outputs the HTML for a text field wrapped in all the GDS specific markup
to allow labels, hints, error messages, etc.

### DIFF
--- a/app/form_builders/adp_text_field.rb
+++ b/app/form_builders/adp_text_field.rb
@@ -1,0 +1,134 @@
+
+# class to create specialized text field wrapped in all the GDS gubbins
+class AdpTextField
+
+  include ExternalUsers::ClaimsHelper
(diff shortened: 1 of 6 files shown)


app/form_builders/adp_text_field.rb
```diff
@@ -0,0 +1,134 @@
...
+  def has_errors?
+    @errors.errors_for?(@anchor_id.to_sym)
+  end
+
+  def to_html
+    result = div_start
+    result += anchor
+    result += label
+    result += hint
+    result += error_message
```
cleargif, line 48: The error message and the span it is wrapped in should only be shown if there is an error.

Currently this element is being shown and makes for the red line

``` html
<span class="error"></span>
```

PR #1569, opened 2016-06-27
Stats

stephenrichards:
This PR implements the first stage of producing a dashboard to track the daily number of claims that are being
submitted.  It models a statistics table to store the results, and the first collector, to store the number of submitted claims.

It does not yet implement the cron job to record this information daily, nor the gecko API.


slorek:
👍  When building the API endpoints bear in mind please that we may wish to pull this information out in the future with other clients - the LAA have a data store called MI which stores a wide array of performance metrics and their Integration Hub is currently pushing data to it from APIs in existing apps.

PR #1635, opened 2016-07-13
UI fixes: Re-/Allocation fixes

cleargif:
https://www.pivotaltracker.com/story/show/126020835

### DIFF
--- a/app/views/case_workers/admin/allocations/_allocation.html.haml
+++ b/app/views/case_workers/admin/allocations/_allocation.html.haml
@@ -20,7 +20,7 @@
         = label_tag :quantity_to_allocate, t('.no_of_claims')
         = number_field_tag :quantity_to_allocate, nil, { style: 'width:100%', min: 0, size: 5, class: 'form-control'}
       .form-col
-        = f.label :case_worker_id, t('dictionary.models.case_worker')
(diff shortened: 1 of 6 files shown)


app/views/case_workers/admin/allocations/_allocation.html.haml
```diff
@@ -62,7 +62,7 @@
                 %td
                   %span.js-test-case-number
                     = link_to claim.case_number , case_workers_claim_path(claim)
-                    %div.unique-id-small= claim.unique_id
+                    %span.unique-id-small= claim.unique_id
```
zheileman, line 65: @cleargif changing to span makes the ID stay in the same line sometimes, instead of jumping to the next line.
This is not really an issue on production because the IDs are high, but will happen frequently on local and dev.

screen shot 2016-07-13 at 17 00 27

cleargif, line 65: Will have to look at it again. Div cannot be side a span. Will have to apply some display styling

cleargif, line 65: @zheileman I fixed it with css but I am to ashamed to actually tag the line where I did it..

Git will give me away anyway...damn you git!

PR #1641, opened 2016-07-14
Link meta data + tab indexes

cleargif:
https://www.pivotaltracker.com/story/show/126303533

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -512,18 +512,18 @@ en:
         scheme: Scheme
         messages: 'Messages'
         defendants: 'Defendant(s)'
-        case_number: 'Case number'
+        case_number_html: 'Case number <span class="visuallyhidden">click to order by this column</span></a>'
         case_type: 'Case type'
-        submission_date: 'Date submitted'
+        submission_date_html: 'Date submitted <span class="visuallyhidden">click to order by this column</span></a>'
         authorised_date: 'Date authorised'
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -512,18 +512,18 @@ en:
...
         case_type: 'Case type'
-        submission_date: 'Date submitted'
+        submission_date_html: 'Date submitted <span class="visuallyhidden">click to order by this column</span></a>'
         authorised_date: 'Date authorised'
-        total: 'Claimed'
-        assessed: 'Assessed'
-        status: 'Status'
+        total_html: 'Claimed <span class="visuallyhidden">click to order by this column</span></a>'
+        assessed_html: 'Assessed <span class="visuallyhidden">click to order by this column</span></a>'
+        status_html: 'Status  <span class="visuallyhidden">click to order by this column</span></a>'
```
cleargif: @zheileman is there a more efficient way to do this?

I mean the repetition

zheileman:
@cleargif I've committed a small refactor to move the visually hidden hint text to the `sortable()` method itself, as otherwise it is going to generate as you said a lot of repetition.

Apart from **codeclimate** complaining the method is now too complex (no worries for now) the drawback is the hint text is not i18n and is hardcoded in the sortable() method, but that is not a huge concern at the moment. Maybe in the future!

PR #1642, opened 2016-07-15
Remove repo date validation on fixed/warned dates for cracked trials.

zheileman:
**PT#126318253**

We have found through litigator use that the fixed and warned dates for cracked trials can be before the date of the rep order.  

We therefore need to remove the validation on these fields that says it must be after.

### DIFF
--- a/app/validators/base_validator.rb
+++ b/app/validators/base_validator.rb
@@ -103,10 +103,6 @@ class BaseValidator < ActiveModel::Validator
     @record.errors.add(attribute, message)
   end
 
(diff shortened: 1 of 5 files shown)


app/validators/base_validator.rb
```diff
@@ -103,10 +103,6 @@ def add_error(attribute, message)
     @record.errors.add(attribute, message)
   end
 
-  def case_type_in(*case_types)
```
zheileman, line 106: I've removed this method as it was not being used in any place.

PR #1652, opened 2016-07-18
DAC: Descriptive links in header

cleargif:
https://www.pivotaltracker.com/story/show/126303869

### DIFF
--- a/app/views/layouts/_user.html.haml
+++ b/app/views/layouts/_user.html.haml
@@ -1,6 +1,6 @@
 #proposition-links
   %ul
     %li
-      = link_to current_user.name, signed_in_user_profile_path, class: cp(signed_in_user_profile_path)
+      = link_to content_tag(:span, 'Your account ', class: 'visuallyhidden') + current_user.name, signed_in_user_profile_path, class: cp(signed_in_user_profile_path)
     %li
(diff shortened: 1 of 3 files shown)


app/views/layouts/_user.html.haml
```diff
@@ -1,6 +1,6 @@
 #proposition-links
   %ul
     %li
-      = link_to current_user.name, signed_in_user_profile_path, class: cp(signed_in_user_profile_path)
+      = link_to ('<span class="visuallyhidden">Your account </span>' + current_user.name).html_safe, signed_in_user_profile_path, class: cp(signed_in_user_profile_path)
```
zheileman: @cleargif - using **html_safe** here will open an **XSS** attack vector, as current_user.name is user-generated data and can be filled with specially crafted html/javascript that will render back on page load.

You can replace the above with the following code and have the same result, without XSS:

``` ruby
= link_to content_tag(:span, 'Your account ', class: 'visuallyhidden') + current_user.name, signed_in_user_profile_path, class: cp(signed_in_user_profile_path)
```

If you expect to repeat the visuallyhidden span in other places, we can expose a little helper to simplify that.

cleargif:
@zheileman this is ready again

PR #1655, opened 2016-07-20
DAC: Table semantics

cleargif:
https://www.pivotaltracker.com/story/show/126305187


cleargif:
@stephenrichards @zheileman 

There is some repetitions in here but we can fix that after the next testing DAC will do

Please check my spelling

cleargif:
@stephenrichards 👍

PR #1663, opened 2016-07-22
DAC: Internal link descriptive text

cleargif:
https://www.pivotaltracker.com/story/show/126858027
- document View, Download and Remove links
- claim list case number aria lable

### DIFF
--- a/app/views/case_workers/claims/_claims_list.html.haml
+++ b/app/views/case_workers/claims/_claims_list.html.haml
@@ -29,8 +29,12 @@
       - present_collection(claims).each do |claim|
         %tr
           %td
-            = link_to claim.case_number , case_workers_claim_path(claim),{:class => 'js-test-case-number-link'}
-            %div.unique-id-small= claim.unique_id
+            = link_to claim.case_number ,
+                      case_workers_claim_path(claim),
+                      {:class => 'js-test-case-number-link',
+                        'aria-label': "View #{claim.state.humanize} Claim, Case number: #{claim.case_number}"}
+            %div.unique-id-small
+              = claim.unique_id
           %td
             = claim.external_user.name
           %td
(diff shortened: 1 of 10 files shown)


app/views/case_workers/claims/_claims_list.html.haml
```diff
@@ -29,8 +29,9 @@
       - present_collection(claims).each do |claim|
         %tr
           %td
-            = link_to claim.case_number , case_workers_claim_path(claim),{:class => 'js-test-case-number-link'}
-            %div.unique-id-small= claim.unique_id
+            = link_to claim.case_number , case_workers_claim_path(claim),{:class => 'js-test-case-number-link', 'aria-label': "#{claim.state.humanize}, claim I.D. #{claim.unique_id}"}
```
zheileman: Hey @cleargif, wouldn't be better to use the case number instead of the database ID in the aria-label?
The database ID (unique_id) is not going to be very meaningful to a screen reader user.

cleargif: @zheileman I think I need to put both in there and see what it tests like

stephenrichards: @cleargif I think the user is only interested in the case number, not the database id (we only ask them to specify that if there is an issue with the claim, so we can live without it in the case of visually impaired caseworker/external user).

Also, this field can be empty (you can save a draft claim without specifying a case number) - so we need to say something like casenumber not specified yet in this case.  JLO or I can write a presenter method to do that if you want.

app/views/shared/_evidence_list.html.haml
```diff
@@ -36,6 +36,13 @@
               = document.created_at
             %td
               - if document.converted_preview_document.present?
-                = link_to t('common.view'), document_path(document), target: '_blank', class: 'view'
+                = link_to t('common.view'),
+                          document_path(document),
+                          target: '_blank',
+                          class: 'view',
+                          'aria-label': "Evidance: View document #{document.document_file_name}, opens in new window"
```
zheileman: @cleargif little typo 'Evidance' should read 'Evidence' (there is another occurrence of this few lines down).

PR #1690, opened 2016-08-02
Ensuring we catch early json import failures before trying to process the claim(s).

zheileman:
Some bits of code have been refactor to make it easier to validate against 2 different schemas,
one is a basic version to fail as early as possible, the other is a full schema to validate
against each of the claims found in the JSON imported.


slorek:
Good job on the refactoring; just a small comment would be useful to explain the use of the basic schema as it wasn't immediately clear.

zheileman:
@slorek sure thing. I've added some comments.

PR #1692, opened 2016-08-04
Sidekiq will use REDIS_URL env var if set.

zheileman:
Moved Sidekiq config to their own initializer to follow the conventions and to add some explanation about Redis configuration.


lukaszraczylo:
It's default sidekiq's behaviour. No need for change :)

PR #1693, opened 2016-08-04
Disallow lowercase letters in supplier numbers and better exception handling

zheileman:
Make sure all current supplier numbers in database are uppercase and do not
allow future ones to contain lower case letters.

Also, a not handled exception could be raised when allocating claims with
validations not met (like in this case supplier numbers). Now we handle
the exception and gives the user feedback (no more 500 error page).

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -85,10 +85,12 @@ class Allocation
 
   def allocate_or_error_claim!(claim)
     if claim.case_workers.exists?
-      errors.add(:base,"Claim #{claim.case_number} has already been allocated to #{claim.case_workers.first.name}")
+      errors.add(:base, "Claim #{claim.case_number} has already been allocated to #{claim.case_workers.first.name}")
     else
       allocate_claim! claim
     end
+  rescue => ex
+    errors.add(:base, "Claim #{claim.case_number} has errors: #{ex.message}")
   end
 
   def rollback_all_allocations!
--- a/app/models/supplier_number.rb
+++ b/app/models/supplier_number.rb
@@ -8,7 +8,7 @@
 #
 
 class SupplierNumber < ActiveRecord::Base
-  SUPPLIER_NUMBER_REGEX ||= /\A[0-9][a-zA-Z][0-9]{3}[a-zA-Z]\z/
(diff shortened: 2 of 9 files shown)


app/models/allocation.rb
```diff
@@ -85,10 +85,12 @@ def allocate_all_claims_or_none!(claims)
 
   def allocate_or_error_claim!(claim)
     if claim.case_workers.exists?
-      errors.add(:base,"Claim #{claim.case_number} has already been allocated to #{claim.case_workers.first.name}")
+      errors.add(:base, "Claim #{claim.case_number} has already been allocated to #{claim.case_workers.first.name}")
     else
       allocate_claim! claim
     end
+  rescue => ex
```
slorek, line 92: Consider specifying the class of exceptions we're looking out for. See http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

app/models/supplier_number.rb
```diff
@@ -8,7 +8,7 @@
 #
 
 class SupplierNumber < ActiveRecord::Base
-  SUPPLIER_NUMBER_REGEX ||= /\A[0-9][a-zA-Z][0-9]{3}[a-zA-Z]\z/
+  SUPPLIER_NUMBER_REGEX ||= /\A[0-9][A-Z][0-9]{3}[A-Z]\z/
```
slorek, line 11: Is this a different supplier number than above? If not, the regex should be the same (and ideally not duplicated).

app/models/allocation.rb
zheileman, line 92: In this case we don't really care about the exception, just any (StandardError) exception will do, as it is going to rollback the whole transaction. We are not recovering from it so I don't see the point of specifying for example StateMachines::InvalidTransition (that's the one being raised) when some other might be raised down the line.

app/models/supplier_number.rb
zheileman, line 11: Yes they are different.

app/models/allocation.rb
slorek, line 92: Rescuing `StandardError` is much more preferable than all `Exception` classes.

PR #1694, opened 2016-08-04
Restoring contact us page and linking it in the footer.

zheileman:
**PT#127449803**

### DIFF
--- a/app/views/grape_swagger_rails/application/index.html.erb
+++ b/app/views/grape_swagger_rails/application/index.html.erb
@@ -263,6 +263,9 @@
             <li>
               <a href="http://www.gov.uk/help/cookies">Cookies</a>
             </li>
+            <li>
+              <a href="/contact_us">Contact</a>
+            </li>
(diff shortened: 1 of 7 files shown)


app/views/grape_swagger_rails/application/index.html.erb
```diff
@@ -263,6 +263,9 @@
             <li>
               <a href="http://www.gov.uk/help/cookies">Cookies</a>
             </li>
+            <li>
+              <a href="/contact_us">Contact</a>
```
slorek, line 267: Is it possible to use `contact_us_page_path` here?

zheileman, line 267: @slorek I tried, but it raises an exception, undefined local variable or method `contact_us_page_path'
I don't think that page (it is swagger documentation for the API) has visibility of the route helpers apart from other quirkiness judging by how it is written and the comments in there. It doesn't use our application layout, etc.

PR #1698, opened 2016-08-09
CLI task to test the notify service API and ruby client.

zheileman:
The purpose of this is to provide a simple way to test the
formatting and sending of emails via the GOV.uk Notify service,
and to test the ruby library available from GDS.

It is customizable to a certain degree but it is not intended to
be used other than for test purposes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem 'kaminari',               '~> 0.16.2'
 gem 'libreconv',              '~> 0.9.0'
 gem 'logstasher',             '0.6.2'
 gem 'logstuff',               '0.0.2'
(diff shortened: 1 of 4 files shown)


zheileman:
@slorek I created this PR yesterday but forgot to tell. It is ready.

slorek:
LGTM but could you please add some documentation in the README.

zheileman:
@slorek it is not really intended to stay in the codebase, it was just part of the chore/spike, to help to test the ruby library.
The idea is to remove it once we implement the emails, as it is useless otherwise. Once we implement the emails in the app, the app itself will send those emails without need for a rake task.

It is internal, for us lets say. Not to be exposed in the README.

PR #1699, opened 2016-08-10
New ActionMailer delivery method for GOV.UK Notify service.

zheileman:
This PR prepare the work to be able to send emails via an external
service (GOV.UK Notify) maintaining the same interface as any other
standard ActionMailer.
Also it will use ActiveJob, and the standard delivery configuration
for enqueued deliveries.

The way it is done will simplify extracting to a gem as there is no
coupling with our own app.

An example 'NotifyMailer' is included with a test email method.

### DIFF
--- a/app/mailers/govuk_notify_mailer.rb
+++ b/app/mailers/govuk_notify_mailer.rb
@@ -1,0 +1,37 @@
+# Extract to gem and inherit in the app from this class
+#
+class GovukNotifyMailer < ActionMailer::Base
+  default delivery_method: :govuk_notify
+
+  attr_accessor :govuk_notify_template
+  attr_accessor :govuk_notify_personalisation
+
+  protected
+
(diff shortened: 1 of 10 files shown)


app/mailers/govuk_notify_mailer.rb
```diff
@@ -0,0 +1,37 @@
...
+    _validate!
+
+    headers[:body] ||= _default_body
+
+    message = super(headers)
+    message.govuk_notify_template = govuk_notify_template
+    message.govuk_notify_personalisation = govuk_notify_personalisation
+  end
+
+  def set_template(template)
```
slorek, line 21: Might be useful to do type checking here and in `set_personalisation`. Not a blocker.

slorek:
LGTM just one suggestion. Please merge when ready.

PR #1700, opened 2016-08-10
Remove the option 'Travel costs' from the disbursements options.

zheileman:
As there are claims using this disbursement and we can't reassign those
to another disbursement, a temporal workaround was discussed to 'hide'
the option from the dropdowns and API but to continue in the database so
old claims will continue working and validating, but new ones will not
be able to select 'Travel costs'.

Sometime from now it will be safe to remove the option from the database.

PT#126855639

### DIFF
--- a/spec/models/disbursement_type_spec.rb
+++ b/spec/models/disbursement_type_spec.rb
@@ -16,4 +16,12 @@ RSpec.describe DisbursementType, type: :model do
 
   it { should validate_presence_of(:name) }
   it { should validate_uniqueness_of(:name) }
+
+  describe '.allowable_types' do
+    let!(:travel_costs) { create(:disbursement_type, name: 'Travel costs') }
+
+    it 'should exclude "Travel costs" from the result set' do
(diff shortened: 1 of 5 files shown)


spec/models/disbursement_type_spec.rb
```diff
@@ -16,4 +16,12 @@
 
   it { should validate_presence_of(:name) }
   it { should validate_uniqueness_of(:name) }
+
+  describe '.filtered_all' do
+    let!(:travel_costs) { create(:disbursement_type, name: 'Travel costs') }
+
+    it 'should exclude "Travel costs" from the resulset' do
```
slorek: *result set

slorek:
Perhaps a more useful name than `filtered_all` would be something along the lines of `allowable_types`. It can be maintained going forward and is more descriptive I think.

slorek:
Please ensure there's a ticket in the backlog to revisit (as per your code comment) if this is indeed necessary.

zheileman:
@slorek yep, you're right. My naming is awful.
I will change it and push again.

PR #1711, opened 2016-08-17
Add deleted_at to User and CaseWorker

stephenrichards:
This PR sets up the framework for soft delete of caseworkers and users.

It adds a 'deleted_at' field to CaseWorker and User models, and sets the
default scope to exclude any records where deleted at is not null.  Also
adds an including_deleted scope which returns all records.

This has no effect on the current deletion process as that hasn't yet been
changed to use the deleted_at fields.

### DIFF
--- a/app/models/case_worker.rb
+++ b/app/models/case_worker.rb
@@ -21,7 +21,9 @@ class CaseWorker < ActiveRecord::Base
   has_many :case_worker_claims, dependent: :destroy
   has_many :claims, class_name: Claim::BaseClaim, through: :case_worker_claims, after_remove: :unallocate!
 
-  default_scope { includes(:user) }
+  default_scope { includes(:user).where(deleted_at: nil) }
+
+  scope :including_deleted, -> {unscoped.includes(:user) }
 
   validates :location, presence: {message: 'Location cannot be blank'}
   validates :user, presence: {message: 'User cannot be blank'}
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,6 +27,10 @@
 
 class User < ActiveRecord::Base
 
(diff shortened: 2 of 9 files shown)


app/models/user.rb
```diff
@@ -27,6 +27,10 @@
 
 class User < ActiveRecord::Base
 
+  default_scope { where(deleted_at: nil) }
+
+  scope :including_deleted, -> { unscoped }
```
slorek, line 32: Consider changing this implementation to not use `unscoped`, because this would override all scopes when chained, including all the relation scopes (associations). You could use `unscope(where: :deleted_at)` to constrain the behaviour to that column only.

app/models/case_worker.rb
```diff
@@ -21,7 +21,9 @@ class CaseWorker < ActiveRecord::Base
   has_many :case_worker_claims, dependent: :destroy
   has_many :claims, class_name: Claim::BaseClaim, through: :case_worker_claims, after_remove: :unallocate!
 
-  default_scope { includes(:user) }
+  default_scope { includes(:user).where(deleted_at: nil) }
+
+  scope :including_deleted, -> {unscoped.includes(:user) }
```
slorek, line 26: Consider changing this as below.

slorek:
Consider using an explicit scope/class method rather than `default_scope`, in my experience this causes a mess of problems. See http://weblog.jamisbuck.org/2015/9/19/default-scopes-anti-pattern.html

slorek:
I recommend making the relevant implementation changes in this PR as well so we can see the full effect of the changes on the codebase.

slorek:
Looks like the migration hasn't run on CI 😢

PR #1715, opened 2016-08-18
Implement Soft Deletions for caseworkers

stephenrichards:
Caseworkers are no longer deleted from the database because of claims and messages that
may reference the deleted caseworker.  Now a new 'deleted_at' field on CaseWorker and User
is given a timestamp, and any record where that field is not null is considered to be deleted.

A new scope active on CaseWorker and User will only return records where deleted_at is nil,
and Caseworker queries have been updated to use that scope where approperiate.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -89,4 +89,16 @@ class User < ActiveRecord::Base
   def soft_delete
     update(deleted_at: Time.zone.now)
   end
+
+  def active?
+    self.deleted_at.nil?
+  end
+
+  def active_for_authentication?
+    super && active?
+  end
+
+  def inactive_message
+    active? ? super : 'This account has been deleted.'
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,6 +27,9 @@
 
 class User < ActiveRecord::Base
 
+  scope :active, -> { where(deleted_at: nil) }
+  scope :deleted, -> { where.not(deleted_at: nil) }
+
   auto_strip_attributes :first_name, :last_name, :email, squish: true, nullify: true
 
   # Include default devise modules. Others available are:
@@ -41,7 +44,7 @@ class User < ActiveRecord::Base
 
   belongs_to :persona, polymorphic: true
   has_many :messages_sent, foreign_key: 'sender_id', class_name: 'Message'
-  has_many :user_message_statuses, dependent: :destroy
+  has_many :user_message_statuses
 
   validates :first_name, :last_name, presence: true
   validates :email, confirmation: true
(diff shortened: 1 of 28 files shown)


db/migrate/20160111111547_change_advocate_persona_type_to_external_user.rb
```diff
@@ -1,9 +1,5 @@
 class ChangeAdvocatePersonaTypeToExternalUser < ActiveRecord::Migration
   def change
-    User.all.each do |user|
-      if user.persona_type == 'Advocate'
-        user.update_column(:persona_type, 'ExternalUser')
-      end
-    end
+    execute(%q{UPDATE users SET persona_type = 'ExternalUser' WHERE persona_type = 'Advocate'} )
```
slorek: I think this should be an `up` method and a corresponding `down` method needs to be implemented. Otherwise Rails won't be able to reverse this migration.

slorek:
LGTM apart from the comment on the migration implementation.

zheileman:
LGTM. In another PR we can stop deleted users from logging. And when we do the soft delete for claims we can move all the boilerplate code to a concern, kind of 'SoftDeletable' or whatever, as the methods, scopes, attributes, etc are going to be the same.

In the past I've implemented soft-deletes overriding the destroy method, to actually perform the soft-delete, and having another separate method for performing the hard delete, because I think that's safer. Just a thought.

Also note destroy callbacks will not trigger, although we are not using any at the moment.

slorek:
Good point about the callbacks. We could use the `run_callbacks` block to do this. I'm not sure overriding `destroy` is the safest solution, however.

db/migrate/20160111111547_change_advocate_persona_type_to_external_user.rb
stephenrichards: Well I could do, but I only changed the old version so that the migrations would work now that I've added a new field to the users table - it runs all the migrations from scratch as part of the CI suite, which is probably wrong - it should be loading the database schema from schema.rb.  The old version didn't allow backwards migration and we're not going to a migration from January.

app/models/user.rb
```diff
@@ -82,4 +85,20 @@ def save_settings!(data)
...
+  def active?
+    self.deleted_at.nil?
+  end
+
+  def active_for_authentication?
+    super && active?
+  end
+
+  def inactive_message
+    active? ? super : 'This account has been deleted.'
```
zheileman, line 102: We can move this string to devise.en.yml and replace it with a symbol, i.e. :account_deleted

features/step_definitions/sign_in_steps.rb
```diff
@@ -95,3 +103,12 @@ def make_accounts(role, number = 1)
 Given(/^I sign out$/) do
   click_link 'Sign out' rescue nil
 end
+
+When(/^I should be on the Allocation page$/) do
+  expect(find('header.main-header')).to have_content('Allocation')
+end
+
+Then(/^I should get a page telling me my account has been deleted$/) do
+  save_and_open_page
```
zheileman, line 112: save_and_open_page

PR #1716, opened 2016-08-18
Soft deletable concern for CaseWorker and User

stephenrichards:
Caseworkers are no longer deleted from the database because of claims and messages that
may reference the deleted caseworker. Now a new 'deleted_at' field on CaseWorker and User
is given a timestamp, and any record where that field is not null is considered to be deleted.

A new scope active on CaseWorker and User will only return records where deleted_at is nil,
and Caseworker queries have been updated to use that scope where approperiate.

This is implemented as a Concern as it may well be used on Claims as well.

### DIFF
--- a/app/models/case_worker.rb
+++ b/app/models/case_worker.rb
@@ -15,13 +15,18 @@ class CaseWorker < ActiveRecord::Base
   ROLES = %w{ admin case_worker }
 
   include Roles
+  include SoftlyDeletable
 
   belongs_to :location
   has_one :user, as: :persona, inverse_of: :persona, dependent: :destroy
   has_many :case_worker_claims, dependent: :destroy
   has_many :claims, class_name: Claim::BaseClaim, through: :case_worker_claims, after_remove: :unallocate!
 
+
+
   default_scope { includes(:user) }
+  scope :active, -> { where(deleted_at: nil) }
+  scope :deleted, -> { where.not(deleted_at: nil) }
 
   validates :location, presence: {message: 'Location cannot be blank'}
   validates :user, presence: {message: 'User cannot be blank'}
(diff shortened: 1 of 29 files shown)


app/models/case_worker.rb
```diff
@@ -15,13 +15,18 @@ class CaseWorker < ActiveRecord::Base
...
 
   belongs_to :location
   has_one :user, as: :persona, inverse_of: :persona, dependent: :destroy
   has_many :case_worker_claims, dependent: :destroy
   has_many :claims, class_name: Claim::BaseClaim, through: :case_worker_claims, after_remove: :unallocate!
 
+
+
   default_scope { includes(:user) }
+  scope :active, -> { where(deleted_at: nil) }
```
zheileman: these scopes can be removed as they are now in the concern isn't it?

app/models/user.rb
```diff
@@ -82,4 +84,9 @@ def save_settings!(data)
   def unauthenticated_message
     override_paranoid_setting(false) { super }
   end
+
+  def active?
```
zheileman: same for this method, it is in the concern

app/models/concerns/softly_deletable.rb
```diff
@@ -0,0 +1,41 @@
...
+        update(deleted_at: Time.zone.now)
+        self.after_soft_delete if self.respond_to?(:after_soft_delete)
+      end
+    end
+
+    def active?
+      self.deleted_at.nil?
+    end
+
+    def active_for_authentication?
```
zheileman: Not sure about including these 2 devise methods in the concern. After all, they just need to be in User, and if we include the concern in several models, all of them will get polluted with the devise methods, and not having any advantage. What do you think?

stephenrichards: yep, I think you're right.  I'll take them out and put them back in User.

PR #1723, opened 2016-08-19
Add contact us line to bottom of external user pages

stephenrichards:
PT https://www.pivotaltracker.com/story/show/128651399


stephenrichards:
We need to style it in such a way as it appears at the bottom of the page in all cases.

screenshot 2016-08-25 15 16 19

cleargif:
I have moved the element in and styled it correctly. It fixes the layout issue above

PR #1725, opened 2016-08-22
Claim scheme filters in the main claim list for external users.

zheileman:
The filters will only appear when the user has the ability to create AGFS
and LGFS claims, otherwise it doesn't make sense to show the filters.

When LGFS filter is selected, the JSON importer is hidden, as we don't
support yet LGFS claim imports and it creates a confusing experience.
- ALL/AGFS selected (JSON importer available)
  screen shot 2016-08-22 at 17 26 08
- LGFS selected (JSON importer hidden)
  screen shot 2016-08-22 at 17 26 43

### DIFF
--- a/app/models/expense.rb
+++ b/app/models/expense.rb
@@ -102,8 +102,7 @@ class Expense < ActiveRecord::Base
   end
 
   def allow_reason_text?
-    return false if self.reason_id.nil?
-    expense_reason.allow_explanatory_text?
+    !!expense_reason&.allow_explanatory_text?
   end
 
   def expense_reasons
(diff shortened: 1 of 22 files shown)


zheileman:
Awaiting user research from Chris.

app/models/expense.rb
```diff
@@ -102,8 +102,7 @@ def expense_reason
   end
 
   def allow_reason_text?
-    return false if self.reason_id.nil?
-    expense_reason.allow_explanatory_text?
+    !!expense_reason&.allow_explanatory_text?
```
zheileman, line 105: This will fix sentry exception https://sentry.service.dsd.io/mojds/private-beta/group/1499/
I run into that error when uploading a JSON claim to test the scheme filter so I fixed it here.

PR #1741, opened 2016-08-30
Remove the time element in the dates on the management information re…

stephenrichards:
…port


slorek:
Consider using I18n for the date presentation formatting, see http://stackoverflow.com/questions/4592235/rails-how-to-make-date-strftime-aware-of-the-default-locale (last comment)

PR #1767, opened 2016-09-08
Case Details / Summary UI update

cleargif:
https://www.pivotaltracker.com/story/show/130010317

### DIFF
--- a/app/views/shared/_certification.html.haml
+++ b/app/views/shared/_certification.html.haml
@@ -10,7 +10,7 @@
         %th{scope: "col"}
           = t('.instructed_heading')
         %td
-          = claim.certification.certified_by
+          = claim.external_user.name
 
       - if @claim.agfs?
         %tr
--- a/app/views/shared/_message_controls.html.haml
+++ b/app/views/shared/_message_controls.html.haml
@@ -12,7 +12,7 @@
     .grid-row.js-test-send-buttons
       Enter a message below and optionally attach a file (maximum 1 file per message).
       .message-column
-        = f.text_area :body, as: :text, rows: 5, placeholder: t('.message_placeholder'), label: false, class: 'form-control'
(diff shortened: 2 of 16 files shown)


app/views/shared/_certification.html.haml
```diff
@@ -10,7 +10,7 @@
         %th{scope: "col"}
           = t('.instructed_heading')
         %td
-          = claim.certification.certified_by
+          = claim.external_user.name
```
cleargif, line 13: @stephenrichards @zheileman I need a hand updating the spec for this..

app/views/shared/_summary.html.haml
```diff
@@ -90,11 +90,13 @@
...
               %tr.no-border
                 %td
-                  = expense.date
+                  = expense.date.strftime(Settings.date_format)
                 %td{scope:'row'}
                   = expense.name
                 %td.numeric
                   = expense.distance
+                  - if expense.distance?
+                    = expense.mileage_rate_id == 1 ? '(25p)' : '(45p)'
```
cleargif: @zheileman @stephenrichards 
Is this the best way..?

app/views/shared/_message_controls.html.haml
```diff
@@ -12,7 +12,7 @@
     .grid-row.js-test-send-buttons
       Enter a message below and optionally attach a file (maximum 1 file per message).
       .message-column
-        = f.text_area :body, as: :text, rows: 5, placeholder: t('.message_placeholder'), label: false, class: 'form-control'
+        = f.text_area :body, as: :text, rows: 10, placeholder: t('.message_placeholder'), label: false, class: 'form-control'
```
zheileman, line 15: I think 10 rows might be too big. It hinders the ability of the user to see the feedback of the message sent/failed, as the text area will fill half of the window.
But probably we should also rethink how we show this feedback.

PR #1781, opened 2016-09-15
Add Provider reference to Claims

stephenrichards:


config/locales/en.yml
```diff
@@ -104,6 +104,7 @@ en:
         offence: 'Offence'
         court: 'Court'
         case_number: 'Case number'
+        providers_ref: 'Provider reference number'
```
zheileman: Looks like this i18n is not being used in the code.

config/locales/en.yml
```diff
@@ -460,6 +461,7 @@ en:
           advocate_category: 'Advocate category'
           case_offence: 'Case & offence'
           case_number: 'Case number'
+          providers_ref: 'Provider reference number'
```
zheileman: Also, we could use a 'common.external_user. providers_ref' key to not duplicate the same string.

app/views/external_users/claims/case_details/_fields.html.haml
```diff
@@ -22,6 +25,12 @@
           = f.collection_select :external_user_id, @provider.advocates, :id, :name_and_number, { include_blank: '&#160;'.html_safe }, {class: 'form-control autocomplete'}
           = validation_error_message(@error_presenter, :external_user)
 
+        .form-col
+          = f.adp_text_field :providers_ref, label: t('.providers_ref'), errors: @error_presenter
+
+    - else
+      = f.adp_text_field :providers_ref, label: t('.providers_ref'), errors: @error_presenter
```
zheileman: I think we might need a .form-col here as it looks weird and too big otherwise and probably we can make it so there is no need to duplicate this line of code.

screen shot 2016-09-15 at 13 40 41

zheileman (changes requested):

config/locales/en.yml
stephenrichards: it is  - the tests complain about there being a missing key.  However, I've put them into common.external_user as suggested

app/views/external_users/claims/case_details/_fields.html.haml
stephenrichards: Agreed.  Done

zheileman (approved):

PR #1784, opened 2016-09-16
Spike remote CaseType dropdowns

stephenrichards:
this is a spike to investigate how we would populate dropdown boxes using cached
data from an api call, rather than from DB

You can see it in action by creating or editing an Advocate Claim.

The CaseTypes fetched remotely have had two asterisks appended to the name in
order to easily distinguish them from DB-populated case types.

To use it, you will have to change the api key in the settings.yml to one that exists on your db, and run a separate server answering on port 3001 (webrick is single threaded, so the api calls will time out).  Alternatively, you could run it in devunicorn enironment which will be multi threaded, but won't get nice error pages if there's an exception.

Comments welcome.

### DIFF
--- a/app/services/remotes/remote_case_type.rb
+++ b/app/services/remotes/remote_case_type.rb
@@ -1,0 +1,38 @@
+module Remotes
+  class RemoteCaseType < RemoteObject
+
+    @@case_types = nil
+
+    def self.all
+      @@case_types ||= get_case_types
+    end
+
+    def self.lgfs
+      all.select { | ct| ct.roles.include?('lgfs') }
+    end
+
+    def self.agfs
+      all.select { |ct| ct.roles.include?('agfs') }
+    end
+
+    def self.interims
+      all.select { |ct| ct.roles.include?('interim') }
+    end
+
+    def self.find(id)
+      all.detect { |ct| ct.id == id }
+    end
+
+    private
+
+    def self.get_case_types
+      case_types_hash_array = api_get('case_types')
+      ostructs = case_types_hash_array.map{ |hash| OpenStruct.new(hash) }
+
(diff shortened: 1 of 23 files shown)


app/services/remotes/remote_case_type.rb
```diff
@@ -0,0 +1,38 @@
...
+  class RemoteCaseType < RemoteObject
+
+    @@case_types = nil
+
+    def self.all
+      @@case_types ||= get_case_types
+    end
+
+    def self.lgfs
+      all.select { | ct| ct.roles.include?('lgfs') }
```
slorek, line 11: A small performance improvement could be gained by using a private class instance variable here. E.g. `@lgfs ||= ...`

slorek:
Good start, thanks Stephen!

Before we can extend this out we should consider:
- Cache invalidation (how does this work e.g. if new case types are added - maybe the API sends a TTL?)
- What happens when the API doesn't respond successfully or is unavailable
- Payload validation. I've found in the past that validating both sent and received payloads against a strict schema can simplify error detection considerably by surfacing errors early

I think adding `remote` to the name of everything might become redundant quite quickly, but whatever is best for you guys.

app/services/remotes/remote_case_type.rb
stephenrichards: Good point

stephenrichards:
The comment about adding Remote to everything: we need to be able to distinguish between CaseType returned from the DB and those returned by API, at least while it's all part of the same app, so that was the reasoning behind that.

I like the idea of adding a  TTL to the API response - we can iterate on that.

Payload validation: good idea.

Unavailable API:  Dunno what to do there - if the Claim Details service is not responding, then the whole system is down, no?  There's not much we can do, unless we go for a different approach and just cache the whole database locally.

Let's discuss.

PR #1785, opened 2016-09-19
Spike to implement a backend-agnostic caching wrapper.

zheileman:
This will allow us to cache API requests, using as unique keys the endpoints (urls)
including parameters. Future improvement involve filtering out some parameters to
make the hit/miss ratio better (i.e. api_key should be filtered out).

As an example, there is a **Redis** cache and a **Memory** cache. Swapping between them
is done via an initialization file.

A caching policy can be provided to enable the cached content to stale after N mins, for example.

**To test it, run the API smoke tests (unicorn must be running).** `rake api:smoke_test`

The dropdowns will be cached. Run it several times with Redis server started to see the caching in action.
With Memory caching, each run will wipe the cache so the hit/miss is worse.

Some `puts` left on purpose to see the caching/reading easily.

**UPDATE (20/09/2016)**
- Support for custom TTL (in seconds) - default is 900 (15 mins).
- Support for filtering out query params - api_key is filtered by default.
- Improved URL digest: downcase, param sorting and fragment removal.

Example:

``` ruby
CachedApiRequest.cache('url', ttl: 60, ignore_params: ['api_key', 'test']) do
   ... request ...
end
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem 'paperclip',              '~> 4.2.2'
 gem 'paper_trail',            '4.0.2' # version locked, https://github.com/airblade/paper_trail/issues/738
 gem 'pg',                     '~> 0.18.2'
 gem 'rails',                  '~> 4.2.7.1'
+gem 'redis',                  '~> 3.3.1'
 gem 'config',                 '~> 1.2.1' # this gem provides our Settings.xxx mechanism
 gem 'remotipart',             '~> 1.2'
 gem 'rest-client',            '~> 1.8' # needed for scheduled smoke testing plus others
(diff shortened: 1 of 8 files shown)


slorek:
Looks like a good start Jesus. I think we will definitely need to be able to cache more complex/larger data structures than can be achieved with in-process instance variables so using our ElastiCache Redis instance makes sense.

I would recommend looking at Moneta, it's a gem that abstracts different KV store backends: https://github.com/minad/moneta - I've used this before with success.

slorek:
Also consider https://github.com/mloughran/api_cache - I've also used this for a big project in the past. It integrates with Moneta.

PR #1789, opened 2016-09-22
Defining a potential Remote objects interface making use of the API cache

zheileman:
This uses as a base the previous spike https://github.com/ministryofjustice/advocate-defence-payments/pull/1784 and improves upon it, but we don't use it yet.

**Safe to merge.**

Some TODO's and comments in the code so we know what to focus on when we actually start using it for real.

OStructs replaced by ActiveModel which gives us for free some useful functionality and a common interface. It doesn't use DB.

It already wraps the API Cache.

TTL is configurable per resource as we might have some data that barely change and other data that is prone to change more easily.
There is no memoization `(@bla ||= do_stuff)` as this is provided by the API Cache.


slorek:
I wonder if wrapping API responses in `ActiveModel` object might create a lot of extra work, and adapting to change more difficult. Do we need the extra functionality over `HashWithIndifferentAccess`?

I think TTL should be controlled by the data source rather than the client.

zheileman:
We can go with TTL at the source but some of the solutions (ETag and If-Modified-Since) need to always make the request, whether with the caching at the client level will never make the request unless the cached data is stale:

http://www.apiacademy.co/how-to-http-caching-for-restful-hypermedia-apis/

Of course we can go with an approach where the server respond with a TTL and we store it and never hit the server again until it expires but then if the data at source changes (or the source TTL changes), we will not see the changes until our stored TTL expires. We can of course invalidate the cached data if we control both, the producer and the consumer, but this might not be the case.

slorek:
I don't think Etags etc. are the correct solution for us; but we could consider them later to help efficiency. We jsut need the server to send the TTL time in seconds and the client doesn't attempt to query the API endpoint again until that's expired.

PR #1799, opened 2016-09-30
Sample XML schema for injecting into CCR/CCLF

stephenrichards:


spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
+<claim>
+    <claim_details>
+        <claim_uuid>2161f858-cb16-4c77-9792-28defcbc5688</claim_uuid>
+        <!-- type can be Advocate, Litigator Final, Litigator Interim, Litigator Transfer -->
+        <type>Advocate</type>
+        <!--provider_key is the key that CCR/CCLF use to identify provider-->
+        <provider_key>2161f858-cb16-4c77-9792-28defcbc5688</provider_key>
+        <!--We can either send an id that identifies the creator to CCCD, or an id used byCCRR/CCLF -->
```
slorek, line 8: Let's remove any data we don't currently have for now, and then figure out what they need from and can send to us.

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+        <supplier_number>1B222Z</supplier_number>
+        <created_by_uuid>2161f859-cb16-4c77-9792-28defcbc5688</created_by_uuid>
+        <!--External_user_uuid identifies the advocate or litigator who is making the claim-->
+        <external_user_uuid>2161f859-cb16-4c77-9792-28defcbc5688</external_user_uuid>
```
slorek: I wonder if we would need to send them more details, as they won't have the corresponding data for these UUIDs.

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+                We also enclose an amended more detailed invoice from the Private Investigator.\r                                                                                                                                                                                   +
+                We hope the above is now sufficient to be able to pay these invoices..
+            </body>
+            <created_at>2016-09-29T13:11:43Z</created_at>
+            <!--Optional attached document - UUID refers to a document specified in the document section above-->
+            <document_uuid>345fa455-2526-47d9-9ca0-bcc4afddfa68</document_uuid>
+        </message>
+    </messages>
+
+    <determinations>
```
zheileman: This might be confusing, being a collection of 'determinations' but the elements called 'assessment' or 'redetermination'.
I wonder if we could simplify this or in the worst case create 2 different sections, one for the assessment and another for the 'determinations'?

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+        <originally_submitted_at>2016-09-29T13:11:43Z</originally_submitted_at>
+        <state>Authorised</state>
+        <!--Time claim authorised in UTC-->
+        <authorised_at>2016-09-29T13:11:43Z</authorised_at>
+    </claim_details>
+
+
+    <case_details>
+        <!--Case type is a code which identifies the case type to the CCR/CCLF systems -->
+        <case_type>XX1111</case_type>
```
slorek, line 48: We will need to agree common identifiers for the case types with LAAD.

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+            <!--'LAC1 - memo of conviction'-->
+            <!--'Committal bundle front sheets'-->
+            <!--'A copy of the indictment'-->
+            <!--'Order in respect of judicial apportionment'-->
+            <!--'Expenses invoices'-->
+            <!--'Hardship supporting evidence'-->
+            <!--'Details of previous fee advancements'-->
+            <!--'Justification for out of time claim'-->
+            <doc_type>Representation order</doc_type>
+            <doc_type>Expenses invoices</doc_type>
```
slorek, line 93: Can we supply the URLs to the uploaded evidence?

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+        </evidence_checklist>
+
+        <offence>
```
slorek, line 96: Just checking; is it possible that multiple offences can be supplied for a claim?

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+            <category>Abandonment of children under two</category>
+            <class>C: Lesser offences involving violence or damage and less serious drug offences</class>
+        </offence>
+    </case_details>
+
+
+    <defendants>
+        <!--One or more defendants-->
+        <defendant>
+            <defendant_first_name>John</defendant_first_name>
```
slorek: `defendant_*` is probably redundant here.

spikes/xml/claim.xml
slorek: Update; just noticed this is provided below, but no link from doc type here to document below. Do we have this data?

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+        </disbursement>
+    </disbursements>
+
+    <documents>
+        <!--Zero or more documents - always pdf-->
+        <document>
+            <uuid>345fa455-2526-47d9-9ca0-bcc4afddfa68</uuid>
+            <file_name>Hogdson_-_re_order.pdf</file_name>
+            <!--URI refers to location where document is stored in Amazon S3 bucket-->
+            <uri>documents/000/044/782/Hogdson_-_rep_order.pdf</uri>
```
slorek: We should supply fully-qualified URI I think.

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+        </document>
+    </documents>
+
+
+
+    <messages>
+        <!--Zero or more messages-->
+        <message>
+            <!--UUID of message sender - either caseworker or external user-->
+            <sender_uuid>345fa455-2526-47d9-9ca0-bcc4afddfa68</sender_uuid>
```
slorek, line 227: I don't know if this UUD will be useful on its own, we might need to supply further details.

spikes/xml/claim.xml
```diff
@@ -0,0 +1,237 @@
...
+            <!--Optional attached document - UUID refers to a document specified in the document section above-->
+            <document_uuid>345fa455-2526-47d9-9ca0-bcc4afddfa68</document_uuid>
+        </message>
+    </messages>
+
+    <determinations>
+        <!--One assessment must exist. There may be zero or more redeterminations-->
+        <assessment>
+            <created_at>2016-09-29T13:11:43Z</created_at>
+            <caseworker_uuid>2161f859-cb16-4c77-9792-28defcbc5688123</caseworker_uuid>
```
slorek: Same comment on user IDs above.

slorek:
- Where user IDs (caseworker/external user) are provided I don't think this is meaningful on its own to the consumer, we will probably need to supply other details such as their name at minimum, perhaps attaching relevant users to a separate root-level node referenced by their IDs.
- We have some work to do to identify if we need to establish lookup tables for courts, offence types, classes, etc. etc. remembering that CCR/CCLF may not be the sole consumers of these messages.
- We'll also need to create a schema document, which is similar to a JSON schema - https://www.w3.org/standards/xml/schema

spikes/xml/claim.xml
stephenrichards: Well this is the nub of what Phil Eades was talking about.  We need to agree a set of keys to identify each provider (and court, offence class, offence catetory etc).  What I'm trying to convey here, is that we send that key.

spikes/xml/claim.xml
stephenrichards: agreed, will change

spikes/xml/claim.xml
stephenrichards: the above section is just a checklist of attached evidence types.  We probably don't need to send that.

The actual evidence documents  are in  the documents section, which does contain URI

spikes/xml/claim.xml
stephenrichards: nope

PR #1806, opened 2016-10-05
UI: Reimplement GDS core styles

cleargif:
PT: https://www.pivotaltracker.com/story/show/131717007

Current:
mobile-current

Fixed:
mobile-fixed


zheileman:
Remember that changing layout or core styles might have an effect in the API documentation page:

http://localhost:3000/api/documentation

We need to ensure markup and styles are in sync. It is mostly static html markup.

cleargif:
@aliuk2012 

Fixed this:
screen shot 2016-10-06 at 14 48 34

PR #1879, opened 2016-11-25
validate Fee types added to a Litigator Final Claim

stephenrichards:
Litigator Final claims for fixed fee case types can only have fixed fees added to them, not graduated.  Similarly, claims for non-fixed fee case types can only have graduated fees added to them, not fixed.  

This PR adds that validation, which prevents fees of the wrong type being added through the API.

### DIFF
--- a/app/validators/fee/fixed_fee_validator.rb
+++ b/app/validators/fee/fixed_fee_validator.rb
@@ -11,6 +11,16 @@ class Fee::FixedFeeValidator < Fee::BaseFeeValidator
 
   private
 
+  def validate_claim
+    super
+    if @record.claim
+      if @record.claim.final?
+        add_error(:claim, 'Graduated fee invalid on fixed fee case types') unless @record.claim.case_type.is_fixed_fee?
+      end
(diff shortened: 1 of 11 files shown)


app/validators/fee/fixed_fee_validator.rb
```diff
@@ -11,6 +11,16 @@ def self.fields
 
   private
 
+  def validate_claim
+    super
+    if @record.claim
```
slorek, line 16: Would be nicer as `if @record.claim && @record.claim.final? && !@record.claim.case_type.is_fixed_fee?` to reduce nesting and complexity, I think.

app/validators/fee/graduated_fee_validator.rb
```diff
@@ -7,6 +7,17 @@ def self.fields
     ] + super
   end
 
+  private
+
+  def validate_claim
+    super
+    if @record.claim
```
slorek, line 14: Same as above.

slorek (approved):

PR #1890, opened 2016-12-09
Remove n+1 query from claims

stephenrichards:
There was an after_initialize callback on Claim::BaseClaim that instantiated an
assessment if not already there.  This meant that when finding a collection of
claims, an query was made to get the assessment for each claim in the collection.

This PR replace that with an after_save hook to create a blank assessment when
the claim is created, and rewrite the assessment factory to cater for this.

### DIFF
--- a/app/models/determination.rb
+++ b/app/models/determination.rb
@@ -57,6 +57,17 @@ class Determination < ActiveRecord::Base
     !blank?
   end
 
+  def to_s
+    "  id:            #{id}\n" +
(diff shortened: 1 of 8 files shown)


slorek (approved):

app/models/determination.rb
```diff
@@ -57,6 +57,17 @@ def present?
     !blank?
   end
 
+  def to_s
```
slorek, line 60: If this is debug code, consider removing it.

PR #1915, opened 2016-12-23
Rake task to recalculate VAT on all expenses

stephenrichards:


stephenrichards:
There are AGFS claims where VAT hasn't been applied to expenses.  We need to:
a) confirm they should be when the advocate is VAT registered
b)decide which claim states to examine and update

PR #1935, opened 2017-01-04
Add Fee Scheme Manager to determine which fee scheme version to use

stephenrichards:
The FeeSchemeManager class provides a central point in which to decide which fee scheme to use.

Whilst the new fee scheme is under development, on gamma, it will only ever return the old fee scheme, :afgs_v9 or :lgfs_v6, otherwise it takes the earliest reporder date into account

### DIFF
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -285,6 +285,10 @@ module Claim
       end.first
     end
 
+    def earliest_representation_order_date
+      earliest_representation_order.try(:representation_order_date)
+    end
+
     def evidence_doc_types
       DocType.find_by_ids(evidence_checklist_ids)
     end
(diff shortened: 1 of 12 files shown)


app/models/claim/base_claim.rb
```diff
@@ -285,6 +285,10 @@ def earliest_representation_order
       end.first
     end
 
+    def earliest_representation_order_date
+      earliest_representation_order.try(:representation_order_date)
```
jsugarman, line 289: do you know in ruby 2.3.1+ you can use &.method as an alternative to .try(:method) 
e.g.
```
earliest_representation_order&.representation_order_date
```

jsugarman:
👍  LGTM..and brings back memories.

app/models/claim/base_claim.rb
jsugarman: since 2.3.0+ actually
http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

PR #1937, opened 2017-01-06
Unarchive to deal with previous versions with nil vat totals

stephenrichards:
The unarchive controller action does not unarchive by changing the state of the claim, but by reinstating the previous version of the record from the paper_trail gem's versions table.  This can lead to claim records being reinstated with nils in the fees_vat, expenses_vat, disbursement_vat fields.  This PR ensures that those fields are set to 0.0 if nil.


slorek:
Can this not be achieved by changing the DB schema to not allow `nil` values on those columns and default to `0.0` instead?

slorek (approved):

PR #1940, opened 2017-01-06
Enable case sensitivity on email addresses

stephenrichards:


slorek (commented):
Need to understand what issue this is fixing. This could result in some extra frustration for users

PR #1941, opened 2017-01-09
Force email confirmation field to lower case

stephenrichards:
Devise will automatically downcase the email field.  Unless we do the
same for the email  confirmation field, then the the user creation will
fail because the email does not equal the email confirmation.

The devise configuration is set to not check case sensistivity on email,
so the user will be able to log in with any combination of upper or lower case
in theemail.

### DIFF
--- a/app/controllers/super_admins/external_users_controller.rb
+++ b/app/controllers/super_admins/external_users_controller.rb
@@ -17,6 +17,8 @@ class SuperAdmins::ExternalUsersController < ApplicationController
   end
 
   def create
+    # downcase email_confirmation - devise will downcase the email
+    params[:external_user][:user_attributes][:email_confirmation].downcase!
     @external_user = ExternalUser.new(params_with_temporary_password.merge(provider: @provider))
 
     if @external_user.save
--- a/spec/controllers/super_admins/external_users_controller_spec.rb
+++ b/spec/controllers/super_admins/external_users_controller_spec.rb
@@ -82,6 +82,7 @@ RSpec.describe SuperAdmins::ExternalUsersController, type: :controller do
       post :create,
             provider_id: provider,
             external_user: { user_attributes: {  email: 'foo@foobar.com',
+                                                 email_confirmation: 'foo@foobar.com',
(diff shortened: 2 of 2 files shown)


app/controllers/super_admins/external_users_controller.rb
```diff
@@ -17,6 +17,8 @@ def new
   end
 
   def create
+    # downcase email_confirmation - devise will downcase the email
+    params[:external_user][:user_attributes][:email_confirmation].downcase!
```
slorek, line 21: Perhaps it would be worth checking the Devise runtime config at this point before downcasing.

spec/controllers/super_admins/external_users_controller_spec.rb
```diff
@@ -97,6 +98,11 @@ def post_to_create_external_user_action(options={})
         post_to_create_external_user_action
         expect(response).to redirect_to(super_admins_provider_external_user_path(provider, ExternalUser.last))
       end
+
+      it 'downcases the email address'do
```
slorek: I don't think we should test this. If you changed the Devise config it would still pass but users couldn't log in. I think the more useful test would be to check that login can be successful given the current runtime config. That way, if the config changes, it should always pass, and you're testing what is intended.

slorek (dismissed):

app/controllers/super_admins/external_users_controller.rb
stephenrichards: Why would we do that?  The configuration option to specify case insensitivity for this field is hard coded in config/devise.rb:44 - I don't think it can be changed mid-stream.

spec/controllers/super_admins/external_users_controller_spec.rb
stephenrichards: OK

app/controllers/super_admins/external_users_controller.rb
slorek: You're coupling code implementation to configuration. It's not going to be changed during runtime, but can be changed nevertheless. With this change, it will work whichever way it is configured.

PR #1950, opened 2017-01-12
Allow database dumps to be taken of api-sandbox

stephenrichards:


csutter:
Decided to close this – will pick it up again if we ever need it.

PR #1952, opened 2017-04-12
Clean up stats report

stephenrichards:
If a management information report is in the middle of being generated when the server is restarted, the record will be left in the database with a status of 'started'.  This currently prevents subsequent management information generators from running (it checks to see if other reports are in the process of being generated, and immediately exits if that is the case).

This PR cleans up old reports (older than 2 months), and stalled reports that were started more than 2  hours ago and never finished.


jsugarman:
👍  LGTM

PR #1955, opened 2017-04-26
Switch to dotenv

colinbruce:
Switch to using the dotenv gem

This allows users to store local env files and prevent bloating their environment store


csutter (approved):
:+1:

PR #1956, opened 2017-04-26
Geckoboard stats update

colinbruce:
Passing the state to be returned was omitted from the refactoring

This meant that gecko board was showing a blank report


csutter (approved):
:+1:

PR #1958, opened 2017-04-28
Update the dotenv instructions

colinbruce:
The `export` was only for the original instructions and actually broke the dotenv implentation


csutter (approved):
:+1:

PR #1959, opened 2017-05-02
Gitignore ENV files

csutter:
We've added `.env.development` and `.env.test` files for ENV management
in #1955, these should be gitignored so they don't accidentally get
checked in.


colinbruce (approved):
👍

PR #1961, opened 2017-05-02
MI stats limiter

colinbruce:
To reduce time and memory usage, limit the results to the last 6 months
pivotal ticket
[Finishes #144363179]

### DIFF
--- a/app/presenters/claim_csv_presenter.rb
+++ b/app/presenters/claim_csv_presenter.rb
@@ -11,7 +11,7 @@ class ClaimCsvPresenter < BasePresenter
   end
 
   def sorted_and_filtered_state_transitions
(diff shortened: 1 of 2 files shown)


app/presenters/claim_csv_presenter.rb
```diff
@@ -11,7 +11,10 @@ def journeys
   end
 
   def sorted_and_filtered_state_transitions
-    claim_state_transitions.sort.reject { |transition| (transition.to == 'draft' || transition.to == 'archived_pending_delete') }
+    claim_state_transitions.sort.reject do |transition|
+      %w(draft archived_pending_delete).include?(transition.to) ||
+        transition.created_at < Time.now - 6.months
```
csutter, line 16: When we move this out, it should probably be configurable somewhere – but for now, fair enough.

csutter (approved):
:+1:

PR #1964, opened 2017-05-03
Fix intermittently broken Remote::CaseWorker spec

csutter:
Depending on spec run order this spec would fail because it doesn't
absolutely reference the class it's setting expectations on, and there
is another class of the same name in a different module namespace.


colinbruce (approved):
👍

PR #1965, opened 2017-05-04
Add cycling as expense type 🚴

csutter:
Adds "Bike travel" as an additional expense type, with separate mileage
rates.

Requires re-seeding of expense types: `rake data:migrate:all`


csutter:
Requires FE work to show/hide mileage rates as appropriate. Currently it shows all three mileage rates (20p, 25p, 45p) – it should only show 25/45p for car travel, and only 20p for bike travel.

**Not ready to merge until that is done.**

PR #1967, opened 2017-05-10
Feat: Removal of Judicial Apportionment

cleargif:
PT:
https://www.pivotaltracker.com/story/show/144433507

Why are we doing this:
Judicial Apportionment is no longer a required metric to be collected

What does this PR do:
We are removing the view layer inputs and outputs
All BE methods / models / presenters  are unchanged


colinbruce (approved):
👍

PR #1968, opened 2017-05-11
PPE field for LGFS transfer fee

cleargif:
PT:
https://www.pivotaltracker.com/story/show/144704837

Why are we doing this:
On litigator transfer claims there is no input for PPE

What does this PR do:
Refactor to the view to expose the input for the model

## before
screen shot 2017-05-11 at 11 51 26

## after
screen shot 2017-05-11 at 11 51 06


colinbruce (approved):
:+1:

PR #1969, opened 2017-05-15
Change 25k in claim value filter to 20k

cleargif:
PT:
https://www.pivotaltracker.com/story/show/145430905

Why are we doing this:
CCR has not yet been updated to reflect this change. Reverting back until we sync with CCR

What this PR does:
Update the value bands used for allocation filters


csutter (approved):
:+1:

PR #1971, opened 2017-05-17
PPE review

cleargif:
PT:
https://www.pivotaltracker.com/story/show/145589211

Why we doing this:
PPE is collected and shown in a few different ways depending on AGFS vs LGFS and other metrics like Fee type.

What this PR does:
Updating the summary and overview pages with the PPE labels and columns


colinbruce (approved):
👍

PR #1972, opened 2017-05-19
Feat: Allocation bands agfs / lgfs specific

cleargif:
PT:
https://www.pivotaltracker.com/story/show/145430905

Why are we doing this:
Changes in CCR and CCLF are cuasing workarounds in CCCD Allocation.

What does this PR do:
Adding in AGFS and LGFS specific band filters to allow for easier allocation


cleargif:
CLOSING AS THIS DOES NOT SOLVE THE PROBLEM

Band ids are assigned at the time a claim is created. Results in allocation is not being filtered based on the band amount - but rather on the ID of the band at the time the claim was made..

#wat

PR #1973, opened 2017-05-22
GA Tracking housekeeping

cleargif:
PT:
https://www.pivotaltracker.com/story/show/145853977

Why are we doing this?
GA is implemented differently in pre-production environments. It uses Google Tag Manager. In order to review the analytics strategy and test it we want a consistent implementation on ALL pre-prod env's

What this PR does?
- [x] Remove the GTM implementation
- [x] Enable GA on all pre-prod env's
- [x] Fix the double page view call

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -71,7 +71,11 @@ module ApplicationHelper
   end
 
   def ga_outlet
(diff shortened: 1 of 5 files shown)


csutter (approved):
:+1:

PR #1975, opened 2017-05-25
Reverse the RAG status generation sequence

colinbruce:
# What
The geckoboard RAG output was displayed incorrectly
image 
This was due to the ordering of the generated data, successful was returned first so was displayed as Red, rather than green.

# How
This PR reverses (and tests) the sequence returns Rejected, partial and Authorised; so Red-Amber-Green


csutter (approved):
:shipit:

PR #1978, opened 2017-05-31
Litigant supplier number validation on create only

colinbruce:
Why are we doing this:
> Allocation of cases was being blocked in certain circumstances (i.e. a supplier had been removed from a provider) because `supplier_number` was validated on update

What have we done: 
> Apply a separate validator ruleset and apply it on create of litigator_claims.  For other claim types, added `supplier_number` to the validation rulesets

Bonus
> This also fixes the deletion of records that are `archive_pending_delete`  this has been blocked for a number of months, testing on a local, anonymised, copy of the DB, this is no longer the case

### DIFF
--- a/app/validators/claim/litigator_supplier_number_validations.rb
+++ b/app/validators/claim/litigator_supplier_number_validations.rb
@@ -1,0 +1,38 @@
+module Claim
+  module LitigatorSupplierNumberValidations
+
+    def self.included(base)
+      base.class_eval do
+        def self.first_step_common_validations
+          [
+            :supplier_number
+          ]
+        end
+      end
+    end
+
+    private
+
+    def validate_supplier_number
+      validate_presence(:supplier_number, 'blank')
+
+      return unless @record.supplier_number.present?
+
+      validate_pattern(:supplier_number, supplier_number_regex, 'invalid')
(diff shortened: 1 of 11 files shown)


app/validators/claim/litigator_supplier_number_validations.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      validate_pattern(:supplier_number, supplier_number_regex, 'invalid')
+      validate_inclusion(:supplier_number, provider_supplier_numbers, 'unknown') unless @record.errors.key?(:supplier_number)
+    end
+
+
+    # local helpers
+    # ---------------------------
+
+
```
csutter: Whitespace 😱

app/validators/claim/litigator_supplier_number_validator.rb
```diff
@@ -0,0 +1,9 @@
+class Claim::LitigatorSupplierNumberValidator < Claim::BaseClaimValidator
+  include Claim::LitigatorSupplierNumberValidations
+
+  def self.fields_for_steps
+    [
+        [].unshift(first_step_common_validations)
```
csutter:
```ruby
[first_step_common_validations]
```

csutter (dismissed):
Let's fix the (already existing, admittedly) little issues in this and then :shipit: !

csutter (approved):
:+1:

PR #1979, opened 2017-06-02
Improve handling of provider LGFS supplier numbers

csutter:
At the moment, providers can delete any of their LGFS supplier numbers
which breaks the claim state machine if they still have claims that
haven't been archived pending delete.

- Supplier numbers should not be editable once added
- Supplier numbers should not be deletable if they still have claims
  ongoing


colinbruce (commented):
👍

colinbruce (approved):
👍

PR #1980, opened 2017-06-05
Improve user satisfaction reporting

colinbruce:
# Why?
A separate aggregation program used to run and provide an end point for the [gov.uk performance platform](https://www.gov.uk/performance/crown-court-defence-claims) to access data.  This broke in October 2016 and no-one was available to correct it.

# Fixes
This PR makes the application submit an event to google analytics that the PP team can use to aggregate the data instead.  

# Still to do
A separate event will need tobe undertaken to provide them with static data from Oct 2016 to date.

### DIFF
--- a/lib/google_analytics/api.rb
+++ b/lib/google_analytics/api.rb
@@ -1,0 +1,28 @@
+require 'rest_client'
+
+module GoogleAnalytics
+  class Api
+    def self.event(category, action, client_id = '555')
+      return unless tracker_id.present?
+      params = { v: version, tid: tracker_id, cid: client_id, t: 'event', ec: category, ea: action }
+      begin
+        RestClient.get(endpoint, params: params, timeout: 4, open_timeout: 4)
+        return true
+      rescue RestClient::Exception
+        return false
+      end
+    end
+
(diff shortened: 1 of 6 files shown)


lib/google_analytics/api.rb
```diff
@@ -0,0 +1,28 @@
+require 'rest_client'
+
+module GoogleAnalytics
+  class Api
+    def self.event(category, action, client_id = '555')
```
csutter: 555? 🤔

colinbruce: Ah, I knew there was a note I should add... https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide#overview
client_id has a default value of 555 and all examples I found used it, so I followed the herd! :)

It's overwritten by the form anyway, but it's there as a fallback if it fails

csutter: Let's chuck it into a constant (`FALLBACK_CLIENT_ID`?) with a quick comment and then get this out :)

csutter (dismissed):
:+1:

csutter (approved):
:+1:

PR #1981, opened 2017-06-06
GA Custom Dimensions

cleargif:
PT:
https://www.pivotaltracker.com/story/show/146690531

Why are we doing this:
We want to track a few extras

What this PR does:
Custom Dimensions for: Offence Category and Case type


csutter (approved):
:shipit:

cleargif:
All good ideas end up in heaven.

PR #1985, opened 2017-06-20
Rubocop fixes

colinbruce:
Over time, this will allow us to incorporate rubocop checks on merge, but for now this aims to reduce the noise by addressing some of the issues flagged

### DIFF
--- a/spikes/stats/collector/draft_to_submitted_time.rb
+++ b/spikes/stats/collector/draft_to_submitted_time.rb
@@ -25,7 +25,7 @@ module Stats
 
       def calculate_time_draft_to_submitted(claim_id)
         claim = Claim::BaseClaim.active.find claim_id
-        return unless claim.state != 'draft'
+        return if claim.draft?
         period_in_secs = first_submitted_at(claim) - claim.created_at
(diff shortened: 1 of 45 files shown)


spikes/stats/collector/draft_to_submitted_time.rb
```diff
@@ -25,11 +25,10 @@ def first_submitted_at(claim)
 
       def calculate_time_draft_to_submitted(claim_id)
         claim = Claim::BaseClaim.active.find claim_id
-        if claim.state != 'draft'
-          period_in_secs = first_submitted_at(claim) - claim.created_at
-          period_in_days = (period_in_secs / SECONDS_IN_DAY).to_i
-          update_totals(period_in_days)
-        end
+        return unless claim.state != 'draft'
```
csutter:
```ruby
return if claim.state == draft
```

... or even (this should work?)

```ruby
return if claim.draft?
```

csutter (dismissed):
One tiny point, otherwise 🎉

csutter (approved):
:+1:

PR #1986, opened 2017-06-20
Update GA satisfaction events

colinbruce:
Add a label containing the satisfaction text to the GA event handler

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -12,7 +12,7 @@ class FeedbackController < ApplicationController
   def create
     @feedback = Feedback.new(merged_feedback_params)
     if @feedback.save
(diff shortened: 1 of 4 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -12,7 +12,7 @@ def new
   def create
     @feedback = Feedback.new(merged_feedback_params)
     if @feedback.save
-      GoogleAnalytics::Api.delay.event('satisfaction', merged_feedback_params[:rating], params[:ga_client_id])
+      submit_feedback_google_event if @feedback.is?'feedback'
```
csutter: What's Rubocop saying about that? 🤖

csutter (dismissed):
:+1:

app/controllers/feedback_controller.rb
colinbruce: image
I checked first! :)

csutter (approved):
:+1:

PR #1988, opened 2017-06-26
Allow allocation regardless of validation

colinbruce:
Users were unable to allocate claims where linked tables had been invalidated since creation, e.g. the creator of a claim had changed roles from `litigator` to `advocate`.

This removes validation when allocating cases.

It also removes the validation when the timed transitioner runs overnight to archive the old records


csutter (approved):
Looks good. Here's to hoping we can tear that stuff out entirely soon! 👌

PR #1989, opened 2017-06-28
Update allocation to pass correct number

colinbruce:
Previously it always passed the page_size limit, regardless of the quantity to allocate

Personally, there is far to much logic in this controller and it should be extracted to an `allocation_service` it would allow for much clearer execution and testing... but, for now, so that case workers can actually use it...


csutter (approved):
🤦‍♂️

PR #1990, opened 2017-06-30
State machine validation

colinbruce:
This PR (hopefully) addresses the issues that users (mainly case-workers) have been having with transitioning state when validations failed.

This leaves validation in place while external users are creating the claim `state: nil > draft > submitted`

It allows state_machine to set two values for a `claims.disable_for_state_transition`; `:all` and `:leave_amount`

When set to 
`:all` it completely disables the validation of a claim;
`:leave_amount` it leaves validation in place for a claims assessment account.


csutter (approved):
:+1:

PR #1991, opened 2017-07-10
Validation state machine updates

colinbruce:
There were some missing states in the constant used to turn off validation in state_machine transitions.

This creates a new constant that includes all of the states that are post external-user.

### DIFF
--- a/spec/models/claims/state_machine_spec.rb
+++ b/spec/models/claims/state_machine_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe Claims::StateMachine, type: :model do
   describe 'NON_VALIDATION_STATES' do
     subject { Claims::StateMachine::NON_VALIDATION_STATES }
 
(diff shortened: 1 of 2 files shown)


spec/models/claims/state_machine_spec.rb
```diff
@@ -24,6 +24,13 @@
     it('exist') { expect(Claim::BaseClaim.active.state_machine.states.map(&:name).sort).to eq(states.sort) }
   end
 
+  describe 'NON_VALIDATION_STATES' do
+    subject { Claims::StateMachine::NON_VALIDATION_STATES }
+
+    let(:states) { Claim::BaseClaim.active.state_machine.states.map(&:name).sort}
```
jsugarman: space between sort and }

jsugarman (dismissed):
I trust the business logic is sound. Other than one rubocopesque comment LGTM 👍

jsugarman (approved):
LGTM 👍

PR #1992, opened 2017-07-11
add rb-readline to gemfile

jsugarman:
byebug has a OS library dependency that is, nonentheless, not a
gem dependency. This results in rails server or console start raising
an error
```
Sorry, you can't use byebug without Readline. To solve this, you need to
rebuild Ruby with Readline support. If using Ubuntu, try `sudo apt-get
install libreadline-dev` and then reinstall your Ruby.
```

Easiest fix to prevent the error is to add rb-readline as dev/test
dependency in the Gemfile. See [byebug issue](https://github.com/deivid-rodriguez/byebug/issues/289)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,6 +88,7 @@ group :development, :devunicorn, :test do
   gem 'guard-rspec'
   gem 'net-ssh'
   gem 'net-scp'
+  gem 'rb-readline'
 end
 
 group :test do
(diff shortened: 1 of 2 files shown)


colinbruce (dismissed):
👍

colinbruce (commented):
:+1:

colinbruce (approved):
👍

PR #1995, opened 2017-07-11
prevent rspec warnings for shared example group

jsugarman:
Warning is distracting and indicative of unnecessary loading. 
Was being displayed because the shared example group was 
named *_spec.rb and was therefore being loaded for testing 
and required by other files.

see [rspec-core issue](https://github.com/rspec/rspec-core/issues/2277)


colinbruce (approved):
👍  Nice work!

PR #2000, opened 2017-07-13
Add initial JSON Schema for full claim

csutter:
This adds a basic JSON Schema for CCCD claims as exposed by the
`FullClaim` entity.

This doesn't fix any of the underlying issues with the JSON, just
specifies what already exists.


jsugarman (approved):
👍

PR #2003, opened 2017-07-18
remove rb-readline gem

jsugarman:
The rb-readline gem fixes a library dependency on OSX but introduces
a key mapping bug when using the rails console on ubuntu (i.e. on the servers).

Instead local OSX development can be fixed using one of the
options here:

https://github.com/guard/guard/wiki/Add-Readline-support-to-Ruby-on-Mac-OS-X


colinbruce (approved):
:+1: thanks for fixing

PR #2004, opened 2017-07-18
add colored text output to travis console

jsugarman:
makes travis output easier to read.

adds colour output and coloured spec output to the travis console to make it
easier to read.


colinbruce (approved):
:+1:

PR #2005, opened 2017-07-19
Further anonymize data dump

jsugarman:
**What?**
ensure data is anonymous

**Why?**
other fields could potentially contain sensitive information and with a 
view to putting anonymous data on demo for testing against CCR.


colinbruce (approved):
👍  Looking good!

PR #2007, opened 2017-07-21
use single ssh session and refactor dump script

jsugarman:
Two SSH sessions (one for running dump and another for scp'ing)
is unnecessary overhead and causes problems if you try to
use first session after second has started.


colinbruce (approved):
👍  Looking good

PR #2008, opened 2017-07-24
Add a search endpoint to the API

colinbruce:
## Why
A new, refined end point was needed for the prototype allocation

### DIFF
--- a/app/interfaces/api/helpers/search_result_helpers.rb
+++ b/app/interfaces/api/helpers/search_result_helpers.rb
@@ -1,0 +1,43 @@
+module SearchResultHelpers
+  private
+
+  def fee_is_interim_type
+    object.fees.map do |fee|
+      [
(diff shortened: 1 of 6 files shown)


app/interfaces/api/helpers/search_result_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+      [
+        fee.quantity.between?(1, 50),
+        fee.fee_type.description == 'Guilty plea',
+        fee.fee_type.is_a?(::Fee::GraduatedFeeType)
+      ].all?
+    end.any?
+  end
+
+  def interim_claim?
+    object.is_a?(::Claim::InterimClaim)
```
jsugarman, line 32: can do just object.interim? for this kind of logic - on second thoughts this is probably clearer regardless

jsugarman (approved):
👍

PR #2015, opened 2017-07-25
fix missing sample size

jsugarman:
Previous PR inadvertently removed a sample size from
the redcentric:ccr_claims_json rake task.


colinbruce (approved):
👍

PR #2016, opened 2017-07-27
add store of and retrieval method for CCR bill scenario

jsugarman:
**What**
adds a mapping for CCR bill scenario IDs
to CCCD case types and modifies endpoint to supply:

* bill scenario
* number of cases
* number of prosecution witnesses

**Why**
CCR needs to know what bill scenario to apply
to determine the relevant validation and calculations
to apply. The number of cases and prosecution witnesses
is additional information that is used to calculate the Advocate
Fee in CCR.

### DIFF
--- a/app/interfaces/api/entities/ccr_claim.rb
+++ b/app/interfaces/api/entities/ccr_claim.rb
@@ -89,7 +89,6 @@ module API
 
       def scenario
         object.case_type.bill_scenario
-        # 'AS000004' # Hardcoded for "trial" case type
       end
 
       def estimated_trial_length_or_one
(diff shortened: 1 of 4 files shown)


app/interfaces/api/entities/ccr_claim.rb
```diff
@@ -104,16 +114,45 @@ def offence_code_id
         ('A'..'K').zip(501..511).to_h[offence_class_code]
       end
 
-      def wrapped_bill
-        [{
+      def number_of_witnesses
+        object.fees.find_by(fee_type_id: 10)&.quantity&.to_i || 0
```
colinbruce: The `object.fees.find_by(fee_type_id: `xx`)&.quantity&.to_i` is repeated multiple times, this could be extracted to a method for re-use.  

Also, as discussed, using constants for the fee_type_ids might make reading the code easier e.g. 
```ruby
WITNESSES = 10
CASES = 9
def count_of(fee_type)
  object.fees.find_by(fee_type_id: fee_type)&.quantity&.to_i
end
```

```ruby
count_of(CASES)
count_of(WITNESSES)
```

colinbruce (approved):
👍

PR #2017, opened 2017-07-27
Refactor search api endpoint

colinbruce:
In an effort to reduce the significant load times from the Rails loading, I've take n the following steps

* Switch to raw SQL 
* re-merge the helper back into the search_result object (this may need splitting out again in the future) but for now... it works


app/interfaces/api/entities/search_result.rb
```diff
@@ -37,81 +35,104 @@ class SearchResult < BaseEntity
 
       private
 
+      def id
```
jsugarman: i'd be tempted to put in some method missing other forwarding/delegation here for  these repeated object calls:
e.g. 
```ruby
   def method_missing(method_name, *args, &block)
      if method_name.in?(%w(id uuid scheme scheme_type state court_name case_type total))
        object[method_name]
      else
        super
      end
    end

    def respond_to_missing?(method_name, include_private = false)
      object.respond_to?(method_name.to_sym, include_private) || super
    end
````

**Review by jsugarman — DISMISSED** [review]
👍  aside from the possible refactoring for repeated object[''attribute'] calls and the unallocated_sql being a possible candidate for converting to a query object pattern (although it does not, currently, need to be reused?! I am assuming)

File: app/interfaces/api/entities/search_result.rb (hunk 1 above)
Comment (jsugarman) [reply app/interfaces/api/entities/search_result.rb]: respond_to_missing? would need to be something like below actually (not tested)

```ruby 
def respond_to_missing?(method_name, include_private = false)
  object[method_name] || super
end
```

app/interfaces/api/v2/search.rb
```diff
@@ -19,7 +19,8 @@ class Search < Grape::API
         helpers do
           def claims
             built_sql = unallocated_sql.gsub(/REPLACE_MATCHER/, scheme.eql?('agfs') ? ' = ' : ' != ')
-            ActiveRecord::Base.connection.execute(built_sql).to_a
+            result = ActiveRecord::Base.connection.execute(built_sql).to_a
+            JSON.parse(result.to_json, object_class: OpenStruct)
```
jsugarman: excellent 👍

jsugarman (dismissed):
👍

jsugarman (approved):

PR #2018, opened 2017-07-27
add uuid to case types with consistent values across environments

jsugarman:
In order for CCR to validate CCCD claims it 
must determine what "billing scenario"  applies
to it. Case types and billing scenarios are virtually
synonymous!

The intention is that CCR will store a translation/mapping
between the case type's uuid and the billing scenario.

The full claim api endpoint will now need to return the UUID
for consumption by CCR.


colinbruce (approved):
👍 
As discussed, I thought your YAML version made more sense, but this will work too

PR #2019, opened 2017-07-28
add uuid for hearing subsequent to sentence

jsugarman:
another uuid is needed for case type migration.  hearing subsequent to sentence will not otherwise get one.


colinbruce (approved):
👍

PR #2020, opened 2017-07-28
Add debug features

colinbruce:
# Add fuubar gem
Add a nicer formatter (fuubar) for the rspec tests this way it will give you the error messages as they occur, but leave the tests running.
```bash
$ rspec --format Fuubar --color
```
 or change your `.rspec` file

# Implemented rubocop_todo
I've commented out the LineLength cop in rubocop so that the `ignored` value in the todo file is used, when the problems are resolved, we can restore the 120 max again


jsugarman (approved):
👍 sweet

PR #2023, opened 2017-07-28
add progress bar to db dump file upload/download

jsugarman:
**What** 
adds a progress bar to the upload and download of database files 
via the db dump and restore (ETL) scripts

**Why**
The large files can take some time to transfer and the previous method of displaying progress was 
unreliable...and not as pred'y


colinbruce (approved):
nice, big 👍  for visual indication of progress

PR #2024, opened 2017-07-31
Implement allocate API endpoint

colinbruce:
+ add endpoint to API root
+ create tests for the handling of the POST
+ implement the endpoint

### DIFF
--- a/app/interfaces/api/v2/case_workers/allocate.rb
+++ b/app/interfaces/api/v2/case_workers/allocate.rb
@@ -1,0 +1,39 @@
+module API
+  module V2
+    module CaseWorkers
+      class Allocate < Grape::API
+        before_validation do
+          authenticate_user_is?('CaseWorker')
+        end
+
+        namespace :case_workers do
+          params do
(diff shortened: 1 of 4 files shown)


slorek:
🤔 is not /claims/allocate-to-caseworker ? 👹

colinbruce:
Who else would we allocate to @slorek ?
**Update:** I see what you mean now
But we are `POST`ing claim ids to `api/case_workers/allocate`
Why not `api/case_workers/allocate-claims`?

slorek:
Just my pedantic side - the action of allocating is a modification against a claim resource, you're not modifying the case worker. And it should be a PUT/PATCH rather than POST ;) The allocate-to-caseworker makes it more obvious what it means, as otherwise you assume a bit of domain knowledge.

That said, you could do /case_worker/:id/allocate-claims. That would make sense. Apart from the mix of underscores and hyphens ;)

slorek:
Another option - assuming de-allocation and listing allocations is useful, might be to expose `allocations` as a resource, allowing CRUD actions against that. E.g. POST to `/allocations` creates a new allocation, DELETE `/allocations/:id` de-allocates.

PR #2025, opened 2017-08-01
amend full claim endpoint to return case type uuid

jsugarman:
**What**
 returns case type uuid in full claim json 

**Why**
Rather than storing bill scenario logic in CCCD  it 
is expected that CCR will store a mapping/translation
of CCCD case type UUIDs against CCR bill scenario keys/ids.


colinbruce (approved):
👍

PR #2026, opened 2017-08-02
add optional tidy up of remote dumpfile

jsugarman:
leaving the remote dump file in place can cause subsequent
dump files from container to not overwrite the users version. However,
leaving it in place means you could download existing
dump files to save time, so deletion is given as an option.


colinbruce (approved):
👍

PR #2027, opened 2017-08-02
Improve POSTing to root handling

colinbruce:
**Why** 
Currently a user is randomly POSTing to the root url
**What** 
Add a specific route to prevent the server throwing 500 errors.
This will prevent the platforms team being alerted to multi 500 errors out of hours.
It also includes logging of the api_key used so that we can identify the user and discuss how to prevent them posting to the wrong endpoint.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -3,6 +3,7 @@ class ErrorsController < ApplicationController
   protect_from_forgery except: [:not_endpoint, :not_found, :internal_server_error]
 
   def not_endpoint
+    logger.debug("Data POSTed to root with API key: #{not_endpoint_params[:api_key]}") if params.present?
     render status: 422, text: 'Not a valid api endpoint'
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/errors_controller.rb
```diff
@@ -1,6 +1,11 @@
 class ErrorsController < ApplicationController
-  skip_load_and_authorize_resource only: [:not_found, :internal_server_error, :dummy_exception]
-  protect_from_forgery except: [:not_found, :internal_server_error]
+  skip_load_and_authorize_resource only: [:not_endpoint, :not_found, :internal_server_error, :dummy_exception]
+  protect_from_forgery except: [:not_endpoint, :not_found, :internal_server_error]
+
+  def not_endpoint
+    logger.debug("Data POSTed to root with API key: #{not_endpoint_params[:api_key]}") if params.present?
+    render status: 422, text: 'Not a valid api endpoint'
```
jsugarman: be tempted to go with a 403 forbidden but otherwise 👍

jsugarman (dismissed):

PR #2028, opened 2017-08-02
Add CCR claim endpoint entities and tests

jsugarman:
**What** 
- add CCR required details of each defendant to api endpoint
- add tests for added entities
- add JSON schema for current api endpoint (for iteration)
- add test to validate JSON against schema

**Why**
This is to form the basis of iteration on
the existing CCR Claim endpoint to the point
where it can become a full claim end point.


colinbruce (commented):
👍

colinbruce (approved):
👍

PR #2029, opened 2017-08-03
Address vulnerabilities

colinbruce:
gemnasium has highlighted some vulnerabilities in the code base
Therefore we had to bump the ruby version, and therefore rubocop.

There are still `Fixnum` deprecation warnings from ShouldaMatchers when running tests, but this is a known issue and will be fixed in a future update (https://github.com/thoughtbot/shoulda-matchers/milestone/13)


jsugarman (approved):
I would get latest master on prod before merging this. Otherwise 👍

PR #2031, opened 2017-08-03
enable json schema publishing

jsugarman:
**What**
enable viewing of existing JSON schemas 
*e.g. root/json_schemas/ccr_schema*

**Why**
To assist conversations and represent latest state of various schemas
but CCR related in particular.


colinbruce (approved):
:+1:

PR #2034, opened 2017-08-07
Fix vulnerabilities

colinbruce:
## Why
Brakeman warned of a severe vulnerability in the geckoboard controller
Because of the inheritance of `ActionController` and `ActionController::Base` the protect from forgery flag was not turned on

## Fix
Turning on the `protect_from_forgery` address should close that warning

Bonus - Fixed a couple of minor Rubocop warnings


jsugarman (approved):
👍

PR #2037, opened 2017-08-08
Brakeman update

colinbruce:
Brakeman alerted on a small number of changes.

This PR reduces the brakeman alert to zero, this would allow us to turn on brakeman tests in CI

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -34,17 +34,19 @@ class MessagesController < ApplicationController
   end
 
   def download_attachment
-    @message = Message.find(params[:id])
 
-    raise 'No attachment present on this message' if @message.attachment.blank?
+    raise 'No attachment present on this message' if message.attachment.blank?
 
-    send_file Paperclip.io_adapters.for(@message.attachment).path, type:        @message.attachment_content_type,
-                                                                   filename:    @message.attachment_file_name,
-                                                                   x_sendfile:  true
+    send_file Paperclip.io_adapters.for(message.attachment).path, type:        message.attachment_content_type,
+                                                                  filename:    message.attachment_file_name,
+                                                                  x_sendfile:  true
   end
 
   private
 
+  def message
+    @message ||=  Message.find(params[:id])
+  end
   def send_email_if_required
     return unless current_user.persona.is_a?(CaseWorker)
     return unless @message.claim.creator.send_email_notification_of_message?
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -34,7 +34,6 @@ class MessagesController < ApplicationController
   end
 
   def download_attachment
-
     raise 'No attachment present on this message' if message.attachment.blank?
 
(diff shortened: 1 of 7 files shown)


app/models/claims/sort.rb
```diff
@@ -37,34 +37,44 @@ def sort(column, direction)
   #
 
   def nulls_at_top_asc(direction)
```
jsugarman: not your name but perhaps this should be name sort_nulls or nulls_direction.  kind of like the NULLS text being in the method too
```
def sort_nulls(direction)
  'NULLS ' + (direction == 'asc' ? 'FIRST' : 'LAST')
end
```

app/controllers/messages_controller.rb
```diff
@@ -34,17 +34,19 @@ def create
...
+                                                                  filename:    message.attachment_file_name,
+                                                                  x_sendfile:  true
   end
 
   private
 
+  def message
+    @message ||=  Message.find(params[:id])
+  end
   def send_email_if_required
```
jsugarman: line break missing

app/views/user_message_statuses/index.html.haml
```diff
@@ -13,4 +13,4 @@
         %td
           = user_message_status.message.created_at.strftime(Settings.date_time_format)
         %td
-          = link_to 'View claim', __send__("#{current_user.persona.class.to_s.pluralize.underscore}_claim_path", user_message_status.message.claim) + "#messages"
+          = link_to 'View claim', __send__(current_user.persona.message_claim_path, user_message_status.message.claim, anchor: 'messages')
```
jsugarman, line 16: I'd be tempted to put a message_claim_path on the base class that does this dynamically (albeit that class names are not likely to change (though these they did once).
```
def message_claim_path
  "#{class.to_s.pluralize.underscore}_claim_path"
end
```

...unless brakeman no-likey

jsugarman (dismissed):
barring suggestions comments

app/models/claims/sort.rb
colinbruce: How about...
```ruby
  def sort_nulls_by(direction)
    "NULLS #{direction == 'asc' ? 'FIRST' : 'LAST'}"
  end

  def sort_field_with_nulls(field, direction)
    "#{sort_field_by(field, direction)} #{sort_nulls_by(direction)}, id desc"
  end
```

app/controllers/messages_controller.rb
colinbruce: d'oh

app/views/user_message_statuses/index.html.haml
colinbruce: Unfortunately there isn't a base class. :( They both just inherit from `ActiveRecord::Base` and I don't fancy implementing an entire class to inherit this method!

app/models/claims/sort.rb
jsugarman: sweet 👍

jsugarman (approved):
👍

PR #2039, opened 2017-08-09
Modify CCR api JSON response to return null instead of empty string

jsugarman:
Nulls are easier to handle on the CCR side and is preferred
in JSON in any event. Also, modify schema to allow string
or null for fields that could be either.


cibuanokpe (approved):
:thumbs:

PR #2041, opened 2017-08-10
re-allocation fix

colinbruce:
Use the original scheme_filters for re-allocation, the new scheme_filters replaced the old and broke re-allocation


cleargif (commented):
👍

cleargif (approved):
👍

PR #2043, opened 2017-08-11
Specs: allocation + housekeeping

cleargif:
A bit of housekeeping with specs and file names.
Few more specs on allocation


app/views/case_workers/admin/allocations/_scheme_filters.html.haml
```diff
@@ -1,15 +1,27 @@
-#allocation-filters.form-group
-  %fieldset.inline
-    %legend.visuallyhidden
-      Choose your Scheme
+- unless params[:tab] == 'allocated'
```
colinbruce: Was this the fix for re-allocation? 

I'd say keep the code separate `_orig_scheme_filters` and `_scheme_filters` it means we won't have to roll this back, just delete the un-needed version.

Unless this is expressly needed that is?!

colinbruce (approved):
👍

PR #2044, opened 2017-08-11
Fix allocate filter

colinbruce:
## Why? 

Issues with the return of the filters have been raised by users.

This PR addresses the points raised (although they will still not match exactly due to `disk evidence` now being a separate filter


app/interfaces/api/entities/search_result.rb
```diff
@@ -60,31 +60,31 @@ def redetermination
       end
 
       def fixed_fee
-        object.is_fixed_fee.eql?('t')
+        object.is_fixed_fee.eql?('t') && is_submitted?
```
jsugarman, line 63: i assume there is some mapping to the case_type.is_fixed_fee attribute going on here. Is there no `def fixed_fee?` method that can be used?

app/interfaces/api/entities/search_result.rb
```diff
@@ -60,31 +60,31 @@ def redetermination
...
       end
 
       def guilty_plea
-        object.case_type.eql?('Guilty plea')
+        ['Discontinuance', 'Guilty plea'].include?(object.case_type) && is_submitted?
       end
 
       def graduated_fees
-        object.fee_type_code&.in?(graduated_fee_codes).eql?(true)
+        (object.fee_type_code&.in?(graduated_fee_codes).eql?(true) || allocation_type_is_grad?) && is_submitted?
```
jsugarman, line 83: not a change here,  but gotta remember to use `.in?` myself 👍. not for now but can it be used like `'include?` above but the reverse logic
e.g.
```
object.case_type.in? %w[Trial Retrial]
```
which i prefer over:
```
%w[Trial Retrial].include?(object.case_type)
```

jsugarman (approved):
Have not looked but I assume the SQL optimisation you did on this precludes using certain methods on the `object` like `.submitted?` etc. anyway don't want to delay and urgent fixage.

app/interfaces/api/entities/search_result.rb
colinbruce: I think it can be used like that... refactoring is due when we get to re-allocation! :)

PR #2048, opened 2017-08-14
The Lost Specs

cleargif:
These are specs for the second page of the form. Lost in a branch and then found over the weekend :)

The communication and interaction between fee blocks and the side bar.


colinbruce (approved):
👍

PR #2050, opened 2017-08-15
Sort CCR API defendants and rep orders JSON output to have earliest created first

jsugarman:
**What**
Sorts the JSON output from the CCR API endpoint such that
the defendant created first is returned first and the earliest
representation order (per defendant) is returned first.

**Why**
CCR needs to know who the "main" defendant is as 
the main defendants earliest representation order is used order to determine
what fee scheme to apply in calculations. The main defendant is considered
to be the one entered first onto the old paper form. This therefore maps to the 
defendant first created in the CCCD form. The earliest rep order is determined 
by the rep order date attribute.


colinbruce (approved):
:+1:

PR #2052, opened 2017-08-15
Clarify the question posed by db_dump script

jsugarman:
Question is ambiguous and therefore concerning to users.


colinbruce (approved):
:+1:

PR #2053, opened 2017-08-16
Modify change daily attendance calculation

jsugarman:
**What**
- move claim uuid to top level of CCR API JSON and remove claim id
- Daily attendances calculation changes. Prevents calculation result
being greater than the actual trial length or 2, when no Daily attendance
uplifts applied.
- extract daily attendance calculation to its own adapter
- organize CCR specific adapters as namespaced service objects

**Why**
Previous calculation added 2 regardless, but this can lead to invalid
daily attendances where it is higher than the trial length. This, in turn,
can lead to validation errors in CCR during injection.


colinbruce (approved):
:+1:

PR #2061, opened 2017-08-18
Map CCCD claim case types to CCR bill sub types

jsugarman:
**What**
Send relevant bill subtype to CCR for the Advocate fee

**Why**
Successful injection of certain claim case types requires
that the appropriate bill sub type for the advocate fee
is mapped. This is a CCCD side mapping which may eventually
be replaced by a CCR side mapping.


colinbruce (approved):
:+1:

PR #2062, opened 2017-08-21
Gzip allocation api return - Allocation JSON test #3

colinbruce:
# Why?
The current Json size is large, and will only get larger as mandation requires more cases are added.  To get ahaed of the the curve a number of steps have been identified that could reduce the overheard of returning a large JSON object over the wire
### Gzip return
Turn on gzip compression for endpoints
This will need proper benchmarking... ad-hoc tests implies that the file size reduces from ~1.5mb to ~0.3mb, but the time taken to de-compress it in the browser is similar to that taken in downloading the uncompressed file, YMMV

### Note
This includes the optimisation in PR #2060


config/initializers/true_and_false.rb
```diff
@@ -0,0 +1,11 @@
+class TrueClass
```
jsugarman: can this go in the extension folder. we already have a `StringExtension` module for example, so could add a TrueExtension/FalseExtension modules.

lib/extensions/boolean_extension.rb
```diff
@@ -0,0 +1,13 @@
+module BooleanExtension
```
jsugarman, line 1: 👍 sweet

jsugarman (approved):
Ace 👍

PR #2063, opened 2017-08-21
Allocation pagination css

colinbruce:
The pagination of the new allocation table footer meant that it always wrapped on to two lines.

This corrects the split


jsugarman (approved):
👍

PR #2066, opened 2017-08-23
Update geckoboard to post, not put, data

colinbruce:


jsugarman:
As a solution here this is 👍 . If it were a gem this would be useful as a separate feature (i.e. an `publish!(append: true)` type functionality

jsugarman (approved):

PR #2070, opened 2017-08-30
Make offence seeding of unique codes idempotent

jsugarman:
Previous version included the existing record in its
checks for whether the offence code was unique or not.
This prevents taking the current record into account
when determining whether the code is unique. The
seed files find_or_create_by will then not need to
update the record for existing valid unique codes.


colinbruce (approved):
👍

PR #2071, opened 2017-08-31
AGFS / LGFS indicator on CW claims list

cleargif:
https://www.pivotaltracker.com/story/show/150680317

Caseworker needs to tell the difference between AGFS / LGFS claims.

### DIFF
--- a/app/views/case_workers/claims/_claims_list.html.haml
+++ b/app/views/case_workers/claims/_claims_list.html.haml
@@ -8,6 +8,8 @@
         = claims.total_count
     %thead
       %tr
+        %th{scope:'col'}
+          = sortable 'type', t('.type')
         %th{scope: 'col'}
           = sortable 'case_number', t('.case_number')
         %th{scope: 'col'}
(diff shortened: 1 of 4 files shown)


app/views/case_workers/claims/_claims_list.html.haml
```diff
@@ -45,6 +45,8 @@
                         'aria-label': "View #{claim.state.humanize} Claim, Case number: #{claim.case_number}"}
             %span.unique-id-small
               = claim.unique_id
+              %br/
+              = claim.type == 'Claim::AdvocateClaim' ? 'AGFS' : 'LGFS'
             - if claim.disk_evidence == true
```
cleargif: This is likely not the right / most reliable way to do this... but to show the requirement it works

@colinbruce @jsugarman

colinbruce: try 
```ruby
= claim.agfs? ? 'AGFS' : 'LGFS'
```

cleargif:
@colinbruce @jsugarman @slorek 

This is the first stab at this... we can make it better once y'all back from holiday

PR #2074, opened 2017-09-11
Update AWS-SDK gem

colinbruce:
Aamazon updated the SDK so that individual requirements could be loaded.
This reduces the overhead of loading the entire SDK when only SQS is needed


jsugarman (approved):

colinbruce:
Not merging for now... S3 buckets are used to upload documents, not including the aws-sdk-s3 gem didn't break any tests (?!) but we should probably check the impact before merging this PR

PR #2075, opened 2017-09-13
Remove original allocation

colinbruce:
Remove obsolete partials, locale entries and the calling of the removed partial


cleargif (commented):
🥇

cleargif (approved):
👍

PR #2076, opened 2017-09-13
Update the code climate/rubocop interaction

colinbruce:
We are losing 'points' on CC because it is using an older version
of Rubocop, e.g.
rubocop 46: `%i-literals should be delimited by ( and )`
rubocop 49: `%i-literals should be delimited by [ and ]`


jsugarman (approved):
Excellent catch 👍

PR #2077, opened 2017-09-13
Cracked trial validation transition

colinbruce:
Users cannot allocate claims that were created before the new cracked trial validation rules were implemented.

This re-uses the `only_amount_assessed` validation flag on trial_fixed_notice_at, trial_fixed_at and trial_cracked_at while allocating claims to ensure that the fields are not re-validated on authorisation


jsugarman (approved):
LGTM 👍

PR #2082, opened 2017-09-18
remove cracked trial validation

colinbruce:
This reverts the validation change that was implemented due to an inconsistency in CCR and CCCD validations...

it turns out that *we* were right.


jsugarman (approved):
LGTM 👍

PR #2084, opened 2017-09-20
Use GovUKNotify as email sender

colinbruce:
## Why?
Issues with previous provider caused issues with third party AV/firewalls

## How?
Switching to GovUK Notify will save money and allow simpler email patterns


app/controllers/concerns/password_helpers.rb
```diff
@@ -19,7 +19,14 @@ def update_password
...
   def deliver_reset_password_instructions(user)
-    user.send_reset_password_instructions
+    token, enc = Devise.token_generator.generate(user.class, :reset_password_token)
+    user.reset_password_token   = enc
+    user.reset_password_sent_at = Time.now.utc
+    user.save(validate: false)
+    message = DeviseMailer.reset_password_instructions(user, token, current_user.name)
+    message.deliver_now
+  rescue => e
+    Rails.logger.error("DEVISE MAILER ERROR: '#{e.message}' while sending reset password mail")
```
jsugarman: should we raise the exception to allow sentry to pick it up, or are our logs being alerted on

config/environments/development.rb
```diff
@@ -73,8 +73,8 @@
 
 
   # normal dev mail configuration
-  config.action_mailer.perform_deliveries = false
-  config.action_mailer.delivery_method = :file
+  config.action_mailer.perform_deliveries = Settings.govuk_notify.api_key.present?
+  config.action_mailer.raise_delivery_errors = false
```
jsugarman: a discussed, for development we probably want to see the error

config/initializers/config.rb
```diff
@@ -1,31 +1,15 @@
 Config.setup do |config|
   # Name of the constant exposing loaded settings
   config.const_name = 'Settings'
+  config.use_env = true
+  config.env_prefix = 'SETTINGS'
+  config.env_separator = '__'
```
jsugarman, line 6: is the double underscore deliberate?

spec/mailers/previews/notify_mailer_preview.rb
```diff
@@ -1,8 +1,8 @@
 class NotifyMailerPreview < ActionMailer::Preview
```
jsugarman, line 2: should this whole file not be removed?

jsugarman (approved):
Aside from some little comments/queries/double-checks. it looks good

Also you can probably amend the `Gemfile` to remove, at least, `gem 'premailer-rails',         '~> 1.9'`

spec/mailers/previews/notify_mailer_preview.rb
jsugarman: ..and 
- app/assets/stylesheets/email.css
- app/views/layouts/email.html.haml

config/initializers/config.rb
colinbruce: Yes, it's to allow users to use underscores in env vars (if they need to?!) while allowing config to split a double underscore into dotted strings e.g. test__value becomes `Settings.test.value`, while test_value would remain `Settings.test_value`

PR #2087, opened 2017-09-21
Remove invalid offences

jsugarman:
**What**
- Delete invalid offences
- Update claims with invalid offences to have correct class of the same offence

**Why**
Legislation stipulates that certain offences must always be of a particular class (band value) regardless. Data Injection into CCR (which applies this legislative logic) requires these invalid
CCCD offences to be removed as they have no mapping/equivalent in CCR.

**References**
[Pivotal Tracker ticket](https://www.pivotaltracker.com/n/projects/1985371/stories/151245975) for ***ID #151245975***
Legslative reference showing offences that must be of a particular class (banding value):
screen shot 2017-09-20 at 11 00 25


zheileman (approved):
👍

PR #2091, opened 2017-09-22
Remove offence id from CCR JSON API response

jsugarman:
No longer needed as unique_code is
exposed and will be used by consumers
including CCR.


colinbruce (approved):
Looks good

PR #2092, opened 2017-09-25
Change new claim page title

cleargif:
PT: https://www.pivotaltracker.com/story/show/149840901

Why are we doing this:
The word “gradutated” is not always the correct context for a claim

What this PR does:
Removes the “graduated” from the main title


colinbruce (approved):
👍

PR #2094, opened 2017-09-25
Prevent sending advocate fee when no basic fees claimed.

jsugarman:
**What**
Stop sending Advocate fee "bill" when no CCCD fee equivalents being claimed 

**Why**
Advocate Fee was being sent as a CCR "Bill" even when no CCCD fees that make it up
were part of the claim. This causes a validation error in CCR when the Advocate fee is provided with no PPE (at least) since this is mandatory in CCR. This PR prevents the CCR JSON API endpoint from adding the Advocate Fee "bill" if no "basic fees" are being claimed (i.e. if their quantity/rate/amount are all 0).

### DIFF
--- a/app/interfaces/api/entities/ccr_claim.rb
+++ b/app/interfaces/api/entities/ccr_claim.rb
@@ -112,8 +112,13 @@ module API
 
       def bills
         @bills ||= [].tap do |arr|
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
👍

PR #2095, opened 2017-09-26
Send defendant details in JSON API for CCR consumption

jsugarman:
**What**
Adds defendant details to JSON API for CCR consumption

**Why**
CCR allows any bill type to specify that the MAAT reference is NOT required.
It then requires the user to enter the defendants first name, surname and DOB
to be supplied. In addition, and specifically, a MAAT reference is NOT required
in CCCD for "Breach of crown court order" claims. There may be others...

Pivotal Tracker Ticket: **#151434930**


colinbruce (approved):
👍

PR #2096, opened 2017-09-27
Add case numbers to AGFS number of cases uplift (NOC)

jsugarman:
Adds the ability to specify case numbers against Number of case uplift fees (both basic and fixed fee incarnations)

PT:
https://www.pivotaltracker.com/story/show/150128862

- [X] additional case numbers logic for basic fee type - BANOC
- [X] additional case numbers logic for fixed fee type - FXNOC
- [x] view formatting/styling (DP to action)
- [X] expose case numbers for CCR api endpoint as part of advocate fee


colinbruce (approved):
LGTM, It would be nice if we could get rid of the code climate warnings *cough* @cleargif ;)

cleargif:
@colinbruce your wish is my command...

PR #2098, opened 2017-10-03
Correct email recipient

colinbruce:
The message notifictaions should go to the creator, not the user.

In the factory they were both the same person, so when I duplicated that it was harder to test the required need.


jsugarman (approved):
👍

PR #2099, opened 2017-10-04
Prevent nilClass error on querying CCR API for unsubmitted claims

jsugarman:
Enables testing of CCR API endpoint for all claim statuses


colinbruce (approved):
👍

PR #2100, opened 2017-10-04
Validate the presence of case number lists

jsugarman:
**What**
Enforce the presence of a matching number of case numbers for particular fees

**Why**
Various Basic and Fixed fees described as "uplifts" refer to additional
cases for which the claimant is claiming reimbursement. CCR will not
accept such uplifts without the case numbers of the additional cases.

**API considerations**
To prevent breaking current API consumers by enforcing this validation across the board 
we are not enforcing this on the API fee endpoint (for now). API vendors have been informed
and the release notes updated to give them chance to amend and release their applications to match this logic.

***Pivotal ticket:*** https://www.pivotaltracker.com/story/show/151841573


features/page_objects/sections/fee_summary_section.rb
```diff
@@ -0,0 +1,14 @@
+class FeeSummarySection < SitePrism::Section
+  element :captionna, '.table-cation visuallyhidden'
+  elements :names, 'tbody > tr > th'
+  elements :values, 'tbody > tr > td'
+  # sections :rows, 'tbody > tr' do
```
jsugarman: remove!

colinbruce (commented):
👍  Technically looks good

Maybe we should add a hard date for enforcing the validation in the api release notes, e.g. 1 April 2018

colinbruce (approved):

slorek:
🛑 Have we checked whether this data is actually required by caseworkers, and if not, whether CCR's validation can be relaxed?

jsugarman:
@slorek it is required by the system - you cannot claim an uplift unless you provide a valid case number to which it applies. relaxing CCR validation on injection is an option but then data would have to be provided at some point to complete the claim.

PR #2101, opened 2017-10-06
map and expose miscellaneous fees to CCR bills

jsugarman:
**What**
Add CCR miscellaneous fee equivalents to the JSON api output 

**Why**
For injection of equivalent CCCD fees into CCR.

**NOTE**
CCR miscellaneous fees are not synonymous with CCCD miscellaneous fees. Some
Basic Fees and Fixed fees in CCCD are handled as miscellaneous fees in CCR.


app/interfaces/api/entities/ccr/adapted_advocate_fee.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        expose :case_numbers
+
+        # NOTE: for possible use in comparisons between CCCD and CCR calculation comparision
+        # expose :calculated_fee, as: :calculatedFee
+
+        private
+
+        def quantity
+          1.0
```
colinbruce, line 27: Why deliberately set this to 1?  Wouldn't the app/api force a value?

Or, does the fact that the fee exist often mean the quantity is one?

I'll keep reading... :)

config/schemas/ccr_claim_schema.json
```diff
@@ -95,40 +95,56 @@
           },
           "number_of_cases": {
             "id": "/properties/bills/items/properties/number_of_cases",
-            "type": "integer"
+            "type": "string"
```
colinbruce, line 98: Is this because CCR needs strings?

colinbruce (approved):
A couple of small questions, but if the answers are obvious, feel free to merge.

app/interfaces/api/entities/ccr/adapted_advocate_fee.rb
jsugarman: In CCR there is only ever one advocate fee. Ideally CCR should just assume this and the other hardcoded values and we can then do away with exposing quantity, rate and amount entirely. 

will speak to @cibuanokpe about this, but this has been going exposed for a while now and am just exposing the same thing in a different way here.

config/schemas/ccr_claim_schema.json
jsugarman: Had a long slack chat with clem about this and we have settled on using strings for any number. BigDecimals, in particular, default to strings via the BigDecimal.to_json method. There is quite alot of chatter about this stuff e.g. https://github.com/rails/rails/issues/25017 . rails used to have an override but it was removed for sensible reasons. I made integers strings too firstly for consistency (CCR can just assume strings and handled appropriately) but also was reading that the JSON datatype integer is not parsed consistently by some libraries in any event - meaning you could put a float/decimal in and might get away with it.

colinbruce: 👍

app/interfaces/api/entities/ccr/adapted_advocate_fee.rb
colinbruce: :+1:

colinbruce:
I restarted that build after a flickering test... Other than than that good to merge 👍

PR #2103, opened 2017-10-11
Bump rubocop to 0.50

colinbruce:
# Why?
Trying to jump many versions of rubocop is a pain.  Trying to keep up to date makes it easier


app/models/claim/base_claim.rb
```diff
@@ -414,14 +414,14 @@ def opened_for_redetermination?
       return true if redetermination?
 
       transition = filtered_last_state_transition
-      transition && transition.to == 'redetermination'
+      transition&.to == 'redetermination'
```
jsugarman, line 417: 👍

app/form_builders/adp_form_builder.rb
```diff
@@ -45,7 +45,7 @@ def full_anchor_name_for(_object, anchor_name)
 
   def make_object_name
     klass_name = object.class.to_s
-    klass_name = 'Claim' if klass_name =~ /^Claim::/
+    klass_name = 'Claim' if klass_name.match?(/^Claim::/)
```
jsugarman, line 48: 🤔 Was this a cop suggestion? - i ask because match? is clearer but =~ is more performant from my understanding. 
...
Just noticed the question mark  - match? - looks like introduced in 2.4 regexp and assume it is performant 👍

app/models/claim/transfer_brain_data_item_collection.rb
```diff
@@ -69,7 +69,7 @@ def valid_transfer_stage_ids(litigator_type, elected_case)
       transfer_stages = @collection_hash.fetch(litigator_type).fetch(elected_case)
       ids = []
       transfer_stages.each do |transfer_stage_id, result_hash|
-        result_hash.each do |_case_conclusion_id, result|
+        result_hash.each_value do |result|
```
jsugarman, line 72: 👍

jsugarman (commented):
looking good so far - you still working on this i guess

jsugarman (approved):
👍

PR #2104, opened 2017-10-13
Refresh gems

colinbruce:
AIM: Reduce the number of outdated gems in the project

At the start of this work, according to Gemnasium, there were 49 outdated gems.

Some cannot be updated because of breaking changes, but many can and should be updated


jsugarman (commented):
looks good to me but obviously needs some manual testing and test suite runs.

Because i am adding quite a few more JSON API specific tests would you mind holding on with merging this til mine is merged so you can run against my new stuff.

jsugarman (approved):
good to go

PR #2106, opened 2017-10-16
Fix the agfs and lgfs data generators.

jsugarman:
A previous rubocop fix had improperly applied
a new helper, introducing an error.


cleargif (approved):
👍  if tests pass

PR #2107, opened 2017-10-17
Remove unnecessary exposures from adapted advocate fee entity

jsugarman:
The CCR adapted advocate fee entity does not need to expose the quantity, rate and amount as they 
are irrelevant for injection of an advocate fee of type advocate fee into CCR.


cibuanokpe (approved):
👍

PR #2110, opened 2017-10-19
Download all link

colinbruce:
# What

Add a download all files link for caseworkers

# Why

To allow caseworkers quick downloads of all files, rather than downloading the files individually

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -716,6 +716,7 @@ DEPENDENCIES
   rspec-rails (~> 3.4)
   rubocop (~> 0.50)
   ruby-progressbar
+  rubyzip
   sass-rails (~> 5.0.6)
   scheduler_daemon!
   sdoc (~> 0.4.0)
--- a/app/services/s3_zip_downloader.rb
+++ b/app/services/s3_zip_downloader.rb
@@ -6,21 +6,20 @@ class S3ZipDownloader
     @folder = Dir.mktmpdir("#{@claim.case_number}-", './tmp')
     move_files
     @files_to_zip = Dir.entries(@folder) - %w[. ..]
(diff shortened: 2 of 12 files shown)


Gemfile.lock
```diff
@@ -104,10 +104,9 @@ GEM
       descendants_tracker (~> 0.0.1)
     colorize (0.8.1)
     concurrent-ruby (1.0.5)
-    config (1.5.0)
+    config (1.2.1)
```
jsugarman: is this the result of needing a rebase on the gem refresh PR?

app/services/s3_zip_downloader.rb
```diff
@@ -0,0 +1,33 @@
+class S3ZipDownloader
+  require 'zip'
+
+  def initialize(claim)
+    @claim = claim
+    @folder = Dir.mktmpdir("#{@claim.case_number}-", './tmp')
```
jsugarman, line 6: cool, perhaps a
```
Rails.root.join('tmp') => Pathname object
```
or
```
File.join(Rails.root, 'tmp') => string
```

jsugarman: but just a suggestion

jsugarman (approved):
👍 couple of comments

Gemfile.lock
colinbruce: Hmm Good question... rolling back seems a bad idea... I thought I had rebased, but I'll double check.

app/services/s3_zip_downloader.rb
colinbruce: I just used the standard ruby invocation of https://apidock.com/ruby/Dir/mktmpdir/class

PR #2111, opened 2017-10-20
FIX: Add handler for vat_amount being nil

colinbruce:
# Why
The front end allows caseworkers to submit a nil value for VAT amount
this causes the redetermination presenter to blow up because it
can't add nil to the total.
This prevented the display of the claim entirely.

# How
This ensures the vat amount will add zero instead of nil and allow the
display of the claim


jsugarman (approved):
LGTM 👍

PR #2112, opened 2017-10-20
Ensure CCR JSON returns empty array when no fees/bills exist

jsugarman:
**What**
Ensure that the bills in the JSON output always return an array 
even if it is empty, and never nil or {}.

**Why**
CCR assumes it receives an array, not nil. Returning
an empty array instead of nil, or JSON {}, is more
predictable for consumers generally.


colinbruce (commented):
👍

colinbruce (approved):
👍

PR #2113, opened 2017-10-20
Add additional case numbers to all summary pages

jsugarman:
**What**
displays additional case numbers added to fees in summary pages

**Why**
The case workers need this information for entry into CCR. The 
external users draft claim summary and pre submission check
list/summary page should also display this for consistency.


colinbruce (approved):
LGTM 👍

PR #2114, opened 2017-10-23
Hide download all link when no documents

cleargif:
Why are we doing this:
- the download all link should only be visible if there are downloads

What this PR does:
- hides the link when no docs


jsugarman (approved):
👍

PR #2115, opened 2017-10-25
Messaging update

colinbruce:
# Why
Users needed to request redeterminations and send messages when their claims where in additional states.

# How
Extend the #show_message_controls? method to better define the rules around when to display the message partial


jsugarman (approved):
LGTM 👍

colinbruce:
For reference, the new rules on message display are linked from the pivotal ticket 
https://www.pivotaltracker.com/story/show/152107041

[finishes #152107041]

PR #2116, opened 2017-10-25
Fix message_added_email

colinbruce:
# Why?
The mailer was attempting to send emails to a softly-deleted user
As this suffixes `.deleted.{id}` to the users email, this would fail.

# What?
Add tests to replicate the situation and then handle the situation by
not attempting to send messages to softly-deleted external users


jsugarman (approved):
👍

PR #2119, opened 2017-10-26
Add find user

colinbruce:
### Why?
As a caseworking superuser, I'd like to be able to find an existing user by email.
Sometimes when adding a new user, I am told that the email address is already in use but I can't find it without manually searching through the providers.

### DIFF
--- a/app/views/layouts/_primary_navigation.html.haml
+++ b/app/views/layouts/_primary_navigation.html.haml
@@ -7,6 +7,8 @@
         %li
           = link_to 'Add a provider', new_super_admins_provider_path, class: cp(new_super_admins_provider_path)
         %li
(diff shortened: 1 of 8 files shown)


app/views/layouts/_primary_navigation.html.haml
```diff
@@ -7,6 +7,8 @@
         %li
           = link_to 'Add a provider', new_super_admins_provider_path, class: cp(new_super_admins_provider_path)
         %li
+          = link_to 'Manage user by email', super_admins_external_users_find_path, class: cp(new_super_admins_provider_path)
```
jsugarman: Does this link name come from design. Thinking that "Find user by email" or similar is clearer

jsugarman (approved):
Just a comment relating to the terminology for the link and page title versus not found messages (provider/user)

PR #2120, opened 2017-10-27
Reduce technical debt - cyclomatic complexity

colinbruce:
### Why?

We agreed that reducing the technical debt was a good thing and that we would aim to remove the rubocop_todo file.  This removes the cyclomatic complexity cop and addressed the issues that we weer ignoring

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -36,12 +36,12 @@ class Allocation
 
   def save
     return false unless valid?
-    if allocating?
-      allocate_all_claims_or_none! if claims_in_correct_state_for?(:allocation)
-    elsif deallocating?
-      deallocate_claims if claims_in_correct_state_for?(:deallocation)
-    elsif reallocating?
-      reallocate_claims if claims_in_correct_state_for?(:reallocation)
+    if should_allocate_claims
+      allocate_all_claims_or_none!
+    elsif should_deallocate_claims
+      deallocate_claims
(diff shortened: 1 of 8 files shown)


app/models/claims/sort.rb
```diff
@@ -13,17 +13,8 @@ def sort(column, direction)
...
-    when 'advocate'
-      sort_advocates(direction)
-    when 'case_type'
-      sort_case_type(direction)
-    when 'total_inc_vat'
-      sort_total_inc_vat(direction)
-    when 'amount_assessed'
-      sort_amount_assessed(direction)
+    if META_SORT_COLUMNS.include?(column)
+      send("sort_#{column}", direction)
```
jsugarman, line 17: :thumbsup: cool

app/models/allocation.rb
```diff
@@ -68,11 +68,19 @@ def case_worker
     # deallocation will have a nil case worker id
   end
 
-  def allocating?
-    @allocating
+  private
+
+  def should_reallocate_claims
```
jsugarman: Just a naming suggestion: personally like the boolean `?` thing
```
can_allocate_claims?
```

jsugarman (approved):
Good stuff

app/models/allocation.rb
colinbruce: Good call... 
> naming things is hard

:(

PR #2121, opened 2017-10-30
Messaging extend allowed statuses

colinbruce:
As a provider, I want to be able to send a message after I have requested a redetermination/written reasons so that if i have forgotten something or if I need to attach more than one document I can.

As a provider, I want to be able to request for redetermination/written reasons and send a message if my claim has been authorised, so that if a caseworker accidentally clicked authorised instead of part-authorised, I can still apply for redetermination.

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -3,7 +3,8 @@ module ClaimsHelper
     submitted
     allocated
     part_authorised
-    rejected refused
+    rejected
+    refused
     authorised
     redetermination
     awaiting_written_reasons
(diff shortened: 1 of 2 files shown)


app/helpers/claims_helper.rb
```diff
@@ -1,12 +1,23 @@
 module ClaimsHelper
+  EXTERNAL_USER_MESSAGING_ALLOWED = %w[
+    submitted
+    allocated
+    part_authorised
+    rejected refused
```
jsugarman: linebreak missing ?!

jsugarman (approved):
👍  one teensy comment

app/helpers/claims_helper.rb
colinbruce: D'oh

PR #2122, opened 2017-10-30
Skip flickering test

jsugarman:
Skip a test that flickers alot, and which is on a class that is not currently used in any event.


colinbruce (approved):
yay, go! 👍

PR #2123, opened 2017-10-30
Fix rescue error violations

jsugarman:
Replace unspecific rescues with either a more specific error to rescue or alternative to a rescue, such as using a safe operator `&.`

***Note:*** where StandardError is rescued this is an explicit rescue of what is actually the default (when no named error class is rescued) so **just** fixes the violation. Where replacements
of the rescue have been written with a safe operator or specific error we may end up getting sentry errors which we can then iterate on.

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -36,17 +36,21 @@ class FeedbackController < ApplicationController
   end
 
   def merged_feedback_params
-    feedback_params.merge(email: (begin
-                current_user.email
-              rescue
-                (email_from_user_id || 'anonymous')
-              end),
+    feedback_params.merge(email: user_email_or_anonymous,
                           user_agent: request.user_agent)
   end
 
+  def user_email_or_anonymous
+    if current_user
+      current_user.email
+    else
+      email_from_user_id || 'anonymous'
+    end
+  end
+
   def email_from_user_id
-    User.active.find(params[:user_id]).try(:email)
-  rescue
+    User.active.find(params[:user_id])&.email
+  rescue ActiveRecord::RecordNotFound
     nil
   end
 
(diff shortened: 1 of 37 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -36,17 +36,21 @@ def referrer_path
...
-    feedback_params.merge(email: (begin
-                current_user.email
-              rescue
-                (email_from_user_id || 'anonymous')
-              end),
+    feedback_params.merge(email: user_email_or_anonymous,
                           user_agent: request.user_agent)
   end
 
+  def user_email_or_anonymous
```
colinbruce, line 43: Nice 👍

app/models/fee/base_fee.rb
```diff
@@ -116,9 +116,7 @@ def perform_validation?
     end
 
     def calculated?
-      fee_type.calculated?
-    rescue
-      true
+      fee_type&.calculated?.nil? ? true : fee_type&.calculated?
```
colinbruce, line 119: Could be simplified with the OR operator
```ruby 
fee_type.calculated? || true
```
if the call on the left fails, it will return true

app/services/claims/case_worker_claim_updater.rb
```diff
@@ -80,8 +80,8 @@ def update_and_transition_state
           update_assessment if @assessment_params_present
           add_redetermination if @redetermination_params_present
           @claim.send(event, audit_attributes.merge(reason_code: @transition_reason)) unless @state.blank? || @state == @claim.state
-        rescue => ex
-          add_error ex.message
+        rescue StandardError => err
```
colinbruce, line 83: +1 for consistent naming of error objects!

colinbruce (approved):
Nice work, one minor observation.

LGTM 👍

app/models/fee/base_fee.rb
jsugarman: tried this and unfortunately this does not handle the `false` possibility. 
```
false || true => true
```
The code is, nonetheless, not a very nice way of defaulting nil to true.

PR #2124, opened 2017-10-30
Shorten long blocks to new standard of 80 lines

jsugarman:
Reduce block length requirement to 30 lines and fix or exclude violations


colinbruce (approved):
👍

PR #2125, opened 2017-10-31
Improve method missing usage

jsugarman:
Adds respond_to_missing? methods
to ensure any calls to respond_to?
for missing methods, behaves appropriately.


.rubocop_todo.yml
```diff
@@ -124,8 +124,8 @@ Style/IfInsideElse:
 # Offense count: 5
 Style/MethodMissing:
   Exclude:
```
colinbruce: can you delete the whole block rather than just comment out the affected lines?

colinbruce (commented):
:+1: LGTM with one minor fix

colinbruce (approved):

PR #2126, opened 2017-10-31
Update `spend to date` counting

colinbruce:
Why? The previous method was dropping figures as old claims were deleted

How? This version sums the daily spends to date and adds it to today as total


jsugarman (commented):
Are you going to delete this branch?

PR #2127, opened 2017-10-31
Require and validate defendant DoB for all case types

jsugarman:
**What**
Enforces Defendant DoB requirements for all case types.

**Why**
Previously defendant DoB was not required (and field hidden) for 
Breach of crown court order case types (Fixed fees). This is the
opposite of what CCR requires where a DoB is required for these
types of case specifically because the defendant will/may not 
have a rep order MAAT reference for such cases.


colinbruce (approved):
Simplifying code? 👍

PR #2128, opened 2017-11-01
Improve class length violations

jsugarman:
- [X] use exclusions instead of increasing max allowed
- [X] Reduce base claim presenter class length
- [X] Add missing specs and refactor specs


colinbruce (approved):
:+1: LGTM

PR #2134, opened 2017-11-03
Factory bot tweaks

jsugarman:


Gemfile
```diff
@@ -9,6 +9,7 @@ gem 'cancancan',              '~> 1.15'
 gem 'cocoon',                 '~> 1.2.6'
 gem 'devise',                 '~> 3.5.1'
 gem 'dropzonejs-rails',       '~> 0.7'
+gem 'factory_bot_rails',     '~> 4.7'
```
colinbruce, line 12: Nice, as per my original commit, I couldn't see why this was loaded on the live server...
D'oh.

colinbruce (approved):
Great, thanks for fixing this!

Gemfile
jsugarman: yeah, its a pretty abstract problem (reload does not fail locally because you are in development env by default) in which i was digging around changing files on the container itself and re running db:reload. couple of points worth noting, while we are here:
- do we really need to reload dev - there are plusses and minuses
- ideally the deploy should fail more obviously (i.e. return a failure/red in jenkins) with output in the jenkins log as rake tasks can break and no one would notice - its happened quite a few times in the past i recall and its almost always the reload task.

will force merge this later, in your absence.

have a good day anyway.

jsugarman: getting another prob with reload now so will PR another change later, but are you aware of any paperclip changes that have gone in recently?

```
Errno::ENOMEM: Cannot allocate memory - file -b --mime '/tmp/8fa40b07a81c316ceaaadd06689c840020171103-352-sywnvm.pdf'
```

jsugarman: @colinbruce ^^^

colinbruce: No, I'm not aware of any changes with paperclip...
Could rebuilding docker lots of times swallowed memory?

PR #2136, opened 2017-11-03
Replace hotfix with transition based validation disabling

jsugarman:
**What**
Prevent validating of claim components during transitions
which should not be affected by them.

**Why**
With the implementation of new validations
existing submitted claims may suddenly become
invalid thereby preventing case workers from
processing them - and advocates/litigators
cannot amend by this stage.

### DIFF
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -349,7 +349,7 @@ module Claim
       force_validation? || validation_required?
     end
 
-    # we must validate
+    # validation required when
     # being created from api (as draft)
     # or is in state of archived_pending_delete or draft (not from api)
     # or is in a state deleted (old statement????)
@@ -360,13 +360,10 @@ module Claim
       return false if draft? || archived_pending_delete?
       return false if disabled_for_transition?
       true
-
-      # from_api? ||
-      # !(draft? || archived_pending_delete?)
     end
 
     def disabled_for_transition?
-      disable_for_state_transition.eql?(:all)
+      disable_for_state_transition.present?
     end
 
     def step?(num)
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -349,9 +349,24 @@ module Claim
       force_validation? || validation_required?
     end
 
-    # we must validate unless it is being created as draft from any source except API or is in state of archive_pending_delete or deleted
+    # we must validate
+    # being created from api (as draft)
+    # or is in state of archived_pending_delete or draft (not from api)
+    # or is in a state deleted (old statement????)
+    # or is transitioning via certain events
+    #
     def validation_required?
-      from_api? || !(draft? || archived_pending_delete?)
+      return true if from_api?
+      return false if draft? || archived_pending_delete?
+      return false if disabled_for_transition?
+      true
+
+      # from_api? ||
+      # !(draft? || archived_pending_delete?)
+    end
+
+    def disabled_for_transition?
+      disable_for_state_transition.eql?(:all)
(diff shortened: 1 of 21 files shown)


spec/validators/claim/base_claim_validator_spec.rb
```diff
@@ -25,69 +25,196 @@
...
+        claim.force_validation = false
+        nulify_fields_on_record(claim, :case_type, :court, :case_number, :advocate_category, :offence, :estimated_trial_length, :actual_trial_length)
+        claim.defendants.destroy_all
+        claim.fees.destroy_all
+        claim.expenses.destroy_all
+      end
 
-      before(:each) do
-        claim.force_validation=false
+      it 'validation is performed' do
```
jsugarman: should probably be `it 'claim is not valid' do`

app/models/claim/base_claim.rb
```diff
@@ -349,9 +349,21 @@ def perform_validation?
...
+    # validation required when
+    # being created from api (as draft)
+    # or is in state of archived_pending_delete or draft (not from api)
+    # or is in a state deleted (old statement????)
+    # or is transitioning via certain events
+    #
     def validation_required?
-      from_api? || !(draft? || archived_pending_delete?)
+      return true if from_api?
+      return false if draft? || archived_pending_delete?
```
colinbruce: Could this be
``` ruby
return false if draft? || archived_pending_delete? || disabled_for_transition?
```
Or was it separated for a reason?

app/models/claims/state_machine.rb
```diff
@@ -79,7 +79,7 @@ def self.included(klass)
         before_transition on: %i[reject refuse], do: :set_amount_assessed_zero!
 
         around_transition any => NON_VALIDATION_STATES.map(&:to_sym) do |claim, transition, block|
-          validation_state = %i[authorise part_authorise].include?(transition.event) ? :only_amount_assessed : :all
+          validation_state = %i[authorise authorise_part].include?(transition.event) ? :only_amount_assessed : :all
```
colinbruce, line 82: Yikes.. good catch!

jsugarman:
@colinbruce can you take a look at this PR which is aiming to prevent unnecessary validations from firing for transition events triggered by case workers.

This was the fundamental cause of the errors encountered on friday by case workers, whereby, they were attempting to allocate!, authorise!, authorise_part! etc. claims that violated a new validation (defendant dob must always exist). 

To my mind case workers should not be impacted by validation violations that are beyond their control.

The existing logic already prevented validation of the claim itself but NOT of the sub models of claim (defendant, representation_orders, fees etc.). This does not make sense to me. so i have changed it so that the `disable_for_state_transition` attribute set in the `around_transition` block in the state_machine is used to determine whether to validate via `base_claim#perform_validation?` method. This method is delegated to by its various sub models.

spec/validators/claim/base_claim_validator_spec.rb
colinbruce: I think it's okay in the context of `Validation is performed here` versus not being performed later

colinbruce (approved):
👍  Nice

app/models/claim/base_claim.rb
jsugarman: purely seaprate as the disabled for transition was my main focus - I can't see any reason not to  make it as you suggest, will try!

app/models/claims/state_machine.rb
jsugarman: indeed, it means that part authorise may not have required an assessment in the past...doh!

PR #2137, opened 2017-11-09
disallow external_users messaging on a rejected claim

colinbruce:
Why?
As per ticket #152604192, do not allow external_users to send messages
or apply for redetermination


jsugarman (approved):
👍

PR #2142, opened 2017-11-17
Update Dockerfile

colinbruce:
**What?** remove WORKDIR instruction
**Why?** It looks like a typo from https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/b4c2012550076862fb68927a2c8857696c9e396a anyway

removing it still allows the build to proceed


jsugarman (approved):
LGTM

PR #2145, opened 2017-11-23
Add expenses to CCR api end point

colinbruce:
### What?
Enable injection of  CCCD expenses into CCR

### Why?
So that caseworkers can avoid re-keying expenses into CCR

### Todo

- [ ] Speak to CCR devs about adding/handling bike travel claims
- [x] re-factor the test set-up in `app/services/ccr/expenses_adapter.rb`
- [x] delete the original spike of expenses adapter


spec/fixtures/ccr_test_expenses_fixtures.rb
```diff
@@ -0,0 +1,87 @@
+module CCR
```
jsugarman, line 1: 👍

spec/services/ccr/expenses_adapter_spec.rb
```diff
@@ -0,0 +1,113 @@
+require 'rails_helper'
+
+require('support/shared_examples_for_expenses')
+require('fixtures/ccr_test_expenses_fixtures')
```
jsugarman, line 4: does this always guarantee the same as 
```
require Rails.root.join('fixtures', 'ccr_test_expenses_fixtures')
```

app/services/ccr/expenses_adapter.rb
```diff
@@ -0,0 +1,109 @@
+module CCR
+  class ExpensesAdapter
+    attr_reader :expense
+
+    TRANSLATION_TABLE = {
```
jsugarman, line 5: is there anyway we could refactor this to remove the repetition of the string values?

jsugarman: ...although the repetition is clear 🤔

jsugarman (approved):
👍 looks good.

PR #2149, opened 2017-11-27
LAC check list option handling

cleargif:
Why are we doing this:
- This option is redundant

What this PR does:
- JS will remove the option unless it is in use by a claim
- As the option is empty we simply remove the input


jsugarman (approved):
LGTM 👍

PR #2152, opened 2017-11-27
Add unicorn worker killer

jsugarman:
**What**
Kill Unicorn workers that exceed a maximum memory limit

**Why**
Memory leak on prod


colinbruce (approved):
LGTM 👍

PR #2156, opened 2017-11-29
Modify fixed fee number of defendants calculation

jsugarman:
**What**
Modify calculation for number of defendants on a fixed fee

**Why**
CCR expects the total number of defendants and any above
one is treated as an uplift. In CCCD this equates to
the sum of all quantities for Number of defendants uplift fees 
FXNDR plus one for the main defendant.


colinbruce (approved):
:+1:

PR #2157, opened 2017-11-29
Advocate fee defendant uplift calc

jsugarman:
**What**
Send number of defendants for advocate fee 

**Why**
Previously this was determined by CCR from
the actual of number of defendants on the claim
but this means CCR could end paying a fee when
one was not claimed. Now this number is determined
from the defendant uplift quantity plus one for the
main defendant.

**TODO**
CCR will need to implement this change


colinbruce (approved):
👍

PR #2158, opened 2017-11-30
Add unicorn worker killer

jsugarman:
**What**
Kill unicorn worker processes that exceed given memory limit

**Why**
Out of memory errors on prod


colinbruce (approved):
LGTM

colinbruce (approved):
👍

PR #2159, opened 2017-11-30
Mock misc fee defendant uplifts

jsugarman:
So CCR can implement use of number of
defendants attribute for all bills (
advocate-fee/fixed-fee/misc-fee).


colinbruce (approved):
LGTM 👍

PR #2160, opened 2017-11-30
Add misc fee defendant uplift specs

jsugarman:
**What**
Consolidates miscellaneous fees of the uplift variety into their "parent" misc fee 
as a defendant count, for injection into CCR.

**Why**
CCR accepts only the "parent" varieties of miscellaneous fee
with (defendant) uplifts applied via a "number of defendants" field
on the bill/fee.


colinbruce (approved):
👍

PR #2161, opened 2017-12-01
Updated link for special prep

cleargif:
Why are we doing this:
- new forms live 1st Dec

What this PR does:
- updated link to download page

PT:
https://www.pivotaltracker.com/n/projects/1985371/stories/149888625


colinbruce (approved):
👍

PR #2162, opened 2017-12-01
Add error logging on redraft fails

colinbruce:
### What?

Add Clone failing output to the logs

### Why?

We are getting lots of error reports on redraft from users, but because
the errors seem to be duplication related, anonymising and duplicating
the database to local doesn't replicate the S3 doc storage.

This step is designed to log better errors for us, so that we can handle
actual failures better in future and give users better feedback


jsugarman (approved):
Good idea. every loves logs 👍

PR #2163, opened 2017-12-04
Generate Disc Evidence PDF

colinbruce:
### What?

Replace the static PDF with a fillable PDF form 

### Why?

This will allow:
a) us to pre-fill the form before rendering to the user, and;
b) the user to complete the remaining fields on screen 

### ToDo

- [x] Update readme to cover dev installing of pdftk


.travis.yml
```diff
@@ -4,8 +4,12 @@ addons:
   postgresql: '9.3'
   code_climate:
     repo_token: c94f8c6df526a441d103d0c6d54623174af5eb6ccbfe5f3e7c7dbce734ebfa98
+env:
+  -PDFTK_PATH=/usr/bin/pdftk
 before_install:
-- export TZ=Europe/London
+  - export TZ=Europe/London
+  - sudo apt-get -qq update
```
jsugarman, line 11: found this while looking up the -qq switch
```
 you should never use -qq without a no-action modifier such as -d, --print-uris or -s as APT may decided to do something you did not expect.
```

README.md
```diff
@@ -68,6 +68,16 @@ To import JSON claims, or import via the API, you need to run a multi-threaded s
 rails server -e devunicorn
 ```
 
+### Download and install pdftk
+
```
jsugarman, line 72: worked for me on El Capitan too 👍

jsugarman (approved):
looks good 👍 

one comment regarding use of the -qq switch which may or may not be an issue

.travis.yml
colinbruce: Cheers, but I took the line from travis... https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Standard-Infrastructure 

So I guess they are happy with silent fails on update?!

PR #2164, opened 2017-12-06
Update the travel_time quantity

colinbruce:
CCR expects travel time to be in quarter amounts e.g. 1.25, 1.5, 1.75, 2.0

Currently case-workers will round up the provider stored values, this commit
ensures that when providing data for the CCR endpoint it does this automatically
this makes the injection more robust and will reduce the re-keying needed


jsugarman (approved):
Cool 👍

PR #2165, opened 2017-12-06
Judicial apportionments help text

cleargif:
PT: https://www.pivotaltracker.com/story/show/153171646

Why are we doing this:
External users are not clear about Judicial apportionments.

What this PR does:
Adds in some help text to explain the context where Judicial apportionment might be applicable


colinbruce (approved):
:+1: Awesome... next best thing to removing it completely

PR #2166, opened 2017-12-07
Map discontinuance claim fees to CCR Advocate fee bill

jsugarman:
Discontinuance is a bill scenario in CCR that
accepts the AGFS_FEE, AGFS_FEE bill type/sub-type.

The CCR fee itself only accepts PPE, defendant uplift and
case uplift fields, which are already being passed.


colinbruce (approved):
:+1:

PR #2167, opened 2017-12-07
Add errors when allocation fails

colinbruce:
As part of the allocation process, the system logs the claim as allocated
and _then_ links the case worker.  Twice we have seen a situation where
part 1 happens but part 2 fails.

This PR should log errors that can be monitored in kibana to give us some
details to debug this issue.

~~No tests could be written because I cannot determine the state of the data
before the circumstances occur.~~

I spent far to long writing the tests to make code climate happy


jsugarman (approved):
👍 LGTM

PR #2169, opened 2017-12-07
Expose retrial details

jsugarman:
**What**
expose details needed for injection of retrials

**Why**
So retrials can be injected with all data required
by CCR, to avoid rekeying by case workers.


app/interfaces/api/entities/ccr_claim.rb
```diff
@@ -35,12 +44,24 @@ def defendants_with_main_first
         object.defendants.order(created_at: :asc)
       end
 
+      def dummy_retrial_reduction
+        true
```
colinbruce: A little worried about exposing a dummy field... 
I take it you have a plan to remove this?

colinbruce (approved):
Other than one concern 👍

colinbruce (approved):
Cool 👍

PR #2171, opened 2017-12-11
MI fixes - Remove the TimeFromRejectToAuthCollector

colinbruce:
This was causing issues on the overnight stats run
Removing it as the graph it was used for is no longer required


jsugarman (approved):
👍 delete, delete, delete! :)

PR #2173, opened 2017-12-11
Prevent attempted duplication of injection_attempts

colinbruce:
This should prevent redrafts attempting to duplicate the source claims Injection attempts for the new clone claim


jsugarman (approved):
Ahh, yes. good catch 👍

PR #2174, opened 2017-12-11
Fix deploy breakage

jsugarman:
Found in `/var/log/upstart/advocatdefencepayments-container.log`
```
bundler: failed to load command: unicorn (/usr/local/bundle/bin/unicorn)
ArgumentError: tried to create Proc object without a block
  /usr/src/app/app/interfaces/api/entities/message.rb:7:in `lambda'
  /usr/src/app/app/interfaces/api/entities/message.rb:7:in `<class:Message>'
```


jsugarman:
Hotfix so am force merging this in colin's absence

PR #2175, opened 2017-12-11
Display retrial reduction

jsugarman:
**What**
Enable users to specify they are opting for a 
reduced rate on a retrial claim

**What**
When a trial goes to retrial remuneration may be requested
at a reduced rate. If not reduced the original trial remuneration
may be reduced retrospectively. However, reduced rates need
not be applied if the advocate did not work on the original trial
or if they prefer to reduce the rate on the trial.

### DIFF
--- a/features/page_objects/sections/retrial_section.rb
+++ b/features/page_objects/sections/retrial_section.rb
@@ -1,0 +1,10 @@
+require_relative 'common_date_section'
+
+class RetrialSection < SitePrism::Section
+  section :retrial_started_at, CommonDateSection, '#retrial_started_at'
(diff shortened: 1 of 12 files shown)


features/page_objects/sections/retrial_section.rb
```diff
@@ -0,0 +1,10 @@
+require_relative 'common_date_section'
+
+class RetrialSection < SitePrism::Section
+  section :retrial_started_at, CommonDateSection, '#retrial_started_at'
+  section :trial_concluded_on, CommonDateSection, '#retrial_concluded_at'
```
jsugarman: rename

colinbruce (approved):
:+1:

PR #2176, opened 2017-12-12
API V2 security

colinbruce:
### What?
Ensure that all API v2 endpoints are only accessible by case_worker api_keys

### Why?
At present, and until the API v2 is expanded, all end points should only be accessed by case_workers, previously any valid api_key could access them.


jsugarman (approved):
👍

PR #2177, opened 2017-12-13
Validate defendant uplift fees against number of defendants

jsugarman:
**What**
Ensure advocates cannot claim defendant uplift
miscellaneous fees that exceed the number of defendants.

**Why**
To ensure that no miscellaneous fees can be created that would
error when injected into CCR. CCR requires that the number of
defendant uplifts is less than or equal to the number of defendants 
on the claim. CCCD handles defendant uplifts via separate records
where the quantity is the number of defendants. A sum of these quantities
per uplift fee type must therefore not exceed the number of defendant
records on the claim.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,7 @@ Metrics/LineLength:
 Metrics/BlockLength:
   ExcludedMethods:
     - included
+    - class_methods
     - namespace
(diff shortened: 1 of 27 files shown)


colinbruce (approved):
:+1: LGTM

PR #2178, opened 2017-12-14
Improve Redraft logging

colinbruce:
Why?  We have some logging info stored, but insufficient to identify the issue

How?  Add error traps at all levels of the clone path and consistently pass them up the chain


jsugarman (approved):
👍 
will need to remove some of these test later i assume

PR #2179, opened 2017-12-14
Allocation fix

colinbruce:
Although this should not occur in live it is a fix for allocation where fees are missing

On a local, dev machine an ajax error was raised when trying to filter claims with null fee types

Uses `&.` to check, and exclude, null fees in all fee types


jsugarman (approved):
👍 LGTM

PR #2183, opened 2017-12-18
Trap redraft timeout error explicitly

colinbruce:
Add an explicit timeout error trap for when cloning takes too long

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -87,6 +87,8 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
     LogStuff.send(:error, 'ExternalUsers::ClaimsController',
                   action: 'clone',
                   claim_id: @claim.id,
+                  documents: @claim.documents.count,
+                  total_size: @claim.documents.sum(:document_file_size),
                   error: error.message) do
       'Failed to clone'
     end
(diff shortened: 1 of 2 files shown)


app/controllers/external_users/claims_controller.rb
```diff
@@ -78,14 +78,27 @@ def confirmation
...
+      draft = claim_updater.clone_rejected
+    end
+    LogStuff.send(:info, 'ExternalUsers::ClaimsController',
+                  action: 'clone',
+                  claim_id: @claim.id,
+                  documents: @claim.documents.count,
+                  total_size: @claim.documents.sum(:document_file_size)) do
+      'Redraft succeeded'
+    end
+
```
jsugarman, line 92: up to you but a bit of DRYing out if worthwhile

```
def logit error=nil, &block
 LogStuff.send(:info, 'ExternalUsers::ClaimsController',
                   action: 'clone',
                   claim_id: @claim.id,
                   documents: @claim.documents.count,
                   total_size: @claim.documents.sum(:document_file_size)
                   error: error
                  ) do
       yield if block_given?
     end
end
```

and call...

```
def clone_rejected
  clone stuff
  logit 'redraft succeeded'
rescue...
  logit error.message, 'redraft failed'
end
```

jsugarman (approved):
just a DRY suggestion but may not be worth it

PR #2185, opened 2017-12-18
Use retrial length for injection of retrials

jsugarman:
Retrials have separate trial length fields
against which basic fees for daily attendance
are validated. In addition, for injection, it
is the retrial length that should be used in
calculating the daily attendances sent to CCR.


colinbruce (approved):
👍

PR #2186, opened 2017-12-19
Update rejection handling

colinbruce:
- [x] update reject reasons list with revised list when a caseworker selects 'rejected' for AGFS .
- [x] Add validation to ensure case workers have selected a reason when submitting a `reject` response
- [x] display a mandatory a free form text box when 'other' is selected
- [x] NEW: Show other reasons on claim view 
image

- [x] Add interim disbursement choices for LGFS and make them conditional so that they only show for interim claims.
- [x] make ALL choices checkboxes so that more than one option can be selected.
- [x]  Update MI report to add reason_text after `other`
- [ ] consider mapping to 'improved' messages for providers, giving them clear instructions on what to do next. i.e. 'no indictment attached' would map to 'please redraft your claim and attach an indictment'. This can happen in a future iteration.

## Github Todo

- [x] design review
- [x] remove un-necessary migration `reasons`
- [x] extend tests, the coverage is at 94%
- [x] rebase PR

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -103,7 +103,7 @@ Metrics/MethodLength:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 193
+  Max: 196
 
 # Offense count: 7
 Metrics/PerceivedComplexity:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -100,10 +100,13 @@ Metrics/MethodLength:
     - 'scheduled_tasks/daily_stats_generator.rb'
     - 'spikes/supplier_import/migrate/20161020133346_laa_imported_suppliers.rb'
 
-# Offense count: 2
+# Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 196
+  Exclude:
+    - 'app/helpers/application_helper.rb'
+    - 'app/models/claims/cloner.rb'
+    - 'app/models/claims/state_machine.rb'
 
 # Offense count: 7
 Metrics/PerceivedComplexity:
--- a/app/controllers/case_workers/claims_controller.rb
+++ b/app/controllers/case_workers/claims_controller.rb
@@ -79,6 +79,7 @@ class CaseWorkers::ClaimsController < CaseWorkers::ApplicationController
   def claim_params
     params.require(:claim).permit(
       :state,
+      :reason_text,
       :additional_information,
       assessment_attributes: %i[
         id
@@ -94,7 +95,7 @@ class CaseWorkers::ClaimsController < CaseWorkers::ApplicationController
         disbursements
         vat_amount
       ]
-    ).merge(params.permit(:state_reason, :reason_text))
+    ).merge(params.permit(:state_reason))
   end
 
   def set_claims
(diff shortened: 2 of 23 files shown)


cleargif (approved):
👍

app/views/shared/_determinations_form.html.haml
```diff
@@ -1,54 +1,59 @@
...
-            = render partial: 'case_workers/claims/determination_fields', locals: { f: rf, claim: claim}
+          - elsif claim.redeterminations.any?
+            // REDETERMINATION
+            = render partial: 'shared/determination_amounts', locals: { claim: claim, determination: claim.redeterminations.last }
 
-        - elsif claim.redeterminations.any?
-          // REDETERMINATION
-          = render partial: 'shared/determination_amounts', locals: { claim: claim, determination: claim.redeterminations.last }
+          - else
+            // ELSE
```
jsugarman, line 34: leftover??

app/controllers/case_workers/claims_controller.rb
```diff
@@ -94,7 +95,7 @@ def claim_params
         disbursements
         vat_amount
       ]
-    ).merge(params.permit(:state_reason))
+    ).merge(params.permit(state_reason: []))
```
jsugarman, line 98: can this be added to main permit? i.e.

```
params.require(:claim).permit(
  :state,
  :state_reason: []
  ...
```

colinbruce: Good spot...
I just changed the data type and didn't look at the weird merging... :\

.rubocop_todo.yml
```diff
@@ -103,7 +103,7 @@ Metrics/MethodLength:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 193
+  Max: 196
```
jsugarman: maybe put a file exclude in for this file specifically. but either way.

jsugarman:
it look good but just thinking about whether or not it is worth converting to array datatype for the reason code and possible rename of column. This would convert existing data too.

```
def change
  change_column :claim_state_transitions, :reason_code, :text, array: true, default: [], using: "(string_to_array(reason_code, ','))"
end
```

jsugarman (approved):
👍 as discussed, for consistency with how other serialized data is handled, continue to rely on rails serialized array (YAML storage), rather than using Postgres array type.

app/controllers/case_workers/claims_controller.rb
colinbruce: reverted, but added reasons for the inconsistent approach

app/presenters/claim_csv_presenter.rb
```diff
@@ -96,7 +96,11 @@ def current_or_end_state
   end
 
   def state_reason_code
-    @journey.last.reason_code
+    @journey.last.reason_code.flatten.join(', ')
```
jsugarman, line 99: you could put a .humanize in here for the report, if this is the output for the report?! to change no_indictment to "No indictment" etc. or perhaps a `ClaimStateTransitionReason.get()` call for each

```
@journey.last.reason_code.humanize.flatten.join(', ')
```


```
@journey.last.reason_code.flat_map do |reason| 
  ClaimStateTransitionReason.get(reason) 
end.join(',')
```

colinbruce: They've always had the raw values in the MI report, I'm just showing multiples now!

PR #2187, opened 2017-12-21
Reset Sequence Ids to max id for table

jsugarman:
**What**
Synchronise all sequences for tables with id sequence

**Why**
Restoring a full dump of a database
can result in db sequence IDs being thrown
out of sink. This addition to the restore
task ensures that all tables that rely on
nextval to give them an ID will not break
after restore.


colinbruce (approved):
:trophy:

PR #2189, opened 2017-12-22
Replace use of Webrick with puma as local development server

jsugarman:
See [CVE-2009-4492](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2009-4492)


colinbruce (approved):
:+1:

PR #2192, opened 2018-01-04
Ensure that an assessment is always returned

jsugarman:
[Sentry errors](https://sentry.service.dsd.io/mojds/private-beta/issues/26495/) have shown that assessments are not
always being created and can raise an error
on summary pages as a result. Rather than workaround
this in the presenter this change ensures an assessment
is always returned (being built if necessary).


colinbruce (approved):
LGTM

PR #2194, opened 2018-01-05
Add basic files needed for CCLF endpoint

jsugarman:
Expose API endpoint for data to be injected into CCLF:

- [x] pattern
- [x] basic claim attributes
- [x] codeclimate


colinbruce (approved):
Looks like a good start! :)
👍

PR #2196, opened 2018-01-08
Fix MI Generator for legacy records

colinbruce:
There was a data-type mismatch in the tests.

This now covers the legacy data type and corrects the output

### DIFF
--- a/app/services/slack_notifier.rb
+++ b/app/services/slack_notifier.rb
@@ -1,9 +1,9 @@
 class SlackNotifier
-  def initialize
+  def initialize(channel = Settings.slack.channel)
     @slack_url = Settings.slack.bot_url
     @ready_to_send = false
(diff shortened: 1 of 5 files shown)


app/services/slack_notifier.rb
```diff
@@ -1,9 +1,9 @@
 class SlackNotifier
-  def initialize
+  def initialize(channel = Settings.slack.channel)
```
jsugarman, line 2: is this set to injectotron channel?
  
Ahh, you override anyway with cccd_development. perhaps that channel should be the default, but for another time perhaps

jsugarman (approved):
Cool 👍  - ideally we need to have reason code field respond as an array for legacy data as well but another TODO. Slack notifier could do with some more genericising, but another TODO

PR #2197, opened 2018-01-09
Refuse reasons update

colinbruce:
## What?
Add reasons to the refuse option for case_workers

## Acceptance criteria
- [x] Introduce a checklist of options when a caseworker has selected 'refused'
- [x] error messaging/handling for when nothing is entered in to the 'other' text field when 'other' is selected

## Next steps
* There will be a subsequent spike to assess the handling of this process and break it out into a view model (potentially) to better handle the error messages
* Work is underway to test merging the Refuse/Reject lists, therefore the technical debt of duplicating the JS code can be removed when the design is ready and the User research is complete.

### DIFF
--- a/spec/presenters/claim_csv_presenter_spec.rb
+++ b/spec/presenters/claim_csv_presenter_spec.rb
@@ -249,7 +249,7 @@ RSpec.describe ClaimCsvPresenter do
           end
         end
 
-        context 'refused with a multiple reasons' do
+        context 'refused with multiple reasons' do
           before do
             claim.refuse!(reason_code: ['no_rep_order', 'wrong_case_no'])
           end
--- a/spec/presenters/claim_csv_presenter_spec.rb
+++ b/spec/presenters/claim_csv_presenter_spec.rb
@@ -223,6 +223,56 @@ RSpec.describe ClaimCsvPresenter do
             end
           end
         end
+
+        context 'refused with a single reason as a string ' do
+          before do
+            claim.refuse!(reason_code: ['no_rep_order'])
+          end
+
+          it 'the refusal reason code should be reflected in the MI' do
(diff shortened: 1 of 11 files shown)


cleargif (approved):
The js looks fine as is.

spec/presenters/claim_csv_presenter_spec.rb
```diff
@@ -223,6 +223,56 @@
...
+          end
+
+          it 'the refusal reason code should be reflected in the MI' do
+            ClaimCsvPresenter.new(claim, view).present! do |csv|
+              expect(csv[0][11]).to eq('no_rep_order')
+            end
+          end
+        end
+
+        context 'refused with a multiple reasons' do
```
jsugarman: `with [a] multiple`

spec/services/claims/case_worker_claim_updater_spec.rb
```diff
@@ -113,7 +146,7 @@ module Claims
           expect(updater.claim.assessment.disbursements).to eq 0.0
         end
 
-        it 'if no state_reason are supplied' do
+        it 'if rejected and no state_reason are supplied' do
```
jsugarman, line 134: could use a local shared_examples with a parameter to remove the duplication here:
e.g. for stuff i am doing
```
#near top of spec block
shared_examples "creates injection attempts" do
  it 'returns true' do
    is_expected.to be true
  end

  it 'creates an injection attempt' do
    expect{ run! }.to change(InjectionAttempt, :count).by(1)
  end
end
```

and use:

```
    context 'when injection succeeded' do
      let(:json) { valid_json_on_success }
      include_examples 'creates injection attempts'
    end

    context 'when injection failed' do
      let(:json) { valid_json_on_failure }
      include_examples 'creates injection attempts'
    end

PR #2198, opened 2018-01-09
Add retrial started at

jsugarman:
**What**
Add retrial started at date.

**Why**
Is equivalent of first day of trial but
for retrials. Need only for LGFS interim claims
of grad fee type "retrial start".


colinbruce (approved):
👍

PR #2203, opened 2018-01-15
Litigator message queuing

colinbruce:
Implement a separate queue (using SNS rather than SQS) for Litigator claims

This allows the two separate backends to poll separate queues and not block each other.

Long term goal:  Use a single SNS point for both AGFS and LGFS and use a lambda query to update to SQS queues, this will give a single endpoint for users interested in stats around claims issued, while allowing individual end points for the back end importers


jsugarman (approved):
👍

PR #2205, opened 2018-01-16
Fix for refusing a claim with other, but no text

colinbruce:
There was an omission in the code that meant the wrong
values where being checked.


jsugarman (approved):
👍

PR #2206, opened 2018-01-17
Fix reject other

colinbruce:
The second, `refuse` text field was blanking the initial, `reject` text input field

This PR ensures they have unique ID and names on the page and then processes them into a single field for validation and storage


jsugarman (approved):
👍 tricky catch

PR #2207, opened 2018-01-17
Enable Ajax data injection error dismissing

jsugarman:
**What**
Enable case workers to dismiss/ignore individual data injection errors

**Why**
The fact an injection failed is primarily to inform
case worker decision making about work loads.
The time taken to process a claim that does not inject
is significantly longer.

### DIFF
--- a/app/presenters/concerns/injection_attempt_errorable.rb
+++ b/app/presenters/concerns/injection_attempt_errorable.rb
@@ -2,6 +2,14 @@ module InjectionAttemptErrorable
   extend ActiveSupport::Concern
 
   included do
+    def last_injection_attempt
+      claim.injection_attempts&.last
+    end
+
+    def injection_errors
+      last_injection_attempt.error_messages if last_injection_attempt&.active?
+    end
+
     def injection_error
       messages = injection_errors
       return unless messages.present?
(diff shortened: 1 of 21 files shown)


app/presenters/concerns/injection_attempt_errorable.rb
```diff
@@ -2,6 +2,14 @@ module InjectionAttemptErrorable
   extend ActiveSupport::Concern
 
   included do
+    def last_injection_attempt
+      claim.injection_attempts&.last
```
colinbruce, line 6: Should this be: `claim&.injection_attempts&.last`?
could this be called on a claim that has not been injected?

colinbruce (approved):
One minor observation but feel free to merge if you have considered it
:+1:

app/presenters/concerns/injection_attempt_errorable.rb
jsugarman: 👍 good spot, but as discussed, typically has_many relationships return empty array for non-existent children which would also respond to last with a nil return. One aside on this is that the presenter may be working on a Remote::Claim object which should 'behave' similarly on the has many relation but the `.last` does not - hence the `&.` on last

PR #2208, opened 2018-01-18
Working errors on refuse reject

colinbruce:
### What?
Display better errors to users when refusing or rejecting claims

### Why?
Case workers can reject or refuse claims, validation ensures this is completed, but currently the form does not re-render properly, selections are not completed and context errors are not displayed

### How? 
Extend the CaseWorkerClaimUpdater, add better granularity of error naming and descriptions
Update the view to display the users' choices and display context errors next to incorrect fields


jsugarman (approved):
👍

PR #2210, opened 2018-01-19
Add injection errors search filter value spec

jsugarman:
**What**
Expose active injection errors to case worker allocation page

**Why**
So the datatables jQuery plugin can use this data
to visually display to case workers which claims
did not inject into CCR/CCLF success fully and
enable (in a later PR) filtering by those claims.

### DIFF
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -108,7 +108,7 @@ module API
       end
 
       def injection_errored
-        object.injection_errors.blank? ? 0 : 1
+        object.injection_errors.present?.to_i
       end
     end
   end
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -18,6 +18,7 @@ module API
       expose :last_submitted_at_display
       expose :defendants
       expose :maat_references
+
       expose :filter do
         expose :disk_evidence
         expose :redetermination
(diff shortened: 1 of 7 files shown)


app/interfaces/api/entities/search_result.rb
```diff
@@ -98,6 +106,10 @@ def interim_disbursements
       def risk_based_bills
         ((risk_based_class_letter && contains_risk_based_fee).eql?(true) && is_submitted?).to_i
       end
+
+      def injection_errored
+        object.injection_errors.blank? ? 0 : 1
```
colinbruce: I'm pretty sure you could use `object.injection_errors.blank?.to_i` to match the other filters

colinbruce (approved):
Minor observation on setting the 0/1 value... but otherwise good to go. :+1:

app/interfaces/api/entities/search_result.rb
jsugarman: :thumbsup: will give it a go

jsugarman: yep, looks good
```ruby
[121] Pry:Claim-for-Crown-Court-Defence(main)> 'whatevs'.present?.to_i
=> 1
[122] Pry:Claim-for-Crown-Court-Defence(main)> ''.present?.to_i
=> 0
[123] Pry:Claim-for-Crown-Court-Defence(main)> nil.present?.to_i
=> 0
[124] Pry:Claim-for-Crown-Court-Defence(main)> false.present?.to_i
=> 0

```

PR #2212, opened 2018-01-22
Allow "from" attribute in injection response

jsugarman:
CCR and CCLF will identify themselves using this
attribute in the injection attempt responses. CCCD
will use this to output the responders name to 
slack. Eventually may add attribute to injection attempt 
to persist this data.


cibuanokpe (approved):
Looks good.

PR #2213, opened 2018-01-22
Display injection errors in allocation queue

cleargif:
- Datatables row callback
- amended styling to include borders on table
- specs


jsugarman (approved):
great 👍

PR #2216, opened 2018-01-25
Final Claim Litigator fee mapping

jsugarman:
**What**
Exposing JSON API bill for a CCLF litigator fee

**Why**
The Litigator Fee is one of 6 kinds of bills accepted by CCLF.


colinbruce (approved):
👍

PR #2218, opened 2018-01-26
Splitting submit claim pages

lostie:
[Pivotal Tracker](https://www.pivotaltracker.com/n/projects/2145949/stories/154602080)

These changes, to start with, only split case details and defendant details.

**NOTE:** This is an interim step to move away from numbered steps for the claim submission flows without interfering too much with the other existent flows that remain untouched.

**NOTE 2:** I've done a bit of tidy up in specs so all the validator specs are "tagged" as `validator` specs and automatically make use of the validation helper methods, mostly for DRY purposes.


colinbruce (approved):
Looks good 👍

PR #2219, opened 2018-01-29
Feedback email if anonymous

colinbruce:
[Pivotal Ticket](https://www.pivotaltracker.com/story/show/153989596)
### Why?
As a provider who is struggling to log in to the service, I need to be able to provide my email address on the feedback form, so that I can be contacted with help and assistance. 

### How?
Add an optional field to the feedback, and bug report views that, if we cannot identify the user by other means, prompts the user to add an email, if they wish.  This is the incorporated into the zendesk ticket

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -86,6 +86,16 @@ en:
     not_provided: *not_provided
     not_applicable: *not_applicable
 
+  activemodel:
+    attributes:
+      feedback:
+        heading: 'Help us improve this service'
+        callout: 'We will use your answers to the questions below to help us build a better service.'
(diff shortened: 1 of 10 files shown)


lostie:
LGTM 👍

lostie (approved):

config/locales/en.yml
```diff
@@ -86,6 +86,16 @@ en:
     not_provided: *not_provided
     not_applicable: *not_applicable
 
+  activemodel:
```
jsugarman, line 89: 👍

jsugarman (approved):
👍

PR #2220, opened 2018-01-30
Disk evidence on MI

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/152515358)

### DIFF
--- a/lib/extensions/boolean_extension.rb
+++ b/lib/extensions/boolean_extension.rb
@@ -3,11 +3,19 @@ module BooleanExtension
     def to_i
       1
     end
+
+    def to_yesno
+      'Yes'
+    end
   end
 
   module False
     def to_i
       0
+    end
+
+    def to_yesno
+      'No'
     end
   end
 end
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
👍

lib/extensions/boolean_extension.rb
```diff
@@ -3,11 +3,19 @@ module True
     def to_i
       1
     end
+
+    def to_yesno
```
lostie: **[not a blocker]** This seems more like a representation of a boolean value for the views rather than an extension for the `Boolean` type. Also, we're doing an implicit translation here, so if we needed to have this is _Cymraeg_ or something like that it would be a bit more difficult.

I would suggest for us to use a presenter where necessary.

lib/extensions/boolean_extension.rb
```diff
@@ -3,11 +3,19 @@ module True
...
+      'Yes'
+    end
   end
 
   module False
     def to_i
       0
     end
+
+    def to_yesno
```
lostie: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2220/files#r164720255)

lostie:
Apart from that minor comment LGTM 👍

PR #2222, opened 2018-01-30
Update AWS notifications

colinbruce:
Send Advocate claim confirmation messages to both
old SQS and new SNS until CCR has updated


jsugarman (approved):
👍 cool

PR #2226, opened 2018-02-01
Fix multi-page form continue flicker

jsugarman:
Add sleep before continue button click

Seems to be the most reliable way to ensure we move to 
the next page. Tried extending default max wait time and
using other wait methods from capybara with poor results.


lostie (approved):
LGTM 👍

PR #2227, opened 2018-02-01
Force redetermination before AWR

colinbruce:
[Pivotal Ticket](https://www.pivotaltracker.com/story/show/154280789)

**AS A** senior caseworker
**I WANT** written reasons claims to have first been through the redetermination process
**SO THAT** I don't have to spend lots of time rejecting these claims.

### acceptance criteria
only display the 'awaiting written reasons' radio button when the claim has already had the state 'redetermination'

redetermination-written-reasons-flow

### DIFF
--- a/app/services/claims/external_user_actions.rb
+++ b/app/services/claims/external_user_actions.rb
@@ -3,5 +3,9 @@ module Claims
     def self.all
       Settings.claim_actions
     end
+
+    def self.available_for(claim)
+      [] << Settings.claim_actions[claim.applicable_for_written_reasons? ? 1 : 0]
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


colinbruce:
Putting on hold for now -- awaiting business clarification on next steps

app/services/claims/external_user_actions.rb
```diff
@@ -0,0 +1,15 @@
...
+  class ExternalUserActions
+    def self.all
+      Settings.claim_actions
+    end
+
+    def self.available_for(claim)
+      if claim.applicable_for_written_reasons?
+        Settings.claim_actions
+      else
+        [] << Settings.claim_actions[0]
```
jsugarman: **Not a blocker.** Is it worth removing the specific reason that is not applicable, i.e. Request written reasons from the existing array, rather than select the one which is applicable. making it more flexible in future, as and if we ever need more reasons

```
Settings.claim_actions.reject{|action| action.match? 'written reasons')
```

jsugarman (approved):
one possible suggestion, otherwise 👍

PR #2228, opened 2018-02-02
Subsequent defendants do not prompt for Rep Order details

cleargif:
PT:
https://www.pivotaltracker.com/story/show/154057203

Why we doing this:
See PT.

What this PR does:
Binds a listner to trigger the `Add rep order` link after insert. nasty :(


jsugarman (approved):
Was looking up cocoon to see if they have a way that this should just work. when you add a defendant.

will check that out but feel free to get this in if tests pass

PR #2230, opened 2018-02-02
Add disbursement bills to CCLF JSON API

jsugarman:
Add disbursement bills to CCLF JSON API

To enable injection of disbursements into CCF


app/services/cclf/simple_bill_adapter.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    private
+
+    def case_type_adapter
+      @adapter ||= ::CCLF::CaseTypeAdapter.new(object.claim.case_type)
+    end
+
+    # delegate missing methods to object if it can respond
+    def method_missing(method, *args, &block)
```
lostie: I was going to suggest if we could use [SimpleDelegator](http://ruby-doc.org/stdlib-2.4.2/libdoc/delegate/rdoc/SimpleDelegator.html) here to the same effect, but I've seen [some comments](https://github.com/rails/rails/issues/23824#issuecomment-221171899) by DHH on the subject.

lostie (approved):
LGTM 👍

PR #2231, opened 2018-02-05
Remove SQS for AGFS

colinbruce:
We were temporarily sending both SQS and SNS notifications on completion of an AGFS claim
This is no longer needed


jsugarman (approved):
👍

PR #2232, opened 2018-02-07
Soft delete Cost judge type disbursements

jsugarman:
These disbursement types have been softly deleted on all environments
already as they are handled as LGFS miscellaneous fees in the app
and are handled as "OTHER" type bills in CCLF. They are not applicable
to LGFS Interim claims either.


colinbruce (approved):
:+1:

PR #2233, opened 2018-02-07
Update supplier number on disc_evidence_coversheet_builder

colinbruce:
## What? 
It has been highlighted by a user that, in certain circumstances, the disc evidence coversheet can be generated with the incorrect supplier number.

## How?
Update the generation code to use `claim.supplier_number` instead of the external_user's supplier number


jsugarman (approved):
👍

PR #2234, opened 2018-02-07
Injection vat considerations

jsugarman:
Supply fees and disbursements with or without vat and corresponding flag

CCLF adds VAT or not based on a flag/option given
to the case worker. CCCD injection into CCLF needs
to mimic this behavior.


jsugarman:
replaced with renamed and squashed version of branch

PR #2235, opened 2018-02-08
Apply agreed handling of VAT for CCLF

jsugarman:
What:
provide required CCCD JSON for CCLF to handle VAT

Why:
 - CCCD fees - CCCD will only supply net amounts (without VAT).
   CCLF will assume that VAT is not included and therefore apply VAT
   if the supplier number is vat registered.

- CCCD disbursements and travel expenses (i.e. CCLF disbursement
   equivalents)  - CCCD will supply the net amount and VAT amount separately.
   CCLF will use these to determine whether it needs to handle VAT as a
   separate disbursement, as below:

   If VAT amount equals the VAT rate % (20% currently) of the net amount
   then it can create one disbursement with a value of the net amount and
   mark the bill as VAT not included.

   If VAT amount is not equal the VAT rate % of the net amount then it
   needs to create 2 disbursements:
   * to represent the part on which VAT is not claimed.
   * to represent the part on which VAT is claimed.

### DIFF
--- a/app/interfaces/api/entities/cclf/adapted_disbursement.rb
+++ b/app/interfaces/api/entities/cclf/adapted_disbursement.rb
@@ -2,10 +2,7 @@ module API
   module Entities
     module CCLF
       class AdaptedDisbursement < AdaptedBaseBill
-        # NOTE: CCLF only requires the net amount and calculates vat based on rep order date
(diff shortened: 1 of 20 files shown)


colinbruce (approved):
:+1:

PR #2237, opened 2018-02-09
Provide apply vat flag to CCLF

jsugarman:
What: provides a flag indicating whether to apply vat or not

Why:
  `claim/apply_vat [boolean]`
  CCLF does not dynamically apply VAT based on supplier
  vat registration (unlike CCR), but purely based on a "VAT included"
  checkbox ticked, or not, by the case worker. This CCLF tick box should
  therefore be markable for any bill, other than disbursements, 
  based on the reverse logic of this flag - i.e. `!apply_vat`.


colinbruce (approved):
:+1: Will merge when green

cibuanokpe:
Passed

cibuanokpe (approved):
Passed

PR #2241, opened 2018-02-15
Spitting AGFS fee pages

lostie:
[Pivotal Tracker](https://www.pivotaltracker.com/story/show/154964860)

This includes the following changes:

- Split the advocate fees pages into `basic and fixed fees`, `misc fees`, `travel expenses`, `supporting evidence` and `additional information`
- Updates the JS sidebar component to calculate the total amount taking into account current stored values for fees, vat (before was just re-setting the calculator to 0 and calculating everything from scratch)
- Re-calculates total in advocate `claim` before validation if `fees` or `expenses` have been modified.


app/models/claims/calculations.rb
```diff
@@ -2,9 +2,15 @@ module Claims::Calculations
   def calculate_fees_total(category = nil)
     fees.reload
     if category.blank?
-      fees.map(&:amount).compact.sum
+      fees.map do |fee|
```
lostie: This piece of code is far from the ideal, but the whole fee amount calculation is a bit convoluted, so this was the only way I managed to do to calculate the fee amount:

- it tries to calculate the fee amount (there's cases where it does not calculate, but when it does, the method also updates the amount attribute. Arguably it should potentially just calculate an amount and that amount would have to be assigned out it, but I left it unchanged out of fear of affecting other parts of the process :S )
- it always returns the fee amount, mainly because there's times where the calculate_amount does not execute, but the fee amount itself is already set.

This should be revisited to try to simplify the convoluted way the way it works atm.

app/validators/claim/advocate_claim_validator.rb
```diff
@@ -72,11 +75,21 @@ def validate_defendant_uplifts
...
-    @record
-      .fees
-      .where.not(id: @record.misc_fees.select(&:marked_for_destruction?).map(&:id))
-      .defendant_uplift_sums
-      .values
-      .map(&:to_i).any? { |sum| sum + 1 > no_of_defendants }
+    defendant_uplifts_counts.map(&:to_i).any? { |sum| sum + 1 > no_of_defendants }
+  end
+
+  def defendant_uplifts
```
lostie, line 81: 📝 This piece of code was refactored as well, mainly because the way this validation was being performed was to retrieve all the _database_ **fees** associated with the claim.

What I found out once I've separated the fees pages was that when I was submitting the **fixed fees** they were assigned (in memory) to the `fixed_fees` association on the claim, so calling `fees` on the claim was returning only the pre-setup basic fees and not the fixed fees.

This new approach retrieves all the types of fees from memory, while maintaining the same check to verify the number of uplifts.

features/page_objects/claim_form_page.rb
```diff
@@ -121,7 +121,7 @@ def add_misc_fee_if_required
   end
 
   def attach_evidence(count = 1)
-    available_docs = Dir.glob "spec/fixtures/files/*.pdf"
+    available_docs = Dir.glob "#{Rails.root}/spec/fixtures/files/*.pdf"
```
lostie, line 124: 📝  Tiny change to support other browser drivers (namely `chromedriver`, which requires an absolute path to the files.

There's one more change across the features steps that would greatly improve the support for `chromedriver`:

`.trigger 'click'` is not supported in `chromedriver`. `.click` should be used instead (or redefine the click method to attempt `trigger 'click' first).

app/controllers/external_users/advocates/claims_controller.rb
```diff
@@ -13,6 +14,12 @@ def create
 
   private
 
+  def set_form_step
```
lostie, line 17: The intent of this method is to ensure the user is always sent to a valid page, rather than getting a 500 `Bad range` error when trying to submit a page for a step that no longer exists.

colinbruce (approved):
Looks good, small fix suggestion by code climate, but otherwise :+1:

app/models/claims/calculations.rb
jsugarman: I'd create an issue or put a `#TODO/FIXME` in the code to cover this

app/validators/claim/advocate_claim_validator.rb
jsugarman: does it handle fixed fees that are marked for destruction?

jsugarman: it does, just saw that below!

app/validators/claim/advocate_claim_validator.rb
```diff
@@ -72,11 +75,21 @@ def validate_defendant_uplifts
...
+    (@record.basic_fees + @record.fixed_fees + @record.misc_fees).select do |fee|
+      !fee.marked_for_destruction? &&
+        fee.fee_type &&
+        Fee::BaseFeeType.defendant_uplift_unique_codes.include?(fee.fee_type.unique_code)
```
jsugarman, line 85: There is a `defendant_uplift?` method that could be used here - see `DefendantUpliftable` module

app/validators/claim/advocate_claim_validator.rb
```diff
@@ -72,11 +75,21 @@ def validate_defendant_uplifts
...
-      .defendant_uplift_sums
-      .values
-      .map(&:to_i).any? { |sum| sum + 1 > no_of_defendants }
+    defendant_uplifts_counts.map(&:to_i).any? { |sum| sum + 1 > no_of_defendants }
+  end
+
+  def defendant_uplifts
+    (@record.basic_fees + @record.fixed_fees + @record.misc_fees).select do |fee|
+      !fee.marked_for_destruction? &&
+        fee.fee_type &&
```
jsugarman, line 84: is this check needed?

app/views/external_users/claims/supporting_documents/_disk_evidence.html.haml
```diff
@@ -1,5 +1,5 @@
 %h2.bold-medium
-  = "Evidence supplied on disk"
+  = t('.disk_evidence_title')
```
jsugarman, line 2: 👍

spec/validators/claim/base_claim_sub_model_validator_spec.rb
```diff
@@ -27,7 +27,7 @@
       @misc_fee = FactoryBot.create :misc_fee,:with_date_attended, claim: claim
       FactoryBot.create :date_attended, attended_item: @misc_fee
       claim.fees.map(&:dates_attended).flatten      # iterate through the fees and dates attended so that the examples below know they have been created
-      claim.form_step = :fees
+      claim.form_step = :fixed_fees
```
jsugarman: I think this spec is going to need work to make sure validation of each fee association is covered - currently this looks like it is testing fixed fee validation yet it has basic fees (fees of these two fee types should not co-exist).

app/validators/claim/advocate_claim_sub_model_validator.rb
```diff
@@ -14,19 +15,14 @@ def has_one_association_names_for_steps
...
-      fees: %i[
-        basic_fees
-        misc_fees
-        fixed_fees
-        expenses
-        messages
-        redeterminations
-        documents
-      ]
+      fixed_fees: %i[basic_fees fixed_fees],
```
jsugarman: i would rename this to something more unequivocal - e.g. basic_or_fixed_fees, main_fees - because it gets confusing to look at a form step that is also an association

features/page_objects/claim_form_page.rb
jsugarman: i'd create an issue and/or ticket for this

spec/validators/claim/base_claim_sub_model_validator_spec.rb
jsugarman: actually, i recommend above to change this step name which would resolve this

jsugarman (approved):
Aside from the name of the step for basic/fixed fees it looks good. A lot of work 👍 . I would get our testers to exercise the flow as much as possible since our happy path features tests are not sufficient - as discussed.

app/assets/javascripts/modules/external_users/claims/SideBar.js
```diff
@@ -49,11 +49,11 @@ moj.Modules.SideBar = {
     var self = this;
 
     this.totals = {
```
lostie, line 51: @cleargif could you review this JS changes and let me know if you're happy with them?

features/page_objects/claim_form_page.rb
lostie: I've created an issue for it: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2248

jsugarman:
superseded by LGFS branch

PR #2242, opened 2018-02-16
MI fix

colinbruce:
An edge case has occurred where all claim_state_transitions are ineligible, this meant that MI failed

This allows the `parsed_journey` method to return nothing and handle that in `present!` and then skip that row in `management_information_generator`

### DIFF
--- a/app/models/stats/management_information_generator.rb
+++ b/app/models/stats/management_information_generator.rb
@@ -24,10 +24,8 @@ module Stats
         csv << Settings.claim_csv_headers.map { |header| header.to_s.humanize }
         Claim::BaseClaim.active.non_draft.find_each do |claim|
           ClaimCsvPresenter.new(claim, 'view').present! do |claim_journeys|
-            if claim_journeys.any?
-              claim_journeys.each do |claim_journey|
-                csv << claim_journey
-              end
+            claim_journeys.each do |claim_journey|
+              csv << claim_journey
             end
           end
         end
--- a/app/models/stats/management_information_generator.rb
+++ b/app/models/stats/management_information_generator.rb
@@ -24,8 +24,10 @@ module Stats
         csv << Settings.claim_csv_headers.map { |header| header.to_s.humanize }
         Claim::BaseClaim.active.non_draft.find_each do |claim|
           ClaimCsvPresenter.new(claim, 'view').present! do |claim_journeys|
-            claim_journeys.each do |claim_journey|
-              csv << claim_journey
+            if claim_journeys.any?
+              claim_journeys.each do |claim_journey|
+                csv << claim_journey
+              end
             end
           end
         end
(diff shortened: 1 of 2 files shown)


app/models/stats/management_information_generator.rb
```diff
@@ -24,8 +24,10 @@ def generate_new_report
         csv << Settings.claim_csv_headers.map { |header| header.to_s.humanize }
         Claim::BaseClaim.active.non_draft.find_each do |claim|
           ClaimCsvPresenter.new(claim, 'view').present! do |claim_journeys|
-            claim_journeys.each do |claim_journey|
-              csv << claim_journey
+            if claim_journeys.any?
+              claim_journeys.each do |claim_journey|
```
lostie: **[nit pick/not a blocker]** If `claim_journeys` is an empty array I believe the `claim_journeys.any?` is irrelevant as the `.each` with immediately end. The only case where I can see this failing would be if `claim_journeys` was `nil`, but in that case even `.any?` would fail.

What are we trying to achieve with that extra check (I might lack context :P )?

colinbruce: Good points...
Claim journeys is returning `[nil]` but the `if` code block could be leftover from a previous iteration... I'll test and update

app/presenters/claim_csv_presenter.rb
```diff
@@ -22,7 +22,7 @@ def sorted_and_filtered_state_transitions
   def parsed_journeys
     journeys.map do |journey|
       @journey = clean_deallocations(journey)
-      Settings.claim_csv_headers.map { |method_call| send(method_call) }
+      Settings.claim_csv_headers.map { |method_call| send(method_call) } if @journey.any?
```
lostie, line 25: **[not a blocker]** Re-setting this instance variable per iteration of `journeys` array suggests (to me) that we should probably have some sort of `ClaimCSV::Journey` presenter that handles all _THE_ journey presentation, leaving this one to be something like:

```ruby
def parsed_journeys
  journeys.map do |journey|
    ClaimCSV::Journey.new(clean_deallocations(journey)).to_csv
  end
end
```

This is probably more an improvement in the design of the presenter to isolate the claim presentation from the journey presentation.

app/models/stats/management_information_generator.rb
lostie: I would probably just use `claim_journeys.compact` then if that is the case, but again, no blockers, your solution works as well ;)

lostie (approved):
LGTM 👍

app/presenters/claim_csv_presenter.rb
colinbruce: Good suggestion, I've added an issue as we need to get this bugfix out so they can generate their stats

PR #2247, opened 2018-02-20
Add PR template

jsugarman:
To assist standardising PRs.

Please make comments/suggestions on what you think the PR template should look like.


colinbruce (approved):
👍  
Good idea, lets merge it as is and we can amend once we start using it

PR #2250, opened 2018-02-21
Error messages: new style radios

cleargif:
PT:
https://www.pivotaltracker.com/story/show/155396815
https://www.pivotaltracker.com/story/show/154276163

Why are we doing this?
- new style radio / check boxes are not dispayed correctly
- all select dropdowns have `readonly` styles applied

What this PR does:
- error issues is related to the `field_with_errors` that rails adds in
- CSS specific to inputs

### DIFF
--- a/config/initializers/config.rb
+++ b/config/initializers/config.rb
@@ -13,4 +13,12 @@ Config.setup do |config|
   # Parse numeric values as integers instead of strings.
   #
   # config.env_parse_values = false
+
+  # This removes the wrapper `div.fields_with_errors` around
+  # elements where validations fail
+  # This wrapper breaks the new style Radio / Check boxes
+  ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
+    html_tag
+  end
+
 end
--- a/app/assets/stylesheets/moj/_forms.scss
+++ b/app/assets/stylesheets/moj/_forms.scss
@@ -4,6 +4,7 @@ form {
       border-left-width: 3px;
       .form-control{
         border-width: 2px;
+        border-color: $error-colour;
       }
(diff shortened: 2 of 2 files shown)


config/initializers/config.rb
```diff
@@ -13,4 +13,8 @@
   # Parse numeric values as integers instead of strings.
   #
   # config.env_parse_values = false
+  ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
+    html_tag.html_safe
```
jsugarman: just remove the `.html_safe` as discussed

jsugarman: also should have a go at running through negative/unhappy path manual tests as current feature tests do not cover errors to any great degree.

jsugarman (dismissed):

config/initializers/config.rb
lostie: Could you leave a `# NOTE:` in this change indicating why this is being made (basically that we don't want to rely on Rails form fields error display, but use the GDS one instead that is managed in ...[point the classe(s) where it is]).

Also, FYI, [here it is](http://guides.rubyonrails.org/configuring.html#configuring-action-view) the default implementation for Rails field error.

config/initializers/config.rb
```diff
@@ -13,4 +13,8 @@
   # Parse numeric values as integers instead of strings.
   #
   # config.env_parse_values = false
+  ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
```
jsugarman: Needs a note, as @lostie recommended

jsugarman (changes requested):

PR #2253, opened 2018-02-22
Evidence provision fees

colinbruce:
[Pivotal ticket](https://www.pivotaltracker.com/story/show/155067395)
The EPF has only two allowable amounts that are claimable. 45 and 90 (ex VAT). CCLF enforces this logic.

**As a litigator**
**When** i am claiming an evidence provision fee
**I should** be presented only with the valid amounts that are claimable
**So** i do not make claims for erroneous amounts

### DIFF
--- a/spec/factories/fee_types.rb
+++ b/spec/factories/fee_types.rb
@@ -135,7 +135,7 @@ FactoryBot.define do
         description 'Evidence provision fee'
         code 'XEVI'
         unique_code 'MIEVI'
-        quantity_is_decimal true
+        quantity_is_decimal false
       end
     end
 
--- a/spec/factories/fee_types.rb
+++ b/spec/factories/fee_types.rb
@@ -129,6 +129,14 @@ FactoryBot.define do
         unique_code 'MIUPL'
         quantity_is_decimal true
       end
+
+      trait :mievi do
+        lgfs
+        description 'Evidence provision fee'
+        code 'XEVI'
+        unique_code 'MIEVI'
+        quantity_is_decimal true
+      end
     end
 
     factory :fixed_fee_type, class: Fee::FixedFeeType do
--- a/spec/validators/fee/misc_fee_validator_spec.rb
+++ b/spec/validators/fee/misc_fee_validator_spec.rb
@@ -45,6 +45,34 @@ RSpec.describe Fee::MiscFeeValidator, type: :validator do
       end
     end
 
+    context 'when misc fee is an evidence provision fee' do
+      subject { fee.valid? }
+
+      let(:fee_type) { build :misc_fee_type, :mievi }
+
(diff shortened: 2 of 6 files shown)


colinbruce:
[finishes ##155067395]

colinbruce:
[finish #155067395]

spec/validators/fee/misc_fee_validator_spec.rb
```diff
@@ -45,6 +45,36 @@
       end
     end
 
+    context 'when misc fee is an evidence provision fee' do
+      let(:fee_type) { build :misc_fee_type, :mievi }
+
+      before { allow(fee).to receive(:fee_type).and_return(fee_type) }
+
+      context 'when the amount is set to' do
```
jsugarman: if you put a `subject { fee }` here, you could use `is_expected.to be_valid` below, but either way

spec/validators/fee/misc_fee_validator_spec.rb
```diff
@@ -45,6 +45,36 @@
...
+            let(:amount) { value }
+
+            it { expect(fee).to be_valid }
+          end
+        end
+
+        context 'a decimal' do
+          let(:amount) { '45.10' }
+
+          it { should_error_if_equal_to_value(fee, :amount, amount, 'incorrect_epf') }
```
jsugarman: just a suggestion, for consistency more than anything, you could get rid of the `context` and `let` and just do

```
it 'should error for decimals' do
  should_error_if_equal_to_value(fee, :amount, '45.10', 'incorrect_epf')
end
```

spec/validators/fee/misc_fee_validator_spec.rb
```diff
@@ -45,6 +45,36 @@
...
+        end
+
+        context 'zero' do
+          let(:amount) { '0' }
+
+          it { should_error_if_equal_to_value(fee, :amount, amount, 'incorrect_epf') }
```
jsugarman: as  comment above

spec/factories/fee_types.rb
```diff
@@ -129,6 +129,14 @@
         unique_code 'MIUPL'
         quantity_is_decimal true
       end
+
+      trait :mievi do
+        lgfs
+        description 'Evidence provision fee'
+        code 'XEVI'
+        unique_code 'MIEVI'
+        quantity_is_decimal true
```
jsugarman: `quantity_is_decimal ` should be false for this fee type

jsugarman (approved):
All good aside from some spec'ing suggestions. Also, as discussed, I think some api release documentation should go in at the same time [as here on sandbox](https://api-sandbox-adp.dsd.io/api/release_notes)

spec/factories/fee_types.rb
colinbruce: Good catch! copy-paste error :(

PR #2256, opened 2018-02-26
Certify and submit

colinbruce:
#### What
AS A provider who uses an API from my case management system to transfer cases into CCCD
I WANT to be able to 'certify and submit' my claim straight away,

#### Why
SO THAT I do not have to go through a lot of pages of my claim to be able to submit.

#### How
Added an optional button to the external_user display view for api claims
When a claim is from the API it should appear:
image

When it is web sourced, it should still look like:
image

#### Ticket
[Add `certify and submit` option for providers](https://www.pivotaltracker.com/story/show/155429494)

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -132,9 +132,6 @@ form{
 }
 
 .button-secondary.blue {
-  margin: 0;
-  padding-left: $gutter;
-  padding-right: $gutter;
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
LGTM 👍 on the view code.

You could possibly add a button presence check if there is already a feature test that follows the api --> submit route

cleargif (approved):
👍

PR #2257, opened 2018-02-26
New style radio button fixes

cleargif:
#### What
Implementing new style radio buttons on Claim overview page

#### Ticket
PT:
[Fix Claims list scheme select radio buttons]( https://www.pivotaltracker.com/story/show/155518693)

#### Why
The semantics of the radio buttons is tied to the display. 

#### How
Correctly implement the radio buttons as per GDS standards


colinbruce (approved):
:+1:

PR #2260, opened 2018-02-28
Add expenses as disbursement bills to interim claims

jsugarman:
#### What
Add expenses as disbursement bills to interim claims for a warrant fee

#### Why
Interim warrants do NOT have disbursements in CCCD,
but they do have expenses, and vice versa for other
interim fee types. CCLF conflates the two types and
treats all as disbursements.

#### How
Add adapted expenses to all interim claim bills so
that they will be exposed if they exist.


colinbruce (approved):
👍

PR #2261, opened 2018-03-01
Fix issue with interim fee validation

lostie:
#### What

The validation was being triggered incorrectly on a step of the submission flow where the interim fee is not required (e.g. case details).

This ensures the validation is only triggered when the claim is in a stage where that validation is actually required.

#### Ticket

[PT](https://www.pivotaltracker.com/story/show/155593565)

#### Why

The interim fee that was saved as a draft was being validated in the submission of the case details, which is the first step of the submission process and should not require the interim fee to be validated.

#### How

By ensuring that the interim fee is only validated at a stage in the submission where its expected to already been submitted (currently the **fees** page).

### DIFF
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -213,6 +213,15 @@ module Claim
       @form_step = step.nil? ? nil : step.to_sym
     end
 
+    def steps_range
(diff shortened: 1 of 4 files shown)


colinbruce (approved):
👍

jsugarman (approved):
👍

PR #2262, opened 2018-03-01
Ensure transfer claims have a transfer fee

jsugarman:
#### What
Ensure transfer claims have a transfer fee - confirmed by BA

#### Why
This validation logic seems to have gone missing
(not fired) in the page break up.

#### Ticket

[bug ticket 🐛 ](https://www.pivotaltracker.com/story/show/155628354)

#### How
The transfer claim validator should have a transfer_fee in its
fee step array to ensure presence of a transfer fee. This in addition
to the transfer fee itself being validated via the transfer claim submodel validator.


lostie (approved):
LGTM 👍

PR #2263, opened 2018-03-02
[AGFS Fee reform] Bill type selection page

lostie:
#### What

👨‍🏭  Introduce a new type of advocate claim (**advocate interim claim**) that can be selected from the 'What would you like to do?' page.

**NOTE:** ⚠️  This feature is controlled by a flag that allows us to deploy the code to production without actually making it available to the user until we decide to switch it. For the time being every single environment apart from **gamma/production** should have the feature available.

Also, as part of this work, took the opportunity to refactor the **ClaimTypesController** to simplify its business logic. 

#### Ticket

[PT](https://www.pivotaltracker.com/story/show/155393013)

#### How

Re-use the existing behaviour for starting a claim:
- the user is sent to the _claim types/schemes/bill types_ selection page
- whereas before, for advocate claims, the user was instantly redirected to the new advocate claim page (since there was only one type of claim for advocates), now the user is presented with a list of advocate claim types.
- when the feature for the :agfs_fee_reform is not active, the current behaviour was preserved and the user is instantly redirected to the new advocate claim page.


app/assets/stylesheets/moj/_HOTFIX_KILLME.scss
```diff
@@ -16,9 +16,11 @@
   @include clearfix;
 }
 .grid-row {
-  @include container;
-  margin-top:$gutter-half;
-  margin-bottom:$gutter-half;
+  &:not(.grid-row-fix) {
```
lostie, line 19: 🗒  Reason for this change, is mainly todo with the fact that the design seem to be using CSS directly from the GOV CSS, whereas we seem to be patching, so in order to preserve the current behaviour for all the pages except the one I have edited, I've come up with this solution.

app/assets/stylesheets/moj/_buttons.scss
```diff
@@ -37,3 +37,9 @@ p.file-upload-name {
 a.delete-draft {
   color: $red;
 }
+
+a.link-secondary, a.fix {
```
lostie, line 41: 🗒  Directly taken from the designs CSS.

app/assets/stylesheets/moj/_forms.scss
```diff
@@ -14,6 +14,11 @@ form {
       margin: 20px 0 !important;
     }
   }
+
+  .form-buttons * {
```
lostie, line 18: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2263/files#r171911547)

app/controllers/external_users/claim_types_controller.rb
```diff
@@ -1,53 +1,51 @@
 class ExternalUsers::ClaimTypesController < ExternalUsers::ApplicationController
   skip_load_and_authorize_resource
 
-  before_action :init_claim_types
-  before_action :set_claim_types_for_provider, only: [:selection]
+  before_action :set_available_fee_schemes_for_provider, only: %i[selection]
+  before_action :disable_primary_navigation, only: %i[selection]
```
lostie: 🗒  In the new design, the navigation is not shown for this page. Every other page still needs to show the menu. This is not an ideal solution IMO but that's the one that I came up with to avoid embracing a major refactor of the views just to achieve this. Hopefully this will be removed entirely once all the other pages also get updated.

app/helpers/external_users/claim_types_helper.rb
```diff
@@ -0,0 +1,15 @@
+module ExternalUsers
+  module ClaimTypesHelper
+    def bill_type_prompt_heading_for(user)
```
lostie: 🗒 This helper is not very pretty, a lot of elsifs is a sign, but based on the very specific requirement of displaying a given heading for each role, this is the best I came up with for now.

app/models/claim/advocate_interim_claim.rb
```diff
@@ -0,0 +1,5 @@
+module Claim
```
lostie, line 1: 📓  Literally a place holder, for now, since there isn't yet any logic associated with it apart from being able to select it in the "claim type selection page"

app/services/claims/context_mapper.rb
```diff
@@ -22,9 +22,25 @@ def available_schemes
       end
     end
 
+    def available_compreensive_schemes
```
lostie: 🗒  As I've mentioned before, I'm not particularly attached to these name, so happy to change this to something that makes more sense to everyone.

app/views/layouts/_contact_us_link.html.haml
```diff
@@ -1,3 +1,7 @@
 - if show_contact_us_link?
-  %p
-    = link_to("Should you require help with completing this form you can contact us.", contact_us_page_path)
\ No newline at end of file
+  %details
```
lostie, line 2: 🗒  I don't know if as a team we're aware of this, but this template seems to be rendered even in the main claims page, where's no form to complete, which looks weird to me 😅 .

config/settings.yml
```diff
@@ -93,6 +97,9 @@ remote_api_url: <%= (ENV['GRAPE_SWAGGER_ROOT_URL'] || 'http://localhost:3001') +
 
 # contact email address
 laa_contact_email: crowncourtdefence@legalaid.gsi.gov.uk
+advocates_email: advocates-fee@legalaid.gsi.gov.uk
```
lostie, line 100: Moved some static settings to configurations because they were set in a view ( 😱  ) until now.

This also makes it re-usable across the app.

lib/feature_flag.rb
```diff
@@ -0,0 +1,15 @@
+class FeatureFlag
```
lostie, line 1: 🗒  Very dumb feature flag management. Ideally we would be able to persist the features and their state in the database, so that we can switch them on demand (as an admin user). For now, until such requirement is raised, I've kept it as simple as possible.

spec/views/external_users/claim_types/selection_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-describe 'external_users/claim_types/selection.html.haml', type: :view do
+RSpec.describe 'external_users/claim_types/selection.html.haml', type: :view do
```
lostie, line 3: I'm unsure of the value this kind of specs brings us 😰 . Happy to discuss it with the team though.

colinbruce: I've never seen the value in view specs either...

colinbruce (approved):
👍

spec/support/test_helpers.rb
```diff
@@ -37,4 +37,12 @@ def no_error_for(record, attribute)
     return true if record.errors[attribute].empty?
     return false
   end
+
+  def with_env(env)
```
jsugarman, line 41: not really a blocker but could just mock the ENV
```
allow(ENV).to receive(:[]).with('ENV').and_return 'staging'
```

jsugarman (approved):
👍 as discussed - i think we should retain the terminology ClaimType for pure CCCD type code logic,  and keep bill_type terminology for the injection adapters and entities, scheme terminology no longer seems applicable beyond applying roles to user, providers (and perhaps some other objects).

If you are happy that the view specs testing of the available options for various roles is covered elsewhere i'd be tempted to remove the selection view spec entirely.

Also, as discussed, I'd consider stubbing the ENV var instead of changing the value in the relevant tests

app/controllers/external_users/claim_types_controller.rb
```diff
@@ -1,48 +1,48 @@
...
     else
       redirect_to external_users_claims_url, alert: error_message(:invalid_claim_types)
     end
   end
 
-  def set_available_fee_schemes_for_provider
+  def set_available_claim_types_for_provider
     context = Claims::ContextMapper.new(current_user.persona)
-    @available_fee_schemes = context.available_compreensive_schemes
+    @available_claim_types = context.available_compreensive_claim_types
```
jsugarman: available_compreensive_claim_types -> available_comprehensive_claim_types

app/services/claims/context_mapper.rb
```diff
@@ -22,8 +22,8 @@ def available_schemes
       end
     end
 
-    def available_compreensive_schemes
-      available_claim_types.map { |claim_type| compreensive_scheme_for(claim_type) }.compact
+    def available_compreensive_claim_types
```
jsugarman: available_compreensive_claim_types - typo

PR #2265, opened 2018-03-05
Increase allocation logging

colinbruce:
#### What
We are seeing a small number of cases that, despite the error handling in allocate!, end up in a state where the claim is allocated, but no link is made to a case worker.

#### Why
This results in the claim disappearing from the re/allocation tables and , co-incidentally, MI breaking.

#### How
Despite our best efforts, the error handling is not trapping this, so this increase in logging will hopefully allow us to track down the circumstances where this occurs.


jsugarman (approved):
👍

PR #2267, opened 2018-03-06
add interim fee validation for quantity

jsugarman:
#### What
Modify interim fee validation for quantity

#### Ticket

[pivotal ticket](https://www.pivotaltracker.com/story/show/155513265)

#### Why
Pages of prosecution evidence (PPE) is the quantity 
attribute for interim fees. There should be a minimum of 1
PPE both logically and for injection if a fee is to be claimed except
for the 'effective PCMH' and 'retrial new solicitor' interim fee. 
  ##### Injection *
  CCLF permits 0 PPE/quantity for  `Effective PCMH` and `Retrial new solicitor` but requires 
  a minimum of 1 for `re/trial start`.

#### How
Modify existing numericality check from 0 to 99999 to 1 to 99999

### DIFF
--- a/app/validators/fee/interim_fee_validator.rb
+++ b/app/validators/fee/interim_fee_validator.rb
@@ -11,7 +11,7 @@ class Fee::InterimFeeValidator < Fee::BaseFeeValidator
     if @record.is_disbursement? || @record.is_interim_warrant?
       validate_absence_or_zero(:quantity, 'present')
     else
-      validate_numericality(:quantity, 'numericality', 0, 99_999)
+      validate_numericality(:quantity, 'numericality', 1, 99_999)
     end
   end
 
--- a/app/validators/fee/interim_fee_validator.rb
+++ b/app/validators/fee/interim_fee_validator.rb
@@ -8,10 +8,12 @@ class Fee::InterimFeeValidator < Fee::BaseFeeValidator
   end
 
   def validate_quantity
-    if @record.is_disbursement? || @record.is_interim_warrant?
+    if quantity_must_be_zero?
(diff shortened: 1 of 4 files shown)


app/validators/fee/interim_fee_validator.rb
```diff
@@ -8,9 +8,11 @@ def self.fields
   end
 
   def validate_quantity
-    if @record.is_disbursement? || @record.is_interim_warrant?
+    if qauntity_must_be_zero?
```
lostie: **[typo]** _quantity_ instead of _qauntity_

spec/validators/fee/interim_fee_validator_spec.rb
```diff
@@ -1,7 +1,28 @@
 require 'rails_helper'
 
+# note: before validation hook sets nil to zero
+RSpec.shared_examples 'quantity numericality between' do |min, max|
```
lostie: **[not a blocker/nit pick]** If you're only using this shared example in this spec, you could just have them inside the top level of `RSpec.describe`:

```ruby
RSpec.describe Fee::InterimFeeValidator, type: :validator do
  shared_examples 'quantity numericality between' do |min, max|
    ...
  end

  ...
end

lostie (approved):
LGTM 👍

app/validators/fee/interim_fee_validator.rb
jsugarman: ta

spec/validators/fee/interim_fee_validator_spec.rb
jsugarman: cool, 👍

PR #2268, opened 2018-03-06
Delete old unused form builder class for date fields

jsugarman:
#### What
Delete old unused form builder class for date fields

#### Why
This class has been superceded by gov uk form
elements gems incarnation of date fields -
gov_uk_date_field.


colinbruce (approved):
👍 for removing obsolete code!

PR #2269, opened 2018-03-06
[AGFS fee reform] Case details page updates

lostie:
#### What

- Remove advocate category from case details page
- Change display of transfer court details:
  
  User needs to select if the case was transferred from another court, if yes, further transfer court details are displayed for the user to fill in (they're mandatory as long as it answered to the former question as **yes**)

- Implement a **Back** button for the claim submission forms

- Instructed advocates form field is displayed as a radio if there's 5 or less advocates for the current provider (in case there's only one it's automatically selected). If there's more than 5, a text field with autocomplete is displayed instead.

#### Ticket

💎  [AGFS final fee: case details page](https://www.pivotaltracker.com/story/show/155367082)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -141,10 +141,6 @@ module ApplicationHelper
     [number[0..3], number[4..6], number[7..10]].join(' ')
   end
 
-  def display_primary_navigation?
-    @primary_navigation_disabled != true
-  end
-
   def yes_no_options
     [%w[Yes true], %w[No false]]
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -144,4 +144,8 @@ module ApplicationHelper
   def display_primary_navigation?
     @primary_navigation_disabled != true
   end
+
+  def yes_no_options
+    [%w[Yes true], %w[No false]]
+  end
 end
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -223,6 +223,10 @@ module Claim
       steps_range.include?(step_index)
     end
 
+    def step_back?
+      previous_step.present?
+    end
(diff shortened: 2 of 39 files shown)


colinbruce (approved):
o_O :+1:

app/helpers/application_helper.rb
```diff
@@ -141,7 +141,7 @@ def format_phone_number(number)
     [number[0..3], number[4..6], number[7..10]].join(' ')
   end
 
-  def display_primary_navigation?
-    @primary_navigation_disabled != true
+  def yes_no_options
+    [%w[Yes true], %w[No false]]
```
jsugarman, line 145: **not a blocker** - could apply translations in the dictionary and use these

app/models/claim/base_claim.rb
```diff
@@ -188,6 +188,7 @@ def assessment
     before_validation do
       errors.clear
       destroy_all_invalid_fee_types
+      reset_transfer_details unless case_transferred_from_another_court
```
jsugarman: I would not call this that because there is a Claim::TransferDetail model that it could get confused with e.g. claim.transfer_detail

perhaps the rather unpleasantly long `reset_transfer_court_details`

app/models/claim/base_claim.rb
```diff
@@ -569,6 +607,11 @@ def set_allocation_type; end
 
     def destroy_all_invalid_fee_types; end
 
+    def reset_transfer_details
```
jsugarman: As naming comment above

jsugarman (approved):
👍 just the naming suggestion we discussed

app/helpers/application_helper.rb
lostie: I've thought about it and will probably change it in the future to do that 👍

PR #2271, opened 2018-03-07
Add disbursement VAT amount validation

jsugarman:
#### What
Add disbursement VAT amount validation

[ticket](https://www.pivotaltracker.com/story/show/155461780)

#### Why
To ensure VAT amount does not exceed 20%
(or current VAT rate) of the NET amount.

CCLF will reject such fees for injection.

#### How
extend existing validators

### DIFF
--- a/app/validators/base_validator.rb
+++ b/app/validators/base_validator.rb
@@ -181,10 +181,16 @@ class BaseValidator < ActiveModel::Validator
     validate_amount_less_than_item_max(field)
   end
 
-  def validate_vat_numericality(field, lower_than_field:, allow_blank: true)
-    validate_presence_and_numericality(field, minimum: 0, allow_blank: allow_blank)
-    validate_amount_greater_than(field, lower_than_field, 'greater_than')
-    add_error(:vat_amount, 'max_vat_amount') if (@record.vat_amount || 0) > max_vat_amount
+  def validate_vat_less_than_max(vat_attribute, net_attribute)
+    vat_amount = @record.send(vat_attribute) || 0
+    net_amount = @record.send(net_attribute) || 0
+    add_error(vat_attribute, 'max_vat_amount') if vat_exceeds_max?(vat: vat_amount, net: net_amount)
+  end
+
+  def validate_vat_numericality(attribute, lower_than_field:, allow_blank: true)
+    validate_presence_and_numericality(attribute, minimum: 0, allow_blank: allow_blank)
+    validate_amount_greater_than(attribute, lower_than_field, 'greater_than')
+    validate_vat_less_than_max(attribute, lower_than_field)
   end
 
   def validate_two_decimals(field)
@@ -193,11 +199,8 @@ class BaseValidator < ActiveModel::Validator
     add_error(field, 'decimal') unless value == rounded
   end
 
-  def max_vat_amount
-    VatRate.vat_amount(net_amount || 0, @record.claim.vat_date, calculate: true)
-  end
-
-  def net_amount
-    @record.respond_to?(:net_amount) ? @record.net_amount : @record.amount
+  def vat_exceeds_max?(vat:, net:)
+    max_vat = VatRate.vat_amount(net, @record.claim.vat_date, calculate: true)
+    vat.round(2) > max_vat.round(2)
   end
 end
--- a/app/validators/base_validator.rb
+++ b/app/validators/base_validator.rb
@@ -184,11 +184,20 @@ class BaseValidator < ActiveModel::Validator
   def validate_vat_numericality(field, lower_than_field:, allow_blank: true)
     validate_presence_and_numericality(field, minimum: 0, allow_blank: allow_blank)
     validate_amount_greater_than(field, lower_than_field, 'greater_than')
+    add_error(:vat_amount, 'max_vat_amount') if (@record.vat_amount || 0) > max_vat_amount
   end
 
   def validate_two_decimals(field)
     value = @record.__send__(field)
     rounded = value.round(2)
     add_error(field, 'decimal') unless value == rounded
+  end
+
+  def max_vat_amount
+    VatRate.vat_amount(net_amount || 0, @record.claim.vat_date, calculate: true)
+  end
+
+  def net_amount
+    @record.respond_to?(:net_amount) ? @record.net_amount : @record.amount
   end
 end
--- a/spec/controllers/vat_rates_controller_spec.rb
+++ b/spec/controllers/vat_rates_controller_spec.rb
@@ -13,13 +13,10 @@ require 'rails_helper'
 
 RSpec.describe VatRatesController, type: :controller do
 
-  before(:all) do
-    @vr1 = FactoryBot.create :vat_rate, effective_date: Date.new(2000, 1, 1),  rate_base_points: 1750
-    @vr2 = FactoryBot.create :vat_rate, effective_date: Date.new(2011, 4, 1),  rate_base_points: 2000
-  end
-
-  after(:all) do
-    VatRate.destroy( [ @vr1.id, @vr2.id ] )
+  before do
+    VatRate.delete_all
+    create(:vat_rate, effective_date: Date.new(2000, 1, 1), rate_base_points: 1750)
+    create(:vat_rate, effective_date: Date.new(2011, 4, 1), rate_base_points: 2000)
   end
 
   describe 'GET vat' do
--- a/spec/models/claims/totals_spec.rb
+++ b/spec/models/claims/totals_spec.rb
@@ -70,8 +70,6 @@ RSpec.describe Claim, type: :model do
       claim.reload
     end
 
-    # before { claim.reload }
-
     context 'AGFS claim' do
       subject(:claim) { create(:claim) }
 
--- a/spec/models/claims/totals_spec.rb
+++ b/spec/models/claims/totals_spec.rb
@@ -1,114 +1,118 @@
 require 'rails_helper'
 
 RSpec.describe Claim, type: :model do
-  subject { create(:claim) }
-  let(:expenses) { [3.5, 1.0, 142.0].each { |amount| create(:expense, claim_id: subject.id, amount: amount) } }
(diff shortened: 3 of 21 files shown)


spec/validators/expense_validator_spec.rb
```diff
@@ -43,17 +43,30 @@
...
+        expense.vat_amount = 2
         expect(expense).to be_valid
       end
 
       it 'is invalid if greater than the total amount' do
-        expense.amount = 11
-        expense.vat_amount = 11.5
+        expense.amount = 100
+        expense.vat_amount = 100.01
         expect(expense).not_to be_valid
```
jsugarman: `be_invalid`

spec/presenters/assessment_presenter_spec.rb
```diff
@@ -1,20 +1,21 @@
...
-  let(:claim)       { FactoryBot.create :claim, apply_vat: true }
-  
+  let(:claim) { FactoryBot.create :claim, apply_vat: true }
+  let(:presenter) { AssessmentPresenter.new(claim.assessment, view) }
+
   context 'currency fields' do
-    it 'should format currency amount' do
-      claim.assessment.update_values(1452.33, 2455.77, 1505.24)
-      presenter = AssessmentPresenter.new(claim.assessment, view)
+    let(:thousand_currency_regex) { /£\d,\d{3}\.\d{2}/ }
```
jsugarman: `currency_pattern`

app/validators/base_validator.rb
```diff
@@ -181,14 +181,26 @@ def validate_presence_and_numericality(field, minimum: 0, allow_blank: false)
...
 
   def validate_two_decimals(field)
     value = @record.__send__(field)
     rounded = value.round(2)
     add_error(field, 'decimal') unless value == rounded
   end
+
+  def vat_exceeds_max?(vat:, net:)
+    max_vat = VatRate.vat_amount(net, @record.claim.vat_date, calculate: true)
+    vat > max_vat
```
jsugarman: check rounding issues, might need...
```
vat.round(2) > max_vat.round(2)
```

spec/models/claims/totals_spec.rb
```diff
@@ -1,114 +1,118 @@
...
+      VatRate.delete_all
+      create(:vat_rate, :for_2011_onward)
+      [3.5, 1.0, 142.0].each do |amount|
+        create(:expense, claim_id: claim.id, amount: amount, vat_amount: 2)
+      end
+      claim.reload
+    end
 
-    before { subject.reload }
+    # before { claim.reload }
```
jsugarman: remove

spec/controllers/vat_rates_controller_spec.rb
```diff
@@ -13,13 +13,10 @@
...
 
-  before(:all) do
-    @vr1 = FactoryBot.create :vat_rate, effective_date: Date.new(2000, 1, 1),  rate_base_points: 1750
-    @vr2 = FactoryBot.create :vat_rate, effective_date: Date.new(2011, 4, 1),  rate_base_points: 2000
-  end
-
-  after(:all) do
-    VatRate.destroy( [ @vr1.id, @vr2.id ] )
+  before do
+    VatRate.delete_all
```
lostie, line 17: ❓  Don't have much knowledge or background on this (so bare with me 😅 ), but since this is now a `before(:each)` block what's the reason for deleting all vat rates each time? (they should be transitional)

jsugarman: No worries, there is one seeded VatRate for the whole suite (see before(:suite) in rails helper) so to make the test more explicit/clear in this spec, in particular,  I have removed it so you can be sure you only have the VAT rates the test defines.

I have removed the before(:all) because it is [dodgy](https://makandracards.com/makandra/11507-using-before-all-in-rspec-will-cause-you-lots-of-trouble-unless-you-know-what-you-are-doing) and causes flickers from my experience. could put the `delete_all` in a before(:all), but i think any performance/speed boost we may get from it carries the aforementioned risk

spec/models/claims/totals_spec.rb
```diff
@@ -127,27 +129,27 @@
           claim.fees << create(:misc_fee, rate: 10.5, quantity: 3)
           claim.fees << create(:misc_fee, rate: 6.25, quantity: 1)
           claim.save!
-          expect_totals_to_be(claim, 0.0, 0.0, 37.75, 6.61, 0.0, 0.0)  # VAT £6.61 calculated using the test VAT rate of 17.5%
+          expect_totals_to_be(claim, 0.0, 0.0, 37.75, 7.55, 0.0, 0.0) # VAT £6.61 calculated using the test VAT rate of 20.0%
```
lostie, line 157: 💭  I know this is not your initial change, but the first feeling I get looking at this expectation is that I don't have a clue what is being asserted. Having looked at the actual implementation of the  expectation, I would argue that `DRY` in this case its not necessary good or clear.

I would suggest, at least, changing the interface to be something more expressive, for example (not necessary this, but you get the idea):

```ruby
expect(claim).to have_totals(
  expenses_total: 9.99,
  expenses_vat:  2.0,
  fees_total: 37.75,
  fees_vat: 7.55,
  disbursements_total: 0.0,
  disbursements_vat: 0.0,
  total: 9.99 + 37.75 + 0.0,
  vat_amount: 2.0 + 7.55 + 0.0
)
```

spec/models/claims/totals_spec.rb
```diff
@@ -1,114 +1,116 @@
...
       end
     end
   end
 
   context 'expenses vat' do
-    let!(:expenses) { [3.5, 1.0, 142.0].each { |amount| create(:expense, claim_id: subject.id, amount: amount, vat_amount: 2) } }
-
-    before { subject.reload }
+    before do
+      VatRate.delete_all
```
lostie, line 89: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2271/files#r173426782)

spec/models/vat_rate_spec.rb
```diff
@@ -11,16 +11,12 @@
...
-    @vr1 = FactoryBot.create :vat_rate, effective_date: 1.year.ago, rate_base_points: 2225
-    @vr2 = FactoryBot.create :vat_rate, effective_date: 3.years.ago, rate_base_points: 800
-    @vr3 = FactoryBot.create :vat_rate, effective_date: 10.years.ago, rate_base_points: 1750
-  end
-
-  after(:all) do
-    VatRate.destroy( [ @vr1.id, @vr2.id, @vr3.id ] )
+RSpec.describe VatRate do
+  before do
+    VatRate.delete_all
```
lostie, line 16: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2271/files#r173426782)

spec/models/claims/totals_spec.rb
jsugarman: indeed 😢 , i changed to get it passing (should have been a separate commit to get tests passing) and have done some spec refactoring generally. It would be good to implement as you  suggest as above, but started to get refactoring creep

spec/presenters/assessment_presenter_spec.rb
```diff
@@ -1,20 +1,21 @@
 require 'rails_helper'
 
 RSpec.describe AssessmentPresenter do
-  
-  let(:claim)       { FactoryBot.create :claim, apply_vat: true }
-  
+  let(:claim) { FactoryBot.create :claim, apply_vat: true }
+  let(:presenter) { AssessmentPresenter.new(claim.assessment, view) }
```
lostie: **[not a blocker/nit pick]** `AssessmentPresenter` is the class being tested so this could just be:

```ruby
subject(:presenter) { described_class.new(claim.assessment, view) }
```

spec/models/claims/totals_spec.rb
jsugarman: ...and ideally it should test the individual values as separate tests

spec/presenters/redetermination_presenter_spec.rb
```diff
@@ -1,19 +1,20 @@
 require 'rails_helper'
 
 describe RedeterminationPresenter do
-
-  let(:claim)           { FactoryBot.create :claim, apply_vat: true }
-  let(:rd)              { FactoryBot.create :redetermination, fees: 1452.33, expenses: 2455.77, disbursements: 2123.55, claim: claim }
-  let(:presenter)       { RedeterminationPresenter.new(rd, view) }
+  let(:claim) { create :claim, apply_vat: true }
+  let(:rd) { create :redetermination, fees: 1452.33, expenses: 2455.77, disbursements: 2123.55, claim: claim }
+  let(:presenter) { RedeterminationPresenter.new(rd, view) }
```
lostie: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2271/files#r173431420)

spec/presenters/assessment_presenter_spec.rb
jsugarman: 👍

spec/presenters/redetermination_presenter_spec.rb
jsugarman: 👍

spec/rails_helper.rb
```diff
@@ -128,7 +128,7 @@
 
   config.before(:suite) do
     DatabaseCleaner.clean_with(:truncation)
-    FactoryBot.create :vat_rate, effective_date: Date.new(1960, 1, 1), rate_base_points: 1750
+    FactoryBot.create(:vat_rate, :for_2011_onward)
```
lostie, line 131: Just out of interest, could you elaborate on the reason for requiring this VAT rate to be set for the suite? 🙇

jsugarman: I did not put it in, but it does affect a lot of specs as fee hooks require it for calculating totals.

lostie: 😱  => 🏃 😅

lostie (approved):
LGTM

PR #2273, opened 2018-03-09
Update Sidekiq gem

lostie:
#### What

The new version [no longer requires](https://github.com/mperham/sidekiq/blob/v4.2.9/Changes.md#420) `Sinatra` for the Web UI, it runs directly on `Rack`, hence the reason for removing `Sinatra` from the `Gemfile`, and updating the gem to its most recent **MINOR** version.

#### Why

Github had flagged a security vulnerability with the version of `rack-protection` used by `Sinatra`, which was being used by `Sidekiq::Web`.

screen shot 2018-03-08 at 12 39 34


colinbruce (approved):
:+1:
Removing dependencies... Nice!

lostie:
🔬 Tested in dev 💚

PR #2274, opened 2018-03-12
[AGFS fee reform] Add controller and views for advocate interim claim

lostie:
#### What

Also:

- Updated cancancan authorizations by adding the new `AdvocateInterimClaim` to the types of claim an advocate or admin can have access to
- Refactored base claims controller for external users to:
  - initialize the claim for new/create actions without needing the inherited controllers todo so
  - ensure the new/create actions are both authorized (until now a user knowing the URL could still access some of those pages)
  - update edit/update before_action to ensure only editable claims are able to be updated (so far only edit had this check, if anyone submitted an update request directly without using the web interface, it would still update an un-editable claim)

#### Ticket

[AGFS interim fee: case details page](https://www.pivotaltracker.com/story/show/155395810)


app/controllers/external_users/claims_controller.rb
```diff
@@ -16,6 +18,7 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
 
   before_action :set_and_authorize_claim, only: %i[show edit update unarchive clone_rejected destroy summary
                                                    confirmation show_message_controls messages disc_evidence]
+  before_action :redirect_unless_editable, only: %i[edit update]
```
lostie, line 21: Previously this was only checked for the `:edit` action, anyone that tried to update the claim directly using any HTTP client would be able to update it. This prevents that and keeps the behaviour in line between `edit` and `update`

app/controllers/external_users/claims_controller.rb
```diff
@@ -132,18 +135,15 @@ def unarchive
   end
 
   def new
+    @claim = resource_klass.new
+    authorize! :new, @claim
```
lostie, line 139: Previously this action was not authorised which meant that any user logged in would be able to access this page as long as they knew the URL. This ensures the action is correctly authorised and only users with permissions to manage this type of claim can do so.

app/controllers/external_users/claims_controller.rb
```diff
@@ -155,6 +155,8 @@ def edit
   end
 
   def create
+    @claim = resource_klass.new(params_with_external_user_and_creator)
+    authorize! :create, @claim
```
lostie, line 159: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2274/files#r174404793)

app/controllers/external_users/advocates/claims_controller.rb
```diff
@@ -1,15 +1,7 @@
...
-  def new
-    @claim = Claim::AdvocateClaim.new
-    super
-  end
-
-  def create
-    @claim = Claim::AdvocateClaim.new(params_with_external_user_and_creator)
-    super
-  end
+  resource_klass Claim::AdvocateClaim
```
lostie, line 4: There was a lot of copy/paste behaviour among all of the external users claims controllers only to initialise the claim with the appropriate claim type.

This change allows us to define the type of class for the controller's resource (mainly because both model and controller fall out of the normal `Rails` conventions and the controller can't determine the model class correctly), meaning the `ClaimsController` can use that definition to initialised the resource for all the claim types.

app/models/ability.rb
```diff
@@ -97,11 +97,11 @@ def can_administer_provider(persona)
   end
 
   def can_manage_litigator_claims(persona)
-    can_manage_own_claims_of_class(persona, [Claim::LitigatorClaim, Claim::InterimClaim, Claim::TransferClaim])
+    can_manage_own_claims_of_class(persona, Claim::BaseClaim.lgfs_claim_types)
```
lostie, line 100: Delegate the available `agfs/lgfs` claim types to the `Claim` class itself so we don't need to repeat this list all over the code.

config/locales/en.yml
```diff
@@ -138,6 +138,10 @@ en:
               blank: "can't be blank"
             claim_ids:
               blank: "can't be blank"
+        claim/advocate_interim_claim:
```
jsugarman: As discussed, despite not being a rails convention for handling error message translation I think we should adhere to it for claim errors.

spec/validators/claim/advocate_interim_claim_web_validations_spec.rb
```diff
@@ -0,0 +1,390 @@
...
+      let(:one_other_representation_order_attrs) {
+        valid_one_other_representation_order_attrs.merge(
+                                         representation_order_date_dd: (release_date - 1.day).day.to_s,
+                                         representation_order_date_mm: (release_date - 1.day).month.to_s,
+                                         representation_order_date_yyyy: (release_date - 1.day).year.to_s)
+      }
+
+      specify {
+        is_expected.to be_invalid
+        expect(claim.errors[:base]).to match_array(['You cannot submit an interim bill for claims with the earliest representation order before 1 April 2018.'])
```
jsugarman: suggestions: could use include to make the test a little less rigid

```
include 'You cannot submit an interim bill for claims with the earliest representation order before 1 April 2018.'
```

spec/presenters/error_presenter_spec.rb
```diff
@@ -1,24 +1,26 @@
...
         end
       end
 
       context 'fieldname not present' do
-        it 'should return 99999 ' do
-          expect(presenter.send(:generate_sequence, 'nokey')).to eq 99999
+        let(:field) { 'nokey' }
+
+        # NOTE: no idea what this expected value is!?
+        it 'returns 99999' do
```
jsugarman: i think this was to ensure error sequences get a value even if there is no such message in the error_messages.yml

better description for test might be
```
it 'returns default sequence' do
```

app/controllers/external_users/advocates/claims_controller.rb
jsugarman: 👍

jsugarman (approved):
👍 controllers needed a good clean, nice! just the error message convention issue discussed

config/locales/error_messages.en.yml
```diff
@@ -52,6 +52,10 @@ base:
     long: The quantity of defendant uplifts exceeds the number of defendants
     short: Too many defendant uplifts
     api: Defendant uplift sums must match number of defendants
+  unclaimable:
```
lostie, line 55: 👹  ♻️  @jsugarman

jsugarman: 👍 me-no-speak-emoji :)

PR #2278, opened 2018-03-14
Transfer fee ppe validation

jsugarman:
#### What
Apply PPE requirement validations

#### Ticket

[Conditional requirement for PPE](https://www.pivotaltracker.com/story/show/155773272)

#### Why
Like interim fees, some transfer fees require PPE to be provided - while
for others its optional - in order to successfully inject into CCLF, because
the calculation of the bill amount is based on it.

#### How
Use `transfer brain` virtual model

### DIFF
--- a/spec/validators/fee/transfer_fee_validator_spec.rb
+++ b/spec/validators/fee/transfer_fee_validator_spec.rb
@@ -3,10 +3,10 @@ require 'rails_helper'
 RSpec.describe Fee::TransferFeeValidator, type: :validator do
   include_context 'force-validation'
 
-  let(:claim) { build :transfer_claim }
-  let(:fee) { build :transfer_fee }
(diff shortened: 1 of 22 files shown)


spec/validators/fee/transfer_fee_validator_spec.rb
```diff
@@ -27,21 +27,43 @@
...
+    it 'validates absence of case numbers' do
       fee.case_numbers = 'T20150111,T20150222'
-      expect(fee).not_to be_valid
+      expect(fee).to be_invalid
+    end
+  end
+
+  describe '#validate_quantity' do
+    let(:fee) { build(:transfer_fee, claim: claim) }
+
```
jsugarman, line 53: extra whitespace

colinbruce (approved):
👍

PR #2279, opened 2018-03-15
Bump the version of postgres to 9.6

colinbruce:
#### What
Bump the version of postgres to 9.6
#### Why
9.4 was no longer included in Ubuntu and, after the update on gamma,
the builds would no longer install it

### DIFF
--- a/script/db_dump.rb
+++ b/script/db_dump.rb
@@ -40,7 +40,7 @@ end
 
 def install_postgres(ssh)
   ssh.exec! 'sudo docker exec advocatedefencepayments apt-get update'
-  ssh.exec! 'sudo docker exec advocatedefencepayments apt-get -y install postgresql-9.4'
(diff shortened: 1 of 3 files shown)


lostie:
🔔  Seems like the following files are also trying to install it (even though I'm not sure what they're used for):

- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/script/db_static_data_upload.sh#L42
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/script/db_upload.rb#L43

colinbruce:
Good catch, thanks @lostie.  I just fixed the one that was broken for me! :(

lostie (approved):
LGTM 👍 

(Seems odd though that only those scripts have a specific reference to the `postgresql` version we're using, given that we have `CI`, `Docker` for example that should require it to be setup with a specific version for consistency)

PR #2281, opened 2018-03-19
Implement the warrant fee step for advocates interim claims

lostie:
#### What

- Enhances the warrant fee validator to verify that:
  - warrant needs to have been issued at least 3 months ago in order to be claimed
  - issued date needs to be on or after the earliest representation order
- Eligible advocate categories are now conditional on the claim being under the new AGFS fee reform or not. If they are, a different list of advocate categories is used. If they're not the existent list is used as default.
- Preserve existent behaviour with checks for `interim?` (which until now meant `lgfs? && interim?` given that the only claim that was considered _interim_ was a LGFS one). Until we are clear of what behaviour should apply to the new _advocate interim claim_, this changes preserve the existent behaviour by also performing and extra check to verify if the claim is _LGFS_.

#### Ticket

[AGFS interim fee: Interim fee page](https://www.pivotaltracker.com/story/show/155396675)

### DIFF
--- a/app/controllers/external_users/advocates/interim_claims_controller.rb
+++ b/app/controllers/external_users/advocates/interim_claims_controller.rb
@@ -2,4 +2,9 @@ class ExternalUsers::Advocates::InterimClaimsController < ExternalUsers::ClaimsC
   skip_load_and_authorize_resource
 
   resource_klass Claim::AdvocateInterimClaim
+
+  def build_nested_resources
+    @claim.build_warrant_fee if @claim.warrant_fee.nil?
+    super
+  end
 end
--- a/app/models/claim/advocate_interim_claim.rb
+++ b/app/models/claim/advocate_interim_claim.rb
@@ -6,6 +6,10 @@ module Claim
                    unless: proc { |c| c.disable_for_state_transition.eql?(:all) }
     validates_with ::Claim::AdvocateInterimClaimSubModelValidator
 
+    has_one :warrant_fee, foreign_key: :claim_id, class_name: 'Fee::WarrantFee', dependent: :destroy, inverse_of: :claim
+
+    accepts_nested_attributes_for :warrant_fee, allow_destroy: false
+
     before_validation do
       set_supplier_number
     end
@@ -63,6 +67,22 @@ module Claim
 
     def requires_case_type?
       false
+    end
+
+    def agfs?
+      true
+    end
+
+    def final?
+      false
+    end
+
+    def interim?
+      true
+    end
+
+    def eligible_advocate_categories
+      Claims::FetchEligibleAdvocateCategories.for(self)
     end
 
     private
--- a/app/presenters/claim/advocate_interim_claim_presenter.rb
+++ b/app/presenters/claim/advocate_interim_claim_presenter.rb
@@ -10,4 +10,9 @@ class Claim::AdvocateInterimClaimPresenter < Claim::BaseClaimPresenter
   def can_have_disbursements?
     false
   end
+
+  def raw_warrant_fees_total
+    claim.warrant_fee&.amount || 0
+  end
+  present_with_currency :warrant_fees_total
 end
--- a/app/validators/claim/advocate_interim_claim_validator.rb
+++ b/app/validators/claim/advocate_interim_claim_validator.rb
@@ -10,12 +10,8 @@ class Claim::AdvocateInterimClaimValidator < Claim::BaseClaimValidator
         supplier_number
       ],
       defendants: %i[earliest_representation_order],
-      offence_details: %i[offence],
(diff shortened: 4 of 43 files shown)


app/controllers/external_users/advocates/interim_claims_controller.rb
```diff
@@ -2,4 +2,9 @@ class ExternalUsers::Advocates::InterimClaimsController < ExternalUsers::ClaimsC
   skip_load_and_authorize_resource
 
   resource_klass Claim::AdvocateInterimClaim
+
+  def build_nested_resources
```
lostie, line 6: I'm not a big fan of this approach of ensuring the associated resources are initialised for the purposes of form rendering, but for now I've decided to keep it similarly to the other controllers for consistency.

app/validators/claim/advocate_interim_claim_validator.rb
```diff
@@ -10,12 +10,8 @@ def self.fields_for_steps
         supplier_number
       ],
       defendants: %i[earliest_representation_order],
-      offence_details: %i[offence],
-      fees: %i[
-        advocate_category
-        total
-        defendant_uplifts
-      ]
+      # offence_details: %i[offence],
```
lostie, line 13: This is just commented out, as the offences work is not completed and was preventing me from reaching the next step (interim fees).

app/validators/representation_order_validator.rb
```diff
@@ -40,7 +40,7 @@ def claim
   end
 
   def earliest_permitted
-    if claim.try(:interim?)
+    if claim&.lgfs? && claim&.interim?
```
lostie, line 43: There's quite a lot of changes like this one in this PR, and as stated in the description and the commit message that includes these changes, the main reason behind this is to simply preserve the existent behaviour we had so far (which only applied to **LGFS** interim claims).

app/models/claim/advocate_interim_claim.rb
```diff
@@ -65,6 +69,22 @@ def requires_case_type?
       false
     end
 
+    def agfs?
```
lostie, line 72: ⚠️  The introduction of these methods (`agfs?`, `final?` and `interim?`) caused all kind of side effects, because some checks in the code were, for example, using only `interim?` to check for a LGFS interim.

For `agfs?` ones, there's quite a lot of checks on this but I'm not quite sure if they're necessary to be changed to ensure they're `agfs? && !interim?` 😰  (I'm expecting these ones to be less problematic as the new type of claim has a brand new flow, so any unexpected behaviour will be fixed when we come to make the changes required and realise its not returning or showing what we expect it to be.

Also, initially, for consistency purposes, I've introduced the method `final?` to the existent `AdvocateClaim` but that was triggered a bunch of unexpected behaviours and failing validations on things like fees. I've decided to revert that change as I don't own enough knowledge right now to be able to make changes to the existent code and ensure the existent behaviour was preserved.

app/presenters/claim/advocate_interim_claim_presenter.rb
```diff
@@ -10,4 +10,9 @@ def type_identifier
   def can_have_disbursements?
     false
   end
+
+  def raw_warrant_fees_total
+    claim.warrant_fee&.amount || 0
```
colinbruce, line 15: This doesn't seem to have test coverage (according to code climate)

app/validators/claim/advocate_interim_claim_validator.rb
colinbruce: :(

app/validators/claim/advocate_interim_claim_validator.rb
```diff
@@ -30,4 +26,14 @@ def validate_earliest_representation_order
     return unless date.present?
     add_error(:base, 'unclaimable') unless date >= Date.parse(Settings.agfs_fee_reform_release_date.to_s)
   end
+
+  def validate_offence
+    validate_presence(:offence, 'blank')
```
colinbruce, line 31: Also not tested

colinbruce (changes requested):
A couple of missing tests, but otherwise 👍

app/validators/claim/advocate_interim_claim_validator.rb
lostie: This one will get addressed when I come to do the work on the offences themselves ;)

app/presenters/claim/advocate_interim_claim_presenter.rb
lostie: ⏩  🔬

lostie:
@colinbruce ♻️

colinbruce (approved):
LGTM 👍

PR #2283, opened 2018-03-20
Scheme 10 API

colinbruce:
#### What
Extend the API to allow querying and submission of scheme 10 offences

#### Ticket
[Changes to software vendor API ](https://www.pivotaltracker.com/story/show/141298257)

--------

#### TODO
- [x] create rake task to inject data in to newly created migration
- [x] release notes
--------

#### New tickets?
- [ ] Identify the CCR requirements for the Injection API endpoint
- [ ] Extend the Scheme 10 tests
- [ ] Determine requirements for display of new offences (e.g. description + contrary + year&chapter)


app/interfaces/api/entities/ccr/offence.rb
```diff
@@ -3,7 +3,9 @@ module Entities
     module CCR
       class Offence < API::Entities::BaseEntity
         expose :unique_code
-        expose :offence_class, using: API::Entities::CCR::OffenceClass
+        expose :offence_class,
+               if: ->(instance, _opts) { instance.fee_schemes.first.number.eql?(9) },
```
lostie: **[not a blocker/tiny nit pick]** If `9` means something here, I would consider setting it in some constant `AGFS_SCHEME9` (maybe in the `FeeScheme` class) so it's a bit more clear what it is and can be re-used across different places.

app/interfaces/api/entities/offence.rb
```diff
@@ -9,8 +9,13 @@ class Offence < BaseEntity
       expose :full_format, merge: true, unless: export_format? do
         expose :id
         expose :description
-        expose :offence_class_id
-        expose :offence_class, using: API::Entities::OffenceClass
+        expose :offence_class_id, if: ->(instance, _opts) { instance.fee_schemes.first.number.eql?(9) }
+        expose :offence_class,
+               if: ->(instance, _opts) { instance.fee_schemes.first.number.eql?(9) },
```
lostie: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2283/files#r175754917)

app/models/fee_scheme.rb
```diff
@@ -1,4 +1,18 @@
-class FeeScheme
+class FeeScheme < ActiveRecord::Base
+  validates :start_date, :number, :name, presence: true
+
+  has_many :offence_fee_schemes
+  has_many :offences, through: :offence_fee_schemes
+
+  scope :agfs, -> { where(name: 'AGFS') }
+  scope :lgfs, -> { where(name: 'LGFS') }
+  scope :current, lambda {
```
lostie: **[not a blocker/nit pick]** This returns a _list_ of fee schemes that are currently active or in use, correct? Maybe it's just my own interpretation, but the first time I've looked at the method it suggested it would return **the** current **one**.

app/models/offence.rb
```diff
@@ -14,12 +14,18 @@ class Offence < ActiveRecord::Base
   auto_strip_attributes :description, squish: true, nullify: true
 
   belongs_to :offence_class
+  belongs_to :fee_band
   has_many :claims, -> { active }, class_name: Claim::BaseClaim, foreign_key: :offence_id, dependent: :nullify
+  has_many :offence_fee_schemes
+  has_many :fee_schemes, through: :offence_fee_schemes
```
lostie: ❓ Can a given offence be associated with multiple fee schemes?

app/models/offence.rb
```diff
@@ -28,4 +34,9 @@ class Offence < ActiveRecord::Base
   def offence_class_description
     offence_class.letter_and_description
   end
+
+  def offence_class_xor_fee_band
+    return if offence_class.present? ^ fee_band.present?
+    errors[:base] << 'Specify an Offence or a FeeBand, not both'
```
lostie: **[not a blocker]** 💄   Consider moving the error message to the locales.

db/offence_code_seeder_scheme_ten.rb
```diff
@@ -0,0 +1,29 @@
...
+    def initialize(description, class_letter, contrary_to)
+      @description = description
+      @class_letter = class_letter
+      @contrary_to = contrary_to
+    end
+
+    def unique_code
+      modifier = 0
+      unique_code = code
+      unique_code = code(modifier += 1) while exists?(unique_code)
```
lostie, line 13: 💭  I lack context here (so bare with me on it 😅 ) but would it be better for the modifier to be a `SecureRandom` of some sort (most likely to be unique, could even be time based if required)?

db/seeds/scheme_ten.rb
```diff
@@ -0,0 +1,67 @@
+require 'csv'
+require_relative '../offence_code_seeder_scheme_ten.rb'
+
+FeeScheme.create(number: 9, name: 'LGFS', start_date: '2014-03-20 00:00:00')
```
lostie: 💭  Consider using `find_or_create_by` instead so that in theory this seed can be executed multiple times with no risk of creating duplicates. Also, by doing that, we can instantly assign it to a variable and avoid lines 8 and 9.

db/seeds/scheme_ten.rb
```diff
@@ -0,0 +1,67 @@
...
+FeeScheme.create(number: 9, name: 'AGFS', start_date: '2012-04-01 00:00:00', end_date: '2018-03-31 23:59:59')
+FeeScheme.create(number: 10, name: 'AGFS', start_date: '2018-04-01 00:00:00')
+
+agfs_scheme_nine = FeeScheme.find_by(name: 'AGFS', number: '9')
+lgfs_scheme_nine = FeeScheme.find_by(name: 'LGFS', number: '9')
+
+Offence.find_each do |offence|
```
lostie: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2283/files#r175762038)

db/seeds/scheme_ten.rb
```diff
@@ -0,0 +1,67 @@
...
+file_path = Rails.root.join('lib', 'assets', 'data', 'scheme_10_offences.csv')
+csv_file = File.open(file_path, 'r:ISO-8859-1')
+csv = CSV.parse(csv_file, headers: true)
+
+csv.each do |row|
+  # row desc : Category,Band,Offence,Contrary to,Year and Chapter
+  description = row[2].strip
+  fee_category = FeeCategory.find_by(number: row[0])
+  fee_band = FeeBand.find_by(fee_category: fee_category, description: row[1])
+  unique_code = OffenceCodeSeederSchemeTen.new(description, fee_band.number.to_s.each_byte.to_a.join(''), row[3]).unique_code
```
lostie: **[not a blocker/nit pick]** (Being really picky here 😅 ) If the intent of this class is just to generate a unique code for the new offence, consider renaming the class to something like `GenerateOffenceUniqueCode` and calling returns the unique code itself.

Another option would be to have the whole row passed to the class `OffenceMapper` which could have a `to_h` method in it that returns all the mapped attributes required to find/create the offence.

As a side note question, (independently from the code implementation) what data makes a given offence unique?

app/interfaces/api/entities/offence.rb
lostie: **[not a blocker]** Also, 💭  consider moving the responsibility of checking the fee scheme to the offence itself:

```ruby
# made up name, feel free to come up with a better one :P
if: ->(instance, _opts) { instance.on_scheme?(FeeScheme::AGFS_SCHEME9) },
```

lostie (changes requested):
No real blockers really, although it seems that _scheme 9/10_ is used across the set of changes without any scope, but aren't these changes specific to AGFS only? Or do they apply to LGFS as well? If they don't I would consider to re-enforce that scope.

app/models/offence.rb
jsugarman: i agree

db/seeds/scheme_ten.rb
jsugarman: yeah, need to make it idempotent, so `find_or_create_by sounds` good

app/interfaces/api/entities/ccr/offence.rb
jsugarman: in terms of helpers and clarity it would be nice to have some delegation allowing `instance.for_agfs_scheme?(9)` or `for_scheme?(AGFS_SCHEME9)` - but this can be iterated on.

app/interfaces/api/entities/offence.rb
jsugarman: agree 👍

app/interfaces/api/v1/dropdown_data.rb
```diff
@@ -50,13 +50,16 @@ def role
...
+                     desc: 'OPTIONAL: Date of representation order in YYYY-MM-DD',
+                     standard_json_format: true
           end
           get do
-            offences = if params[:offence_description].present?
-                         Offence.where(description: params[:offence_description])
-                       else
-                         Offence.all
-                       end
+            scheme_date = params[:rep_order_date] || '2018-01-01'
```
jsugarman, line 59: I'd be tempted to put a scheme selector option on the grape documentation here, either in addition or instead of the date sensitive option above

e.g. 
```
params :offence_filter do
  optional :offence_scheme, type: String, desc: I18n.t('api.v1.dropdown_data.params.scheme_filter'), values: %w[9 10]
end
```

and use


```
# ineffecient
offences.select { |o| o.fee_schemes.first.number.eql?(offence_scheme) }
```

app/models/fee_scheme.rb
jsugarman: perhaps `:active` instead but either way

app/models/offence.rb
jsugarman: @lostie should discuss as i had a similar question but looks like a given offence can be applicable to a scheme and version and at different `active` dates I.e. AGFS 9 and LGFS 9. and I am guess we may eventually have LGFS scheme 10

db/seeds/scheme_ten.rb
jsugarman: NOTE: This generator has a pl/sql equivalent in CCR DB that was used to generate the same keys and provide the common language used to inject offences in both CCR and CCLF

spec/factories/claim/advocate_claims.rb
```diff
@@ -83,6 +83,7 @@
     end
 
     factory :authorised_claim do
+      offence { FactoryBot.create :offence, :with_fee_scheme, offence_class: FactoryBot.create(:offence_class) }
```
jsugarman: does this require the `FactoryBot` prefix? have been trying to remove these unless namespacing is causing an issue

lostie: **[not a blocker]** As discussed offline, consider using the following instead (not actually tested 😅 ):

```ruby
factory :authorised_claim do
  association :offence, factory: :offence, :with_fee_scheme, :with_offence_class
  ...
end
```

where `with_offence_class` would be a _trait_ on the **offence** factory itself.

colinbruce: I've removed the FactoryBot calls.  @lostie Any objection to me creating an Issue for the `association` changes? I don't want to delay this further! :)

db/migrate/20180301113027_create_fee_schemes.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFeeSchemes < ActiveRecord::Migration
+  def change
+    create_table :fee_schemes do |t|
+      t.string :name
+      t.integer :version
```
jsugarman, line 5: 👍

db/offence_code_seeder_scheme_ten.rb
```diff
@@ -0,0 +1,29 @@
...
+      unique_code
+    end
+
+    private
+
+    def code(modifier = nil)
+      description.abbreviate +
+        modifier.to_s +
+        '_' +
+        class_letter
```
jsugarman, line 23: not a blocker as can be done later but will need to be `band.to_s --> 1_1` or somesuch

jsugarman: ...as discussed

app/interfaces/api/entities/ccr/offence.rb
```diff
@@ -3,7 +3,9 @@ module Entities
     module CCR
       class Offence < API::Entities::BaseEntity
         expose :unique_code
-        expose :offence_class, using: API::Entities::CCR::OffenceClass
+        expose :offence_class,
+               if: ->(instance, _opts) { instance.scheme_nine? },
```
jsugarman, line 7: 👍 much clearer

jsugarman (approved):
👍 A good start to the fee scheme changes.

spec/factories/offence_classes.rb
```diff
@@ -21,10 +21,10 @@
 
   trait :with_lgfs_offence do
     after(:create) do |record|
-      create :offence, :miscellaneous, offence_class: record
+      offence = create :offence, :miscellaneous, offence_class: record
+      offence.fee_schemes << (FeeScheme.first || create(:fee_scheme, :agfs_nine))
```
lostie: Shouldn't the fee scheme here (`create(:fee_scheme, :agfs_nine))`) be for `LGFS` ? Also, `FeeScheme.first` should probably be `FeeScheme.lgfs.first` to return the first LGFS rather than the first one it finds (that could be LGFS).

lostie (approved):

db/offence_code_seeder_scheme_ten.rb
colinbruce: It is actually passed the band
```ruby
unique_code = OffenceCodeSeederSchemeTen.new(description, offence_band.number.to_s.each_byte.to_a.join(''), row[3]).unique_code
```
but I left the signature of the method the same. Lazy copypasta :(

Renaming the parameter to `suffix` in both cases would probably be clearer

jsugarman: oh right, yes, sounds good

lib/tasks/scheme_ten.rake
```diff
@@ -0,0 +1,19 @@
+namespace :scheme_ten do
```
jsugarman: not a blocker. I'd namespace this to db with task name, or sub-namespace, of :seed
```
namespace :db
  namespace :scheme_ten
    namespace :seed
```

and use

```
rake db:scheme_ten:seed
```
or alternative
```
rake db:seed:scheme_ten
```
others could be namepaced therein in future

jsugarman: rails adds a `db:seed` task so the alternative above extends that notion

lostie (approved):
LGTM 👍  => :shipit:  => 🎉  🎈

PR #2284, opened 2018-03-20
Sidebar: Show / Hide on pages as required

cleargif:
#### What
- Introducing a new partial to cut down on duplication
- Apply logic to show / hide the sidebar as required

#### Ticket
[Sidebar: Show / Hide on pages as required](https://www.pivotaltracker.com/story/show/156087810)

#### Why
The side bar should only be shown on some pages.

### DIFF
--- a/app/views/external_users/claims/interim_fee/_fields.html.haml
+++ b/app/views/external_users/claims/interim_fee/_fields.html.haml
@@ -3,6 +3,7 @@
   %h2.bold-medium
     = t('.interim_fees')
   %p.form-hint.xsmall
+    = t('.interim_fees_prompt_text_vat')
     = t('.interim_fees_prompt_text_html')
 
   = f.fields_for :interim_fee do |inf|
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -479,7 +479,7 @@ en:
         lgfs_transfer_rb: Litigator transfer fee
     claims:
       hints:
-        fees_header_prompt_text_html: All fees should be entered exclusive of VAT. The system will add VAT to fees for VAT registered users.
+        fees_header_prompt_text_html: &fees_vat_prompt All fees should be entered exclusive of VAT. The system will add VAT to fees for VAT registered users.
       redraft:
         error_html: 'An error is preventing this claim from being redrafted.  The problem has been logged and is being investigated. To continue please start a new claim.'
       additional_information:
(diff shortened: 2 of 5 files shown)


app/views/external_users/claims/interim_fee/_fields.html.haml
```diff
@@ -3,6 +3,7 @@
   %h2.bold-medium
     = t('.interim_fees')
   %p.form-hint.xsmall
+    = "All fees should be entered exclusive of VAT. The system will add VAT to fees for VAT registered users."
```
lostie: ❓ isn't this the same as the line below (that is using the locales)? 😕  (💭  If it's not, I would still consider defining a locale for this, avoiding static wording in the views).

config/locales/en.yml
```diff
@@ -751,7 +751,7 @@ en:
             fixed_fees: "Fixed fees"
             fee_type: "Fee type"
             total: "Total"
-            prompt_text_html: Check the <a href="https://www.gov.uk/government/publications/graduated-fee-calculators" target="_blank" rel="external" title="Graduated Fee Calculator">graduated fee calculators</a> for help
+            prompt_text_html: Check the <a href="https://www.gov.uk/government/publications/graduated-fee-calculators" target="_blank" rel="external" title="Graduated Fee Calculator">graduated fee calculators</a> for help.
```
lostie: **[not a blocker]** Seems like this wording is the same across all different types of claim (we're repeating the same wording for each claim type that requires it).

💭  Consider this instead:

```ruby
prompt_text_html: &graduated_fee_prompt_content Check the <a
href="https://www.gov.uk/government/publications/graduated-fee-calculators" target="_blank" rel="external" title="Graduated Fee Calculator">graduated fee calculators</a> for help.
...
graduated_prompt_text: *graduated_fee_prompt_content
...
interim_fees_prompt_text_html: *graduated_fee_prompt_content
...
transfer_fees_prompt_text_html: *graduated_fee_prompt_content
```

This should work, and avoid repeating the same content across multiple prompts.

lostie (changes requested):
If we could tied up those small nits, would be nice to make things easier to change in the future.

Other than that, LGTM 👍

app/views/external_users/claims/interim_fee/_fields.html.haml
cleargif: ooops.. forgot to wire it up again!

config/locales/en.yml
cleargif: cool .. will do like above.

lostie (approved):
LGTM 👍

PR #2285, opened 2018-03-20
Add brakeman

jsugarman:
#### What
run brakeman as part of suite

#### Why
Gemnasium will go out of business in May 2018
and cease to function. This currently runs brakeman
so to replace this functionality it has
been added into the test suite run.


colinbruce (approved):
👍 LGTM

PR #2286, opened 2018-03-20
Remove rack-protection gem

colinbruce:
#### What
Remove the rack protection gem
#### Why
This was added, and locked at 1.5.4, due to a previous security vulnerability in the locked, outdated, sinatra.  As sinatra has now been removed, and this was a dependency of it, this should be superfluous.


jsugarman (approved):
👍

PR #2290, opened 2018-03-21
[AGFS fee reform] Supporting evidence & Additional information page changes

lostie:
#### What

Introduces the _supporting evidence_ and _additional information_ pages to the **advocate interim claim** submission through the web interface.

Also:

- Presents a different set of document types the user is able to _check_ if the type of claim is **advocate interim**, preserves the existent list for all other types.

#### Ticket

- 💎  [AGFS final and interim fees: supporting evidence page](https://www.pivotaltracker.com/story/show/155396942)
- 💎  [AGFS final and interim fees: additional information page](https://www.pivotaltracker.com/story/show/155396962)

### DIFF
--- a/app/controllers/case_workers/claims_controller.rb
+++ b/app/controllers/case_workers/claims_controller.rb
@@ -1,6 +1,5 @@
 class CaseWorkers::ClaimsController < CaseWorkers::ApplicationController
   include PaginationHelpers
-  include DocTypes
 
   skip_load_and_authorize_resource
   authorize_resource class: Claim::BaseClaim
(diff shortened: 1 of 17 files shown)


app/controllers/case_workers/claims_controller.rb
```diff
@@ -60,7 +56,6 @@ def update
 
   def prepare_show_action
     @claim.assessment = Assessment.new if @claim.assessment.nil?
-    @doc_types = DocType.all
```
lostie, line 63: 🗒  There were cases on this controller where we were setting this _unused_ variable 3 times in the same action! 😰

app/views/external_users/claims/supporting_evidence_checklist/_fields.html.haml
```diff
@@ -3,4 +3,4 @@
     %h3.heading-medium
       = t('.supporting_evidence_checklist')
 
-  = render partial: 'external_users/claims/supporting_evidence_checklist/evidence_checklist', locals: { f: f, collection: DocType.all }
\ No newline at end of file
+  = render partial: 'external_users/claims/supporting_evidence_checklist/evidence_checklist', locals: { f: f, collection: f.object.eligible_document_types }
```
lostie, line 6: This is the point where the `@doc_types` defined in the controller should be used. As you can see by the existent code, it was using `DocType.all` directly in the view, so that variable was not being used at all.

jsugarman (approved):
LGTM 👍

PR #2292, opened 2018-03-22
[AGFS fee reform] Graduated & Miscellaneous fees changes

lostie:
#### What

**Graduated fees changes**

- Display only relevant **advocate categories** for new fee scheme **[already in place from previous iterations]**
- Display a custom set of graduate fees for new fee scheme.
  - The approach was to update the seeds for fee types and leverage the existent of `roles` for a fee type, adding extra roles for both current and new fee schemes (ideally, I would say that would be done using some sort of relation with the fee scheme directly, without that in place, I've taken the decision to go with this instead. 
  - Also as part of this I've done some refactoring to the fee type seeds so it can be executed multiple times without incurring into creating duplicate methods (this is on the assumption that the initial seed was made using a specific `id` provided by the CSV file to create the record in the database using that as the primary key.
- Display only the quantity for `PPE` in the graduates page for advocate claims.

**Miscellaneous fees changes**

- Add new miscellaneous fee types for scheme 10 to the seed file and set them under the appropriate roles.
- Display a different set of miscellaneous fees for AGFS scheme 9 and AGFS scheme 10.

#### Ticket

- 💎  [AGFS final fee: Graduated fees](https://www.pivotaltracker.com/story/show/155569366)
- 💎  [AGFS final fee: miscellaneous fees](https://www.pivotaltracker.com/story/show/155569882)

### DIFF
--- a/app/models/claim/advocate_claim.rb
+++ b/app/models/claim/advocate_claim.rb
@@ -176,8 +176,11 @@ module Claim
     end
 
     def eligible_basic_fee_types
+      # TODO: this should return a list based on the current given fee scheme
+      # rather than conditionally return scheme 10 specifically
+      # TBD once all the fee scheme work is integrated
       return Fee::BasicFeeType.agfs unless fee_scheme == 'fee_reform'
-      Fee::BasicFeeType.agfs.fee_reform
+      Fee::BasicFeeType.agfs_scheme_10s
     end
 
     def eligible_misc_fee_types
@@ -234,9 +237,14 @@ module Claim
     # create a blank fee for every basic fee type not passed to Claim::AdvocateClaim.new
     def instantiate_basic_fees
       return unless case_type.present? && !case_type.is_fixed_fee?
-      # return unless new_record?
+      return unless editable?
 
-      (eligible_basic_fee_types - basic_fees.map(&:fee_type)).each do |basic_fee_type|
+      fee_type_ids = basic_fees.map(&:fee_type_id)
+      eligible_basic_fee_type_ids = eligible_basic_fee_types.map(&:id)
+      not_eligible_ids = fee_type_ids - eligible_basic_fee_type_ids
+      basic_fees.where(fee_type_id: not_eligible_ids).delete_all
+      eligible_basic_fee_types.each do |basic_fee_type|
+        next if fee_type_ids.include?(basic_fee_type.id)
         basic_fees << Fee::BasicFee.new_blank(self, basic_fee_type)
       end
     end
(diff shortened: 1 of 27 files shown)


app/models/claim/advocate_claim.rb
```diff
@@ -232,12 +236,15 @@ def set_supplier_number
 
     # create a blank fee for every basic fee type not passed to Claim::AdvocateClaim.new
     def instantiate_basic_fees
-      return unless new_record?
+      return unless case_type.present? && !case_type.is_fixed_fee?
```
lostie, line 243: This method has been creating all sorts of issues/complications throughout the development process. I've decided to keep it for now as I still lack full understanding of the impact of removing it and managing what is trying to achieve elsewhere.

Still, because of that, to ensure the management of the **basic fees** vs **fixed fees** and **basic fees for scheme 9** vs **basic fees for scheme 10**, I had to make some changes to it to ensure that behaviour was still working.

app/models/fee/base_fee_type.rb
```diff
@@ -25,7 +25,7 @@ def initialize(message = DEFAULT_MSG)
   end
 
   class BaseFeeType < ActiveRecord::Base
-    ROLES = %w[lgfs agfs].freeze
+    ROLES = %w[lgfs agfs agfs_scheme_9 agfs_scheme_10].freeze
```
lostie, line 28: 💭  Potentially remove this in the future ⏩  in favour of association with a given fee scheme.

app/models/claim/advocate_claim.rb
```diff
@@ -232,13 +236,16 @@ def set_supplier_number
...
-      basic_fee_types.each do |basic_fee_type|
-        next if basic_fee_type.id.in?(existing_basic_fee_type_ids)
-        basic_fees << Fee::BasicFee.new_blank(self, basic_fee_type)
+      fee_type_ids = basic_fees.map(&:fee_type_id)
+      eligible_basic_fee_type_ids = eligible_basic_fee_types.map(&:id)
+      not_eligible_ids = fee_type_ids - eligible_basic_fee_type_ids
+      self.basic_fees = basic_fees.reject { |fee| not_eligible_ids.include?(fee.fee_type_id) }
+      eligible_basic_fee_types.each do |basic_fee_type|
+        next if fee_type_ids.include?(basic_fee_type.id)
+        basic_fees.build(fee_type: basic_fee_type, quantity: 0, amount: 0)
```
lostie, line 252: The use of the Rails **conventions** to build a new basic fee for the claim instead of the existent `<< Fee::BasicFee.new_blank` makes **ALL** the different. I've spent hours debugging why I was getting duplicates with this! That's why! 😫

colinbruce (approved):
👍

PR #2294, opened 2018-03-23
Address CVEs

colinbruce:
What
Bump Loofah and rails-html-sanitizer gem version

Why
address CVE-2018-8048


jsugarman (approved):
lgtm 👍 assuming tests pass

PR #2295, opened 2018-03-23
Amend dump scripts to enable dump to new disaster stack

jsugarman:
#### What
Allow dumping of anonymised data to the disaster stack

#### Why
To have a Disaster recovery testing environment 
that matches prod as closely as possible.


colinbruce (approved):
👍 Hopefully we'll never need to dump *from* disaster, but belt and braces!

PR #2297, opened 2018-03-26
Add logging for vat_registered sentry errors

colinbruce:
#### What
Add logging for nil provider errors
#### Why
In sentry we are seeing occasional errors for `undefined method 'vat_registered?' for nil:NilClass`
#### How
This is designed to log when the situation recurs to enable further remediation steps


lostie (approved):
LGTM 👍

PR #2300, opened 2018-03-29
API, AGFS scheme ten rep order validation

colinbruce:
#### What
Add validation of rep orders after scheme ten implementation
#### Ticket
[Changes to software vendor API ](https://www.pivotaltracker.com/story/show/141298257)
#### Why
In the web site, users are only shown offences after entering a defendants representation order so this validation will not fire as the offence list will be shown after the user has entered a date.

In the API, users create a claim with an offence, then a defendant, then a rep order.  At that point an API user can submit a rep order that is invalid for the previously supplied offence_id.
#### How
Add a validation that only fires if a user attempts to create a rep order on a claim where a Scheme Ten offence has already been set.

### DIFF
--- a/app/validators/representation_order_validator.rb
+++ b/app/validators/representation_order_validator.rb
@@ -17,6 +17,9 @@ class RepresentationOrderValidator < BaseValidator
     validate_presence(:representation_order_date, 'blank')
     validate_on_or_before(Date.today, :representation_order_date, 'in_future')
     validate_on_or_after(earliest_permitted[:date], :representation_order_date, earliest_permitted[:error])
+    if scheme_ten?
+      validate_on_or_after(Settings.agfs_fee_reform_release_date, :representation_order_date, 'scheme_ten_offence')
+    end
 
     return if @record.is_first_reporder_for_same_defendant?
     first_reporder_date = @record.first_reporder_for_same_defendant.try(:representation_order_date)
@@ -37,6 +40,10 @@ class RepresentationOrderValidator < BaseValidator
   #
   def claim
     @record.defendant.claim
+  end
+
+  def scheme_ten?
+    claim&.offence&.scheme_ten?
   end
 
   def earliest_permitted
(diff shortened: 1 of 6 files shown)


spec/factories/claim/advocate_claims.rb
```diff
@@ -31,7 +34,7 @@
     sequence(:cms_number) { |n| "CMS-#{Time.now.year}-#{rand(100..199)}-#{n}" }
 
     after(:build) { |claim| post_build_actions_for_draft_claim(claim) }
-    after(:create) { |claim| add_defendant_and_reporder(claim) }
+    after(:create) { |claim, evaluator| add_defendant_and_reporder(claim) if evaluator.create_defendant_and_rep_order }
```
lostie, line 37: 👌

app/validators/representation_order_validator.rb
```diff
@@ -17,6 +17,9 @@ def validate_representation_order_date
     validate_presence(:representation_order_date, 'blank')
     validate_on_or_before(Date.today, :representation_order_date, 'in_future')
     validate_on_or_after(earliest_permitted[:date], :representation_order_date, earliest_permitted[:error])
+    if scheme_ten?
```
lostie, line 20: Will this validation create any duplicate error messages for advocate interim claims (at least through the web flow) since [this](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/app/validators/claim/advocate_interim_claim_validator.rb#L24) is already in place?

colinbruce: Hmm... it hasn't on the tests or in my manual run through.  I'll double check and add extra tests! Thanks for the spot

colinbruce: Okay @lostie, I've run through the scenarios (in my head, no cukes needed!) and this won'r be an issue, either on the web or API.  On the web, the offence is added after this page and we only show the appropriate offences for the (already validated by this) rep order.  In the API, the claim includes the offence, so then we validate the rep order date against that.  

My next job is hooking up the Advocate Interim end point on the API.  I may need to tweak these two rules then...

lostie (approved):
LGTM 👍

PR #2302, opened 2018-04-03
Clarify API dropdown offence tests

colinbruce:
#### What
After the month switch to April, the seeded data for the dropdown tests broke.
#### Why
The seeds meant that `FeeScheme.current_agfs` no longer matched scheme nine 
#### How
Clarified the setup to use `FeeScheme.nine.agfs.first`


jsugarman (approved):
👍  - might be worth having a separate PR to add a  scheme 10 equivalent offence endpoint test

PR #2303, opened 2018-04-03
Formally remove Postgresql adminpack

jsugarman:
Not required and running db:migrate is removing it in any event.


colinbruce (approved):
👍 for removing an oddity...

PR #2304, opened 2018-04-03
Add API advocate interim fee

colinbruce:
#### What
Add API endpoint to allow providers (via software vendors) to submit the new AGFS Interim(warrant) claims
#### Ticket
[Changes to software vendor API - task 3](https://www.pivotaltracker.com/story/show/141298257)
#### ToDo

- [x] Update the new tests to do something valuable
- [ ] Write release notes


app/interfaces/api/v1/claim_params_helper.rb
```diff
@@ -2,6 +2,7 @@ module API
   module V1
     module ClaimParamsHelper
       extend Grape::API::Helpers
+      AGFS_SCOPE = %i[api v1 external_users claims advocate_claim params].freeze
```
lostie, line 5: 👍 

**[not a blocker]** 💭 Consider changing https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/app/interfaces/api/v1/external_users/claims/advocate_claim.rb#L7-L11 to make use of this new constant instead for DRYness.

app/interfaces/api/v1/external_users/claims/advocate/interim_claim.rb
```diff
@@ -0,0 +1,50 @@
+module API::V1::ExternalUsers
+  module Claims
+    module Advocate
```
lostie: 💭 Favour naming resource modules (associated directory as well) with pluralized names, e.g. `Advocates`.

lostie (approved):
LGTM 👍 

One minor thing, the API specs seem to be only asserting **change** when a new claim is created successfully. We should probably assert **not_to change** as well. Asserting that the API returns a non-201 code does not necessarily ensure a record was not created.

lostie (approved):
👍 👍

lostie (approved):
LGTM 👍

PR #2306, opened 2018-04-03
Fix scheme ten seeds

colinbruce:
#### What
After software vendor input, some mis-banded and missing offences have been identified
#### Why
It would have prevented users from submitting claims for the affected expenses


lostie (approved):
LGTM 👍

PR #2307, opened 2018-04-04
Seeding fixes

lostie:
#### What

- Reload column data information in seed services to ensure they're using the most update ones.
- Seed demo data with scheme 9 basic fees only (preserving the existent behaviour)

#### Why

Migrations/seeds load the model multiple times at different stages and they cache the column information, causing errors like these:

```ruby
Caused by:
NoMethodError: undefined method `unique_code=' for #<ExpenseType:0x00007f8604cad3e8>
db/seed_helper.rb:58:in `find_or_create_expense_type!'
db/seeds/expense_types.rb:19:in `block in <top (required)>'
```

### DIFF
--- a/db/seeds/expense_types.rb
+++ b/db/seeds/expense_types.rb
@@ -13,6 +13,7 @@ expense_types = [
 ]
 
 max_id = 0
+ExpenseType.reset_column_information
 expense_types.each do |fields|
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
👍 LGTM

PR #2309, opened 2018-04-04
Order offences by offence category number and band number

lostie:
#### What

Change order of offence search results

#### Why

Business change requirement.

### DIFF
--- a/app/services/fee_reform/search_offences.rb
+++ b/app/services/fee_reform/search_offences.rb
@@ -16,7 +16,7 @@ module FeeReform
                         .joins(offence_band: :offence_category)
                         .includes(offence_band: :offence_category)
                         .group('offence_categories.id, offence_bands.id, offences.id')
-                        .order('offence_categories.description, offence_bands.description, offences.description')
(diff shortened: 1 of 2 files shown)


app/services/fee_reform/search_offences.rb
```diff
@@ -15,8 +15,8 @@ def call
       offences = Offence.unscoped.in_scheme_ten
                         .joins(offence_band: :offence_category)
                         .includes(offence_band: :offence_category)
-                        .group('offence_categories.id, offence_bands.id, offences.id')
-                        .order('offence_categories.description, offence_bands.description, offences.description')
+                        .group('offence_categories.id', 'offence_bands.id, offences.id')
```
colinbruce: [Not a blocker] Does this need to be split?

colinbruce (approved):
👍

app/services/fee_reform/search_offences.rb
lostie: Good catch! No, I probably didn't without even realising... fixing it.

PR #2310, opened 2018-04-04
Display fees in the order of their associated fee types

lostie:
#### What

Fee types have a position column to determine how they should be displayed.

Summary pages should use that order given there's new fee types that no longer match the expected default order (by fee id).

**NOTE:** Seems that there's quite a lot of repetition to display similar content across multiple templates. This PR does **NOT** address that, even though it should probably be re-visited.


colinbruce (approved):
:+1: Last change before we lock the build down for tonight...

PR #2312, opened 2018-04-05
Display PPE field only for claim with offence category 6 or 9

lostie:
#### What

Self-descriptive.

NOTE: This only applies to claims on the new fee reform scheme.

#### Why

Outstanding change that was deliberately left behind on [ticket](https://www.pivotaltracker.com/story/show/155569366) as it was not a real blocker.

### DIFF
--- a/app/presenters/fee/basic_fee_presenter.rb
+++ b/app/presenters/fee/basic_fee_presenter.rb
@@ -10,6 +10,12 @@ class Fee::BasicFeePresenter < Fee::BaseFeePresenter
     I18n.t(key, scope: t_scope)
   end
 
+  def should_be_displayed?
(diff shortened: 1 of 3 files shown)


app/presenters/fee/basic_fee_presenter.rb
```diff
@@ -10,6 +10,12 @@ def prompt_text
     I18n.t(key, scope: t_scope)
   end
 
+  def should_be_displayed?
+    return true unless claim.fee_scheme == 'fee_reform'
+    return true unless FEE_CODES_RESTRITED_DISPLAY.include?(code)
```
colinbruce: typo in name... FEE_CODES_RESTRI**C**TED_DISPLAY

colinbruce (approved):
:+1:

PR #2315, opened 2018-04-09
Replace LGFS case uplifts with defendant uplifts

jsugarman:
#### What
Replace the misc. fee type for LGFS case uplifts with defendant uplifts

[Ticket/bug](https://www.pivotaltracker.com/story/show/154606747)

#### Why
Feedback from BA and learning from Data Injection into CCLF show
that the Miscellaneous fee type "Case uplift" should not exist. 

Defendant uplifts can exist, but data injection handles defendant uplifts
based purely on the actual number of defendants provided in CCCD 
because it is a calculated fee.

#### How
Repurpose the Case uplift fee type (MIUPL) as a defendant uplift

### DIFF
--- a/app/validators/claim/litigator_claim_validator.rb
+++ b/app/validators/claim/litigator_claim_validator.rb
@@ -15,7 +15,26 @@ class Claim::LitigatorClaimValidator < Claim::BaseClaimValidator
       defendants: [],
       offence_details: %i[offence],
       graduated_fees: %i[actual_trial_length],
+      miscellaneous_fees: %i[defendant_uplifts],
       additional_information: %i[total]
     }
+  end
+
+  def validate_defendant_uplifts
+    return if @record.from_api?
+    no_of_defendants = @record.defendants.reject(&:marked_for_destruction?).size
+    add_error(:base, 'lgfs_defendant_uplifts_mismatch') if defendant_uplifts_greater_than?(no_of_defendants)
+  end
+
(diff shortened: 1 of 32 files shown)


app/validators/claim/litigator_claim_validator.rb
```diff
@@ -15,7 +15,35 @@ def self.fields_for_steps
...
+  # we add one because uplift quantities reflect the number of "additional" defendants
+  def defendant_uplifts_greater_than?(no_of_defendants)
+    defendant_uplifts_counts.map(&:to_i).any? { |sum| sum + 1 > no_of_defendants }
+  end
+
+  def defendant_uplifts
+    # had to remove basic_fees
+    (@record.fixed_fees + @record.misc_fees).select do |fee|
+      !fee.marked_for_destruction? &&
+        fee.fee_type &&
```
lostie: **[not a blocker/possible improvement]** 💭 I wonder if we could define a method on the _fee_ itself that answers:

```ruby
fee.fee_type && Fee::BaseFeeType.defendant_uplift_unique_codes.include?(fee.fee_type.unique_code)
```
directly so we avoid the indirection and move the responsibility to check if the fee is _uplifftable_ to the _fee_ itself.

jsugarman: 👍 this actually already existed (and i had put it in 🤦‍♂️  )

lostie (approved):
LGTM 👍

PR #2316, opened 2018-04-09
Optimize case workers archived claims search

lostie:
#### What

Optimize case workers archived claims search

Also:
- Clean up some of the factory creation bad patterns to avoid side
effects on after creation

#### Ticket

[Caseworker admin archive 500 internal server error](https://www.pivotaltracker.com/n/projects/1985371/stories/156224498)

#### Why

As described in the Pivotal Tracker ticket, this page is raising a lot of exceptions, namely `Net::ReadTimeout` mainly todo with un-optimized SQL queries that take too long to return results

#### How

Re-structuring the way the SQL is performed:
- Does not included any filtering when it's not needed (which was creating very un-optmized queries such `field ILIKE '%%'`)
- Replaces the use of `SELECT DISTINCT` in favour of `GROUP BY`, as the `SELECT DISTINCT` produced by Rails was perform `SELECT DISTINCT` on all the claim fields, turning the query plan into a very complicated one. The `GROUP BY` in this case achieves the same by group by claim.id which are already unique anyway.
- Removes duplicate filters injected by status scopes. The `search` method enforces a `states` param to be passed in, which in all the cases, created the same filter as the one defined by the status scope.
- API Eager loads most of the resources associated with the claims to avoid `N+1` queries.

### DIFF
--- a/app/interfaces/api/entities/claim.rb
+++ b/app/interfaces/api/entities/claim.rb
@@ -30,11 +30,11 @@ module API
       private
 
       def messages_count
-        object.messages.count
+        object.messages.size
       end
 
       def unread_messages_count(user)
-        object.unread_messages_for(user).count
+        object.unread_messages_for(user).size
       end
     end
   end
--- a/app/interfaces/api/v2/case_workers/claim.rb
+++ b/app/interfaces/api/v2/case_workers/claim.rb
@@ -52,19 +52,19 @@ module API
             end
 
             def current_claims
-              current_user.claims.caseworker_dashboard_under_assessment.search(
+              current_user.claims.search(
                 search_terms, Claims::StateMachine::CASEWORKER_DASHBOARD_UNDER_ASSESSMENT_STATES, *search_options
               )
             end
 
             def archived_claims
-              ::Claim::BaseClaim.active.caseworker_dashboard_archived.search(
+              ::Claim::BaseClaim.active.search(
                 search_terms, Claims::StateMachine::CASEWORKER_DASHBOARD_ARCHIVED_STATES, *search_options
               )
             end
 
             def allocated_claims
-              ::Claim::BaseClaim.active.__send__(scheme).caseworker_dashboard_under_assessment.search(
+              ::Claim::BaseClaim.active.public_send(scheme).search(
                 search_terms, Claims::StateMachine::CASEWORKER_DASHBOARD_UNDER_ASSESSMENT_STATES, *search_options
               )
             end
@@ -82,7 +82,12 @@ module API
             end
 
             def claims
-              claims_scope.sort(params.sorting, params.direction).page(params.page).per(params.limit)
+              claims_scope
+                .includes(:external_user, :case_type, :injection_attempts,
+                          :case_workers, :court, :messages,
+                          defendants: %i[representation_orders])
+                .sort(params.sorting, params.direction)
+                .page(params.page).per(params.limit)
             end
           end
 
--- a/app/models/claims/search.rb
+++ b/app/models/claims/search.rb
@@ -20,10 +20,18 @@ module Claims::Search
     }
   }.freeze
 
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/MethodLength
   def search(term, states = [], *options)
     raise 'Invalid search option' if (options - QUERY_MAPPINGS_FOR_SEARCH.keys).any?
-    sql = options.inject([]) { |a, e| a << "(#{QUERY_MAPPINGS_FOR_SEARCH[e][:query]})" }.join(' OR ')
-    relation = options.inject(all) { |a, e| a.joins(QUERY_MAPPINGS_FOR_SEARCH[e][:joins]) }
+    term = term.to_s.strip.downcase
+
+    if term.present?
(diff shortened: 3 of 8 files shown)


app/interfaces/api/entities/claim.rb
```diff
@@ -30,11 +30,11 @@ class Claim < BaseEntity
       private
 
       def messages_count
-        object.messages.count
+        object.messages.size
```
lostie, line 33: The use of `size` instead of count is with the intent of avoiding `N+1` queries, since if the messages have been preload for the claims, `size` returns the total messages in memory, whereas `count` performs an actual `count` query on the database for each claim being rendered.

app/interfaces/api/entities/claim.rb
```diff
@@ -30,11 +30,11 @@ class Claim < BaseEntity
...
       end
 
       def unread_messages_count(user)
-        object.unread_messages_for(user).count
+        object.unread_messages_for(user).size
```
lostie, line 37: This was supposed to behave the same way as [messages_count](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2316/files#r180154528), but unfortunately due the having a specific user scope, I was unable to preload this :S.

app/interfaces/api/v2/case_workers/claim.rb
```diff
@@ -52,19 +52,19 @@ def value_band_id
             end
 
             def current_claims
-              current_user.claims.caseworker_dashboard_under_assessment.search(
```
lostie, line 55: `caseworker_dashboard_under_assessment` scope is exactly the same same as the one used by  the search method (state in `Claims::StateMachine::CASEWORKER_DASHBOARD_UNDER_ASSESSMENT_STATES`)

app/interfaces/api/v2/case_workers/claim.rb
```diff
@@ -52,19 +52,19 @@ def value_band_id
...
+              current_user.claims.search(
                 search_terms, Claims::StateMachine::CASEWORKER_DASHBOARD_UNDER_ASSESSMENT_STATES, *search_options
               )
             end
 
             def archived_claims
-              ::Claim::BaseClaim.active.caseworker_dashboard_archived.search(
```
lostie, line 61: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2316/files#r180155331)

app/interfaces/api/v2/case_workers/claim.rb
```diff
@@ -52,19 +52,19 @@ def value_band_id
...
+              ::Claim::BaseClaim.active.search(
                 search_terms, Claims::StateMachine::CASEWORKER_DASHBOARD_ARCHIVED_STATES, *search_options
               )
             end
 
             def allocated_claims
-              ::Claim::BaseClaim.active.__send__(scheme).caseworker_dashboard_under_assessment.search(
```
lostie, line 67: [Ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2316/files#r180155331)

app/interfaces/api/v2/case_workers/claim.rb
```diff
@@ -82,7 +82,12 @@ def claims_scope
             end
 
             def claims
-              claims_scope.sort(params.sorting, params.direction).page(params.page).per(params.limit)
+              claims_scope
+                .includes(:external_user, :case_type, :injection_attempts,
```
lostie, line 86: includes most of the associations that are used by the API claim presenter to avoid `N+1` queries.

app/models/claims/search.rb
```diff
@@ -20,10 +20,18 @@ module Claims::Search
...
 
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/MethodLength
   def search(term, states = [], *options)
     raise 'Invalid search option' if (options - QUERY_MAPPINGS_FOR_SEARCH.keys).any?
-    sql = options.inject([]) { |a, e| a << "(#{QUERY_MAPPINGS_FOR_SEARCH[e][:query]})" }.join(' OR ')
-    relation = options.inject(all) { |a, e| a.joins(QUERY_MAPPINGS_FOR_SEARCH[e][:joins]) }
+    term = term.to_s.strip.downcase
+
+    if term.present?
```
lostie, line 29: Does not include any of the filtering options if no search `term` was provided.

spec/factories/claim/base_claims.rb
```diff
@@ -14,9 +14,11 @@
     providers_ref       { random_providers_ref }
     disable_for_state_transition nil
     after(:create) do |claim|
-      defendant = create(:defendant, claim: claim)
-      create(:representation_order, defendant: defendant, representation_order_date: 380.days.ago)
-      claim.reload
+      unless claim.defendants.present?
```
lostie, line 17: This is a big 🅱️ 👣  for me! Allows us to define a *claim* factory providing the defendants directly, rather than taking the existent convoluted approach of `after_create` having to clear any existent defendants autogenerated by the factory and then assigning the actual ones we want to have.

jsugarman (approved):
👍 looks good. good to go if you are happy existing search functionality across the pages using this will not be changed

just one small suggestion as discussed - possible inject --> each_with_object in the search method itself.

PR #2317, opened 2018-04-10
Change copy for trial dates/ rep order dates error checks

lostie:
#### What

- It only displays a single error instead of the current two (for each of
the trial dates)
- Changed the wording for the error message displayed in both cases

#### Ticket

[Validation error messaging - defendant details page](https://www.pivotaltracker.com/n/projects/2145949/stories/155852933)


jsugarman (approved):
👍

PR #2319, opened 2018-04-11
Don't display offence details for fixed fee case types

lostie:
#### What

Don't display offence details for fixed fee case types in both _summary pages_.

#### Ticket

[Remove offence details from summary when fixed fee](https://www.pivotaltracker.com/story/show/156576898)


jsugarman (approved):
👍

PR #2320, opened 2018-04-12
Fix supplier number generation

lostie:
#### What

The supplier number generation was being done by using the latest value of the primary key for the supplier numbers column.

Due to the formatting of the supplier number in use:

`9X + 3 digits + X`

in test mode there were some occurences where the primary key wasgreater than 3 digits which made the newly generated supplier number fail (as it justs adds +1 to it).

This solution instead, searches for the latest supplier number record in the database with the specified format, retrieves its supplier number (not the primary key) and adds +1 to the 3 digits sequence in it.

In real life there's still a possibility that the +1 will exceed the 3 digit number, but atm, i lack understand as of the reason for the restrictive formatting, so this keeps the same generation approach but makes it more reliable.

### DIFF
--- a/app/controllers/external_users/registrations_controller.rb
+++ b/app/controllers/external_users/registrations_controller.rb
@@ -13,7 +13,7 @@ class ExternalUsers::RegistrationsController < Devise::RegistrationsController
     resource.save
     yield resource if block_given?
     if resource.persisted?
-      create_external_user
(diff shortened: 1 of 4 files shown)


colinbruce (approved):
👍 
If this stops the flicker we owe you 🍺

jsugarman (approved):
👍 you still wanting to get this in?

PR #2324, opened 2018-04-12
Do NOT validate trial dates against rep order for non trial cases

lostie:
#### What

Bug introduced in #2317 

Oversight on my part. The trial dates do not need to checked against the earliest representation order if the case type is not a **Trial** 😰


jsugarman (approved):
LGTM 👍

PR #2325, opened 2018-04-13
Handle deprecation warning on old migrations adding timestamps

jsugarman:
#### What
Removes a deprecation warning relating to Rails 5 upgrade.

#### Why
Preparing for rails 5 upgrade

```
DEPRECATION WARNING: `#timestamps` was called without specifying an option for `null`.
In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to
prevent the behavior of your existing migrations from changing.
```

#### How
explicitly state `null: true` to retain existing functionality.


lostie (approved):
LGTM 👍

PR #2327, opened 2018-04-13
Add message to claim messages on reject/refuse

jsugarman:
#### What
Add a message to the claim with specific content for a given reject/refuse reason

#### Ticket

[Reject/Refused actions](https://www.pivotaltracker.com/story/show/155235798)

#### Why
To clarify reason for rejection or refusal of a claim for providers

#### How
Add a message at point of rejection or refusal


app/models/claim_state_transition_reason.rb
```diff
@@ -2,60 +2,134 @@ class ClaimStateTransitionReason
...
     rejected: {
-      no_indictment: 'No indictment attached',
-      no_rep_order: 'No Magistrates’ representation order attached (granted before 1/8/2015)',
-      time_elapsed: 'Claim significantly out of time with no explanation.',
-      no_amend_rep_order: 'No amending representation order',
-      case_still_live: 'Case still live',
-      wrong_case_no: 'Incorrect case number',
-      wrong_maat_ref: 'Wrong MAAT reference',
-      other: 'Other'
+      no_indictment: {
```
lostie, line 17: 💭 Wouldn't this be better in a configuration file that this class reads from? (Feels like it, at least for me, a lot of it seems to be wording for a specific keyword, that as it stands, can't even be localized).

app/services/claims/state_transition_message_builder.rb
```diff
@@ -0,0 +1,36 @@
...
+    def long_reasons
+      reasons.each_with_object([]) do |reason, arr|
+        details = reason_details(reason, reason_text)
+        arr << "#{details.description}: #{details.long_description}"
+      end
+    end
+
+    def reasons_message
+      <<~MSG
+        Your claim has been #{@state}:
```
lostie, line 30: 💭 Consider defining this copy in the locales with both _state_ and _long_reasons_ as variables that can be interpolated.

lostie (approved):
Codewise LGTM 👍 

Would be nice to have all that configurations somewhere else where can more easily be changed/localised without changing the classes/code every time 🏆

app/models/claim_state_transition_reason.rb
jsugarman: yeah, i did consider taking this whole hash constant out as it is lookup data fundamentally. the options seem to me to be:

1. create a separate class/module just for the has
2. yamlize or other format it and load once on app start
3. store in db

first option the simplest - what do you think?

app/services/claims/state_transition_message_builder.rb
jsugarman: if i had time i had considered I18n'ing all the hash strings. seems more consistent to have one approach.

app/models/claim_state_transition_reason.rb
lostie: 💭 If you move the `short` and `long` to the locales, the list can remain here looking just like this, maybe:

 ```ruby
TRANSITION_REASONS = {
  rejected: %i[no_indictment no_rep_order time_elapsed no_amend_rep_order case_still_live wrong_case_no wrong_maat_ref other],
  disbursement: ...
}
```

File: app/models/claim_state_transition_reason.rb [hunk 3]
```diff
@@ -2,60 +2,140 @@ class ClaimStateTransitionReason
...
-  TRANSITION_REASONS = {
+  def self.klass_scope
+    [to_s.demodulize.underscore.downcase.to_sym]
+  end
+
+  def self.t(local_scope)
+    I18n.t(local_scope, scope: klass_scope)
+  end
+
+  TRANSITION_REASON_KEYS = {
```
lostie, line 15: Given the new changes (with the localised values), couldn't this just be something like?

```ruby
# the file could even be loaded before in some sort of initialiser
TRANSITION_REASON_KEYS = YAML.load_file(<claim_transitions_locale_file>.yml).freeze
```

what's the reason for all the empty strings?

app/models/claim_state_transition_reason.rb
```diff
@@ -72,22 +152,29 @@ def refuse_reasons_for(claim)
       reasons_for("refused_#{claim.class.to_s.demodulize.tableize}")
     end
 
+    def transition_reasons
```
lostie, line 155: **[not a blocker]** Have you considered having this to return the list of `ClaimStateTransitionReason` object instances instead of the so that methods below like `reasons_for(state)` just return the matching instance, rather done initialising one.

app/presenters/claim_state_transition_presenter.rb
```diff
@@ -15,6 +15,20 @@ def audit_users
     format(sentence, author: transition.author.name, subject: transition.subject&.name)
   end
 
+  def reason_header
+    "#{'Reason'.pluralize(transition.reason.size)} provided:"
```
lostie, line 19: **[not a blocker]** This seems to be trying to achieve the same as this: https://robots.thoughtbot.com/pluralizing-rails-i18n-translations. Consider using the Rails one instead.

app/models/claim_state_transition_reason.rb
jsugarman: they are loaded dynamically based on the key paths in the hash and the translation follows that format

lostie (approved):
LGTM 👍 although the transitions bit seems a bit weird.

PR #2329, opened 2018-04-13
Upgrade to Rails 5

lostie:
#### What

- Update gems
- `quiet_assets` gem is no longer required as sprockets-rails has the ability to do so.
(https://rossta.net/blog/quiet-assets-in-rails-5.html#replacing-the-quiet_assets-gem)
- Squeel gem is not compatible with Rails 5, replaced with baby_squeel (https://github.com/rzane/baby_squeel)
- Update all models that inherited from ActiveRecord::Base to now inherit from the new ApplicationRecord
- Update all jobs that inherited from ActiveJob::Base to now inherit from the new ApplicationJob
- Replace uses of before_filter with before_action
- Custom module Sort that implemented the method sort was causing arity conflicts with ActiveRecord/Array collections method sort. Renamed the method name in the module to get around the problem for the time being.
- Update bin/ and config/ files with the new Rails 5 defaults/conventions. (Preserved any custom configuration made before the upgrade)


app/models/claim/base_claim.rb
```diff
@@ -157,8 +157,8 @@ def assessment
 
     scope :dashboard_displayable_states, -> { where(state: Claims::StateMachine.dashboard_displayable_states) }
 
-    scope :total_greater_than_or_equal_to, ->(value) { where { total >= value } }
-    scope :total_lower_than, ->(value) { where { total < value } }
+    scope :total_greater_than_or_equal_to, ->(value) { where(arel_table[:total].gteq(value)) }
```
colinbruce, line 160: [Not a blocker]  Why switch to Arel syntax?  Was there a gem issue?

colinbruce (approved):
Code review... :+1: 
I'd like to run some local machine tests for peace of mind, plus we should definitely deploy to an env and get sign off, but either way, awesome work @lostie

app/models/claim/base_claim.rb
jsugarman: agree with colin, prefer to avoid arel unless necessary

app/models/claims/allocation_filters.rb
```diff
@@ -47,7 +47,7 @@ def all_graduated_fees
 
     def all_risk_based_bills
       where(type: 'Claim::LitigatorClaim')
-        .where(offence_id: Offence.joins(:offence_class).where(offence_class: { class_letter: %w[E F H I] }))
+        .where(offence_id: Offence.joins(:offence_class).where(offence_classes: { class_letter: %w[E F H I] }))
```
jsugarman, line 50: have you got a reference for this.

spec/form_builders/adp_text_field_spec.rb
```diff
@@ -160,7 +160,7 @@ def b100_with_value_with_hint
     context 'errored value with hint' do
       it 'produces error text' do
         resource.case_number = nil
-        resource.errors[:case_number] = 'Validation error here'
+        resource.errors.add(:case_number, 'Validation error here')
```
jsugarman, line 163: this will need changing in the validation helpers to i imagine. which you may have done in later commits (just working through commit by commit atm)

app/interfaces/api/helpers/resource_helper.rb
```diff
@@ -7,11 +7,11 @@ def declared_params
     end
 
     def claim_source
-      params.source || 'api'
+      params[:source] || 'api'
```
jsugarman, line 10: no patch for grape then?

spec/api/v1/dropdown_data_spec.rb
```diff
@@ -127,14 +127,14 @@
     end
 
     it 'should include the offence class as nested JSON' do
-      response = get OFFENCE_ENDPOINT, params
+      response = get OFFENCE_ENDPOINT, params: params
```
jsugarman: something to like in rails 5

app/models/claim/base_claim.rb
lostie: This syntax `{ where { total < value }` (the use of curly block in the `where` clause specifically)  was not been interpreted correctly since the upgrade.

app/models/claims/allocation_filters.rb
lostie: http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-includes (at the end of this section)

> Note that includes works with association names while references needs the actual table name.

lostie: https://stackoverflow.com/questions/16349195/how-to-join-an-associated-table-with-a-has-one-association

PR #2330, opened 2018-04-17
Extend allocation logging

colinbruce:
#### What?
Add a transaction attribute to the allocation handler
#### Why?
This is to allow better atomic filtering in Kibana as multiple, long running,
allocation jobs can overlap in kibana making it awkward to filter


jsugarman (approved):
👍

PR #2332, opened 2018-04-17
Split Ground Hearing fee into whole and half day

lostie:
#### What

**NOTE:** It was discovered post release of the fee reform changes that this fee should be in fact two fees for whole and half day.

#### Ticket

[Ground rules hearing - whole and half day](https://www.pivotaltracker.com/story/show/156687178)

#### How

⚠️ Data migration is required:

```shell
SEEDS_DRY_MODE=false bundle exec rake fee_types:seed
```


colinbruce (approved):
:+1:

The `101` fee has not been used in live, so minimal impact on existing claims

PR #2333, opened 2018-04-18
Offences scheme 10 - design changes

cleargif:
PT:
https://www.pivotaltracker.com/story/show/156577013

Updated design
Removed links
Highlight search terms

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -154,11 +154,6 @@ form{
       text-decoration: none;
     }
 
-    .link-grey:hover {
-      color: #6f777b !important;
-      text-decoration: underline;
-    }
-
     .align-right {
       text-align: right;
     }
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/application.css.scss
```diff
@@ -154,11 +154,6 @@ form{
       text-decoration: none;
     }
 
-    .link-grey:hover {
```
lostie, line 157: Code being removed 😍

app/assets/stylesheets/moj/_autocomplete.scss
```diff
@@ -42,7 +47,7 @@
   }
 
   &:hover {
-    cursor: pointer;
+    // cursor: pointer;
```
lostie: 💭 If we're not using this, we might as well get rid of it rather that leaving it commented out.

lostie (approved):
LGTM 👍 , no blockers.

PR #2334, opened 2018-04-18
Api scheme ten endpoint updates

colinbruce:
#### What
Update the API endpoints to add Scheme Ten specific values
#### Why
Some of the dropdowns now have scheme 10 specific values that were not addressed
#### TODO (wip)
 - [x] write release notes
 - [x] Fee Types
 - [x] Advocate categories

### DIFF
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -9,6 +9,12 @@ module API
         params :role_filter do
           optional :role, type: String, desc: I18n.t('api.v1.dropdown_data.params.role_filter'), values: %w[agfs lgfs]
         end
+        params :scheme_ten_role_filter do
+          optional :role,
+                   type: String,
+                   desc: I18n.t('api.v1.dropdown_data.params.role_filter'),
+                   values: %w[agfs_scheme_9 agfs_scheme_10 lgfs]
+        end
 
         def role
           params.role.try(:downcase).try(:to_sym) || :all
@@ -36,10 +42,15 @@ module API
         end
 
         resource :advocate_categories do
-          # TODO: this might need to be changed given there's a different list
-          # of advocate categories depending on the fee scheme in use
           desc 'Return all Advocate Categories'
-          get { Settings.advocate_categories }
+          params { use :scheme_ten_role_filter }
+          get do
+            if role.eql?(:agfs_scheme_10)
+              Settings.agfs_reform_advocate_categories
+            else
+              Settings.advocate_categories
+            end
+          end
         end
 
         resource :trial_cracked_at_thirds do
@@ -74,10 +85,7 @@ module API
         resource :fee_types do
           params do
             category_types = %w[all basic misc fixed graduated interim transfer warrant].to_sentence
-            optional :role,
-                     type: String,
-                     desc: I18n.t('api.v1.dropdown_data.params.role_filter'),
-                     values: %w[agfs_scheme_9 agfs_scheme_10 lgfs]
+            use :scheme_ten_role_filter
             optional :category,
                      type: String,
                      default: 'all',
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -17,11 +17,11 @@ module API
         end
 
         def role
-          params.role.try(:downcase).try(:to_sym) || :all
+          params[:role]&.downcase&.to_sym || :all
         end
 
         def category
-          params.category.try(:downcase)
+          params[:category]&.downcase
         end
       end
 
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -13,6 +13,10 @@ module API
         def role
           params.role.try(:downcase).try(:to_sym) || :all
         end
+
+        def category
+          params.category.try(:downcase)
+        end
       end
 
       after do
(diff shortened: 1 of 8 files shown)


app/models/concerns/roles.rb
```diff
@@ -8,7 +8,11 @@ module Roles
     klass.validate :roles_valid
 
     klass::ROLES.each do |role|
-      klass.scope role.pluralize.to_sym, -> { matching_role_query([role.to_s]) }
+      if role.split('_').last.numeric?
```
jsugarman: can leave as is or could also do it by overriding/patching inflections. 

```
ActiveSupport::Inflector.inflections do |inflect|
  inflect.irregular 'agfs_scheme_9', 'agfs_scheme_9'
end
```

not sure if you can do this 'locally' or whether you have to use `config/initializers/inflections.rb`

app/models/claim/advocate_claim.rb
```diff
@@ -180,16 +180,16 @@ def eligible_basic_fee_types
...
+      Fee::BasicFeeType.agfs_scheme_9
     end
 
     def eligible_misc_fee_types
       # TODO: this should return a list based on the current given fee scheme
       # rather than conditionally return scheme 10 specifically
       # TBD once all the fee scheme work is integrated
-      return Fee::MiscFeeType.agfs_scheme_10s if fee_scheme == 'fee_reform'
-      Fee::MiscFeeType.agfs_scheme_9s
+      return Fee::MiscFeeType.agfs_scheme_10 if fee_scheme == 'fee_reform'
```
jsugarman: just a comment as i do not understand how this effects the api but i rather like the plural version

app/interfaces/api/v1/dropdown_data.rb
```diff
@@ -13,6 +13,10 @@ class DropdownData < API::Helpers::GrapeApiHelper
         def role
           params.role.try(:downcase).try(:to_sym) || :all
         end
+
+        def category
+          params.category.try(:downcase)
```
jsugarman: not a blocker - alternative `params.category&.downcase`

app/models/concerns/roles.rb
lostie: **[not a blocker]** More of a tiny improvement for DRYness.

💭 Consider this instead:

```ruby
scope_name = role.split('_').last.numeric? ? role.to_sym : role.pluralize.to_sym
klass.scope scope_name, -> { matching_role_query([role.to_s]) }
```

or ultimately, if this is a rule that applies to **all strings** ending in numerics:

```ruby
# config/initializers/inflections.rb
ActiveSupport::Inflector.inflections(:en) do |inflect|
  ...
  inflect.plural /^(.*\d)$/, '\1'
end
```

 and in the latter, the code would remain unchanged.

app/views/pages/api_release_notes.html.haml
```diff
@@ -2,6 +2,47 @@
 
 %section.api-documentation
 
+  %hr
+  %h2 19 April 2018
```
lostie: **[not a blocker]** Maybe a 💭 for a possible feature. Seems like if we had some structured release notes (in either DB or even YAML notation) we could more easily updated them, where this template would just recursively render all of that information.

💡Something to thing about...

jsugarman (approved):
👍 LGTM - just the possible use of inflections for irregular pluralization. as @lostie also suggested.

lib/extensions/string_extension.rb
```diff
@@ -29,6 +29,12 @@ def strftime(format)
     Time.zone.parse(self).strftime(format)
   end
 
+  def numeric?
```
lostie: 💭 Consider renaming this method to something that describes more accurately its behaviour (e.g. `has_numeric_ending?`, `ends_with_numeric?`), as the method name, as is, seems misleading.

app/models/concerns/roles.rb
jsugarman: 👍 @lostie - although i think handling the exceptions explicitly via `.irregular` is clearer, a generic numeric inflector might confuse things with patched rails magic

app/interfaces/api/v1/dropdown_data.rb
```diff
@@ -9,9 +9,19 @@ class DropdownData < API::Helpers::GrapeApiHelper
         params :role_filter do
           optional :role, type: String, desc: I18n.t('api.v1.dropdown_data.params.role_filter'), values: %w[agfs lgfs]
         end
+        params :scheme_ten_role_filter do
+          optional :role,
+                   type: String,
+                   desc: I18n.t('api.v1.dropdown_data.params.role_filter'),
+                   values: %w[agfs_scheme_9 agfs_scheme_10 lgfs]
```
jsugarman: As discussed i would include agfs here to and default this to behaving as agfs_scheme_9 - or vice versa - to maintain current functionality -i think this is probably needed as some vendor client apps might use this filtering option to populate list items

lib/extensions/string_extension.rb
colinbruce: I'm removing it anyway as it's no longer needed but, in truth, it did check the entire string.  I split it in the, also deleted, concern
```ruby
if role.split('_').last.numeric?
```
It was only checking the numeracy of the final string in the array... but anyway, it's going! :)

spec/api/v1/dropdown_data_spec.rb
```diff
@@ -187,6 +187,7 @@ def get_filtered_fee_types(category=nil)
       allow(Settings).to receive(:agfs_reform_advocate_categories).and_return(['QC', 'Leading junior', 'Junior'])
       allow(Settings).to receive(:advocate_categories).and_return(['QC', 'Led junior', 'Leading junior', 'Junior'])
       params.merge!(role: role)
+      binding.pry if role.eql?('agfs_scheme_10s')
```
lostie: ✂️

colinbruce: 🤦‍♂️ thanks

lostie (approved):
LGTM 👍

PR #2335, opened 2018-04-19
Sidebar calculations: VAT vs No VAT

cleargif:
PT:
https://www.pivotaltracker.com/story/show/156912584

VAT is incorrectly calculated on LGFS draft claims.


colinbruce (approved):
:+1:

PR #2339, opened 2018-04-23
MI for scheme ten claims

colinbruce:
#### What
Allow the LAA Service Development team can access the required scheme ten data they need for monitoring 
#### Why
Currently, the team have to manually extract and process the data for sending via email
#### How
Add an MI endpoint to the API.  It allows JSON or CSV output.  We can create an MI user without a login, but with an API  key and allow them to hit the url to download the data


app/interfaces/api/v2/mi_query_helper.rb
```diff
@@ -0,0 +1,40 @@
+module API::V2
+  module MIQueryHelper
+    extend Grape::API::Helpers
```
lostie: I know this is currently being used for the **API** only, but the class itself seems to be defining a query only. I would suggest this is in a called something like `Claims::MI::Queries` (as a module/class) inside `app/models/claims/mi/queries' and you could just call it as:

```ruby
Claims::MI::Queries.scheme_ten_claims(start_date, end_date)
```

Doesn't feel right having all that `SQL` as a Grape helper.

app/interfaces/api/v2/mi_query_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+          INNER JOIN courts ON courts.id = c.court_id
+          LEFT OUTER JOIN offences o on o.id = c.offence_id
+            LEFT OUTER JOIN offence_bands ob ON o.offence_band_id = ob.id
+          INNER JOIN external_users eu ON eu.id = c.external_user_id
+            INNER JOIN users u ON u.persona_id = eu.id
+            INNER JOIN providers p ON p.id = eu.provider_id
+        WHERE
+          c.last_submitted_at BETWEEN '#{start_date}' AND '#{end_date}'
+          AND c.type LIKE 'Claim::Advocate%'
+          AND ro.representation_order_date >= '2018-04-01'
```
lostie: **[not a blocker]** Can this be read from an actual record or setting? It works for now, but again, would be better to have from the actual source.

app/interfaces/api/v2/mi_query_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+          INNER JOIN defendants d ON d.claim_id = c.id
+            INNER JOIN representation_orders ro ON ro.defendant_id = d.id
+          INNER JOIN courts ON courts.id = c.court_id
+          LEFT OUTER JOIN offences o on o.id = c.offence_id
+            LEFT OUTER JOIN offence_bands ob ON o.offence_band_id = ob.id
+          INNER JOIN external_users eu ON eu.id = c.external_user_id
+            INNER JOIN users u ON u.persona_id = eu.id
+            INNER JOIN providers p ON p.id = eu.provider_id
+        WHERE
+          c.last_submitted_at BETWEEN '#{start_date}' AND '#{end_date}'
```
lostie: Be wary of SQL injection, more so when using `execute` statements. 

Consider using 
`ActiveRecord::Base.sanitize(start_date)` 
and 
`ActiveRecord::Base.sanitize(end_date)`

to ensure the values are properly escape before the query is executed.

lostie: This is not strictly related with this PR, but it seems like the claims column **last_submitted_at** is being used in different places for query/filtering purposes. It might be a good candidate for an index 📇 .

app/interfaces/api/v2/mi/scheme_ten.rb
```diff
@@ -0,0 +1,44 @@
+module API
+  module V2
+    module MI
+      class SchemeTen < Grape::API
```
lostie: 💭 Consider renaming the file/class to something like `API::V2::MI::AgfsSchemeTenClaims`, first because the resource appears to be _claims_ and second because _scheme ten_ seems to apply only specifically to *AGFS*

app/interfaces/api/entities/scheme_ten_claim.rb
```diff
@@ -0,0 +1,15 @@
+module API
+  module Entities
+    class SchemeTenClaim < BaseEntity
```
lostie: Following https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2339/files#r183688638, this class should probably be called `AgfsSchemeTenClaim ` or even put that as a namespace (`AgfsSchemeTen::Claim`), up to you ;)

app/interfaces/api/v2/mi_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+          start_date = Date.new(2018, 4, 1).beginning_of_day
+          end_date = Date.yesterday.end_of_day
+        else
+          start_date = params[:date].to_date.beginning_of_day
+          end_date = params[:date].to_date.end_of_day
+        end
+        [start_date.strftime('%FT%T'), end_date.strftime('%FT%T')]
+      end
+
+      def build_csv_from(data)
```
lostie: **[not a blocker]** Consider defining a service that produces the CSV for the MI report and calling the service from the API. The main advantage of it is that all the logic to produce the report is not tied to the API and can be potentially called from else.

app/interfaces/api/v2/mi_query_helper.rb
colinbruce: For now, they specifically want claims after that date... in future it could be made more generic

app/interfaces/api/v2/mi/agfs_scheme_ten_claims.rb
```diff
@@ -0,0 +1,44 @@
...
+        require 'csv'
+        helpers Claims::MI::Queries
+        helpers API::V2::MIHelper
+
+        content_type :csv, 'text/csv; utf-8'
+        default_format :json
+
+        resource :mi, desc: 'MI endpoint' do
+          params do
+            optional :api_key, type: String, desc: 'REQUIRED: The API authentication key of the user'
```
lostie, line 14: **[not a blocker]** Something I've just noticed now, if the description for this field is accurate, we should probably define as such:

```ruby
requires :api_key, type: String, desc: 'REQUIRED: The API authentication key of the user'
```

To enforce that _param_ to actually be mandatory.

lostie (approved):
LGTM 👍 

I think there's still some mixing of API vs database concepts, but we can address later on.

app/interfaces/api/v2/mi/agfs_scheme_ten_claims.rb
colinbruce: I seem to remember that there was an issue with validation (of course!) and setting these fields to `requires`, hence why this pattern of `optional= => REQUIRED` all the way through the API :(

PR #2340, opened 2018-04-23
Check your claim page changes

lostie:
#### What

* Re-styling the "Check your claim" page and add links to each of the sections so users can edit them directly, rather than having to go step by step until reaching the named section.
* When the users are on the claims list page and click on a claim that is in _draft_ the page shown is the "Check your claim" page
* The users are not able to save as draft for _case details_ and _transfer fee details_ as their the most basic information that is required in order for the claim to exist in the system. **NOTE:** This only applies for the web submissions. API submissions remained the same.
* In the web submissions, each form step is now validated independently from any other steps. **NOTE:** This only applies for the web submissions. API submissions remained the same.
* **DRY** form templates. Remove repetition from form templates (first step for this process, that requires further **DRYness**).
* "Check Your Claim" page display a top level banner if the claim is **NOT** valid and informs the user of which sections still need to be completed.
* For each section, if the section cannot be edited, an info banner is displayed informing the user of which other sections need to be completed before being able to edited.
* Assigns total attributes for _litigator final/interim/transfer_ claims before validation so that its possible to correctly validate the associated fees total in each of their own steps.

#### Ticket

[Change links on the 'check your claim page'](https://www.pivotaltracker.com/story/show/156600656)


app/presenters/claim/interim_claim_presenter.rb
```diff
@@ -25,15 +25,18 @@ def type_identifier
     'lgfs_interim'
   end
 
-  def raw_fixed_fees_total
```
lostie, line 28: This doesn't seem to be needed for this type of claim.

app/controllers/external_users/claims_controller.rb
```diff
@@ -426,7 +426,7 @@ def submitting_to_laa?
   end
 
   def continue_claim?
-    params.key?(:commit_continue) || params.key?(:commit_stage_1)
```
lostie, line 429: Logic behind `commit_stage_1` was no different from `commit_continue`

app/controllers/external_users/claims_controller.rb
```diff
@@ -17,8 +17,8 @@ class ResourceClassNotDefined < StandardError; end
 
   before_action :set_and_authorize_claim, only: %i[show edit update unarchive clone_rejected destroy summary
                                                    confirmation show_message_controls messages disc_evidence]
-  before_action :redirect_unless_editable, only: %i[edit update]
   before_action :set_form_step, only: %i[edit]
+  before_action :redirect_unless_editable, only: %i[edit update]
```
lostie, line 21: The order of the before actions was changed as `redirect_unless_editable` now makes use of the _form_step_, so needs to have it defined before hand.

app/controllers/external_users/claims_controller.rb
```diff
@@ -448,7 +448,7 @@ def redirect_to_next_step
                   title: '%{action_t} %{type} claim page %{step}'
                 }, claim_tracking_substitutions)
 
-    redirect_to edit_polymorphic_path(@claim, step: @claim.next_step!)
+    redirect_to edit_polymorphic_path(@claim, step: @claim.next_step)
```
lostie, line 451: This is redirecting, so the `@claim` instance doesn't need to be mutated here to have the next step as the action called with assign the `form_step` with the `@claim.next_step`.

app/models/claim/advocate_claim.rb
```diff
@@ -101,22 +101,32 @@ class AdvocateClaim < BaseClaim
...
         transitions: [
-          { to_stage: :basic_and_fixed_fees }
-        ]
+          { to_stage: :basic_fees }
+        ],
+        dependencies: %i[case_details defendants]
       },
       {
-        name: :basic_and_fixed_fees,
+        name: :basic_fees,
```
lostie, line 118: Splitting **basic_fees** and **fixed_fees** as they should be 2 different steps anyway.

app/models/claim/advocate_interim_claim.rb
```diff
@@ -24,27 +24,23 @@ class AdvocateInterimClaim < BaseClaim
       {
         name: :defendants,
         transitions: [
-          {
-            to_stage: :offence_details,
-            condition: ->(claim) { !claim.fixed_fee_case? }
-          },
-          {
-            to_stage: :basic_and_fixed_fees,
```
lostie, line 32: This was not required for advocate interim claims, probably the result of a bad copy/paste 😓 .

app/models/stage_collection.rb
```diff
@@ -24,7 +24,7 @@ def previous_stage(stage)
   end
 
   def next_stage(stage)
-    find_stage(stage)&.first_valid_transition
+    find_stage(find_stage(stage)&.first_valid_transition)
```
lostie, line 27: Ensure the result of this method is always a `Stage` object instead of a `symbol`.

config/locales/en.yml
```diff
@@ -874,10 +919,12 @@ en:
         heading: Interim fees
         summary:
           caption: 'Interim Fees: This section contains details of any interim fees that have been claimed.'
+          header: Interim fees
           summary: 'Note: To best navigate this table, select the right hand column and then step down through the rows.'
-      tranfer_detail:
+      transfer_detail:
```
colinbruce, line 925: 👍

colinbruce (approved):
Looking good...

Manual testing of API testing starts now

PR #2343, opened 2018-04-24
Fix for the display of reject reasons

colinbruce:
#### What
When a claim was submitted with _no_ fees, the code would error
#### Why
It's a positive check for a single fee type to add the extra value.  So, if there are _no_ fees, we don't need to add it
#### How
Added safe navigation operators (&.) to check for the presence of fees

### DIFF
--- a/app/models/claim_state_transition_reason.rb
+++ b/app/models/claim_state_transition_reason.rb
@@ -64,7 +64,7 @@ class ClaimStateTransitionReason
 
     def reject_reasons_for(claim)
       reasons = reasons_for('rejected')
-      reasons.insert(6, reasons_for(:disbursement)) if claim.fees.first.fee_type.code.eql?('IDISO')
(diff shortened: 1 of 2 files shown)


app/models/claim_state_transition_reason.rb
```diff
@@ -64,7 +64,7 @@ def reasons(state)
 
     def reject_reasons_for(claim)
       reasons = reasons_for('rejected')
-      reasons.insert(6, reasons_for(:disbursement)) if claim.fees.first.fee_type.code.eql?('IDISO')
+      reasons.insert(6, reasons_for(:disbursement)) if claim&.fees&.first&.fee_type&.code.eql?('IDISO')
```
lostie, line 67: *&* 🚂 , choo, choo! 😆

lostie (approved):
LGTM 👍

PR #2345, opened 2018-04-25
Advocate category validation Scheme 10

colinbruce:
#### What
Vendors have been unable to submit claims via the API
#### Why
This is because API submissions don't have a rep-order, but do have an offence.
#### How
Check for a SchemeTen offence as well as the presence of a post S10 rep order date.

----
#### To do
- [x] extend tests to ensure validations are firing properly

### DIFF
--- a/app/services/claims/fetch_eligible_advocate_categories.rb
+++ b/app/services/claims/fetch_eligible_advocate_categories.rb
@@ -10,7 +10,7 @@ module Claims
 
     def call
       return unless claim&.agfs?
-      return fee_reform_categories if using_fee_reform?
+      return fee_reform_categories if using_fee_reform? || Offence.in_scheme_ten.include?(claim.offence)
       default_categories
     end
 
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
👍 makes me think more and more that we need an attribute on claim like `fee_scheme` which can be set internally by the web app, but explicitly by the api submissions.

app/services/claims/fetch_eligible_advocate_categories.rb
```diff
@@ -10,7 +10,7 @@ def initialize(claim)
 
     def call
       return unless claim&.agfs?
-      return fee_reform_categories if using_fee_reform?
+      return fee_reform_categories if using_fee_reform? || Offence.in_scheme_ten.include?(claim.offence)
```
lostie, line 13: 👍 (although, it feels `Offence.in_scheme_ten.include?(claim.offence)` seems to be something that can also determine the claim's fee scheme?)

lostie (approved):
LGTM 👍

PR #2346, opened 2018-04-26
Clarify scheme ten offence api

colinbruce:
#### What
Add extra `act_of_law` field to offence output in API
#### Ticket
[Clarify scheme ten offence data](https://www.pivotaltracker.com/story/show/157116611)
#### Why
Contact from the Vendor Support Slack raised a query about the offence data... there was nothing to distinguish some offences.
> Codes 2129 and 2133 appear to be identical and both have a description of "Distributing, showing or playing a recording". I think one should be “Distributing, showing or playing a recording - Public Order Act 1986, s.21”, and the other “Distributing, showing or playing a recording - Public Order Act 1986, s.29E”. There are several others that have the same problem. Could you confirm please if this is correct ?

and
> there appear to be some duplicates under Category 17 - Band 17.1 - "TEW Offences" - 1953,2001,2037,2042,2047,2097,2122 (also 1065 / 1741 - but these two are under different bands)
#### How
Expose extra field in the API offence entity

### DIFF
--- a/app/interfaces/api/entities/offence.rb
+++ b/app/interfaces/api/entities/offence.rb
@@ -13,6 +13,7 @@ module API
         expose :offence_class,
                if: ->(instance, _opts) { instance.scheme_nine? },
                using: API::Entities::OffenceClass
+        expose :act_of_law, if: ->(instance, _opts) { instance.scheme_ten? }
         expose :offence_band,
                if: ->(instance, _opts) { !instance.scheme_nine? },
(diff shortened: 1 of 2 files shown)


lostie (approved):
LGTM 👍 , only question I have if these changes only apply to the API or should the users submitting claims through the app see the year chapter as well.

colinbruce:
The users do see it via the app, it's in the footer of the offence search result panels

PR #2347, opened 2018-04-26
Prevent message creation for [Part] Authorised claims

jsugarman:
#### What
prevent messages being created for authorised and part authorised claims

#### Why
Should only create messages when there are reasons (rejected/refused transitions)


lostie (approved):
LGTM 👍 , is there any tests that cover this functionality, (if not) maybe would be worth having one that covers this change.

PR #2348, opened 2018-04-26
Clear transition reasons when assessment included

jsugarman:
#### What
Quick fix for an existing bug related to transitioning

[Bug: [Part] Authorise can record reject/refuse reasons](https://www.pivotaltracker.com/story/show/156965581)

#### Why
It is possible to provide reject/refuse reasons and then [part] authorise
which will add messages.


colinbruce (approved):
LGTM 👍

jsugarman:
Implemented validation instead - can revisit if necessary

PR #2349, opened 2018-04-26
Rails 5 - allocation fix

colinbruce:
#### What
Allocation was broken due to updated handling of boolean returns from PSQL/ActionRecord
#### Why
Because it was broken
#### How
Remove test of returned `.('t')`


jsugarman (approved):
LGTM 👍

PR #2350, opened 2018-04-26
Fix where chain exception

lostie:
Sentry Issue: https://sentry.service.dsd.io/mojds/private-beta/issues/30658/

```
undefined method `count' for #<ActiveRecord::QueryMethods::WhereChain
```

#### What

Fix where chain exception

#### How

Use of the normal brackets syntax instead of the curly brackets syntax to perform the query.


colinbruce (approved):
👍

PR #2352, opened 2018-04-27
Refactor case worker claim updater

jsugarman:
#### What
Refactor case worker claim updater service and validate reasons presence

#### Ticket

relates to [Bug: [Part] Authorise can record reject/refuse reasons](https://www.pivotaltracker.com/story/show/156965581)

#### Why
This service object was doing to much and it was possible to 
provide reject/refuse reasons and then authorise the claim,
thereby adding a message to the claim.

#### How
extract validation logic to a validator and share common functionality in concern

#### further actions
ticket to prevent reasons from being sent from form when authorising?!

### DIFF
--- a/app/services/claims/case_worker_claim_updater.rb
+++ b/app/services/claims/case_worker_claim_updater.rb
@@ -22,7 +22,6 @@ module Claims
       event = Claims::InputEventMapper.input_event(state)
 
       claim.class.transaction do
-        # @claim.update(@params)
         update_assessment if assessment_params
         add_redetermination if redetermination_params
         claim.send(event, audit_attributes) unless state_not_updateable?
@@ -34,7 +33,7 @@ module Claims
     end
 
     def state_not_updateable?
-      state.blank? || state == claim.state
+      state.blank? || state.eql?(claim.state)
     end
 
     def update_assessment
@@ -45,11 +44,10 @@ module Claims
       claim.redeterminations.create(redetermination_params)
     end
 
-    # rubocop:disable Metrics/LineLength
     def add_message
-      claim.messages.create(sender_id: current_user.id, body: transition_message) if Release.reject_refuse_messaging_released?
+      return unless Release.reject_refuse_messaging_released?
+      claim.messages.create(sender_id: current_user.id, body: transition_message)
     end
-    # rubocop:enable Metrics/LineLength
 
     def transition_message
       StateTransitionMessageBuilder.new(state, transition_reasons, transition_reason_text).call
--- a/app/services/claims/case_worker_claim_updater.rb
+++ b/app/services/claims/case_worker_claim_updater.rb
@@ -27,7 +27,7 @@ module Claims
         claim.send(event, audit_attributes) unless state_not_updateable?
         add_message if transition_reasons.present?
       rescue StandardError => err
-        add_error(err.message)
+        @result = validator.add_error(err.message)
         raise ActiveRecord::Rollback
       end
     end
--- a/app/services/claims/case_worker_claim_updater.rb
+++ b/app/services/claims/case_worker_claim_updater.rb
@@ -52,16 +52,18 @@ module Claims
 
     def validate_params
       if @assessment_params_present || @redetermination_params_present
-        validate_state_when_value_params_present
+        validate_state_when_assessment_present
+        clear_transition_reasons
       elsif @state.blank?
         add_error 'You should select a status'
+
       else
-        validate_state_when_no_value_params
+        validate_state_when_no_assessment_present
         validate_reason_presence
       end
     end
 
-    def validate_state_when_value_params_present
+    def validate_state_when_assessment_present
       if @state.blank?
         add_error 'You must specify authorised or part authorised if you supply values'
       elsif @state == 'refused'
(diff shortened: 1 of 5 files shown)


app/services/claims/claim_state_transition_validator.rb
```diff
@@ -0,0 +1,52 @@
...
+      common_determination_validations
+    end
+
+    def validate_refused
+      add_error('You cannot specify values when refusing a claim') if determination_present?
+      common_undetermined_validations
+    end
+
+    def validate_rejected
+      add_error('You cannot specify values when rejecting a claim') if determination_present?
```
jsugarman: try adding to common_undetermined_validations

app/services/claims/claim_state_transition_validator.rb
```diff
@@ -0,0 +1,52 @@
...
+    def validate_authorised
+      common_determination_validations
+    end
+
+    def validate_part_authorised
+      common_determination_validations
+    end
+
+    def validate_refused
+      add_error('You cannot specify values when refusing a claim') if determination_present?
```
jsugarman: try adding to common_undetermined_validations

app/services/concerns/case_worker_claim_param_readable.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def determination_present?
+      assessment_params || redetermination_params
+    end
+
+    def reasons_present?
+      transition_reasons.present? || transition_reason_text.present?
+    end
+
+    def add_error(message, attribute = :determinations)
```
jsugarman: should probably belong to the validator and then use `validator.add_error` in updater as required

app/services/claims/case_worker_claim_updater.rb
```diff
@@ -1,161 +1,64 @@
...
-        'fees' => '0.00',
-        'expenses' => '0.00',
-        'disbursements' => '0.00'
-      }.merge(@redetermination_params)
-      @claim.redeterminations << Redetermination.new(params_with_defaults)
+      claim.redeterminations.create(redetermination_params)
     end
 
     # rubocop:disable Metrics/LineLength
     def add_message
```
lostie: 💭 To get rid of `# rubocop:disable Metrics/LineLength`:

```ruby
def add_message
  return unless Release.reject_refuse_messaging_released?
  claim.messages.create(sender_id: current_user.id, body: transition_message)
end
```

jsugarman: 👍

app/services/claims/case_worker_claim_updater.rb
```diff
@@ -1,161 +1,63 @@
...
-      validate_params
-      update_and_transition_state if @result == :ok
+      @result = validator.call
+      update_and_transition_state if result.eql?(:ok)
       self
     end
 
     private
 
-    def extract_transition_params
```
lostie, line 23: ✂️ ✂️ ✂️  => 🏆 🌟

app/services/claims/claim_state_transition_validator.rb
```diff
@@ -0,0 +1,51 @@
+module Claims
+  class ClaimStateTransitionValidator
```
lostie, line 2: Just a 💭 , not a **blocker**: It would be really nice to start using [ActiveModel::Validator](http://api.rubyonrails.org/classes/ActiveModel/Validator.html) (specially for new ones), rather than implementing our own, more so because we don't need to re-implement any of the existent default checks like:

```ruby
validates_presence_of :state

# instead of 
add_error('You must select a status') if state.blank?
```

Same thing applies to the end result:

```ruby
record.valid?

# instead of
result = validator.call
result.eql?(:ok)
```

jsugarman: This was a familiarising refactor and i had thought of completely moving things to a claim state transition validator but did not go down this route here (limited nature of of this PR), but is worth a spike, because i believe this will cause issues with validation of data that may become erroneous for historical claims by the point of claim processing by the case worker.

app/services/concerns/case_worker_claim_param_readable.rb
```diff
@@ -0,0 +1,82 @@
+module CaseWorkerClaimParamReadable
```
lostie, line 1: 💭 **[Just FYI, maybe for a future PR]** Related with https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2352/files#r184657292, if we were to build, let's say for the sake of this example a `UpdatableCaseWorkerClaim` object in `CaseWorkerClaimUpdater` with the params we received, all these methods would be methods of `UpdatableCaseWorkerClaim` instance, so would be the validations, with no need for sharing concerns across classes, since this _claim_ instance would respond to all of it:

```ruby
class UpdatableCaseWorkerClaim
  include ActiveModel::Validations

  validates_with Claims::ClaimStateTransitionValidator

  def initialize(claim, params)
    @claim = claim
    @params = params 
  end

  def current_user
    @current_user ||= params[:current_user]
  end
  
  ...
end

# in  CaseWorkerClaimUpdater
record = UpdatableCaseWorkerClaim.new(claim, params)
record.valid?
```

Just an idea.

lostie (approved):
LGTM 👍 

Could we potentially create some github issues to address some of the things we discussed in the future, like the of `ActiveModel::Validations`, etc?

PR #2354, opened 2018-05-01
Bug fix: CSS leak

cleargif:
#### What
CSS is leaking and adding highlighting where not required

#### Ticket

[Offence Search Highlight leak](https://www.pivotaltracker.com/story/show/157215060)

#### Why
CSS is not scoped to the correct context

#### How
Applied a more specific style selector

### DIFF
--- a/app/assets/stylesheets/moj/_autocomplete.scss
+++ b/app/assets/stylesheets/moj/_autocomplete.scss
@@ -22,9 +22,7 @@
   overflow-y: scroll;
 }
 
-.tt-highlight{
-  background-color: #ffffca !important;
-}
+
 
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/moj/_autocomplete.scss
```diff
@@ -61,7 +59,12 @@
   padding: 0.5em;
 }
 
+#offence-list .tt-highlight{
+    background-color: #ffffca !important;
+  }
```
lostie: **[nit pick/not a blocker]** the indentation doesn't seem right on lines 63/64.

lostie (approved):
LGTM 👍

PR #2356, opened 2018-05-01
Fix the anonymous data dump for rails 5

jsugarman:
#### What
Fix the anonymous data dump for rails 5

#### Why
Silence Noisy deprecation warnings  and 
fix the TypeError for array types on insert statement creation

Arel changes impact the #to_sql method call
on InsertManager objects.

This is more of a patch than anything else
as it will have to be changed again for
further rails/arel updates.

see below for fix inspiration:
  [arel patch commit](https://github.com/rails/arel/commit/6160bfbda1d1781c3b08a33ec4955f170e95be11)
 [example implementation](https://github.com/IFTTT/polo/commit/8fba3068ab406c95dd395469b30cebf80d4e978e)

[bug ticket](https://www.pivotaltracker.com/story/show/157214871)

#### How
silence deprecation warning with variable and explicitly cast types on insert statement creation

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -79,10 +79,12 @@ namespace :db do
     end
 
     # The following will export the previously excluded tables data, in an anonymised way
+    $arel_silence_type_casting_deprecation = true
     excluded_tables.each do |table|
       task_name = "db:dump:#{table}"
       Rake::Task[task_name].invoke(filename)
     end
+    $arel_silence_type_casting_deprecation = false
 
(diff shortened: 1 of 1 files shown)


lostie (approved):
LGTM 👍

lostie:
💭 For these long lived features, might be worth thinking about moving the data extraction logic to some sort of service that can have some basic testing, to ensure changes on its dependencies don't break its functionality (we already discovered a couple of these issues only post-mortem, as there was no coverage for them).

PR #2357, opened 2018-05-03
Scheme 10 offence di

jsugarman:
#### What
Modify seed and migrators for handing scheme 10 offence data injection (CCR)

#### Ticket

[INJECTION: CCR scheme 10 offences](https://www.pivotaltracker.com/story/show/156533369)

#### Why
Ensure offences in CCCD and CCR have parity and can be identified by a unique code.

#### How
Modify scheme 10 offence data in CCCD and CCR to match and
generate unique identifiers that are known to both.

--------

#### TODO (wip)
 - [x] create and analyse diffs
 - [x] amend seed data in CCCD and/or CCR
 - [x] create generator for unique codes in scheme 10


db/migrate/20180503141759_modify_scheme10_offence_data.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    relation.
+      where(description: 'Offences under regulations 6 (Borrowing and banking of landfill allowances) and 7 (Trading and other transfer of landfill allowances).').
+      update_all(description: 'Offences under regulations under s.6 and s.7')
+
+    relation.
+      where(description: 'Possession or supply of apparatus for use in dishonestly obtaining an electronic communication service.').
+      update_all(description: 'Possession or supply of apparatus etc for contravening s.125')
+
+    Rake::Task['data:migrate:offence_unique_code_scheme_10'].invoke
```
lostie: 🙅‍♂️ Mixing migrations with rake task IMO is **not** a good idea and will come back to bite us eventually.

This ties that rake task as a dependency of the migration, they should not be related/dependent on each other. 

💭 Consider having the rake task behaviour as a service under `<RAILS_ROOT>/db/data` so it remains completely under the data migration scope.

db/migrate/20170824203630_add_unique_code_to_offences.rb
```diff
@@ -1,7 +1,7 @@
 class AddUniqueCodeToOffences < ActiveRecord::Migration
   def up
     add_column :offences, :unique_code, :string, default: 'anyoldrubbish', null: false
-    Rake::Task['data:migrate:offence_unique_code'].invoke
+    Rake::Task['data:migrate:offence_unique_code_scheme_9'].invoke
```
lostie: I've just noticed this is already in, but the [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2357/files#r186106020) also applies here.

db/seed_helper.rb
```diff
@@ -1,13 +1,25 @@
 module SeedHelper
+  # unique_code has unique, not null contraint but cannot be given
+  # expected value until all offences seeded
+  # - see DataMigrator::OffenceUniqueCodeMigrator
+  def self.find_or_create_scheme_10_offence!(attrs)
```
lostie, line 5: **[nit pick/not a blocker]** If all the methods here are `def self.method_name` wouldn't make more sense for this to be a **class** instead of a **module**?

FYI, alternatively you could consider keeping this as a **module** and allowing the module to be `extended/included`, as the _self_ definitions prevents that (although arguably this _module_ doesn't seem to be required anywhere else, hence preferring my first suggestion), by using [module_function](https://apidock.com/ruby/Module/module_function):

```ruby
module SeedHelper
  def find_or_create_scheme_10_offence
    ...
  end
  module_function :find_or_create_scheme_10_offence
end
```

_module_ seems to be overused in the code 😓 even when it doesn't bring any abstraction/behaviour sharing.

db/seeds/scheme_ten.rb
```diff
@@ -50,18 +53,46 @@
...
+  def contrary_to
+    self['contrary_to']
+  end
+
+  def year_chapter
+    self['year_chapter']
+  end
+end
+
+class CSV::Row
```
lostie, line 78: **[not a blocker]** Even though in this case, this patching seems quite self contained, keep seeing core ruby classes/gem classes patched makes me really 😨 . It's **not** a good practice and its only prone to come and bites us in the future if those core ruby classes/gem classes change their interface.

Personally, to achieve that same behaviour, I would just pass the row to a `OffenceCSVRow` instance and call `to_h` on it to give me the key/value pairs that you need to create the offence in the database.

db/migrate/20180503141759_modify_scheme10_offence_data.rb
jsugarman: its complicated

db/seeds/scheme_ten.rb
```diff
@@ -50,18 +53,46 @@
...
+    contrary: row.contrary_to,
+    year_chapter: row.year_chapter
   )
   OffenceFeeScheme.find_or_create_by(offence: offence, fee_scheme: agfs_fee_scheme_ten)
-end
\ No newline at end of file
+end
+
+# regenerate unique codes based on offence description and band where order is significant
+require Rails.root.join('lib','data_migrator','offence_unique_code_migrator').to_s
```
lostie: ❓ What's the `.to_s` for in this **require** ? 😕

lib/data_migrator/offence_code_generator.rb
```diff
@@ -2,13 +2,38 @@ class OffenceCodeGenerator
   attr_reader :offence
 
   def initialize(offence)
+    Offence.send(:include, OffenceExtensions)
```
lostie, line 5: I'm repeating myself, sorry, but I think you have the same problem here with the use of _modules_. Adding this extensions _module_ to the `Offence` class doesn't seem to bring any value compared to the more standard OO approach that was in place already. 

On top of that, if this code is loaded/executed at least once for the main app, now any `Offence` instance will have those extensions available. If we had some of this methods already defined in the `Offence` class or we were to add them, you would end up having collisions between them, in most cases ruby with return a really weird error that has nothing todo with the duplicate methods, and the developers will spend a whole afternoon trying to figure out what is going on.

lib/data_migrator/offence_unique_code_migrator.rb
```diff
@@ -2,41 +2,50 @@
...
+        set[unique_code(offence)] = { id: offence.id, description: offence.description, unique_code: offence.unique_code, contrary: offence.contrary, band: offence.offence_band&.description, class_letter: offence.offence_class&.class_letter }
         set
       end
     end
 
-    def test
-      @offence_set.each do |code, details|
-        ap "UPDATE offences SET unique_code = #{code} WHERE id = #{details[:id]} \# for #{details[:description]}"
+    def migrate!
+      puts '-- clearing existing offences.unique_code data'
```
lostie, line 24: **[nit pick/not a blocker]** 💭 Consider using the `Rails.logger` instead. You can always configure the logger to be set to STDOUT which will have the same effect as this, or redirect the logs to stdout if it's not by default.

db/seed_helper.rb
jsugarman: its required in quite a few of the existing seed files if you search for `require Rails.root.join('db','seed_helper')`

db/seeds/scheme_ten.rb
jsugarman: the only extension to `CSV::Row` it is performing is calling the hash directly which id be surprised/worried if they changed in the core class. The idea is simply to neaten up the seed file which was/is complex enough. 

i guess your idea sacrifices brevity for not patching

db/seed_helper.rb
jsugarman: will give this ago, but feels like yak-shaving of existing functionality

db/seeds/scheme_ten.rb
jsugarman: copy-paste - i thought it was required but, after trying, it is not.

lib/data_migrator/offence_unique_code_migrator.rb
```diff
@@ -47,5 +56,20 @@ def generator(offence = nil)
         @generator
       end
     end
+
+    # rubocop:disable Metrics/CyclomaticComplexity
+    def output(code, offence, format)
```
lostie, line 61: **[not a blocker]** 💭 Consider the following approach to remove the use of extensive case statements (which are considered a bad practice, hence the `CyclomaticComplexity` alert):

```ruby
def output(code, offence, format)
  message = {
    sql: "UPDATE offences SET unique_code = #{code} WHERE id = #{offence[:id]}".yellow,
    diff: offence[:unique_code].eql?(code) ? nil : offence[:description].concat("\n -#{offence[:unique_code]}".red).concat("\n +#{code} ".green)
  }[format.downcase.to_sym]
    text: "-- [would have] updated #{offence[:description]},#{offence[:band] || offence[:class_letter]}".white.concat(" unique_code: #{offence[:unique_code]} --> #{code}".green)
  puts message unless message.blank?
end
```

lostie (approved):
Overall LGTM, only main concern I have is todo with making _migrations_ dependent on _rake tasks_.

lib/data_migrator/offence_code_generator.rb
jsugarman: This second comment/para. would only be the case if you called the `OffenceCodeGenerator` or `Rake::Task['data:migrate:modify_scheme_10_offences'].invoke` but these are temporary, probably one time, "script" tools.

lib/data_migrator/offence_unique_code_migrator.rb
jsugarman: that looks a lot more complex and more work for the method as it has to process more. A simple case statement may have a few more paths through it re `Metrics/CyclomaticComplexity` but is more humanly readable. for a laugh read http://tmont.com/blargh/2010/11/cyclomatic-complexity-is-worthless

PR #2359, opened 2018-05-09
Add scheme 10 advocate category to adapter

jsugarman:
#### What
Add scheme 10 advocate category to adapter

#### Why
A new advocate category was added for AGFS
claims under scheme 10 and this was not being
adapted for DI into CCR.


colinbruce (approved):
LGTM 👍

PR #2364, opened 2018-05-14
Quieten RSpec output

colinbruce:
#### What
After the rails 5 update, there are a lot of deprecation warnings 

#### Why
This can make it hard to debug errors
#### How
Follow the warnings and update blockers
#### Todo
- [x] OffenceController
- [x] Unused parameters passed to Capybara::Queries::SelectorQuery : ["state state-rejected"]
#### Not doing in this PR
- before_filter is deprecated and will be removed in Rails 5.1. (*possibly* paperclip or papertrail gem)
- Shoulda-matchers - waiting for new build (https://github.com/thoughtbot/shoulda-matchers/pull/1009)

### DIFF
--- a/spec/controllers/case_workers/admin/allocations_controller_spec.rb
+++ b/spec/controllers/case_workers/admin/allocations_controller_spec.rb
@@ -91,8 +91,8 @@ RSpec.describe CaseWorkers::Admin::AllocationsController, type: :controller do
     before(:each) do
       expect(CaseWorkerService).to receive(:new).and_return(case_worker_service_instance)
       expect(Claims::CaseWorkerClaims).to receive(:new).and_return(case_worker_claims_instance)
-      expect(Allocation).to receive(:new).with(ActionController::Parameters.new(expected_params).permit!)
-      expect(Allocation).to receive(:new).with(ActionController::Parameters.new(expected_params_with_user).permit!).and_return(allocation)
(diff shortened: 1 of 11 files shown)


spec/controllers/case_workers/admin/allocations_controller_spec.rb
```diff
@@ -91,8 +91,8 @@
     before(:each) do
       expect(CaseWorkerService).to receive(:new).and_return(case_worker_service_instance)
       expect(Claims::CaseWorkerClaims).to receive(:new).and_return(case_worker_claims_instance)
-      expect(Allocation).to receive(:new).with(expected_params)
-      expect(Allocation).to receive(:new).with(expected_params_with_user).and_return(allocation)
+      expect(Allocation).to receive(:new).with(ActionController::Parameters.new(expected_params).permit!)
```
lostie: **[not a blocker/nice to have]** 💭 Given we're likely to have quite some definitions of params as the above, having some helper (would be included in the `spec/rails_helper.rb`) like the one suggested [here](https://stackoverflow.com/a/45468322) might help DRY things a bit and make the expectations shorter and nicer to read.

lostie (approved):
LGTM 👍

spec/controllers/case_workers/admin/allocations_controller_spec.rb
colinbruce: ooh.  Nice find..

PR #2365, opened 2018-05-15
Add advocate interim claim data injection

jsugarman:
#### What
Add adapters for DI of advocate interim claims and new final claim daily attendance

[INJECTION: CCR advocate interim claims](https://www.pivotaltracker.com/story/show/156533432)

[INJECTION: CCR Daily attendance(2+) considerations](https://www.pivotaltracker.com/story/show/156533457)

#### Why
Advocate interim claim - Scheme 10 introduces the ability to claim 
warrant fees on AGFS claims via a new claim type.

Advocate final claim daily attendance changes - scheme 10 introduces
a new simpler daily attendance fee with no banding.

#### How
Add relevant entities and adapters

### DIFF
--- a/app/interfaces/api/v2/ccr_claim.rb
+++ b/app/interfaces/api/v2/ccr_claim.rb
@@ -5,7 +5,12 @@ module API
 
       helpers do
         def claim
-          ::Claim::AdvocateClaim.find_by(uuid: params[:uuid]) || error!('Claim not found', 404)
+          ::Claim::BaseClaim.agfs.find_by(uuid: params[:uuid]) || error!('Claim not found', 404)
+        end
+
+        def entity_class
+          return API::Entities::CCR::AdvocateInterimClaim if claim.interim?
+          API::Entities::CCR::AdvocateClaim
         end
       end
 
@@ -14,7 +19,7 @@ module API
         params { use :common_injection_params }
 
         get ':uuid' do
-          present claim, with: API::Entities::CCR::AdvocateClaim
+          present claim, with: entity_class
         end
       end
     end
--- a/app/services/ccr/daily_attendance_adapter.rb
+++ b/app/services/ccr/daily_attendance_adapter.rb
@@ -1,12 +1,26 @@
 # CCR requires the total number of daily attendances.
 # This can be derived from CCCD data as follows:
+#
+# 1. For AGFS scheme 9 claims:
+#
 # The number of attendances claimed is equal to
 # the total of quantities claimed for the 3 DAF/H/J
 # basic fees + 2 (included in the basic fee). If there
 # are none claimed then the attendance can be said to be
 # the least of actual trial length or 2 (included in the
-# basic fee) - CCR will ignore this value
-# if it is not approrpiate e.g. for guilty pleas
+# basic fee) - CCR will ignore this value if it is not
+# approrpiate e.g. for guilty pleas
+#
+# 2. For AGFS scheme 10 claims:
+#
+# The number of attendances claimed is equal to
+# the quantities claimed for the single DAT
+# basic fee + 1 (included in the basic fee). If there
+# are none claimed then the attendance can be said to be
+# the least of actual trial length or 1 (included in the
+# basic fee) - CCR will ignore this value if it is not
+# approrpiate e.g. for guilty pleas
+#
 module CCR
   class DailyAttendanceAdapter
     attr_reader :claim
@@ -24,23 +38,28 @@ module CCR
 
     def attendances
       if daily_attendance_uplifts?
-        daily_attendance_uplifts + DAILY_ATTENDANCES_IN_BASIC
+        daily_attendance_uplifts + daily_attendances_in_basic
       else
-        [trial_length, DAILY_ATTENDANCES_IN_BASIC].compact.min
+        [trial_length, daily_attendances_in_basic].compact.min
       end
     end
 
     private
 
-    # The first 2 daily attendances are included in the Basic Fee (BABAF)
-    DAILY_ATTENDANCES_IN_BASIC = 2
+    def daily_attendances_in_basic
+      claim.fee_scheme.eql?('fee_reform') ? 1 : 2
+    end
 
     def trial_length
       claim&.case_type&.requires_retrial_dates? ? claim&.retrial_actual_length : claim&.actual_trial_length
     end
 
+    def eligible_fee_type_unique_codes
+      claim.fee_scheme.eql?('fee_reform') ? 'BADAT' : %w[BADAF BADAH BADAJ]
+    end
+
     def daily_attendance_fee_types
-      ::Fee::BasicFeeType.where(unique_code: %w[BADAF BADAH BADAJ])
+      ::Fee::BasicFeeType.where(unique_code: eligible_fee_type_unique_codes)
     end
 
     def daily_attendance_uplifts
--- a/config/schemas/ccr_claim_schema.json
+++ b/config/schemas/ccr_claim_schema.json
@@ -241,7 +241,7 @@
         }
       },
       "type": "object",
-      "required": ["offence_class","unique_code"]
+      "required": ["unique_code"]
     },
     "advocate_category": {
       "id": "/properties/advocate_category",
--- a/config/schemas/ccr_claim_schema.json
+++ b/config/schemas/ccr_claim_schema.json
@@ -110,6 +110,10 @@
             "id": "/properties/bills/items/properties/amount",
             "type": "string"
           },
+          "warrant_issued_date": {
+            "id": "/properties/bills/items/properties/warrant_issued_date",
+            "type": ["string", "null"]
+          },
           "dates_attended": {
             "id": "/properties/bills/items/properties/dates_attended",
             "items": {
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -88,7 +88,7 @@ active_features:
 reject_refuse_messaging_released_at: <%= DateTime.new(2018, 4, 24, 23, 05, 0).change(offset: '+100') %>
 
 # agfs_fee_reform_release_date: <%= Date.new(2018, 4, 1) %>
-agfs_fee_reform_release_date: <%= ENV["ENV"].eql?('api-sandbox') ? Date.new(2018, 1, 1) : Date.new(2018, 4, 1) %>
+agfs_fee_reform_release_date: <%= ENV["ENV"].in?(['api-sandbox','demo','development']) ? Date.new(2018, 1, 1) : Date.new(2018, 04, 1) %>
 
 # number of weeks in one state before automatic transition to archived pending delete
 timed_transition_stale_weeks: 16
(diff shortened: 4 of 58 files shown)


features/step_definitions/advocate_claim_steps.rb
```diff
@@ -58,6 +61,17 @@
     end
 end
 
+When(/^I enter a scheme 10 defendant, representation order and MAAT reference$/) do
+    using_wait_time(6) do
+      @claim_form_page.wait_for_defendants
+      @claim_form_page.defendants.first.first_name.set "Bob"
+      @claim_form_page.defendants.first.last_name.set "Billiards"
+      @claim_form_page.defendants.first.dob.set_date "1955-01-01"
+      @claim_form_page.defendants.last.representation_orders.first.date.set_date "2018-01-01"
```
jsugarman: make this be the scheme 10 release date of the system

features/page_objects/sections/offence_result_section.rb
```diff
@@ -0,0 +1,4 @@
+class OffenceResultSection < SitePrism::Section
+  element :description, '.column-two-thirds'
```
jsugarman: maybe add a class to be more explicit for the selector

features/step_definitions/advocate_claim_steps.rb
jsugarman: also try to reuse and parametrise existing step definition for non-scheme-10

features/step_definitions/advocate_claim_steps.rb
```diff
@@ -78,6 +92,15 @@
   end
 end
 
+When(/^I search for the scheme 10 offence '(.*?)'$/) do |search_text|
+  @claim_form_page.offence_search.set search_text
+end
+
+Then(/^I click select and continue on the first option$/) do
```
jsugarman: `Then(/^I click select and continue on the first option$/)` -->
`Then(/^I click "select and continue" on the first search result$/)`
`Then(/^I pick the first search result$/)`

features/step_definitions/advocate_interim_claim_steps.rb
```diff
@@ -0,0 +1,11 @@
+Then(/^I should be on the advocate interim new claim page$/) do
+  expect(@advocate_interim_claim_form_page).to be_displayed
+end
+
+And(/^I fill '(.*)' as the warrant issued date$/) do |date|
```
jsugarman: fill in

app/interfaces/api/v2/ccr_claim.rb
```diff
@@ -5,7 +5,12 @@ class CCRClaim < Grape::API
 
       helpers do
         def claim
-          ::Claim::AdvocateClaim.find_by(uuid: params[:uuid]) || error!('Claim not found', 404)
+          ::Claim::BaseClaim.agfs.find_by(uuid: params[:uuid]) || error!('Claim not found', 404)
```
lostie, line 8: Related with this, I've been delaying [this PR](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2323) for a while but I think it would beneficial to get it out ASAP.

❓ If the `find_by(uuid: params[:uuid])` gives you a unique record, what's the need for the `agfs` scope?

config/schemas/ccr_claim_schema.json
```diff
@@ -110,6 +110,14 @@
             "id": "/properties/bills/items/properties/amount",
             "type": "string"
           },
+          "warrant_issued_date": {
```
lostie, line 113: Having this information flat out on the claim schema, won't it have the same problem we were discussing offline: you can technically claim a warrant fee per defendant (not currently officially supported) ?

config/settings.yml
```diff
@@ -88,7 +88,7 @@ active_features:
 reject_refuse_messaging_released_at: <%= DateTime.new(2018, 4, 24, 23, 05, 0).change(offset: '+100') %>
 
 # agfs_fee_reform_release_date: <%= Date.new(2018, 4, 1) %>
-agfs_fee_reform_release_date: <%= ENV["ENV"].eql?('api-sandbox') ? Date.new(2018, 1, 1) : Date.new(2018, 4, 1) %>
+agfs_fee_reform_release_date: <%= ENV["ENV"].in?(['api-sandbox','demo','development']) ? Date.new(2018, 1, 1) : Date.new(2018, 04, 1) %>
```
lostie: **[not a blocker/nit pick]** Rubocop 👮‍♀️: Spaces after commas, or just turn it into a `%w[]` array format instead.

features/step_definitions/advocate_claim_steps.rb
```diff
@@ -1,6 +1,9 @@
-
 include WaitForAjax
 
+Given(/^AGFS reform commenced on "([^"]*)"$/) do |agfs_reform_start_date|
+  allow(Settings).to receive(:agfs_fee_reform_release_date).and_return Date.parse(agfs_reform_start_date)
```
lostie, line 4: Not sure about this, features specs, IMO should not be mocking internal logic of the application.

lostie, line 4: Won't put this as a blocker, but I'm not quite sure about this, features specs, IMO should not be mocking internal logic of the application, but let the app to behave as it should naturally. I think this should probably be managed through configuration, not stubbing.

spec/factories/claim/base_claims.rb
```diff
@@ -1,5 +1,34 @@
...
 
+  trait :advocate_base_setup do
+    # NOTE: this was introduced because it was the only way to get FactoryBot to set
+    # model attributes on initialize (which seems not to be the default behaviour) and
+    # was causing the factory not to assign the appropriate attributes/associations on
+    # initialize which makes after_initialize logic not to behave as expected since the
+    # expected values are not yet set.
+    # More details can be found here:
+    # https://stackoverflow.com/questions/5916162/problem-with-factory-girl-association-and-after-initialize
+    initialize_with { new(attributes) }
```
lostie: Are we sure this is common setup for both advocate claims? The `initialize_with` was set for advocate final claims due to the way we're managing `basic_fees`.

spec/factories/claim/base_claims.rb
```diff
@@ -1,5 +1,34 @@
...
+    external_user
+    source { 'web' }
+    apply_vat false
+    providers_ref { random_providers_ref }
+    case_type
+    offence
+    advocate_category 'QC'
+    sequence(:cms_number) { |n| "CMS-#{Time.now.year}-#{rand(100..199)}-#{n}" }
+
+    after(:build) { |claim| post_build_actions_for_draft_claim(claim) }
```
lostie: Is this also common setup for both advocate claims? I'm personally of the approach of not adding any behaviour unless we're 💯 sure we need it. I don't understand the complexity of `post_build_actions_for_draft_claim` but seems like calling that for *advocate interim claims* after build might just end up creating data that is not necessary and makes things harder to change in the future.

app/interfaces/api/v2/ccr_claim.rb
jsugarman: may be its just "belt and braces" but it would prevent an LGFS claim query being sent to the wrong endpoint (ccr endpoint). Same/reverse logic exists for the cclf endpoint - in such instances it would return claim not found.

config/schemas/ccr_claim_schema.json
jsugarman: this is not on the claim directly but on a bill item, making it possible, potentially to have multiple per claim

config/settings.yml
jsugarman: will do

features/step_definitions/advocate_claim_steps.rb
jsugarman: simplest way too enable a feature test which we did not have before. without mocking we would not be able to test this feature. can you think of another way to simply get around the 3 month rule on warrant fees?

jsugarman: such as..

spec/factories/claim/base_claims.rb
jsugarman: mainly put here for neatness, but could put back into advocate claim

features/step_definitions/advocate_claim_steps.rb
lostie: 💭 Configuring the `agfs_fee_reform_release_date` for the cucumber environment to be at least 3 months old.

For reference: https://github.com/cucumber/cucumber/wiki/Mocking-and-Stubbing-with-Cucumber

spec/factories/claim/base_claims.rb
jsugarman: will look into this...

features/step_definitions/advocate_claim_steps.rb
jsugarman: Hmm, sounds like that would not be very declarative for a  given feature. that link suggests using stubs instead, which from reading on [RSpec mocks](https://relishapp.com/rspec/rspec-mocks/docs) the statement above is a stub. but `Settings.stub(:agfs_fee_reform_release_date).and_return(Date.parse(agfs_reform_start_date))` works just as well

app/services/ccr/daily_attendance_adapter.rb
```diff
@@ -24,23 +38,28 @@ def attendances_for(claim)
...
+        [trial_length, daily_attendances_in_basic].compact.min
       end
     end
 
     private
 
-    # The first 2 daily attendances are included in the Basic Fee (BABAF)
-    DAILY_ATTENDANCES_IN_BASIC = 2
+    def daily_attendances_in_basic
+      claim.fee_scheme.eql?('fee_reform') ? 1 : 2
```
lostie, line 50: Just FYI, going forward I'm thinking about deprecating the use of `fee_scheme.eql?('fee_reform')` and moving towards other approach of determining the fee scheme type.

spec/factories/claim/advocate_interim_claims.rb
```diff
@@ -1,12 +1,20 @@
...
+    after(:build) do |claim|
+      claim.creator = claim.external_user
+    end
 
     trait :submitted do
-      state :submitted
+      after(:create) { |c| c.submit! }
+    end
+
+    trait :without_fees do
```
lostie, line 14: **[not a blocker/remark]** This is one of the reasons I was suggesting not to use common Advocate factory setup as we're adding stuff to it by default that we then need to destroy. At least the advocate interim claim was "clean" enough that we could add things as required, not destroy stuff that was added by default that we don't actually need for a specific scenario.

lostie (approved):
Overall LGTM 👍

app/services/ccr/daily_attendance_adapter.rb
jsugarman: yeah, was not keen on using this myself. `claim.fee_scheme?(10)` or some such would be much clearer.

PR #2367, opened 2018-05-15
Add extra permitted parameters

colinbruce:
#### What
Fix the broken OffenceController
#### Why
In turning on strong parameters, I missed some that were only used downstream
#### How
Added the parameters that are required downstream by `search_offences.rb`
#### Future
See issue #2368 for reasons of not extending tests as part of this PR, updating Gems is well outside the timescale for fixing this!

### DIFF
--- a/app/controllers/offences_controller.rb
+++ b/app/controllers/offences_controller.rb
@@ -29,7 +29,7 @@ class OffencesController < ApplicationController
   private
 
   def permitted_params
-    params.permit(:fee_scheme, :description)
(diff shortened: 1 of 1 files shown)


lostie:
LGTM 👍 

I think the issue we had with this kind of illustrates that, given the controller has no exact information of the filters used by the offences search service, we should probably define a constant in the service that includes the available filters, which we then can reference in the controller to add to the strong params. That way if we are ever to change the search offences filters, the controller strong params will still be alined with it without requiring knowledge about or further development changes.

lostie (approved):

PR #2370, opened 2018-05-16
Warrant fee info in final claims

lostie:
#### What

- Adds a new section to the **final** claims (AGFS and LGFS) misc fees section to provide information about warrant fees previously claimed on interim claims related.
- Displays the new section in the summary pages for external users and case workers.

⚠️ **Requires migration**

#### Ticket

[AGFS and LGFS final fee - warrants](https://www.pivotaltracker.com/n/projects/1985371/stories/155688164)

#### Why

Case workers need this information to match it against the interim claims. Personally I still continue to think the actual solution would be to start linking claims, rather than adding extra (duplicate) information to the final claims.

### DIFF
--- a/app/assets/javascripts/modules/show-hide-content.js
+++ b/app/assets/javascripts/modules/show-hide-content.js
@@ -16,7 +16,7 @@
 
     // Escape name attribute for use in DOM selector
     function escapeElementName (str) {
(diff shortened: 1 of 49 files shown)


colinbruce (approved):
👍

colinbruce (approved):
:+1: Nice DRYing :)

PR #2371, opened 2018-05-18
CSS: incorrect nesting is causing the date field to collapse

cleargif:
A fix for the warrant fee dates that collapse.


cleargif:
@joelsamuelmoj @lostie

jsugarman (approved):
exactly what i did in my PR :). merge away imo

PR #2374, opened 2018-05-21
Upgrade paper_trail gem to most recent version

lostie:
#### What

The [issue](https://github.com/paper-trail-gem/paper_trail/issues/738) that was locking paper_trail to the version we were currently using was solved and its available since 6.X.

For all purposes, this also takes the opportunity to upgrade the version to the most recent one.

I've been through all the deprecation notes in the gem and it doesn't seems there's anything on our code impacted by it.

**NOTE:** This also solves the deprecation notice that was appearing when running **RSpec**:

```
DEPRECATION WARNING: before_filter is deprecated and will be removed in
Rails 5.1. Use before_action instead. (called from <top (required)> at
config/application.rb:8)
```

#### Why

Getting the application the most up to date as possible to upgrade it to Rails 5.1 with minimum impact.

#### TODO (wip)

 - [ ] A quick tests through submitting claims and the interfaces that allow us to see the changes to the models to check nothing unexpected is broken out of the gem upgrade.


colinbruce (commented):
💥 👍

colinbruce (approved):

PR #2377, opened 2018-05-23
Hide additional case numbers on fixed fees

jsugarman:
#### What
Hide additional case numbers on fixed fees

#### Ticket

[Hide additional case numbers](https://www.pivotaltracker.com/story/show/157771129)

#### Why
So user is not burdened with unnecessary fields on the page

#### How
existing enabling disabling of this field swapped out for show/hide


cleargif (approved):
👍

PR #2378, opened 2018-05-23
CBO-28 Mute "attach indictment" pop up for fixed fees

cleargif:
#### What
After breaking the pages up - the indictment  message on supporting evidence is appearing all the time. This should only be shown for fixed fees

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-28
[Remove 'remember to attach indictment'](https://www.pivotaltracker.com/story/show/157771232)

#### Why
Broken feature after pages was split up

#### How
Reimplement the `data-attr` on the supporting evidence page

### DIFF
--- a/app/views/external_users/claims/fixed_fees/_fields.html.haml
+++ b/app/views/external_users/claims/fixed_fees/_fields.html.haml
@@ -1,5 +1,5 @@
 - if @claim.allows_fixed_fees?
-  #fixed-fees.mod-fees{"data-mute-indictment": @claim.allows_fixed_fees? ? "true" : "false"}
+  #fixed-fees.mod-fees
     .fixed-fee-group-wrapper
       = f.fields_for :fixed_fees do |fixed_fee_fields|
         - @fixed_fee_count += 1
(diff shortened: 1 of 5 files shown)


app/views/external_users/claims/fixed_fees/_fields.html.haml
```diff
@@ -1,5 +1,5 @@
 - if @claim.allows_fixed_fees?
-  #fixed-fees.mod-fees{"data-mute-indictment": @claim.allows_fixed_fees? ? "true" : "false"}
+  #fixed-fees.mod-fees.fx-numberedList-wrapper
```
lostie: `fx-numberedList-wrapper` is this part of this PR?

cleargif: It is not.. :)

lostie:
💭 I suggest in this case that you _squash_ your commits so that we don't have one just saying 'removed class' as it's not a new change but something that should just be part of the previous commit ;)

cleargif:
@lostie I squashed this commit and did a rebase.

lostie (approved):
👍

PR #2380, opened 2018-05-24
Tweak error messages for net amounts

jsugarman:
#### What
Change error message text for Fixed and Miscellaneous fee rate

#### Why
The name/label of these fields is net amount but attribute is rate.

#### before:

screen shot 2018-05-24 at 10 22 44

#### after:

screen shot 2018-05-24 at 10 23 50


colinbruce (approved):
LGTM

PR #2381, opened 2018-05-25
Patch autosave associations to support conditional validations

lostie:
#### What

For more details please read NOTE in [config/initializers/01_autosave_association_patch.rb](./config/initializers/01_autosave_association_patch.rb)

Claim associations are autovalidated by default as long as they're defined. With this change, they're only validated if they're required for a specific step/stage or if they're being submitted through the API.

#### Ticket

[Check your claim page fixes](https://dsdmoj.atlassian.net/browse/CBO-137)

#### Why

Rails models that accept nested attributes for some of its associations [autosave and validate those associations by default](http://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html).

What this means in effect is that, even though the application has its own stage/step validation setup for the claims model, as long as an association was previously set in another stage, when submitting data for another stage, that association will still be autosaved and validated. (This is noted more so when one of the stages is submitted as a **draft** with invalid data for one of the associations and other stage is then submitted with valid data but fails due to the previously submitted association being invalid, even though we're not at the step of validating that association).

#### How

- The claim model no longer has a default `form_step`, as that information is specifically related with the claim submission process only, and the `form_step` is now set only during that flow.
- Autosave associations in Rails was _patched_ to evaluate the value set for `validate: <value>` as by default it just checks if is _true_ or not. This way it allows us to control when these associations are actually validated, letting the validations classes behave as expected during the submission flow and only validate the data related with that particular _stage_. (**NOTE:** Obviously this patching will **potentially** have an impact on any future Rails updates)

Also, worth of note:

- Some of the fee factories, now detect if the associated **fee type** already exists instead of creating a brand new one. This was done mainly to avoid flakey tests and unexpected behaviours and should also helps making use of the seed data more reliably going forward.
- There's a known [issue](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2382) with the factories definition that needs to be addressed. The resolution is **not** part of this PR as this would take quite a lot of refactoring. In this PR, the previous behaviour (which was to create a claim with a default _form_step_ was kept).

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -17,7 +17,7 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
 
   before_action :set_and_authorize_claim, only: %i[show edit update unarchive clone_rejected destroy summary
                                                    confirmation show_message_controls messages disc_evidence]
-  before_action :set_form_step, only: %i[edit]
+  before_action :set_form_step, only: %i[edit update]
   before_action :redirect_unless_editable, only: %i[edit update]
   before_action :generate_form_id, only: %i[new edit]
   before_action :initialize_submodel_counts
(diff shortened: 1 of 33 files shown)


colinbruce (approved):
:+1: LGTM

PR #2383, opened 2018-05-29
Add FactoryBot link rake task

lostie:
#### What

To be able to detect factories that are not valid and should be fixed.

Heavely based on https://semaphoreci.com/community/tutorials/working-effectively-with-data-factories-using-factorybot

#### Why

To helps detect what sort of factories are currently defined that do not match to a valid record, so we can fix them.


colinbruce (approved):
:+1: Cool

PR #2384, opened 2018-05-29
Zero cost redeterminations

colinbruce:
#### What
I want to be able to see the previously authorised amount on a claim, if my claim then goes on to be refused at the redetermination stage
#### Ticket
[Add feature test for non-display of empty refusal assessment](https://dsdmoj.atlassian.net/browse/CBO-27)
#### Why
So that I can still see what I am paid.
#### How
Added feature and unit tests for updating the StateMachine so that it doesn't `zeroise` assessment values when refusing a claim after an assessment has previously been made

### DIFF
--- a/features/refuse_claim.feature
+++ b/features/refuse_claim.feature
@@ -23,3 +23,26 @@ Feature: Case worker rejects a claim, providing a reason
     Then the claim I've just updated is no longer in the list
 
     And I eject the VCR cassette
+
+  Scenario: I refuse a claim after a redetermination request providing a reason
+
+    Given a "case worker" user account exists
+    And an "advocate" user account exists
+    And there is a redetermination claim allocated to the case worker with case number 'A20161234'
+
+    And I insert the VCR cassette 'features/case_workers/claims/reject'
+
+    When I am signed in as the case worker
+    And I select the claim
+    And I click the refused radio button
+    And I select the first refusal reason
+    And I click update
+    Then the status at top of page should be Refused
+    Then the messages should not contain 'Total (inc VAT): £0.00'
+    Then the third message contains 'Claim refused'
+    Then the last message contains 'Your claim has been refused'
+
+    When I click your claims
+    Then the claim I've just updated is no longer in the list
+
+    And I eject the VCR cassette
--- a/spec/models/claims/state_machine_spec.rb
+++ b/spec/models/claims/state_machine_spec.rb
@@ -401,22 +401,33 @@ RSpec.describe Claims::StateMachine, type: :model do
 
   describe 'refuse!' do
     let(:reason_codes) { ['wrong_ia'] }
-    let(:last_state_transition) { claim.last_state_transition }
 
     context 'when refused on first assessment' do
-      before { claim.submit!; claim.allocate!; claim.refuse!(reason_code: reason_codes) }
+
+      before do
+        subject.submit!
+        subject.allocate!
+      end
 
       context 'claim state transitions (audit trail)' do
         it 'updates #to' do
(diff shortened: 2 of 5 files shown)


spec/models/claims/state_machine_spec.rb
```diff
@@ -400,17 +400,40 @@
...
+        claim.allocate!
+        claim.assessment.update(fees: 666.00, expenses: 23.45)
+        claim.authorise_part!
+        claim.redetermine!
+        claim.allocate!
+        claim.refuse!(reason_code: reason_codes)
+      end
+
+      it 'does not set the assessment to zero' do
+        expect(claim.assessment.total).to be_positive
```
lostie: **[not a blocker]** 💭 Consider using the RSpec `change` syntax to more accurately express the expectation:

```ruby
# keeping the before block as is with the exception of the last line.
it 'does not set the assessment to zero' do
  expect { claim.refuse!(reason_code: reason_codes) }
    .not_to change { claim.assessment.total }
    .from(<the_total_value>)
end
```

spec/models/claims/state_machine_spec.rb
```diff
@@ -400,17 +400,40 @@
...
   describe 'refuse!' do
-    before { claim.submit!; claim.allocate!; claim.refuse!(reason_code: reason_codes) }
     let(:reason_codes) { ['wrong_ia'] }
     let(:last_state_transition) { claim.last_state_transition }
 
-    context 'claim state transitions (audit trail)' do
-      it 'updates #to' do
-        expect(last_state_transition.to).to eq('refused')
+    context 'when refused on first assessment' do
+      before { claim.submit!; claim.allocate!; claim.refuse!(reason_code: reason_codes) }
```
lostie: **[not a blocker/nit pick]** Avoid multiple inline statements with `;`, specs are supposed to be expressive anyway, so a couple more lines won't matter really ;). Also consider following a similar approach to [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2384/files#r191457874) whereas in this case using the positive expectation:

```ruby
expect { claim.refuse!(reason_code: reason_codes) }.to change { claim.assessment.total }.to(0)
```

features/refuse_claim.feature
```diff
@@ -23,3 +23,26 @@ Feature: Case worker rejects a claim, providing a reason
...
+
+    And I insert the VCR cassette 'features/case_workers/claims/reject'
+
+    When I am signed in as the case worker
+    And I select the claim
+    And I click the refused radio button
+    And I select the first refusal reason
+    And I click update
+    Then the status at top of page should be Refused
+    Then the messages should not contain 'Total (inc VAT): £0.00'
```
lostie, line 41: Should this be asserting the actual value, rather checking that it's not zero?

The negative check tells us the total amount assessed is not zero, but does not guarantee that it preserves whatever the total amount originally assessed.

**NOTE:** I  guess the only thing that might be a bit trickier for that to work is the fact that the claim is created with some sort of pre-set values that we do not have control in this scenario, so we don't really know what value to expect?

lostie:
LGTM 👍 , minor suggestions, no real blockers though.

features/refuse_claim.feature
colinbruce: This is checking that a message is _not_ displayed.  Previously there would have been the first message stating `Total (inc VAT): £123.00` followed by a message that the claim had been assessed with a total of zero.

This was to ensure that the second message is no-longer displayed

lostie:
🏆  => :shipit:

lostie (approved):

PR #2385, opened 2018-05-30
Quantity for a fee needs to be greater or equal than 1

lostie:
#### What

Validations for fee quantities were allowing `0` to be a valid quantity. If the fee exists and is being submitted there's not real reason for the quantity to be `0` (exception are the basic fees which you can find the reason why attached to its own validation changes).

#### Ticket

[Able to create a misc fee without quantity/amount](https://dsdmoj.atlassian.net/browse/CBO-21)


colinbruce (approved):
👍

PR #2386, opened 2018-05-30
HOTFIX: Already submitted claim check should be done on the claim itself

lostie:
Was previously checking for other claims with the same form_id that have not been submitted. It happens production data has quite a lot of claims (mainly API submissions and JSON imports) that do not have a form_id set and have been submitted already, so was causing currently in progress claims from being submitted if they didn't have a form_id set.


colinbruce (approved):
keanu

PR #2387, opened 2018-05-31
Disable NewRelic monitor for non-prod environments

lostie:
Now that production officially started reporting to NewRelic the non-prod environments are not required to report anymore (were mainly for testing reasons) and it does help keep prices cost to a low.


colinbruce (approved):
:+1:

PR #2388, opened 2018-05-31
Add missing index for defendant_id foreign key on rep orders

lostie:
#### Why

Has been noticed that there's tables with missing indexes on foreign keys or columns that are heavily used for querying/filtering data.

This index should help reduce the amount of sequence scans when using the representation orders table to perform joins with other tables.

**Example of query for the archived claims:**

screen shot 2018-05-31 at 12 17 32


colinbruce (approved):
👍

PR #2392, opened 2018-06-04
Add validations to representation order date

lostie:
#### What

So far most of the validations related with representation order dates were being performed on the trial dates, which means the defendants page that holds the information for the representation order dates was not doing its own checks.

Also:
- Remove indirection on methods that can be delegated directly on the claim object such as `trial_length` and `requires_trial_dates?`

#### Ticket

[LGFS and AGFS final fee: rep oder validation (date of rep order cannot be later than date case concluded)](https://dsdmoj.atlassian.net/browse/CBO-187)

### DIFF
--- a/spec/models/claim/base_claim_spec.rb
+++ b/spec/models/claim/base_claim_spec.rb
@@ -534,6 +534,40 @@ RSpec.describe Claim::BaseClaim do
       end
     end
   end
+
+  describe '#trial_length' do
+    let(:actual_trial_length) { 3 }
+    let(:retrial_actual_length) { 5 }
+    let(:claim) { MockSteppableClaim.new(actual_trial_length: actual_trial_length, retrial_actual_length: retrial_actual_length) }
+
+    subject(:trial_length) { claim.trial_length }
+
+    context 'when the claim requires re-trial dates' do
+      before do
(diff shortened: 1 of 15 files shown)


spec/models/claim/base_claim_spec.rb
```diff
@@ -534,6 +534,40 @@ class MockSteppableClaim < Claim::BaseClaim
...
+    let(:claim) { MockSteppableClaim.new(actual_trial_length: actual_trial_length, retrial_actual_length: retrial_actual_length) }
+
+    subject(:trial_length) { claim.trial_length }
+
+    context 'when the claim requires re-trial dates' do
+      before do
+        expect(claim).to receive(:requires_retrial_dates?).and_return(true)
+      end
+
+      specify { is_expected.to eq(retrial_actual_length) }
```
jsugarman, line 550: any particular reason for using `specify` rather than `it`. since the specs generally use it and `it { is_expected.to }` makes for clear english

lostie: In regards to this, I'm happy to create a separate PR to address it, even though they're formally equivalent 👍 .

jsugarman (approved):
👍

PR #2393, opened 2018-06-05
AGFS Warrants filter

jsugarman:
#### What
Add an allocation queue filter for AGFS warrant claims

#### Ticket

[AGFS warrants in allocation queue](https://dsdmoj.atlassian.net/browse/CBO-16)

#### Why
As a caseworker I want to be able to filter by warrants in the AGFS allocation queue
so that I can allocate them to the appropriate caseworker

#### How

- amend query used by search tool to add advocate interim claims
- expose warrants filter value in search results
- add filter to agfs "case type" search options

### DIFF
--- a/app/controllers/case_workers/admin/allocations_controller.rb
+++ b/app/controllers/case_workers/admin/allocations_controller.rb
@@ -6,6 +6,7 @@ class CaseWorkers::Admin::AllocationsController < CaseWorkers::Admin::Applicatio
   before_action :set_summary_values, only: [:new], if: :summary_from_previous_request?
   before_action :process_claim_ids, only: [:create], if: :quantity_allocation?
 
+  # FIXME: is this even used....the partial it is called from does not appear to be used!!
   helper_method :allocation_filters_for_scheme
 
   def new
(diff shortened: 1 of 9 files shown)


app/controllers/case_workers/admin/allocations_controller.rb
```diff
@@ -6,6 +6,7 @@ class CaseWorkers::Admin::AllocationsController < CaseWorkers::Admin::Applicatio
   before_action :set_summary_values, only: [:new], if: :summary_from_previous_request?
   before_action :process_claim_ids, only: [:create], if: :quantity_allocation?
 
+  # FIXME: is this even used....the partial it is called from does not appear to be used!!
```
colinbruce, line 9: Maybe not... check with @cleargif it may have been a legacy helper from the old allocation page... or still used on re-allocation

app/controllers/case_workers/admin/allocations_controller.rb
```diff
@@ -116,6 +117,7 @@ def is_allocating?
     params[:commit] == 'Allocate'
   end
 
+  # FIXME: is this even used....the partial it is called from does not appear to be used!!
```
colinbruce, line 120: ditto

colinbruce (approved):
Generally 👍 but we can ask @cleargif to check the fixme

PR #2394, opened 2018-06-06
Add index to deleted_at column on the claims table

lostie:
#### What

There's quite a lot of scopes and queries relying on the filter "deleted_at IS NULL" which without the index forces the table to be scanned every single time.

This index should optimise all those queries.


colinbruce (approved):
Out of hours deploy - :+1:

colinbruce:
Merged into #2578

PR #2400, opened 2018-06-07
Display only non-zero basic fees in the summary page

lostie:
#### What

AGFS basic fees are pre-initialised with zeros (needs to be addressed atsome point) so any fee that has not been configured by the advocate should not be display if it remains a zero fee.

#### Ticket

[If there is a '0' element in the basic fees, it should not be shown on the CYC page](https://dsdmoj.atlassian.net/browse/CBO-191)


colinbruce (approved):
Looks fine... 👍 
Should it have tests?</not_a_blocker>

PR #2402, opened 2018-06-08
Add claim_id index to determinations table

lostie:
#### What

Self-descriptive.

#### Why

Historically, some tables were created without these indexes that should be set by default on foreign key columns.


colinbruce (approved):
:+1: OoH deploy

jsugarman (changes requested):
There are a lot of determinations so i would apply the concurrent algorithm and disable the ddl transaction on this as covered here [postgres concurrent indexes](https://robots.thoughtbot.com/how-to-create-postgres-indexes-concurrently-in)

colinbruce:
Merged into #2578

PR #2403, opened 2018-06-08
Fix some N+1 queries in the external users claims page

lostie:
#### What

There's quite a lot of associations that were not being preloaded in the page that displays the list of claims.

#### Why

N+1 queries in cases like this (list of objects) are *bad* and should be avoided.


colinbruce (approved):
👍

PR #2404, opened 2018-06-12
Manage provider page redesign

lostie:
#### What

* Re-style the the add/edit provider page for super admins/admin external users
* Add two new fields to a supplier (name and postcode)
* Resets values on the provider if the provider type is changed

#### Ticket

[Implement new 'manage provider' designs](https://dsdmoj.atlassian.net/browse/CBO-200)


app/assets/stylesheets/v2/_links.scss
```diff
@@ -0,0 +1,46 @@
...
+    color: #2b8cc4;
+  }
+  @media (max-width: 640px) {
+    font-size: 16px;
+    line-height: 1.25;
+  }
+}
+
+a[target="_blank"]:after
+/*a.link-open-window:after
```
colinbruce, line 40: can the commented bit be removed?  herokuapp.com seems superflouous

config/locales/en.yml
```diff
@@ -147,24 +157,17 @@ en:
         supplier_number:
           attributes:
             supplier_number:
-              invalid_format: "LGFS supplier number invalid format"
-              not_unique: "LGFS supplier number has already been taken"
+              invalid: invalid format
```
colinbruce, line 160: Is this added to a description string?

colinbruce (approved):
A couple of [not_a_blocker] comments but otherwise LGTM
:+1:

config/locales/en.yml
lostie: Yes, the old behaviour overrode the default Rails behaviour for error messages. This one just sets the locale for the default error code when the `format` is invalid.

PR #2405, opened 2018-06-13
Working un-archiving

colinbruce:
#### What
Providers are currently unable to un-archive claims that were archived using the previous version of the paper-trail gem
#### Ticket
[Un-archiving records raises an error](https://dsdmoj.atlassian.net/browse/CBO-194)
#### Why
Paper trail changed the method of serialising data in an interim version of the gem
#### How
Implement a new class that attempts to load the previous_version of the claim, if it fails, it calls a new block of code that modifies the `Version` record in the database and then return the updated version
The alternative was to loop through the 1.2 million `Version` records in the database and update them, this seems to be more expedient!

### DIFF
--- a/spec/lib/previous_version_of_claim_spec.rb
+++ b/spec/lib/previous_version_of_claim_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 # require File.join(Rails.root, 'lib', 'cleanse_paper_trail_arrays')
 
-describe PreviousVersionOfClaim do
+RSpec.describe PreviousVersionOfClaim do
   subject(:previous_version) { described_class.new(claim) }
 
   let(:claim) { create(:archived_pending_delete_claim, evidence_checklist_ids: [3, 4, 1]) }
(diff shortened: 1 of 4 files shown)


spec/lib/previous_version_of_claim_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+# require File.join(Rails.root, 'lib', 'cleanse_paper_trail_arrays')
```
lostie: ✂️

lostie (approved):
LGTM 👍 , as spoken offline, might be worth leaving a note/issue on further actions to be taken once the current data expires.

spec/lib/previous_version_of_claim_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+describe PreviousVersionOfClaim do
```
jsugarman: `RSpec.describe` - just a best practice that prevents overide

jsugarman (approved):
LGTM 👍

PR #2409, opened 2018-06-15
Display postcode associated with LGFS supplier numbers

lostie:
#### What

In the case details page for claim submissions.

#### Ticket

Outstanding change related with:

[Implement new 'manage provider' designs](https://dsdmoj.atlassian.net/browse/CBO-200)

### DIFF
--- a/app/presenters/supplier_number_presenter.rb
+++ b/app/presenters/supplier_number_presenter.rb
@@ -1,0 +1,6 @@
+class SupplierNumberPresenter < BasePresenter
+  def supplier_label
+    return supplier_number unless postcode.present?
+    "#{supplier_number} (#{postcode})"
+  end
+end
--- a/spec/factories/supplier_numbers.rb
+++ b/spec/factories/supplier_numbers.rb
@@ -10,6 +10,7 @@
 FactoryBot.define do
   factory :supplier_number do
     provider
+    postcode { Faker::Address.postcode }
     supplier_number { [rand(0..9), ('A'..'Z').to_a[rand(0..25)], rand(100..999), ('A'..'Z').to_a[rand(0..25)]].join }
   end
 end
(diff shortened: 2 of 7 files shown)


spec/factories/supplier_numbers.rb
```diff
@@ -10,6 +10,8 @@
 FactoryBot.define do
   factory :supplier_number do
     provider
+    # hacky way to get the locale to use the appropriate region setting for postcodes (for :en will be the UK postcodes)
```
lostie: 😰 Even though this only affects the factories, it exposed something that we have yet to come accross, which is having the application configured to support region locales. By default `:en` is associated with the US region, where as in our application we want that to be UK. Setting the default locale to `:'en-GB'` brings some extra [complications](http://guides.rubyonrails.org/i18n.html)* that need to be addressed (e.g to support `:'en-GB'`, the current locale files would have to be moved/duplicated or possibly introduce the [Globalize gem](https://github.com/globalize/globalize) to deal with the region differences)

*\**
From the Rails documentation:
> The i18n library takes a pragmatic approach to locale keys (after some discussion), including only the locale ("language") part, like :en, :pl, not the region part, like :en-US or :en-GB, which are traditionally used for separating "languages" and "regional setting" or "dialects". Many international applications use only the "language" element of a locale such as :cs, :th or :es (for Czech, Thai and Spanish). However, there are also regional differences within different language groups that may be important. For instance, in the :en-US locale you would have $ as a currency symbol, while in :en-GB, you would have £. Nothing stops you from separating regional and other settings in this way: you just have to provide full "English - United Kingdom" locale in a :en-GB dictionary.

colinbruce (approved):
:+1:

app/presenters/supplier_number_presenter.rb
```diff
@@ -0,0 +1,7 @@
+class SupplierNumberPresenter < BasePresenter
+  presents :supplier_number
+
+  def supplier_label
+    "#{supplier_number} (#{postcode})"
```
colinbruce: Could this be...
 *psuedo-code*
```ruby
"#{supplier_number} (#{postcode ||= 'no postcode set'})"
```

Then users wouldn't be confused by empty brackets and would have a hint on how to complete the data?

PR #2413, opened 2018-06-21
CBO-202: Travel Automation view scaffold

cleargif:
Transferred the prototype view into HAML and did some partials for the `select` boxes.

#### TODO (wip)

 - [ ] Reverse engineer the JS into a moj Module


cleargif:
See: #2424

PR #2414, opened 2018-06-21
Fix 500 error when validating graduated fees with no case type set

lostie:
#### What

https://sentry.service.dsd.io/mojds/private-beta/issues/32535/

#### Ticket

[500 error on CYC page with no case type set](https://dsdmoj.atlassian.net/browse/CBO-224)

#### Why

User's sometimes fill in all the information til the graduated fees, then go back to the case details and remove the case type and save as draft. When they come back to the Check your claim page, they get a 500 error as the claim no longer has a case type that is required to validated the graduated fees.


colinbruce (approved):
:+1:

PR #2415, opened 2018-06-25
Amend special prep form link form link for litigators

kmahern:
Amend special prep form link on Miscellaneous fees screen.

#### What

Amend special prep link on miscellaneous fees screen for litigators

#### Ticket

CBO-205

#### Why

Link was pointing to AGFS form

#### How

Amended link in config file

--------


colinbruce (approved):
:+1:

PR #2418, opened 2018-06-28
Ping issues

kmahern:
#### What
Amend ping and healthcheck

#### Ticket

CBO-229

#### Why

Currently ping is using a large proportion of system resources.

#### How

Move the num_claims json key from ping to healthcheck (and amend load_test_runner.rb to point at this).
--------


colinbruce (approved):
Awesome 👍

PR #2419, opened 2018-07-02
Case workers travel expenses section order

lostie:
#### What

Claim's view for case workers displays the expenses section with order activated so that the case worker can re-order the expenses to better assess them.

#### Ticket

[Travel expense sorting and ordering](https://dsdmoj.atlassian.net/browse/CBO-221)


colinbruce (approved):
:+1:

cleargif (approved):
👍

PR #2420, opened 2018-07-09
CBO-243 create mi model

colinbruce:
#### What
Create MI model
#### Ticket
[CBO-243](https://dsdmoj.atlassian.net/browse/CBO-243)
#### Why
There is an increased demand for larger MI reports, this creates an MI data model that can, initially, be exported to excel and allow users to run their own filters

#### TODO (wip)

 - [X] Create MI model
 - [X] Run MI import on destroy of archived claim
 - [ ] Index the MI Model once we know the filter columns
   - [x] Time frame of created date – to and from
   - [ ] Bill type – i.e grad fee etc
   - [x] Offence type – a, b etc etc
   - [x] Trial type – GP, TRIAL ETC
   - [x] Number of trial days if applicable
   - [x] PPE ranges

### DIFF
--- a/app/models/stats/mi_data.rb
+++ b/app/models/stats/mi_data.rb
@@ -1,0 +1,61 @@
+module Stats
+  class MIData < ApplicationRecord
+    self.table_name = 'mi_data'
+
+    DUPLICATE = %w[disk_evidence retrial_reduction case_concluded_at effective_pcmh_date first_day_of_trial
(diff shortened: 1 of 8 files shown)


lostie (approved):
👍 

As discussed offline, might be worth evaluating in the future if these sort of requirements increase to consider an integration with a data warehouse solution.

PR #2421, opened 2018-07-09
Re-add gross amount to the travel expenses view

lostie:
#### Ticket

[Add gross amount column to travel expenses table](https://dsdmoj.atlassian.net/browse/CBO-249)


cleargif (approved):
👍

PR #2422, opened 2018-07-10
Safeguard for court name not be filled in disk evidence page

lostie:
#### What

If the user saves some of the pages as draft and goes to the disk evidence page and tries to submit it, the PDF generation was exploding as some of the information that is pre-filled is still not set.

#### Ticket

[Disk evidence PDF generation failing](https://dsdmoj.atlassian.net/browse/CBO-252)


colinbruce (approved):
:+1:

PR #2425, opened 2018-07-12
Remove fee reform text

colinbruce:
#### What
Remove the legacy `FeeScheme.for_claim` method of returning string values
#### Ticket
[FeeScheme claim refactoring](https://dsdmoj.atlassian.net/browse/CBO-260)
#### Why
It was unsustainable going forward
#### Next steps?
- add methods to the FeeScheme model that id the exact version
   this would simplify calls allow `.ags_scheme_ten?` checks rather than `.verison.eql?(10)` upcoming changes to the fee schemes may necessitate scheme 11,12,13,etc

### DIFF
--- a/app/models/fee_scheme.rb
+++ b/app/models/fee_scheme.rb
@@ -17,6 +17,10 @@ class FeeScheme < ApplicationRecord
     where('(:date BETWEEN start_date AND end_date) OR (start_date < :date AND end_date IS NULL)', date: check_date)
   }
 
+  def scheme_10?
+    version.eql?(10)
+  end
+
   def self.current_agfs
     agfs.current.order(end_date: :desc).first
   end
--- a/app/models/fee_scheme.rb
+++ b/app/models/fee_scheme.rb
@@ -27,9 +27,8 @@ class FeeScheme < ApplicationRecord
 
   def self.for_claim(claim)
     # TODO: Align this with Fee reform SPIKE
-    return 'default' if claim.lgfs? || !FeatureFlag.active?(:agfs_fee_reform)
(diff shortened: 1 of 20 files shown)


app/models/fee_scheme.rb
```diff
@@ -27,9 +27,11 @@ def self.current_lgfs
 
   def self.for_claim(claim)
     # TODO: Align this with Fee reform SPIKE
-    return 'default' if claim.lgfs? || !FeatureFlag.active?(:agfs_fee_reform)
     date = claim.earliest_representation_order&.representation_order_date
-    return unless date.present?
-    date >= Date.parse(Settings.agfs_fee_reform_release_date.to_s) ? 'fee_reform' : 'default'
+    if date.present?
+      FeeScheme.for(date).find_by(name: claim.agfs? ? 'AGFS' : 'LGFS')
```
lostie: **[nit pick/not a blocker]** 💭 For DRYness and better readability , consider having the `name` defined before the  if statement so that it can be reusable.

lostie (approved):
LGTM 👍 

As spoken offline, a nice to have would be to cleanup the indirection around `claim.fee_scheme.version...`.

Also, another minor thing, there's a bit of repetition in the specs around making sure the fee scheme records exists in the database before the tests execute which raises again the point that maybe they should just be part of the seeds and the specs would be using them (maybe not a problem to tackle right now, just raising awareness on it 👨‍👧‍👧 )

lostie:
🎩 🙇

lostie (approved):
👍 :shipit:

PR #2426, opened 2018-07-12
Fixes bug - validate dates attended without earliest rep order

lostie:
#### What

https://sentry.service.dsd.io/mojds/private-beta/issues/32841/

**NOTE:** This assumes that the representation order is mandatory part of filling a defendant, so the claim won't be valid for submission until that is set.

#### Ticket

[Claim show page 500 when dates attended validation fails](https://dsdmoj.atlassian.net/browse/CBO-264)


jsugarman (approved):
👍 I assume a guard condition would make rubocop complain about length of line

PR #2427, opened 2018-07-16
Laa fee calculator client integration

jsugarman:
#### What
AGFS scheme 9 Fixed fee calculator integration

#### Ticket
A solution for integrating the fee calculator into the fixed fee pages

[CBO-254](https://dsdmoj.atlassian.net/browse/CBO-254)
design ticket [CBO-306](https://dsdmoj.atlassian.net/browse/CBO-306)

#### Why
So that users do not have to manually calculate fixed fees

#### How
First attempt is to calculate the unit prices for each of the fixed fees on the page
and dynamically update them (using JS/Ajax) when changes mean one or more 
will be impacted. The idea is to, at this time, avoid the need for major changes to 
the backend and frontend (sidebar), however, long term it may be that such changes
are required for simplicity (both code wise and for user benefit).

--------

#### TODO (wip)

- [X] create structure for fee calculator service objects
- [X] add unit price service
- [ ] add ajax/js for immediate update of unit price based on "effectors" - ON GOING
- [X] ajax error handling
- [x] uplift fee calculations 
- [x] multiple fee handling
- [x] “Number of cases uplift” handling
- [x] “Number of defendant uplift” handling
- [ ] “Elected case not proceeded” handling
- [ ] “Cracked case discontinued” handling?? - by erroring - DONE??
- [ ] Implement designs from [CBO-294](https://dsdmoj.atlassian.net/browse/CBO-294) and [CBO-306](https://dsdmoj.atlassian.net/browse/CBO-306)
- [X] unit test fee calculator controller testing
- [x] unit test fee calculator service testing
- [ ] unit test fee calculator ajax (jasmine)??
- [ ] handling api web edited claims with fixed fee values
- [x] feature test fixed fee calculations
- [ ] refactoring of backend services
- [ ] refactoring JS (david plews to review/pair-on)


Gemfile
```diff
@@ -61,6 +61,8 @@ gem 'shell-spinner', '~> 1.0', '>= 1.0.4'
 gem 'ruby-progressbar'
 gem 'geckoboard-ruby'
 gem 'posix-spawn', '~> 0.3.13'
+# gem 'laa-fee-calculator-client', :git => 'git@github.com:ministryofjustice/laa-fee-calculator-client.git', :branch => 'master'
```
lostie: **[not a blocker]** ✂️since we're now using the line below that used tagged versions.

app/assets/javascripts/modules/external_users/claims/FeeCalculator.js
```diff
@@ -0,0 +1,149 @@
...
+    var rate = $(context).closest('.fixed-fee-group').find('input.fee-rate').val();
+    var quantity = $(context).closest('.fixed-fee-group').find('input.fee-quantity').val();
+    var value = (rate * quantity);
+    var text = '&pound;' + moj.Helpers.SideBar.addCommas(value.toFixed(2));
+    $el.html(text);
+  },
+
+  displayError: function(response, context) {
+    // only some errors will have a JSON response
+    //TODO:remove the console.log debug
```
lostie, line 78: ☝️ ✂️ ?

app/models/claim/advocate_claim.rb
```diff
@@ -72,7 +72,7 @@ class AdvocateClaim < BaseClaim
              dependent: :destroy,
              inverse_of: :claim,
              validate: proc { |claim| claim.step_validation_required?(:basic_fees) }
-    has_many :fixed_fees,
+    has_many :fixed_fees, -> { order(:created_at) },
```
lostie, line 75: ❓ Changing the default order of the relation might affect how some of the pages during the claim submission process, or the claim summary page display the fees. Did we verify there was no impact there?

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -0,0 +1,161 @@
+require 'ostruct'
```
lostie: Is this `require` needed given we seem to be using the `Struct` [class](https://ruby-doc.org/core-2.5.0/Struct.html) instead of the [OpenStruct one](https://ruby-doc.org/stdlib-2.5.1/libdoc/ostruct/rdoc/OpenStruct.html) ?

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -0,0 +1,161 @@
...
+      end
+
+      private
+
+      def setup(options)
+        @fee_type = Fee::BaseFeeType.find(options[:fee_type_id])
+        @advocate_category = options[:advocate_category] || claim.advocate_category
+        @quantity = options[:quantity] || 1
+        @current_page_fees = options[:fees].values
+      rescue StandardError
```
lostie, line 41: ❓ What's the intent of this? We're catching all exceptions that happen here and raising a very generic error message, loosing the original exception details in the process.

More so, the `call` method is also catching all `StandardError` exceptions (again), which means, we're just rescue/catch exceptions multiple times.

Looking at the code it seems there's no obvious reason to have the rescue over here (but I lack context, so, not sure).

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -0,0 +1,161 @@
...
+          # CCR::Fee::BasicFeeAdapter::BASIC_FEE_BILL_MAPPINGS, # TODO: all are AGFS_FEE
+          CCR::Fee::FixedFeeAdapter::FIXED_FEE_BILL_MAPPINGS,
+          CCR::Fee::MiscFeeAdapter::MISC_FEE_BILL_MAPPINGS
+        ].inject(&:merge)
+      end
+
+      def fee_type_code_for(fee_type)
+        fee_type = case_uplift_parent if fee_type.case_uplift?
+        fee_type = defendant_uplift_parent if fee_type.defendant_uplift?
+        fee_type_mappings[fee_type&.unique_code&.to_sym][:bill_subtype]
```
lostie, line 113: ❓ Is this likely to 💥 if `fee_type_mappings[fee_type&.unique_code&.to_sym]` returns nil?

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -0,0 +1,161 @@
...
+      end
+
+      # TODO: share/take from adapters
+      def orphan_uplift?
+        %w[FXNOC FXNDR].include?(fee_type.unique_code)
```
lostie, line 118: **[not a blocker]** Feels like we should probably have these sort of codes defined somewhere else (FixedFeeType model maybe), which is actually responsible for knowing this information. Having them here has the potentially for us to replicate this behaviour (`orphan_uplift?`) in other classes.

app/services/claims/fee_calculator/unit_price.rb
```diff
@@ -0,0 +1,85 @@
...
+  module FeeCalculator
+    # TODO: this is using:
+    # 1. for primary/main fixed fees it use the calculate endpoint with quantity/days of 1 and no modifiers
+    # 2. for case uplifts it calculates for the matching "primary" fee type with number of cases
+    #    modifier of 2 (1 additional case) and "days/quantity" taken from the primary fee types' quantity.
+    # However, could use the prices endpoint of API directly or request a new endpoint to reduce number
+    # of calls and/or simplify.
+    #
+    class UnitPrice < Calculate
+      def call
```
lostie, line 16: **[DRY]** This call method seem to be exactly the same as its parent class with the exception of the locale for the error message.

app/services/claims/fee_calculator/unit_price.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      def defendant_uplift_unit_price
+        unit_price(:number_of_defendants) - unit_price
+      end
+
+      def amount
+        if fee_type.case_uplift?
+          uplift_unit_price(:number_of_cases)
+        elsif fee_type.defendant_uplift?
+          # TODO: This call replaced uplift_unit_price(:number_of_defendants)
```
lostie, line 76: ❓ Does this require any action, given its description?

app/services/claims/fee_calculator/unit_price.rb
```diff
@@ -0,0 +1,85 @@
...
+        fee_type.case_uplift? ? case_uplift_parent : defendant_uplift_parent
+      end
+
+      def unit_from_parent_or_one
+        parent = parent_fee_type
+        return 1 unless parent
+        current_total_quantity_for_fee_type(parent)
+      end
+
+      # TODO: promote?
```
lostie, line 57: ❓ What is to be promoted?

lostie: ❓ What is to be promoted?

app/views/external_users/claims/fixed_fees/_fixed_fee_fields.html.haml
```diff
@@ -15,37 +15,55 @@
...
+      errors: @error_presenter
+
+  .js-fixed-fee-calculator-effectee
+    .calculated-fixed-fee
+    = f.adp_text_field :rate,
+        label: t('.rate'),
+        input_classes:'rate fee-rate js-fixed-fee-calculator-rate form-input-denote__input form-control-1-4',
+        input_type: 'currency',
+        errors: @error_presenter,
+        input_readonly: (f.object.rate_calculated ||= false),
```
lostie: **[not a blocker]** This seems a bit unconventional to set the value of the field like this directly in the view, more so that we're doing exactly the same 4 lines below. Something to considering refactoring in the future.

config/locales/en.yml
```diff
@@ -297,6 +298,11 @@ en:
   disk_evidence:
     header: 'Evidence supplied on disk'
     question: 'Will you be sending disk evidence for this claim?'
+  fee_calculator:
+    calculate:
+      amount_unavailable: The calculated rate is unavailable, please enter manually.
+    unit_price:
```
lostie, line 304: Given this configuration for the locales, seems like [the call method](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2427/files#r209670597) has no actual difference between the children class and the parent class.

features/step_definitions/advocate_claim_steps.rb
```diff
@@ -91,16 +92,31 @@
...
   @claim_form_page.fixed_fees.last.select_fee_type name
-  @claim_form_page.fixed_fees.last.quantity.set 1
-  @claim_form_page.fixed_fees.last.rate.set "10.00"
+  wait_for_ajax
   @claim_form_page.fixed_fees.last.case_numbers.set "T20170001"
+  @claim_form_page.fixed_fees.last.quantity.set 1
+  wait_for_ajax
+end
+
+Then(/^I add a fixed fee '(.*?)' with defendants/) do |name|
```
lostie: **[nit pick/not a blocker]** Is `with defendants` required for the description? If so, I can't see anything in the actual implementation suggesting we're doing something specific with defendants. 😕

features/support/stub_calculator_requests.rb
```diff
@@ -0,0 +1,6 @@
+require 'webmock/cucumber'
+
+Before('@stub_calculator_request_and_fail') do
+  stub_request(:get, %r{\Ahttps://laa-fee-calculator(.*).k8s.integration.dsd.io/api/v1/.*\z}).
```
lostie, line 4: **[not a blocker]** If possible, it would be better to use the host configuration (presumably configured in the ENV) rather than statically defining here, no?

spec/services/claims/fee_calculator/unit_price_spec.rb
```diff
@@ -0,0 +1,141 @@
+RSpec.configure do |c|
```
lostie, line 1: **[not a blocker]** Any reason for this not to be within the scope of the `shared_examples` that require it, if the intent of this configuration is to apply to them only?

If I'm not mistaken, running the whole suite, with the random seed, all the specs will be loaded (including this one) first, which means this configuration will be active for all of them.

Or alternatively, this configuration would only be active for a metatag, e.g:

```ruby
# from the capybara examples, 
RSpec.configure do |config|
  config.before do
    Capybara.current_driver = Capybara.javascript_driver if example.metadata[:js]
  end
end
```

spec/support/matchers/general_matchers.rb
```diff
@@ -26,3 +26,24 @@
     msg
   end
 end
+
+RSpec::Matchers.define :be_json do
+  match do |actual|
+    JSON.parse(actual)
```
lostie, line 32: **[nit pick/not a blocker]** if the `actual` variable raises some exception different from JSON parsing (e.g calls a method on nil, claim.some_method) this matcher will tell the JSON is not parsable which is not the case. The rescue should probably not rescue from StandardError but from JSON::ParserError.

spec/vcr_helper.rb
```diff
@@ -3,16 +3,47 @@
...
   end
 end
+
+# use `VCR_OFF=true rspec` too turn off vcr
+VCR.turn_off! if ENV['VCR_OFF']
+
+# custom VCR request matcher to match request based on
+# path and query but not host because laa-fee-calculator
+# host could change and responses are path and query specific
+path_query_matcher = lambda do |request_1, request_2|
```
lostie, line 29: **[DRY]** This method seems to be repeated across different files, are they both necessary or can they be DRYied?

lostie (approved):
Codewise overall LGTM, there's some DRY work that can and should probably be done in the future.

There seems to be also a couple of places where we're defining data that its not of the responsibility of those classes, so that's something to keep in my to keep the code DRY and avoid re-defining the same behaviour across different classes.

lostie:
Just noticed as well that there's a LOT of commits and some of them seem to have temporary descriptions, so maybe they could use some `squashing` ?.

app/services/claims/fee_calculator/calculate.rb
jsugarman: This was to specifically handle the situation where insufficient details have been passed to this class to calculate a fee. In particular if, for example, you do not select an advocate category before selecting a fee type you would get an error from API through no fault of the user. I had intended to refactor this to some degree.

app/services/claims/fee_calculator/calculate.rb
jsugarman: 👍 agree

PR #2428, opened 2018-07-17
Fix advocate categories on fixed fee page

jsugarman:
#### What
Fix advocate categories on fixed fee page for new claims

#### Why
Found that for a new AGFS claim the advocate categories
on the fixed fee pages were showing the scheme 9 categories.

Not sure if this came out of the recent refactor of the fee reform
code but the fetch eligible advocate categories services was
still relying solely on the `fee_scheme:string` of `'fee_reform'`.


app/services/claims/fetch_eligible_advocate_categories.rb
```diff
@@ -22,12 +22,12 @@ def default_categories
...
 
-    def fee_reform_categories
+    def agfs_reform_categories
       Settings.agfs_reform_advocate_categories
     end
 
-    def using_fee_reform?
-      claim.fee_scheme == 'fee_reform'
+    def agfs_reform?
+        claim.fee_scheme&.agfs? && claim.fee_scheme.version >= 10
```
lostie: 💄 Indentation on this line looks wrong.

spec/services/claims/fetch_eligible_advocate_categories_spec.rb
```diff
@@ -1,66 +1,78 @@
 require 'rails_helper'
 
 RSpec.describe Claims::FetchEligibleAdvocateCategories, type: :service do
+  before do
+    FeeScheme.find_by(name: 'LGFS', version: 9) || create(:fee_scheme, :lgfs_nine)
```
lostie: **[not a blocker]** I've spoken offline with @colinbruce about this pattern and the more I see it, the more it feels that this should be centralised somewhere so that we avoid replicating the creation of all the fee schemes that should exist in the system in any spec we need them.

💭 Consider having this (at least) in some sort of helper in `spec/support` that we can call in the specs, so that if we decide to add/remove fee schemes, we can just change the helper code and leave the specs unchanged.

lostie (approved):
LGTM 👍 , only nice to have is some DRYness on the fee scheme creation in the specs.

lostie:
For our own reference, I've created an issue with the places I've noticed there's still occurrences of the string approach for checking the fee_scheme: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2433

lostie (approved):

Issue #2431, opened 2018-07-18
Replace DB report storage with S3


lostie:
Currently the `Stats::StatsReport` is persisting the report results (CSV strings) directly into a database column.

We should consider replacing this approach with S3 storage and when a user requests the specified report, a temporary (authorised) link is provided pointing to the requested report file.

mpw5:
Reviewing old issues: S3 is used for storing reports. The `stats_reports` database column is used only for storing a history of report generation, not the reports themselves.

PR #2436, opened 2018-07-20
Replace DB storage with S3 storage for MI reports

lostie:
#### What

Solves: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2431

#### Ticket

[MI reports file storage](https://dsdmoj.atlassian.net/browse/CBO-288)

#### Why

Is preferred to store file contents in a file storage system (e.g. S3) instead of storing them directly in the database as a blob.

### DIFF
--- a/app/controllers/case_workers/admin/management_information_controller.rb
+++ b/app/controllers/case_workers/admin/management_information_controller.rb
@@ -13,7 +13,14 @@ class CaseWorkers::Admin::ManagementInformationController < CaseWorkers::Admin::
 
   def download
     record = Stats::StatsReport.most_recent_by_type(params[:report_type])
-    send_data record.report, filename: record.download_filename, type: 'text/csv'
+    if record.document?
+      data = open(record.document_url).read
+      content_type = record.document_content_type
+    else
+      data = record.report
+      content_type = 'text/csv'
+    end
+    send_data data, filename: record.download_filename, type: content_type
   end
 
   def generate
--- a/app/models/stats/stats_report.rb
+++ b/app/models/stats/stats_report.rb
@@ -25,6 +25,16 @@ module Stats
     scope :management_information, -> { not_errored.where(report_name: 'management_information') }
     scope :provisional_assessment, -> { not_errored.where(report_name: 'provisional_assessment') }
 
+    has_attached_file :document,
+                      { s3_headers: {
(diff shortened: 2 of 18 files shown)


lostie:
screen shot 2018-07-20 at 11 45 13

☝️ After testing in the demo environment.

app/controllers/case_workers/admin/management_information_controller.rb
```diff
@@ -13,7 +13,14 @@ def index
 
   def download
     record = Stats::StatsReport.most_recent_by_type(params[:report_type])
-    send_data record.report, filename: record.download_filename, type: 'text/csv'
+    if record.document?
+      data = open(record.document_url).read
+      content_type = record.document_content_type
+    else
+      data = record.report
```
lostie, line 20: **NOTE:** We will give a transition period for the new reports to be generated and stored in S3. Once all of them are using this new approach, we should be able to remove the old way of storing things.

app/models/stats/stats_report.rb
```diff
@@ -46,16 +56,28 @@ def self.record_start(report_name)
...
     end
 
+    def document_url(timeout = 10_000)
+      return unless document?
+      document.options[:storage] == :filesystem ? document.path : document.expiring_url(timeout)
+    end
+
     def download_filename
-      "#{report_name}_#{started_at.strftime('%Y_%m_%d_%H_%M_%S')}.csv"
+      # TODO: set the appropriate format as required
```
lostie, line 79: **NOTE:** Once we're using the S3 approach for all reports, this might no longer be required as the document already contains this.

colinbruce (approved):
Nice! 
👍

jsugarman (approved):
👍 code LGTM - from a concept point of view other than bucket permissions (which i checked are good) we are retaining a transaction history in the event we want to locate previous reports so good to go.

I assume that costings are miniscule and therefore a non-issue.

lostie:
The bucket is the same already in use for other parts of the application (like the, users document upload, I believe, I'm just using a different location within the bucket).

As for history/cost savings in S3, we can (and should probably do) implement/improve the existent task that deletes old reports, to also delete their associated S3 files.

lostie:
Ive created https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2443 to address the S3 file deletion.

Issue #2433, opened 2018-07-18
Occurrences where fee_reform string is still in use


lostie:
That should be replace with querying the fee scheme object associated with the claim:

- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/app/presenters/fee/basic_fee_presenter.rb#L29
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/app/views/external_users/claims/basic_fees/_basic_fee_fields_primary.html.haml#L31
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/app/views/external_users/claims/basic_fees/_offence_display.html.haml#L6
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/app/services/claims/fetch_eligible_offences.rb#L23
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/spec/services/claims/fetch_eligible_offences_spec.rb#L62
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/spec/presenters/fee/basic_fee_presenter_spec.rb#L137
- https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/fix-eligible-advocate-categories-scheme-10/spec/services/claims/fetch_eligible_offences_spec.rb#L54

PR #2438, opened 2018-07-20
Replace all string fee reform comparisons with FeeScheme instance check

lostie:
#### What

- Removes the feature flag for fee reform since all the AGFS claims are being submitted in the all environments under the fee reform.

Relates to: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2425
Fixes: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2433

#### Ticket

[Addressing some tech debt from fee reform release](https://dsdmoj.atlassian.net/browse/CBO-287)

#### Why

There were some checks against the string version of the fee scheme that were left over from the previous refactor.

#### Next steps

To allow us to not have to do things like:

```
- if claim.scheme_10?
  # show scheme 10 eligible offences
```

in the code is to have a settings/configurations file that determine how things are set up on a given scheme, to turn the above code into something like:

```
= claim.eligible_offences
```

and under the hood we check the fee scheme associated with the claim and what sort of configuration is set for the eligible offences for that fee scheme.

### DIFF
--- a/app/models/fee_scheme.rb
+++ b/app/models/fee_scheme.rb
@@ -21,8 +21,8 @@ class FeeScheme < ApplicationRecord
     name.eql?('AGFS')
   end
 
-  def scheme_10?
-    version.eql?(10)
+  def agfs_reform?
+    agfs? && version >= 10
   end
 
   def self.current_agfs
--- a/app/models/fee_scheme.rb
+++ b/app/models/fee_scheme.rb
@@ -34,7 +34,6 @@ class FeeScheme < ApplicationRecord
   end
 
   def self.for_claim(claim)
-    # TODO: Align this with Fee reform SPIKE
     date = claim.earliest_representation_order&.representation_order_date
     scheme = claim.agfs? ? 'AGFS' : 'LGFS'
     if date.present?
--- a/app/services/claims/fetch_eligible_advocate_categories.rb
+++ b/app/services/claims/fetch_eligible_advocate_categories.rb
@@ -10,7 +10,7 @@ module Claims
 
     def call
       return unless claim&.agfs?
-      return agfs_reform_categories if agfs_reform? || Offence.in_scheme_ten.include?(claim.offence)
+      return agfs_reform_categories if agfs_reform?
       default_categories
     end
 
@@ -27,7 +27,7 @@ module Claims
     end
 
     def agfs_reform?
-      claim.fee_scheme&.agfs? && claim.fee_scheme.version >= 10
+      claim.agfs_reform? || Offence.in_scheme_ten.include?(claim.offence)
     end
   end
 end
--- a/app/services/claims/fetch_eligible_advocate_categories.rb
+++ b/app/services/claims/fetch_eligible_advocate_categories.rb
@@ -26,16 +26,8 @@ module Claims
       Settings.agfs_reform_advocate_categories
     end
 
-    # TODO: to be changed to solely use fee_scheme objects once rest of app cleared up
     def agfs_reform?
-      case claim.fee_scheme
-      when 'default'
-        false
-      when 'fee_reform'
-        true
-      else
-        claim.fee_scheme&.agfs? && claim.fee_scheme.version >= 10
-      end
+      claim.fee_scheme&.agfs? && claim.fee_scheme.version >= 10
     end
   end
 end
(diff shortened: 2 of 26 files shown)


colinbruce (approved):
:+1:

jsugarman (commented):
Looks good, but pending any significant `config` change version (as discussed), can we wrap up the agfs_reform logic into a fee scheme method like below, since and thinking we should probably be assuming agfs_reform is anything from scheme 10 **onwards**

```
# method on fee_scheme
def agfs_reform?
  agfs? && version >= (10)
end
```

and delegated from base_claim
```
delegate :agfs_reform?, to: :fee_scheme, allow_nil: true
```

and then use that everywhere needed

I had a [branch for similar setup](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/compare/remove-fee-reform-string)

Also, All my new feature test pass against this branch.

app/models/fee_scheme.rb
```diff
@@ -21,8 +21,13 @@ def agfs?
     name.eql?('AGFS')
   end
 
-  def scheme_10?
-    version.eql?(10)
+  # remove if unused
```
jsugarman: remove

app/services/claims/fetch_eligible_advocate_categories.rb
```diff
@@ -26,16 +26,8 @@ def agfs_reform_categories
...
     def agfs_reform?
-      case claim.fee_scheme
-      when 'default'
-        false
-      when 'fee_reform'
-        true
-      else
-        claim.fee_scheme&.agfs? && claim.fee_scheme.version >= 10
-      end
+      claim.agfs_reform? || Offence.in_scheme_ten.include?(claim.offence)
```
jsugarman, line 30: check/test whether the Offence check is needed here as claim.agfs_reform should be able to determine this itself via `FeeScheme.for_claim(self)` logic in `base_claim`

spec/models/external_user_spec.rb
```diff
@@ -298,52 +281,18 @@
...
-      context 'when the AGFS fee reform feature flag is active' do
-        before do
-          allow(FeatureFlag).to receive(:active?).with(:agfs_fee_reform).and_return(true)
-        end
-
-        it 'returns the list of available advocate claim types including the feature flagged ones' do
-          expect(user.available_claim_types)
-            .to match_array([Claim::AdvocateClaim, Claim::AdvocateInterimClaim, Claim::LitigatorClaim, Claim::InterimClaim, Claim::TransferClaim])
-        end
+      it 'returns the list of available advocate claim types including the feature flagged ones' do
```
jsugarman: remove reference to 'feature flag' from description

app/services/claims/fetch_eligible_advocate_categories.rb
lostie: Agree, @colinbruce might have more insight on this as this was something that was raised out of some API related submissions `:not_sure_if:` ?!

spec/models/claim/base_claim_spec.rb
```diff
@@ -704,6 +704,18 @@ def make_date_params(date_string)
...
 end
 
+describe '#agfs_reform?' do
+  let(:claim) { MockBaseClaim.new }
+  let(:mock_fee_scheme) { double(:fee_scheme) }
+
+  specify {
+    expect(FeeScheme).to receive(:for_claim).with(claim).and_return(mock_fee_scheme)
+    expect(mock_fee_scheme).to receive(:agfs_reform?)
+    claim.agfs_reform?
```
lostie: These tend to be a bit of dumb specs, but to complement the already existent expectations, consider to assert that the actual return value of `claim.agfs_reform?` is the same as the one returned by `mock_fee_scheme.agfs_reform?`, as the method in the claim could be something like this:

```ruby
def agfs_reform?
  fee_reform.agfs_reform?
  return 'something else'
end
```

(which is not the case, but it's just to illustrate why is probably better to assert that. 

I was actually looking for some RSpec expectations that just said something like:

```
# meta code :P
expect(claim).to_delegate(:agfs_reform?).to(:fee_scheme)
```

jsugarman: because we have gem 'shoulda-matchers', we can, and do, use  `it { should delegate_method(:meth).to(:object) }`

PR #2439, opened 2018-07-20
[CBO-151] API interfaces for travel distance calculation

lostie:
#### What

- Defines a service to retrieve the distance between two points (including alternative route distances)using the Google Maps API
- Define a controller action that returns the **return travel distance** between the LGFS suppliers base location and the destination of travel

#### Ticket

[Automatically calculate distance for travel expenses](https://dsdmoj.atlassian.net/browse/CBO-151)

#### How

The secrets necessary to integrate with the Google Maps API were highstated already: [here](https://github.com/ministryofjustice/advocate-defence-payments-deploy/pull/160)

**NOTE:** Still ⌛️ on setting up an account under the MOJ/LAA for Google MAPS. Not a blocker for now, but it needs to get **DONE**.

#### TODO (wip)

 - [ ] There's still no current integration with the UI


colinbruce (approved):
Looks fine.  Do we need the third party gem?  Is there a risk of him stealing our API key?

cleargif:
see pr: #2497

PR #2447, opened 2018-07-24
Fix MI generation

colinbruce:
#### What
Move the MI job into services and update the calling method from the scheduled task

#### Why
All of the MI generation jobs where refactored to use the same calling pattern and location, the scheduled job that  starts the long running MI task was not updated.


scheduled_tasks/management_information_generation_task.rb
```diff
@@ -6,7 +6,7 @@ class ManagementInformationGenerationTask < Scheduler::SchedulerTask
 
   def run
     log('Management Information Generation started')
-    Stats::ManagementInformationGenerator.new.run
+    Stats::StatsReportGenerator.call('management_information')
```
lostie, line 9: 🏆 , sorry for missing this out 😓

lostie (approved):
LGTM 👍

PR #2450, opened 2018-07-25
Changes to validation for cracked trials

kmahern:
Changes to validation for cracked trials - see CBO-279

#### What

Changes to validation to allow users to enter a rep order date after the "notice of first fixed warned date" and the "first fixed warned date" without the system bringing up error message that prevents this, and to prevent users from entering a rep order date after the date the case cracked

#### Ticket

CBO-279 https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&modal=detail&selectedIssue=CBO-279

#### Why

To allow for pre-rep order PTPH in cracked trials


colinbruce (approved):
:+1:

PR #2451, opened 2018-07-31
Cbo 45 risk based bills filter

kmahern:
#### What
Make changes to Risk based bills filter

#### Ticket

CBO-45 https://dsdmoj.atlassian.net/browse/CBO-45

#### Why

To enable users to carry out risk based billing

#### How

Made changes to filter classes and associated spec classes to implement functionality described in ticket.

### DIFF
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -109,8 +109,7 @@ module API
       end
 
       def risk_based_bills
-        ((risk_based_class_letter && (contains_risk_based_fee ||
-          (contains_risk_based_transfer_fee && up_to_and_inc_pcmh_transfer?))).eql?(true) && is_submitted?).to_i
+        ((risk_based_class_letter && contains_risk_based_fee).eql?(true) && is_submitted?).to_i
       end
 
       def injection_errored
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -110,7 +110,7 @@ module API
 
       def risk_based_bills
         ((risk_based_class_letter && (contains_risk_based_fee ||
-          contains_risk_based_transfer_fee)).eql?(true) && is_submitted?).to_i
+          (contains_risk_based_transfer_fee && up_to_and_inc_pcmh_transfer?))).eql?(true) && is_submitted?).to_i
       end
 
       def injection_errored
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -109,7 +109,8 @@ module API
       end
 
       def risk_based_bills
-        ((risk_based_class_letter && contains_risk_based_fee).eql?(true) && is_submitted?).to_i
+        ((risk_based_class_letter && (contains_risk_based_fee ||
+          contains_risk_based_transfer_fee)).eql?(true) && is_submitted?).to_i
       end
 
       def injection_errored
--- a/app/interfaces/api/v2/query_helper.rb
+++ b/app/interfaces/api/v2/query_helper.rb
@@ -1,3 +1,4 @@
+
 module API::V2
   module QueryHelper
     extend Grape::API::Helpers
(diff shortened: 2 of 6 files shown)


app/interfaces/api/v2/query_helper.rb
```diff
@@ -1,3 +1,4 @@
+
```
jsugarman: style wise best not to leave blank lines at top of page

app/interfaces/api/entities/search_result.rb
```diff
@@ -109,7 +109,8 @@ def interim_disbursements
       end
 
       def risk_based_bills
-        ((risk_based_class_letter && contains_risk_based_fee).eql?(true) && is_submitted?).to_i
+        ((risk_based_class_letter && (contains_risk_based_fee ||
```
jsugarman: I'd refactor to use a couple of methods or more with this method being the parent. just coz all the conditional logic and braces start to hurt peoples heads

kmahern: Thanks Joel, will do.

spec/models/claims/allocation_filters_spec.rb
```diff
@@ -36,24 +36,24 @@ module Claims
...
+        # create :interim_claim
+        # create :litigator_claim, offence: offence_a
+        # claim = create :litigator_claim, offence: offence_f
+        # create :graduated_fee, fee_type: guilty_plea_fee_type, claim: claim
+
+        # expect(Claim::BaseClaim.all_risk_based_bills).to match_array( [ claim ])
+      # end
+    # end
   end
 end
```
jsugarman: styling - add linebreaks at the end of files

app/interfaces/api/entities/search_result.rb
jsugarman: I'd refactor to use methods to isolate the conditional logic here

app/interfaces/api/v2/query_helper.rb
jsugarman: styling - should remove linebreaks from top of files

app/models/claims/allocation_filters.rb
```diff
@@ -45,13 +45,13 @@ def all_graduated_fees
...
     end
 
-    def all_risk_based_bills
-      where(type: 'Claim::LitigatorClaim')
-        .where(offence_id: Offence.joins(:offence_class).where(offence_classes: { class_letter: %w[E F H I] }))
-        .joins(:fees)
-        .where('"fees"."fee_type_id" = ?', Fee::GraduatedFeeType.where(description: 'Guilty plea').pluck(:id).first)
-        .where('"fees"."quantity" between 1 and 50')
-    end
+    # def all_risk_based_bills
```
jsugarman: could this not be amended to do what you need. I don't have a clear idea of everything going on but be wary of N+1 queries resulting from looping over thousands of records and issuing other queries for each one. I appreciate this is a tricky query though.

if this isn't used/required then should delete

spec/models/claims/allocation_filters_spec.rb
```diff
@@ -36,24 +36,24 @@ module Claims
...
-        create :transfer_claim
-        create :interim_claim
-        create :litigator_claim, offence: offence_a
-        claim = create :litigator_claim, offence: offence_f
-        create :graduated_fee, fee_type: guilty_plea_fee_type, claim: claim
-
-        expect(Claim::BaseClaim.all_risk_based_bills).to match_array( [ claim ])
-      end
-    end
+    # describe 'all_risk_based_bills' do
```
jsugarman: remove if not required

jsugarman (commented):
hi, looks ok but i don't have enough insight to check whether there are any N+1 queries going on. I'd refactor that method with lots of conditional logic tho. 👍

kmahern:
Hi Joel, think I have made the changes you suggested, Re the N+1 queries issue, I also don't know for sure but I don't think I have changed the underlying way that the filter is working. I have released it to Staging if you want to take a look, I have had a quick test and the performance doesn't appear any different to me.

colinbruce (approved):
👍

PR #2225, opened 2018-02-01
Splitting litigator claim pages

lostie:
💎  [Page splitting for Litigator final fee](https://www.pivotaltracker.com/story/show/154799214)
💎  [Page splitting for Litigator Interim Fee](https://www.pivotaltracker.com/story/show/154799221)
💎  [Page splitting for Litigator Transfer Fee](https://www.pivotaltracker.com/story/show/154799228)

Also:

* Change from *numbered steps* to **named steps**. Numbered steps at a given time can represent any sort of step, which makes it quite hard to understand what step was taken by a user in the past.

PR #2453, opened 2018-08-01
Working features in browser

colinbruce:
#### What
@lostie implemented chrome based testing as part of #2225

#### Why
Not all tests would run successfully, some because of the use of `.trigger('click')` some because of hidden or hover elements that poltergeist ignores.

#### How
Find all uses of `.trigger('click')` and replace with `.click`, this is honoured by poltergeist and chrome-webdriver

Identified and rectified all other failures.  

#### TODO (wip)
 - [ ] There is a possibility that one of the tests will flicker.  Monitor and address as needed

### DIFF
--- a/features/step_definitions/advocate_claim_steps.rb
+++ b/features/step_definitions/advocate_claim_steps.rb
@@ -40,7 +40,8 @@ end
 
 Then(/^I select the first search result$/) do
   sleep Capybara.default_max_wait_time
-  @claim_form_page.offence_results.first.select_button(visible: false).click
+  find(:xpath, '//*[@id="offence-list"]/div[3]/div').hover
+  find(:xpath, '//*[@id="offence-list"]/div[3]/div/div[2]/a').click
 end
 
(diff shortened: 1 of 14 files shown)


features/step_definitions/advocate_claim_steps.rb
```diff
@@ -40,7 +40,8 @@
 
 Then(/^I select the first search result$/) do
   sleep Capybara.default_max_wait_time
-  @claim_form_page.offence_results.first.select_button(visible: false).trigger('click')
+  find(:xpath, '//*[@id="offence-list"]/div[3]/div').hover
```
jsugarman, line 43: As discussed, can we just put a `# TODO: get working in site-prism` on this.

jsugarman (approved):
👍 just that todo comment

features/step_definitions/advocate_claim_steps.rb
colinbruce: I've added an issue, #2454

lostie:
🏆 😻

PR #2455, opened 2018-08-01
Amend fee scheme specs to use agfs release date

jsugarman:
#### What
Fix relative date tests for fee scheme

#### Why
Spec was using a relative date which has
now resulted in broken specs due to current
date.


kmahern (approved):
Fine all tests passed

PR #2456, opened 2018-08-08
CBO-293 reinstate grad fee calculator link

kmahern:
Changes for CBO-293 to reinstate calculator link for advocates.

#### What

Changes detailed on CBO-293, to add link to fee calculator for advocates on Fees screens, to make it consistent with the similar litigator screens.

#### Ticket

CBO-293

#### Why

So that advocates can easily access the calculators (because fee calculation will not be fully implemented for some time).

#### How

Added requested links to fee screens.


colinbruce (approved):
:+1:

PR #2458, opened 2018-08-14
[CBO-34] injection report successes/errors

lostie:
#### What

- Adds a daily task to produce injections success rate data that is then published to a Geckoboard dataset
- Exposes a new MI endpoint that returns an aggregated view of total number of injection errors per error category. This was done as such, since currently the injection error messages are extremely verbose and are not classified/categorised (which is something we should probably consider).

#### Ticket

[Report of injection errors/successes](https://dsdmoj.atlassian.net/browse/CBO-34)


colinbruce (approved):
:+1:

PR #2459, opened 2018-08-15
Update Injection Attempt error display

colinbruce:
#### What
As the data type of the error return has changed (it used to be a plain string), this needed to be
reflected in the tests and search objects

#### Why 
The previous configuration showed all claims as having an injection error because `errors: []` was counted as being `.present?`

#### How
This checks for the presence of, and parses, the errors and checks for the presence of entries in the array

### DIFF
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -57,7 +57,13 @@ module API
 
       # for display purposes we only want to use the injection error header
       def injection_errors
-        I18n.t(:error, scope: %i[shared injection_errors]) if object.injection_errors.present?
+        I18n.t(:error, scope: %i[shared injection_errors]) if parse_injection_errors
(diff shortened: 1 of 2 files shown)


app/interfaces/api/entities/search_result.rb
```diff
@@ -57,7 +57,13 @@ def last_submitted_at_display
 
       # for display purposes we only want to use the injection error header
       def injection_errors
-        I18n.t(:error, scope: %i[shared injection_errors]) if object.injection_errors.present?
+        I18n.t(:error, scope: %i[shared injection_errors]) if parse_injection_errors
+      end
+
+      def parse_injection_errors
```
jsugarman: I'd just be tempted to call this `injection_errors_present?`

jsugarman (approved):
👍 just a naming suggestion

PR #2460, opened 2018-08-21
Add missing scheme 10 misc fee mappings

jsugarman:
#### What
Add missing scheme 10 misc fee mappings

#### Why
Additional miscellaneous fees were added under AGFS scheme 10

#### How
Extend CCR misc fee adapter mappings to include them


colinbruce (approved):
:+1:

PR #2461, opened 2018-08-22
Extend MI recording

colinbruce:
#### What
Add new fields to MIData object

#### Ticket
[Add data to MIData generation](https://dsdmoj.atlassian.net/browse/CBO-333)

#### Why
The business would like to get the provisional assessment data broken down further than we currently store

#### How
Add the extra fields and populate them when transforming the data, prior to deletion


jsugarman (approved):
👍

PR #2463, opened 2018-08-23
CBO-280: Expenses additional information

cleargif:
#### What
Adding additional info to travel expenses

#### Ticket

CBO-280

#### Why
So users can give contextual information relating to expenses

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180822130536) do
+ActiveRecord::Schema.define(version: 20180823134645) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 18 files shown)


colinbruce:
Looks generally good @cleargif, does the same partial get used to display info for the case_worker?

If not we should probably add the case_worker summary view too...

cleargif:
@colinbruce - no the view is not shared.. I have done the updates

colinbruce (approved):
👍

db/schema.rb
```diff
@@ -254,6 +255,15 @@
     t.index ["external_user_id"], name: "index_documents_on_external_user_id", using: :btree
   end
 
+  create_table "establishments", force: :cascade do |t|
```
jsugarman, line 258: @cleargif do you know if this is a leftover from a migration on another branch

PR #2465, opened 2018-08-24
Provisional Assessment Report - filters

colinbruce:
#### What
Add filters to Provisional Assessment Report

#### Ticket
[CBO311 refine provisional assessment](https://dsdmoj.atlassian.net/browse/CBO-311)

#### Why
The business need to drill into the data further than the original, aggregated, report allowed

#### How
Create a new output that the users can create, for a given date range, that allows them to perform Transactional Profiling

--------

#### TODO (wip)

 - [X] implement new API, MI report
 - [X] allow calling of new report from existing MI page
 - [x] improve test coverage
 - [x] refactor JS
 - [ ] implement re-design of MI page - on hold, will be separate ticket

### DIFF
--- a/app/assets/javascripts/modules/case_worker/admin/management_information.js
+++ b/app/assets/javascripts/modules/case_worker/admin/management_information.js
@@ -10,7 +10,7 @@ moj.Modules.ManagementInformation = {
 
   buildDataArray: function() {
     var data = {};
-    data.api_key='64c9177b-e5a0-4ac5-9f2b-33166597c436';
+    data.api_key=$('#user_api_key').val();
     data.start_date=$('#start_date').val();
     data.end_date=$('#end_date').val();
     data.format='csv';
--- a/app/assets/javascripts/modules/case_worker/admin/management_information.js
+++ b/app/assets/javascripts/modules/case_worker/admin/management_information.js
@@ -1,0 +1,63 @@
+moj.Modules.ManagementInformation = {
+  init: function () {
+    this.bindEvents();
+  },
+
+  bindEvents: function () {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/modules/case_worker/admin/management_information.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+
+  bindEvents: function () {
+    this.provisionalAssessmentsDateFieldChange();
+    this.provisionalAssessmentsDownloadClicked();
+  },
+
+  buildDataArray: function() {
+    var data = {};
+    data.api_key='64c9177b-e5a0-4ac5-9f2b-33166597c436';
```
jsugarman: is this key ok here?

app/services/reports/provisional_assessments_dates.rb
```diff
@@ -0,0 +1,54 @@
+module Reports
+  class ProvisionalAssessmentsDates
```
jsugarman: not a blocker  - `ProvisionalAssessmentsByDate[s]` as a name may be clearer

jsugarman (commented):
Looks good. just want to confirm if the API key needs to be an env var or other kind of secret?

app/assets/javascripts/modules/case_worker/admin/management_information.js
colinbruce: Good catch... that should have been current_user.api_key... I wonder where that went! 😨

jsugarman (approved):
LGTM 👍

PR #2468, opened 2018-08-30
CBO-335 Changes to validation for travel expenses

kmahern:
Changed travel expense validation to mirror CCR

#### What

Changed validation so that date of travel has to be on or after the earliest representation order date

#### Ticket

CBO-335

#### Why

In order to mirror CCR validation, for data injection.


colinbruce (approved):
:+1:

PR #2469, opened 2018-08-30
Add limit from to fixed fee calc

jsugarman:
#### What
Add a limit_to option to the fee calc prices endpoint call

#### Ticket

[Fix scheme 10 fixed standard appearance fee calculator bug](https://dsdmoj.atlassian.net/browse/CBO-343)

#### Why
Scheme 10 Standard appearance fee price has 2 based on limit ranges. This means you get two
prices back from API which cannot currently be handled appropriatley. This fix enforces a filter (`limit_from`) to get just the first price based on limit ranges if there is more than one.

#### How
Add a `limit_from` of 1 option to the fee calc prices endpoint call. 

This is a quickfix for prices with limit ranges (currently only the scheme 10 Standard appearance fixed fee has a range limit). 

long term fixed need to dynamically determine limit boundaries and reflect in the net amount calculated.

stats:
```
# counts of quantities claimed for fixed standard appearance fees on gamma 3888 total fees claimed of this type
[[0.0, 3], [1.0, 2357], [2.0, 807], [3.0, 352], [4.0, 190], [5.0, 85], [6.0, 54], [7.0, 16], [8.0, 6], [9.0, 4], [10.0, 5], [11.0, 6], [13.0, 2], [87.0, 1]]
```


colinbruce (approved):
LGTM 👍

PR #2470, opened 2018-08-30
Chrome fixed fee calculation feature

jsugarman:
#### What
Bring chrome browser feature test window to foreground

#### Why
For demo purposes and to avoid confusion

#### How
use capybara.switch_to_window helper method called via all sign in steps if
feature test is being run with chrome driver

e.g.
```
 BROWSER=chrome cucumber features/fee_calculator/fixed_fee_calculator.feature
```

#### Note
Seems to be consistently passing when running the particular feature but 
JS dependent steps may fail sometimes.


colinbruce (approved):
:+1: Looks fine

Surprised at need for `switch_to_chrome_window` as it seemed to be working fine for me locally.  Is it a remote thing?

PR #2471, opened 2018-08-31
Add safe navigator to collection of case numbers

jsugarman:
#### What
Prevent 500 error on DI as in [sentry](https://sentry.service.dsd.io/mojds/private-beta/issues/33170/)

#### Ticket

[relates to](https://dsdmoj.atlassian.net/browse/CBO-347)

#### Why
Currently it is possible to, as a separate bug,
to step through a path of form submission
which ends up with a valid claim submission but
with a fixed fee case uplift that has no value
and no additional case numbers.

Until this bug is resolved the addition of
the safe navigator will prevent a 500 on
injection of such a claim.


colinbruce (approved):
👍

PR #2472, opened 2018-08-31
Ensure last_injection_attempt is correct

colinbruce:
#### What
Prevent successfully injected claims from being flagged as injection errors when allocating

#### Ticket
[Update injection_error filtering for allocation](https://dsdmoj.atlassian.net/browse/CBO-348)

#### Why
When multiple injection attempts have been made, this ensures that the most recent attempt is returned

#### How
Sorting by creation date and limiting was returning the oldest record, switching to `DESC` ensures that the most recent attempt is returned


jsugarman (approved):
👍

PR #2475, opened 2018-09-03
Add utility class and module for (re)injecting claims

jsugarman:
#### What
Enable re-injection of multiple claims into CCR

#### Ticket

[CBO-351](https://dsdmoj.atlassian.net/browse/CBO-351)

#### Why
Post DI release for CCR some claims were not injecting
because of a CCR bug. To prevent case workers having
to manually create these we are intending to reattempt
the injection of such claims.

#### How
Add utility class for injection multiple claims and module for querying injection attempts
by error.

### DIFF
--- a/lib/data_injection/injector.rb
+++ b/lib/data_injection/injector.rb
@@ -1,0 +1,21 @@
+# A utility class to inject multiple claims.
+# example:
+#  claims = DataInjection::Queries.claims_with_error("The supplier account code: .* is INVALID!")
+#  injector = DataInjection::Injector.new(claims)
+#  injector.call
+#
+module DataInjection
+  class Injector
+    attr_accessor :claims
+
+    def initialize(claims)
+      @claims = claims
+    end
+
+    def call
+      claims.each do |claim|
+        NotificationQueue::AwsClient.new.send!(claim)
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


lostie (approved):
As spoken offline the in statement could probably be simplified with something like:

```
claims = Claim::BaseClaim.arel_table
claims[:type].in(::Claim::BaseClaim.agfs_claim_types).to_sql
>> "\"claims\".\"type\" IN ('Claim::AdvocateClaim', 'Claim::AdvocateInterimClaim')"
```

(not a blocker though)

Would be useful (possible) to have a Rake task that can call the required commands so there's more visibility of how to perform the `Injector` process.

lib/data_injection/injector.rb
```diff
@@ -0,0 +1,21 @@
...
+  class Injector
+    attr_accessor :claims
+
+    def initialize(claims)
+      @claims = claims
+    end
+
+    def call
+      claims.each do |claim|
+        NotificationQueue::AwsClient.new.send!(claim)
```
ddippolito, line 17: Do you need to instantiate a client for every claim?

ddippolito:
Consider moving this to a Rake task **if** we might need to do this again and again.
Not necessarily something to address now, it can be a future ticket

PR #2457, opened 2018-08-09
CBO-300 Travel expenses: Dynamic location/destination field

cleargif:
#### What
AGFS:
- This feature is not active on AGFS at the moment.
- AGFS & LGFS do share the view

LGFS:
- Integrating the `establishment` data
- Introducing new travel reasons that map to particular establishments
- Managing the state and content for each expense block depending on options selected
- (while not breaking AGFS)

#### Ticket

[CBO-300](https://dsdmoj.atlassian.net/browse/CBO-300)

#### Why
- This feature is to enable the distance lookup feature [CBO-151](https://dsdmoj.atlassian.net/browse/CBO-151)

#### How
- Using a number of new `moj.Helpers` & `moj.Modules` all the moving parts have been abstracted and tested seperately
- New cucumber features test the in page states as the user makes changes

#### TODO
 - [ ] Error states  are not entirely handled properly

PR #2480, opened 2018-09-05
Remove establishments table definition from schema.rb

jsugarman:
#### What
remove a schema definition entry for establishment table from schema.rb

#### Why
Was probably added inadvertently by work on another
branch related to expense api work. see PR: 
[CBO-300 Travel expenses: Dynamic location/destination field](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2457)


cleargif (approved):
👍

PR #2481, opened 2018-09-06
Ignore certain DI errors entirely

jsugarman:
#### What
exclude certain DI errors from DI stats

#### Why
Injection of claims that already exist on
CCR results in one of two "already exist"
errors. Since these are more of a warning
the PM would like to exclude them.

#### Ticket

[CBO-361](https://dsdmoj.atlassian.net/browse/CBO-361)

#### How
create injection attempt scope to exclude certain errors
and modify data injection geckoboard report to use this
for totals counts (totals therefore represent only successes
and "true" errors)

### DIFF
--- a/spec/lib/geckoboard_publisher/injections_report_spec.rb
+++ b/spec/lib/geckoboard_publisher/injections_report_spec.rb
@@ -73,6 +73,7 @@ RSpec.describe GeckoboardPublisher::InjectionsReport, geckoboard: true do
       travel_to(Date.parse('19-MAR-2017')) do
         create_list(:injection_attempt, 3, claim: agfs_claim)
         create_list(:injection_attempt, 2, :with_errors, claim: agfs_claim)
+        create_list(:injection_attempt, 1, :with_stat_excluded_errors, claim: agfs_claim)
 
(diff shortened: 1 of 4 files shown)


spec/lib/geckoboard_publisher/injections_report_spec.rb
```diff
@@ -138,6 +151,7 @@
           is_expected.to include item
         end
       end
+
```
jsugarman: remove

spec/lib/geckoboard_publisher/injections_report_spec.rb
```diff
@@ -129,6 +130,18 @@
...
       let(:end_date) { Date.new(2017, 3, 21) }
 
+      let(:total_excluding_error) do
+        InjectionAttempt
+        .where(created_at: start_date)
+        .where.not('coalesce(error_messages::text,\'\') ILIKE ?','%already exist%')
+        .count
+      end
+
+      it 'injection attempt "already exists" are excluded' do
```
jsugarman: redescribe

lostie (approved):
👍 

One more reason for us to think about aggregating the errors with some sort of error codes on the CCR/CCLF or at the last resort when consuming them.

PR #2482, opened 2018-09-06
Adjourned appeals changes

jsugarman:
#### What
Repurpose the Adjourned appeals (MISAF) fee as a fixed fee type 

#### Ticket

[CBO-339](https://dsdmoj.atlassian.net/browse/CBO-339)

#### Why
This fee type is currently a miscellaneous fee but represents a fixed amount and should be claimable on its own - currently users are forced to select a fixed fee and "make one up" just so they can get to the misc fees page to claim for the Adjournment. 

The name should be changed to reflect the fact that it is claimable only for appeals against conviction, committals for sentence and breach of crown court order hearing adjournments.

#### TODO
- [X] seeds
- [X] factories (if necessary)
- [X] specs and features (if necessary)
- [X] rake task to update database as one off
- [X] amend eligible fixed fees
- [X] amend DI mappings and test
- [X] check fee calc works
- [x] check API and JSON uploader works
- [x] amend API documentation (should canvas for impact)
- [x] test update against data dump of live
- [x] renaming of fee

#### future considerations

** also amend CCCD fee calc integration unit price to handle/ignore FXSAF and FXADJ as parents so number of cases and defendant uplifts apply to the main fixed fee only - current BUG for Standard appearance fee in CCCD


colinbruce (approved):
LGTM :+1: apart from the updates needed for features...

PR #2483, opened 2018-09-07
Add eligible fixed fee API release notes

jsugarman:
#### What
Add API release notes for fetch eligible fixed fee types feature

#### Why
There is an impact on API consumers that are currently
injecting ineligible fixed fee types.


jsugarman:
content was reviewed by PM and ok'ed

PR #2486, opened 2018-09-17
Handle elected case not proceeded fixed fee calculation

jsugarman:
#### What
Enable fee calculation of Elected case not proceeded fees

#### Ticket
[CBO-340](https://dsdmoj.atlassian.net/browse/CBO-340)

#### Why
Although ENP is treated as a fixed fee in CCCD it
actually maps to an advocate fee with a fixed price
in the calculator (AGFS_FEE).

The prices endpoint JSON for the price of this
fee stores the value in a `fixed_fee` attribute
instead of the usual `fee_per_unit` attribute.

From a business/fee scheme point of view the 
amount is not affected by either advocate
category or unit multipliers. This last point will need
handling by other means in the current CCCD configuration
(e.g. validation, JS).

#### How
Modify the calculate and unit price service objects used for fixed 
fee calculation to default offences to miscellaneous class/band when
absent on the claim - because, currently, ENP requires
offence for fee calculator (see issue below). In addition modify the Price
object representing an individual price to use the `fixed_fee` attribute
instead of the `fee_per_unit` for applicable fees and scenarios.

--------

#### TODO (wip)

 - [X] implement ENP fee calc in fixed fee page
 - [x] spec and refactor
 - [ ] handle unit multiplier effect (validation and or FE changes)?

#### Other

- [x] fee calculator requires an offence for this fee but it has no impact - see [fee calc issue](https://github.com/ministryofjustice/laa-fee-calculator/issues/65)

### DIFF
--- a/app/assets/javascripts/modules/external_users/claims/FeeCalculator.js
+++ b/app/assets/javascripts/modules/external_users/claims/FeeCalculator.js
@@ -31,10 +31,6 @@ moj.Modules.FeeCalculator = {
   },
 
   // needs to be usable by cocoon:after-insert so can bind to one or many elements
-  // TODO: "Elected case not proceeded" does not have a
-  // unit price but a single fixed amount regardless
-  // of days claimed for (i.e. you cannot claim for more than
-  // one of this fee).
   fixedFeeQuantityChange: function ($el) {
     var self = this;
     var $els = $el || $('.js-fixed-fee-calculator-quantity');
--- a/app/assets/javascripts/modules/external_users/claims/FeeCalculator.js
+++ b/app/assets/javascripts/modules/external_users/claims/FeeCalculator.js
@@ -31,6 +31,10 @@ moj.Modules.FeeCalculator = {
   },
 
   // needs to be usable by cocoon:after-insert so can bind to one or many elements
+  // TODO: "Elected case not proceeded" does not have a
+  // unit price but a single fixed amount regardless
+  // of days claimed for (i.e. you cannot claim for more than
+  // one of this fee).
   fixedFeeQuantityChange: function ($el) {
     var self = this;
     var $els = $el || $('.js-fixed-fee-calculator-quantity');
--- a/app/services/claims/fee_calculator/calculate.rb
+++ b/app/services/claims/fee_calculator/calculate.rb
@@ -11,6 +11,7 @@ module Claims
     class Calculate
       delegate  :earliest_representation_order_date,
                 :agfs?,
+                :agfs_reform?,
                 :case_type,
                 :offence,
                 to: :claim
@@ -93,10 +94,23 @@ module Claims
         end&.first
       end
 
+      # Send a default offence as fee calc currently requires offences
+      # for some prices even though the values are identical for different
+      # offence classes/bands.
+      # TODO: fee calculator API should not require
+      # offences for at least "Elected case not proceeded"
+      #
+      # TODO: "Elected case not proceeded" does not have a
+      # unit price but a single fixed amount regardless
+      # of days claimed for (or you cannot claim for more than
+      # one of this fee. see javascript responsible.
+      #
       def offence_class
-        # some/all fixed fees do not require offences and they have no bearing on calculated fee amount
-        # TODO: make conditional on fee scheme version
-        offence&.offence_class&.class_letter || offence&.offence_band&.description
+        if agfs_reform?
+          offence&.offence_band&.description || '17.1'
+        else
+          offence&.offence_class&.class_letter || 'H'
+        end
       end
 
       def advocate_type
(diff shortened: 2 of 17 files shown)


app/services/claims/fee_calculator/price.rb
```diff
@@ -1,7 +1,10 @@
 # Price: Wraps a fee calc API price object to
-# factor in modifier effects and uplift parent quantities
-# on fee per unit.
-#
+# factor in modifier effects and uplift parent
+# quantities on fee_per_unit or fixed_fee
+# attributes.
+# Note: a fee calc api fixed_fee is a fee
```
jsugarman: remove

app/services/claims/fee_calculator/price.rb
```diff
@@ -13,8 +16,26 @@ def initialize(price, modifier_name = nil, parent_quantity = 1)
         @parent_quantity = parent_quantity
       end
 
+      # TODO: Some prices use a different fee calc attribute - fixed_fee
```
jsugarman: remove

spec/factories/case_types.rb
```diff
@@ -32,6 +32,8 @@
       is_fixed_fee true
     end
 
+
```
jsugarman: remove

app/assets/javascripts/modules/external_users/claims/FeeCalculator.js
```diff
@@ -31,6 +31,10 @@ moj.Modules.FeeCalculator = {
   },
 
   // needs to be usable by cocoon:after-insert so can bind to one or many elements
+  // TODO: "Elected case not proceeded" does not have a
```
jsugarman: either remove comment or apply logic pending conversation with PM

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -83,20 +84,41 @@ def fee_scheme
...
+      def bill_scenario
+        CCR::CaseTypeAdapter::BILL_SCENARIOS[case_type.fee_type_code.to_sym]
+      end
+
       # TODO: consider creating a mapping to fee calculator id's
       # - less "safe" but faster/negates the need to query the API??
       #
       def scenario
-        # TODO: create select/find_by calls to list endpoints in client gem
+        # TODO: create select/find_by calls to retrieve endpoint data by attribute value
```
jsugarman, line 95: relates to [laa-fee-calculator-client PR](https://github.com/ministryofjustice/laa-fee-calculator-client/pull/24)

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -83,20 +84,35 @@ def fee_scheme
...
       def offence_class
-        # some/all fixed fees do not require offences and they have no bearing on calculated fee amount
-        # TODO: make conditional on fee scheme version
-        offence&.offence_class&.class_letter || offence&.offence_band&.description
+        if agfs_reform?
+          offence&.offence_band&.description || '17.1'
+        else
+          offence&.offence_class&.class_letter || 'H'
+        end
       end
```
colinbruce, line 116: Something about this seems wrong but I can't put my finger on it. 
Would (but  with a better name)
```ruby
def default_value_if_offence_class_missing
  agfs_reform? ? '17.1' : 'H'
end
```
 be more declarative?

colinbruce (approved):
One not-a-blocker question but otherwise 👍

app/services/claims/fee_calculator/calculate.rb
jsugarman: will do

jsugarman: @colinbruce or could put the default on the offence class and band models?

jsugarman: can't think of a particularly neat way of abstracting this without adding a lot of code. is renaming sufficient:

```
def offence_class_or_default
  if agfs_reform?
    offence&.offence_band&.description || '17.1'
  else
    offence&.offence_class&.class_letter || 'H'
  end
end
```

The use of the term `offence_class` within this class `LAA::FeeCalculator::Calculate.offence_class` is intended to reflect the fee calculators own parameter argument of `offence_class` rather than CCCDs OffenceClass model (which is only one of two (OffenceBand being the other)

what do you think?

app/services/claims/fee_calculator/calculate.rb
```diff
@@ -107,7 +107,7 @@ def scenario
       # TODO: fee calculator API should not require
       # offences for at least "Elected case not proceeded"
       #
-      def offence_class
+      def offence_class_or_default
```
jsugarman: remove

app/models/offence_class.rb
```diff
@@ -34,4 +34,8 @@ def letter_and_description
   def lgfs_offence_id
     offences.miscellaneous.first.id
   end
+
+  def class_letter_or_default
```
jsugarman: remove

app/services/claims/fee_calculator/calculate.rb
colinbruce: :+1:

As discussed, let's leave as-is for now, this code is pretty volatile at the moment, I can always suggest an improvement... if I can think of it! :)

PR #2487, opened 2018-09-18
Bump rubocop

jsugarman:
#### What
Update rubocop to 0.59.1

#### Why
currently quite a way behind on 0.52.1

Also, have permanently ignored Metric/lineLength cop for inline and block comments 
using an IgnoredPatterns option. But really need some time to look at this properly.


.rubocop_todo.yml
```diff
@@ -1,107 +1,78 @@
...
+# on 2018-09-18 16:02:28 +0100 using RuboCop version 0.59.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 234
+# Cop supports --auto-correct.
+Layout/EmptyLineAfterGuardClause:
+  Enabled: false
```
colinbruce: Is this a temp fix?  If not, shouldn't it be in `rubocop.yml` or just autocorrected?

colinbruce (approved):
Generally good.

The only observation is that by changing e.g. `LineLength` from named files to max:242, we can now create new massively over sized lines.

By having named lists it would prevent us adding new occurrences but ignore the known 'bad` files.

Happy to discuss

.rubocop_todo.yml
jsugarman: not keen on this cop as a stlye issue. as discussed will move into rubocop

jsugarman:
cool, have fixed a few more simple cops and excluded particularly long method length violations individually (there are only two massive ones)

PR #2488, opened 2018-09-18
Update laa-fee-calculator-client gem to 0.2.0 and use improvements

jsugarman:
#### What
Update laa-fee-calculator-client gem to 0.1.4 and use improvements

#### Why
New release enables searching of result sets using
a `find_by` method. This method can query data
based on attribute values in the "records" rather
than being limited to the APIs query parameters.


colinbruce (approved):
👍

PR #2489, opened 2018-09-20
Fix flickering spec and faulty spec

jsugarman:
#### What
Unit price spec was flickering and fixed_fee test was faulty

#### Why
The use of the fxenp_fee factory trait could
result in either use of a persisted fee type
with unique code of FXENP or unpersisted fee
type for the test. This is, in turn, can result
in a different result dependant on order of
test runs.

reproducible with this:
```
rspec --seed 21453
```

The fixed_fee_type_spec for case uplifts
was not testing what was expected.


colinbruce (approved):
:+1: for reducing test flickers

PR #2493, opened 2018-09-24
Reset fee type mapping singleton after spec

jsugarman:
#### What
Prevent flickering unit_price_spec and impact on others.

#### Why
 The fee type mappings spec seeds case types and resets mappings. 
As a singleton class this mapping will "persist" and thereby impact
other tests that may or may not bereliant on it.

#### How
reset the singleton after each example


colinbruce (approved):
👍

PR #2495, opened 2018-09-25
CBO-356: Fix basic fee JS calculation

jsugarman:
#### What
Apply basic fee `quantity x rate` calculation to FE 

#### Ticket

[CBO-356](https://dsdmoj.atlassian.net/browse/CBO-356)

#### Why
Basic fee quantity x rate calculation is only being performed on the
backend. To reflect the simple calculation to the user immediately
we apply the calculation on the front end via JS.

#### How
Add JS to all applicable basic fees to calculate
the net amount (quantity x rate) inline with backend
logic that does this.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -102,6 +102,17 @@ if (!String.prototype.supplant) {
     $(this).trigger('recalculate');
   });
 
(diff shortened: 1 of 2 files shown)


cleargif:
👍

cleargif (approved):

PR #2499, opened 2018-09-30
CCLF injection reporting

colinbruce:
#### What
Allow CCLF injection reports to be logged separately in Slack

#### Why
To allow easier tracking of the CCLF injection development 
#### How
Provide different channels for AGFS and LGFS injection reports
In the event that the claim type cannot be inferred, e.g. bad API key, log the message in the #cccd_development channel to allow quicker resolution of errors when they don't get lost in successful messages in the other channels

### DIFF
--- a/app/services/claims/injection_channel.rb
+++ b/app/services/claims/injection_channel.rb
@@ -2,7 +2,7 @@ module Claims
   class InjectionChannel
     def self.for(claim)
       return 'cccd_development' if claim.nil?
-      claim.agfs? ? Settings.slack.channel : 'cccd_cclf_injection'
+      claim.agfs? ? 'cccd_ccr_injection' : 'cccd_cclf_injection'
     end
   end
 end
--- a/app/services/claims/injection_channel.rb
+++ b/app/services/claims/injection_channel.rb
@@ -1,0 +1,8 @@
+module Claims
+  class InjectionChannel
+    def self.for(claim)
+      return 'cccd_development' if claim.nil?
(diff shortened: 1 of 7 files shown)


app/services/claims/injection_channel.rb
```diff
@@ -0,0 +1,8 @@
+module Claims
+  class InjectionChannel
+    def self.for(claim)
+      return 'cccd_development' if claim.nil?
+      claim.agfs? ? Settings.slack.channel : 'cccd_cclf_injection'
```
jsugarman: As discussed, can we just put the config for the channels in the app entirely. clear up the TD env vars later

jsugarman (approved):
👍 - Just that minor change to specify all relevant channels in the new class

app/services/claims/injection_channel.rb
```diff
@@ -0,0 +1,8 @@
+module Claims
+  class InjectionChannel
+    def self.for(claim)
+      return 'cccd_development' if claim.nil?
+      claim.agfs? ? 'cccd_ccr_injection' : 'cccd_cclf_injection'
```
jsugarman, line 5: does it need to be `#channel_name`? - [the env var](https://github.com/ministryofjustice/advocate-defence-payments-deploy/pull/164) and old settings was - worth deleting that too

colinbruce: No, the slack API documentation is unclear, but submitting messages with just the channel name, not #-prefixed, works fine

PR #2504, opened 2018-10-02
Travel Net/VAT amounts event triggers

cleargif:
#### What
- Better handling of events for NET / VAT amount calculations
- Calculations will only be done if a mileage rate and distance is present

#### Ticket

CBO-416

#### Why
The calculation should not be done till all variables have been selected

#### How
Better targeting even tiggers


colinbruce (approved):
:+1:

PR #2505, opened 2018-10-03
Update the MI display of Claim::AdvocateInterimClaims

colinbruce:
#### What/Why
Due to an oversight when implementing the new AGFS warrant type,
the classification was not extended to the claim_csv_presenter.
Only Claim::AdvocateClaims are set as AGFS, therefore the new
Claim::AdvocateInterimClaim type is classed as unknown

#### Ticket
[CBO-418](https://dsdmoj.atlassian.net/browse/CBO-418)


jsugarman (approved):
👍

PR #2506, opened 2018-10-03
CBO-411 Case details supplier labels

cleargif:
#### What
The supplier name should be displayed if provided

#### Ticket

CBO-411

#### How
- updated the presenter to include a supplier name if supplier


spec/presenters/supplier_number_presenter_spec.rb
```diff
@@ -19,5 +20,13 @@
         expect(presenter.supplier_label).to eq(account_number)
       end
     end
+
+    context 'when the supplier does not have a defined postcode' do
```
colinbruce: I think this test should be described as `supplier does not have a defined name`

colinbruce (approved):
Minor nitpick but, otherwise, :+1:

PR #2507, opened 2018-10-04
CBO-405: Case details travel distance warning messages

cleargif:
#### What
To list a set of warnings to draw the caseworker's attention to the distance calculations on an expense

#### Ticket

[CBO-405](https://dsdmoj.atlassian.net/browse/CBO-405)

#### Why
More clarity for caseworkers

### DIFF
--- a/app/views/shared/summary/_expenses.html.haml
+++ b/app/views/shared/summary/_expenses.html.haml
@@ -1,3 +1,6 @@
+- if current_user_is_caseworker? && claim.expenses.map(&:distance_gt_calculated?).any?
+  = render partial: 'shared/summary/expenses/distance_warnings', locals: { expenses: claim.expenses }
+
 %h3.heading-medium
   = t('.travel_expenses')
 
@@ -7,7 +10,6 @@
       = t('.travel_expense_additional_information')
     %p
       = claim.travel_expense_additional_information
-
 
 - if claim.expenses.with_vat.any?
   = render template: 'shared/summary/expenses/index', locals: { claim: claim, vat: true }
--- a/app/views/shared/summary/expenses/show.html.haml
+++ b/app/views/shared/summary/expenses/show.html.haml
@@ -6,47 +6,7 @@
     %span.bold-normal
       = expense.reason
   %td{ style: 'width: 75%'}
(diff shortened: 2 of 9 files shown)


app/views/shared/summary/_expenses.html.haml
```diff
@@ -1,3 +1,16 @@
+.error-summary{"aria-labelledby" => "error-summary-expenses", :role => "alert", :tabindex => "-1"}
+  %h1#error-summary-heading-example-1.heading-medium.error-summary-heading
+    There are errors with the travel expenses
+  %p The distance claimed by the provider is greater than the calculated distance for the following expenses:
```
jsugarman: unless we want a more specific message i think `greater` will have to be `different`

joelgsamuel:
@cleargif I don't work on CCCD, could you please avoid tagging me in your PRs?

app/views/shared/summary/_expenses.html.haml
cleargif: @jsugarman  - I will raise it in JIRA.. thought that as well when I read the line

cleargif:
@joelgsamuel - apologies, was a miss type.

app/views/shared/summary/expenses/show.html.haml
```diff
@@ -1,52 +1,12 @@
-%tr.no-border
+%tr.no-border{id: "expense_id_#{expense.id}"}
```
jsugarman: remove, no linking added following design review so not needed

config/locales/en.yml
```diff
@@ -1269,6 +1269,13 @@ en:
           destination: Destination
           view_map_link: View on a map
           hours: Hours
+        distance_warnings:
+          distance_warning_header:
```
jsugarman: rename to just `header`

config/locales/en.yml
```diff
@@ -1269,6 +1269,13 @@ en:
...
+            one: There is an error with the travel expenses
+            other: There are %{count} errors with the travel expenses
+          distance_warning_body:
```
jsugarman: rename to just `body`

PR #2513, opened 2018-10-08
origin supplier name and postcode helper

jsugarman:
#### What
Presenter for supplier name and postcode and conditional display

#### Why
Abstract view logic to presenter. Can be used to 
handle AGFS providers/external user postcodes 
in the future.

Do not show origin for AGFS expenses for now.


cleargif (approved):
👍 LGTM

PR #2514, opened 2018-10-08
Add travel distance calc postcode validation.

jsugarman:
#### What
prevent distance calc to destinations by `name`

#### Why
This reverts commit 20d329b9d1fe9f00b37627e91a986a8cd7f95c30.

Is to prevent calculating distances based on a name such as Nottingham.
Google maps API will not complain if given a general destination name
but the resulting distance will be to what google considers to be 
the "center" of the named place.

#### How
add validation to raise error when given a destination that is not
a postcode.


colinbruce (approved):
:+1:

PR #2516, opened 2018-10-09
Clearing form inputs as they change state

cleargif:
#### What
- Removing any user entered values when elements are hidden
- this relates to the current bug where errors are shown on hidden fields

#### Ticket

CBO-415

#### Why
Validation errors are displayed on fields that are hidden at the time of submit.
STEPS:
1. Add an expense
2. Choose car travel and fill in the all the inputs
3. Change the expense type field to parking
4. Save and continue

Currently:
An error relating to distance, hours or mileage rate might be displayed  (depending on your choices)

Should be:
The form should submit as normal and not return you to the expense page


cleargif:
Recreating under the correct ticket

PR #2517, opened 2018-10-10
CBO 397 - Unique maat references

colinbruce:
#### What
Ensure that MAAT reference numbers are unique within a claim

#### Ticket
[MAAT references must be unique for each defendant](https://dsdmoj.atlassian.net/browse/CBO-397)

#### Why
Having duplicated MAAT references causes issues with data injection to legacy applications

#### How
Add a new method to validate that the submitted MAAT reference only occurs once within the submitted claims MAAT references

### DIFF
--- a/app/validators/representation_order_validator.rb
+++ b/app/validators/representation_order_validator.rb
@@ -67,7 +67,7 @@ class RepresentationOrderValidator < BaseValidator
     return unless case_type&.requires_maat_reference?
     validate_presence(:maat_reference, 'invalid')
     validate_pattern(:maat_reference, /^[0-9]{7,10}$/, 'invalid')
-    matt_reference_uniqueness(:maat_reference, 'unique')
+    validate_matt_reference_uniqueness(:maat_reference, 'unique')
   end
 
   # helper methods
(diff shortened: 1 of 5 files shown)


config/locales/error_messages.en.yml
```diff
@@ -701,6 +701,10 @@ representation_order:
       long: "Enter a valid MAAT reference for the #{representation_order} of the #{defendant}"
       short: Enter a valid MAAT reference
       api: "Enter a valid MAAT reference for the representation order of the defendant"
+    unique:
+      long: "Please ensure that MAAT references are unique for each defendant"
```
jsugarman: just a content thing. A content person reviewed our error messages in the distant past and told us command-like sentences should be used, no pleases etc. So something like 'The MAAT references for the defendants are not unique'

jsugarman (approved):
👍 - the error message content should be changed to not have please in it though

app/validators/representation_order_validator.rb
```diff
@@ -67,6 +67,7 @@ def validate_maat_reference
     return unless case_type&.requires_maat_reference?
     validate_presence(:maat_reference, 'invalid')
     validate_pattern(:maat_reference, /^[0-9]{7,10}$/, 'invalid')
+    matt_reference_uniqueness(:maat_reference, 'unique')
```
jsugarman: long winded but i would use `validate_maat_reference_uniqueness` for consistency and clarity

PR #2518, opened 2018-10-11
Move postcode regex to settings.yml

jsugarman:
#### What
Move postcode regex to settings.yml

#### Why
The use of an initialiser to set `Settings` explicitly
outside of the settings.yml file conflicts with the 
Config gems methods providing `Settings`
and caused failure in local development.

Could be related to an initializer race condition.


colinbruce (approved):
👍

PR #2519, opened 2018-10-11
Ignore conference and view fees for DI

jsugarman:
#### What
Do not send adapted conference and view fees regardless

#### Ticket

[CBO-417 Ignore CAVs on a case and inject everything else in to CCR](https://dsdmoj.atlassian.net/browse/CBO-417)

#### Why
Adapted CAV fees currently cause data injection failures
from CCR. To prevent DI failure and enable the majority
of the claims details to be sent over it has been decided
to exclude them from DI altogether

The CAV fees themselves cannot be reliably adapted for
DI because current handling of CAVs in CCCD and CCR is incompatible.
Too many assumptions would have to be made to get them
to inject and result in inaccurate calculated amounts CCR-side.

#### How
Rather than removing the mapping completely, add an exclusion list to be more explicit.


colinbruce (approved):
👍

PR #2521, opened 2018-10-15
CBO-430 travel automation report

colinbruce:
#### What
Create a new report that exports additional travel information where distances where calculated
This should be between two date-times, by default this will be yesterday (start to end), but users can provide additional dates for a wider range if needed

#### Ticket
[Automate travel expense additional information data](https://dsdmoj.atlassian.net/browse/CBO-430)

#### Why
So that we can analyse responses and act accordingly

#### How
Duplicate report structure and API implementation of other reports

--------

#### TODO (wip)

 - [x] Update report to default to yesterday

### DIFF
--- a/app/services/reports/additional_travel_expense_information_by_dates.rb
+++ b/app/services/reports/additional_travel_expense_information_by_dates.rb
@@ -1,0 +1,36 @@
+module Reports
+  class AdditionalTravelExpenseInformationByDates
+    NAME = 'additional_travel_expense_information'.freeze
+    COLUMNS = %w[claim_id	travel_expense_additional_information].freeze
+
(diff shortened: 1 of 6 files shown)


app/services/reports/additional_travel_expense_information_by_dates.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def call
+      Expense.connection.execute(query).to_a
+    end
+
+    private
+
+    def query
+      <<~SQL
```
jsugarman, line 24: squiggly heredoc 👍

jsugarman (approved):
LGTM 👍

PR #2522, opened 2018-10-15
Add maintenance mode

jsugarman:
#### What
Enable in app maintenance mode

#### Ticket

[relates to CBO-195](https://dsdmoj.atlassian.net/browse/CBO-195)

#### Why
rather than rely on DNS changes at the load balancer which are slow and cumbersome

#### How
Place catchall route at top that is activated by an env vars presence. Then rely on update 
of deployment salt stack pillars to activate maintenance mode


jsugarman:
Works locally but not on server hosted environments - probably due to Template Deploy behaviours

PR #2523, opened 2018-10-15
CBO-413: Reinstating expenses on LGFS interim warrant claims

cleargif:
#### What
Expenses are not being displayed on the interim warrant path

#### Ticket

CBO-413

#### How
Removing the conditional check on the presenter


jsugarman (approved):
👍

PR #2526, opened 2018-10-17
nullify cracked dates for non cracked case types

jsugarman:
#### What
Clear/nullify cracked case related fields when saving
non-cracked case claims.

#### Ticket

[CBO-445](https://dsdmoj.atlassian.net/browse/CBO-445)

#### Why
Prevent API vendor CMSs from submitting details that are inapplicable
and not explicitly validated against in CCCD. This relates to Data injection
errors where such inapplicable details are being sent to CCR which validates 
them regardless of case type/scenario

#### How
Use before validation callback/hook to clear inapplicable values


colinbruce (approved):
:+1:

PR #2528, opened 2018-10-18
Add deleted_at to disbursement type seeding

jsugarman:
#### What
Reflect softly deleted disbursements in seeds

#### Ticket

[CBO-447](https://dsdmoj.atlassian.net/browse/CBO-447)

#### Why
This is to ensure any disbursement types softly deleted
on production are reflected on other envs when/if reseeded.

Need to check all softley deleted records and their
respective seeding files too.


cleargif (approved):
👍

PR #2530, opened 2018-10-22
Add seeding for ccr cclf application case worker accounts

jsugarman:
#### What
Adds seeds for CCR and CCLF case worker application users

#### Why
So `db:reseed` and `db:reload` rake tasks
do not break Data injection.

relates to: https://github.com/ministryofjustice/advocate-defence-payments-deploy/pull/167


colinbruce (approved):
:+1:

PR #2531, opened 2018-10-22
Ensure nil bills not included in response

jsugarman:
#### What
Ensure nil bill types on transfer claims are not injected

#### Why
To fix Transfer claim defendant uplifts injection error

#### How
This Just in time logic/method of excluding any unmappable 
bills based on bill_type presence is used by litigator final
and interim already. Applying it to transfers too as fix.


jsugarman:
github issues

PR #2532, opened 2018-10-22
Ensure defendant nil bills not included in response

jsugarman:
#### What 
exclude nil bill types from transfer claims

#### Why
These are not included in litigator final
and interim claims already. applied to transfers
too.


jsugarman:
transfered to another branch in attempt to kick github webhooks into action

PR #2535, opened 2018-10-22
Use commandline for brakeman

jsugarman:
#### What
Fix brakeman run for local and CI test suite runs

#### Why
Brakeman rake task has been broken following a
gem update. This changes to use the commandline
instead as [recommended](https://brakemanscanner.org/docs/rake/).


colinbruce (approved):
:+1:

PR #2537, opened 2018-10-25
CCLF geckoboard reporting fix

colinbruce:
#### What
Ensure the geckoboard report for CCLF injections _also_ ignores "Duplicate case" warnings
This had been implemented on CCR and should have been done for both fee schemes


lostie (approved):
👍 

💭 Consider converting error messages into error codes so they can be more efficiently filtered without using ILIKE sort of queries

PR #2541, opened 2018-10-30
Update brewfile to use cask

kmahern:
#### What

Update brewfile to use cask, so that developers can setup environment on local machine.


jsugarman (approved):
👍

PR #2543, opened 2018-10-31
CBO-454 : Restoring archived claims

colinbruce:
#### What
Allow providers to unarchive claims

#### Ticket
[Allow claims to be restored](https://dsdmoj.atlassian.net/browse/CBO-454)

#### Why
Following changes made to the system earlier in the year, some claims could be archived in a state where they could not be un-archived.  This removes validation on restore... after all, how could data be in the DB if it's not valid!

All subsequent changes should progress as normal

#### How
Remove validation from the save on unarchiving


jsugarman (approved):
👍

PR #2544, opened 2018-11-01
Cbo 455 allow draft deletion

colinbruce:
#### What
Allow providers to delete _all_ draft claims

#### Ticket
[Unable to delete cases from draft](https://dsdmoj.atlassian.net/browse/CBO-455)

#### Why
Providers can save draft claims in any state, the code required to 'delete' them, actually a logical delete rather than a physical delete, still applied validation

#### How
Change the soft delete code to update the delete_at attribute of the claim.  This does not trigger validation and will allow 'invalid' claims to be removed from provider's claim lists

Bonus: updated the controller code to show better errors if the delete or archive methods fail

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -113,16 +113,14 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
   end
 
   def destroy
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
👍

PR #2547, opened 2018-11-02
CBO-444 : CAV warnings for case workers

colinbruce:
#### What
**As a** AGFS caseworker
**I want** to be able to see if conferences and views (CAVs) have not been injected in to CCR

#### Why
**so that** I know I need to investigate the CAVs that have been added to the claim by the provider and decide whether I need to add this to the injected case in CCR.

#### Ticket
[CAVs - warning message to caseworkers](https://dsdmoj.atlassian.net/browse/CBO-444)

#### How
Implement a new check for CAV fees and a partial for the new warning

### DIFF
--- a/app/presenters/claim/base_claim_presenter.rb
+++ b/app/presenters/claim/base_claim_presenter.rb
@@ -345,6 +345,10 @@ class Claim::BaseClaimPresenter < BasePresenter
     supplier_name
   end
 
+  def has_conference_and_views?
+    claim.fees.select { |f| 'BACAV'.include?(f.fee_type.unique_code) }.any? { |x| x.amount.nonzero? }
+  end
+
   private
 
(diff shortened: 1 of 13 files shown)


app/presenters/claim/base_claim_presenter.rb
```diff
@@ -345,6 +345,10 @@ def supplier_name_with_postcode
     supplier_name
   end
 
+  def has_conference_and_views?
+    claim.fees.select { |f| 'BACAV'.include?(f.fee_type.unique_code) }.any? { |x| x.amount.nonzero? }
```
jsugarman: A fee will only have one fee_type
```
f.fee_type.unique_code.eql?('BACAV')
```

jsugarman (approved):
👍 just one suggestion

PR #2548, opened 2018-11-05
Remove explicit rvm call to see if travis is working again

jsugarman:
#### What
Remove explicit rvm call in `.travis.yml`

#### Why
Travis suddenly stopped reading the ruby-version, now it has started again, this should not be needed


jsugarman:
Force merging this to get it into master to test

PR #2551, opened 2018-11-06
CBO-462 : Ensure single rep order for injection

colinbruce:
#### What
Ensure that the CCR/LF endpoints only return a single rep-order.  If there are multiple, it should be the earliest dated one.

#### Ticket
[More than one rep order on a claim injection error](https://dsdmoj.atlassian.net/browse/CBO-462)

#### Why
CCR/LF will reject the injection if there are multiple rep-orders for a defendant

### DIFF
--- a/app/interfaces/api/entities/ccr/defendant.rb
+++ b/app/interfaces/api/entities/ccr/defendant.rb
@@ -6,7 +6,7 @@ module API
         expose :first_name
         expose :last_name
         expose :date_of_birth
-        expose :representation_orders_with_earliest_first,
+        expose :earliest_representation_order,
                using: API::Entities::CCR::RepresentationOrder,
                as: :representation_orders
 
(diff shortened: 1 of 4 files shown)


app/interfaces/api/entities/ccr/defendant.rb
```diff
@@ -16,8 +16,8 @@ def main_defendant
           object.claim.defendants.unscope(:order).order(created_at: :asc)&.first == object || false
         end
 
-        def representation_orders_with_earliest_first
-          object.representation_orders.unscope(:order).order(representation_order_date: :asc)
+        def earliest_representation_order
+          object.representation_orders.unscope(:order).order(representation_order_date: :asc).take(1)
```
jsugarman, line 20: any advantage of `take(1)` rather than `first`?

jsugarman: as discussed, a collection is expected

jsugarman (approved):
👍

PR #2552, opened 2018-11-07
Lgfs fixed fee calc rebased

jsugarman:
#### What
Fixed fee calculation for LGFS

#### Ticket

[CBO-295 LGFS Fixed Fee Calculation](https://dsdmoj.atlassian.net/browse/CBO-295)

#### Why
So users do not have to use calculator spreadsheets to calculate LGFS 
fixed fees

#### How
Re-implement quantity/rate for LGFS fixed fees and pre-populate rate based
on unit prices from fee calculator API.

#### Todo

- [x] (re)implement quantity/rate on views
- [x] (re)implement basic quantity*rate calculation for lgfs fixed fees
- [x] (re)implement quantity/rate validations and basic calculation
- [x] error messaging for quantity/rate
- [x] implement LGFS fixed fee calc API unit price
- [x] implement LGFS fixed fee calc api for defendant uplifts (these are misc fees with no qxr) - separate ticket
- [x] implement fee calc api for case uplifts - n/a to LGFS fixed fees generally
- [x] implement fee calc api for LGFS edge case fixed fees (ENP, hearing subsequent to sentence etc) 
- [x] rejig specs
- [x] rejig/write new features
- [x] impact on DI??
- [x] handle impact on existing data
- [x] handle/document/release notes impact on cccd API


app/models/fee/base_fee.rb
```diff
@@ -129,16 +133,17 @@ def perform_validation?
       claim&.perform_validation? || claim&.from_json_import?
     end
 
+    # where fee type not available, default calculated? to true
     def calculated?
-      fee_type&.calculated?.nil? ? true : fee_type&.calculated?
+      fee_type&.calculated?.nil? ? true : fee_type.calculated?
```
colinbruce, line 138: Would 
```ruby
fee_type&.calculated? ||= true
```
be clearer?

config/locales/en.yml
```diff
@@ -943,6 +943,7 @@ en:
             quantity_hint: 'Enter a quantity'
             rate: *rate
             amount: Total
+            quantity: Quantity
```
colinbruce: previously you had used 
```*quantity```
to avoid the repetition

File: spec/services/claims/fee_calculator/unit_price_spec.rb [hunk 3]
```diff
@@ -57,129 +57,172 @@
...
-        params.merge!(fee_type_id: uplift_fee.fee_type.id)
-        params[:fees].merge!("1": { fee_type_id: uplift_fee.fee_type.id, quantity: uplift_fee.quantity })
+        it_returns 'a successful fee calculator response'
+        it_returns 'a fee calculator response with amount', amount: 87.0
       end
 
-      it_returns 'a successful fee calculator response'
-      it_returns 'a fee calculator response with amount', amount: 26.0
-    end
+      # TODO: deprecated fee type - to be removed
```
colinbruce, line 125: Can it be removed now? Or soon?

colinbruce (approved):
3 minor comments, but otherwise LGTM :+1:

app/models/fee/base_fee.rb
jsugarman: as discussed, this method is used in model validations, presenters and views. The views in particular are not well spec'd, so risky to change.

spec/services/claims/fee_calculator/unit_price_spec.rb
jsugarman: This only appears in the PR because i indented the AGFS claims context. The case type specific fixed fee uplifts themselves still exist (seeds etc). Therefore i'd say that the test should be removed inline with the actual data (in case we need to add the uplifts backin - unlikely)

Will create a tech debt ticket for this

PR #2553, opened 2018-11-08
CAV warning hotfix

colinbruce:
#### What
Sentry has flagged an issue with CAV where the amount is `nil`

#### How
Use a safe operator on amount for nonzero?


jsugarman (approved):
👍

PR #2554, opened 2018-11-08
Fix flickering features

colinbruce:
#### What
After the introduction of scheme 11 offences the features were flickering during test runs

#### Why
The seeds for scheme 11+ offences can be fragile and expect the
database to have been created from an empty start.

#### How
Ensure the database is deleted prior to a cucumber run by running a `DatabaseCleaner.clean_with(:deletion)` before cucumber starts and ensuring the `Offence` primary key has been reset before creating the required seed data

### DIFF
--- a/features/support/hooks.rb
+++ b/features/support/hooks.rb
@@ -25,6 +25,7 @@ end
 Before do
   unless ($seed_done ||= false)
 
+    ActiveRecord::Base.connection.reset_pk_sequence!('offences')
     load "#{Rails.root}/db/seeds/courts.rb"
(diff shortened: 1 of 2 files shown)


features/support/hooks.rb
```diff
@@ -25,6 +25,7 @@
 Before do
   unless ($seed_done ||= false)
 
+    ActiveRecord::Base.connection.reset_pk_sequence!('offences')
```
jsugarman, line 28: i wonder if this 

```
Rake::Task['db:reseed'].invoke
```

could replace all this

jsugarman (approved):

PR #2556, opened 2018-11-08
Default grad fee quantity to one.

jsugarman:
#### What
Fix CCLF data injection bug

#### Ticket

[CBO-466](https://dsdmoj.atlassian.net/browse/CBO-466)

#### Why

CCLF does not accept a quantity of zero
for litigator fees, `LIT_FEE`, but CCCD grad
fees do. This was a bug identified in production.

The ideal fix should be in CCLF but we are blocked
on such changes at the moment.

#### How

return 1 if quantity is `0` or `nil`


colinbruce (approved):
:+1:

PR #2558, opened 2018-11-12
CBO-458 : MAAT number validation

colinbruce:
#### What
Refine the MAAT validation 

#### Ticket

[MAAT reference validation: take 2!](https://dsdmoj.atlassian.net/browse/CBO-458)

#### Why
To try and prevent providers submitting obviously fake MAAT references, we regularly see `00000000` and `123456789`

#### How
Create a `Settings.maat_regexp` entry that will still allow use of test data, but that can be overwritten by env variable on individual environments if we need to refine this further

### DIFF
--- a/spec/models/representation_order_spec.rb
+++ b/spec/models/representation_order_spec.rb
@@ -61,14 +61,12 @@ describe RepresentationOrder do
       context 'our test MAAT reference' do
         before { representation_order.maat_reference = '2320006' }
 
-        context 'when on the live environment' do
+        context 'uses environment configured MAAT regex' do
           before do
-            allow(ENV).to receive(:[]).with('MAAT_REGEXP').and_return /^[4-9][0-9]{6}$/
-            allow(Settings).to receive(:maat_regexp).and_return ENV['MAAT_REGEXP']
+            allow(Settings).to receive(:maat_regexp).and_return /^[4-9][0-9]{6}$/
           end
 
           it 'should error' do
-            expect(Settings.maat_regexp).to eql /^[4-9][0-9]{6}$/
             expect(representation_order).not_to be_valid
           end
         end
--- a/spec/models/representation_order_spec.rb
+++ b/spec/models/representation_order_spec.rb
@@ -64,7 +64,7 @@ describe RepresentationOrder do
       end
     end
 
-    context 'case type does not require maat refrence' do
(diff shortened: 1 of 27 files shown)


spec/models/representation_order_spec.rb
```diff
@@ -52,13 +53,35 @@
...
+
+      context 'our test MAAT reference' do
+        before { representation_order.maat_reference = '2320006' }
+
+        context 'when on the live environment' do
+          before do
+            allow(ENV).to receive(:[]).with('MAAT_REGEXP').and_return /^[4-9][0-9]{6}$/
+            allow(Settings).to receive(:maat_regexp).and_return ENV['MAAT_REGEXP']
+          end
+
```
jsugarman, line 68: just to break the tests down i'd be tempted to have a 

it 'uses environment configured MAAT regex' do
...
end

spec/models/representation_order_spec.rb
```diff
@@ -52,13 +53,35 @@
...
         expect(representation_order).to be_valid
       end
+
+      context 'our test MAAT reference' do
+        before { representation_order.maat_reference = '2320006' }
+
+        context 'when on the live environment' do
+          before do
+            allow(ENV).to receive(:[]).with('MAAT_REGEXP').and_return /^[4-9][0-9]{6}$/
+            allow(Settings).to receive(:maat_regexp).and_return ENV['MAAT_REGEXP']
```
jsugarman: is this second mock required? i know rspec-mocks get funny, but would hope the ENV stub would be sufficient

jsugarman (approved):
👍 - couple of test suggestions/queries

spec/models/representation_order_spec.rb
colinbruce: It was because the Settings gem doesn't actually read the env vars in test, so I had to mock the env var first and then simulate `Settings` reading it.
I have refactored, and renamed the test, as the actual function changed over time.  Does the new version seem better to you?

colinbruce:
Adding a `do-not-merge` badge because this needs more discussions about handling with the business.  See the Jira ticket for more details

spec/models/representation_order_spec.rb
jsugarman: yep 👍

PR #2562, opened 2018-11-14
CAV allocation fix

colinbruce:
#### What
Certain cases would appear in the allocation queue with a CAV warning that would not then appear on the claim view

#### Why
The methods used to calculate the requirement differed between allocation and the view varied

#### How
Add an extra field to the SQL query that can be used to replicate the same definition in both areas


vcr/cassettes/spec/services/claims/fee_calculator/unit_price_spec.yml
```diff
@@ -1604,4 +1604,44 @@ http_interactions:
       string: '{"count":1,"next":null,"previous":null,"results":[{"id":58840,"scenario":5,"advocate_type":null,"fee_type":54,"offence_class":"K","scheme":2,"unit":"DAY","fee_per_unit":"0.00000","fixed_fee":"349.47000","limit_from":0,"limit_to":null,"modifiers":[],"strict_range":false}]}'
     http_version: 
   recorded_at: Sat, 03 Nov 2018 21:49:42 GMT
+- request:
```
jsugarman: not a blocker - odd that this caused a new request recording. Any idea why? The unit_price_spec should be setup to always make the same requests (and therefore existing recordings should cover it)

app/interfaces/api/v2/query_helper.rb
```diff
@@ -79,14 +69,24 @@ def unallocated_sql
...
+              FROM injection_attempts last_ia
+              WHERE last_ia.claim_id = c.id
+                AND last_ia.deleted_at is NULL
+              ORDER BY last_ia.created_at DESC
+              LIMIT 1
+            ) AS last_injection_attempt ON 1=1
         WHERE
           c.deleted_at IS NULL
-          AND c.type IN CLAIM_TYPES_FOR_SCHEME
+          AND c.type IN ('Claim::AdvocateClaim', 'Claim::AdvocateInterimClaim')
```
jsugarman: I believe `CLAIM_TYPES_FOR_SCHEME` is needed for dynamic scheme context via the block below


```
     helpers do
          def in_statement_for(arr)
            arr.map(&:to_s).join('\', \'').prepend('(\'').concat('\')')
          end

          def claim_types_for_scheme
            return in_statement_for(::Claim::BaseClaim.agfs_claim_types) if scheme.eql?('agfs')
            in_statement_for(::Claim::BaseClaim.lgfs_claim_types)
          end

          def claims
            built_sql = unallocated_sql.gsub(/CLAIM_TYPES_FOR_SCHEME/, claim_types_for_scheme)
            result = ActiveRecord::Base.connection.execute(built_sql).to_a
            JSON.parse(result.to_json, object_class: OpenStruct)
          end

          def scheme
            params[:scheme]
          end
        end
```

to discuss, there should be a breaking test from this change....unless i am missing something

colinbruce: Absolutely correct... this was a cut'n'paste error from getting the query updated :(

vcr/cassettes/spec/services/claims/fee_calculator/unit_price_spec.yml
colinbruce: This must have a been a lurking cassette from a different branch... also :(

jsugarman (approved):
👍

PR #2563, opened 2018-11-14
Update the anonymise export function

colinbruce:
#### What
Add `travel_expense_additional_information` to the anonymiser

#### Why
Providers are adding data into the field that could potentially identify clients

#### How
Add the `travel_expense_additional_information` field to those that are anonymised prior to export


jsugarman (approved):
👍 ooh, yes

PR #2565, opened 2018-11-16
Fix AGFS API integration test

jsugarman:
#### What
fix local AGFS API integration test run

#### Why
This was passing upstream but failing locally
when running
`rspec spec/api/v1/external_users/claims/integration/* --format d`

This appears to have been caused by:

1. an expense validation where date must not be before earliest rep order
2. a namespacing/naming clash on the custom matcher be_valid_api_claim

#### How
Fixed by:

1.  basing expense date on rep order date
2.  renaming local rspec matcher to avoid clash


colinbruce (approved):
LGTM 👍

PR #2566, opened 2018-11-19
CBO-477: Update MI report

colinbruce:
#### What
Add defendant and maat_reference columns to the original MI report

#### Ticket
[MAAT reference on CCCD MI report](https://dsdmoj.atlassian.net/browse/CBO-477)

#### Why
Case workers wanted a quicker way to be able to cross reference data from other sources

#### How
Add extra methods to the csv_presenter and call them


jsugarman (approved):
LGTM 👍

PR #2569, opened 2018-11-20
Use unpresented fee amount for field value

jsugarman:
This field was using a presented fee amount that
resulted in, for example, £20.00 in the
value attribute of the tag. Does not
have an impact on value saved when it is
changed but if fee calc is deactivated
(if we need too, by commenting `.calculated-grad-fee`
css class out in the view) it will not render the
presented amount as the input value.


colinbruce (approved):
:+1:

PR #2570, opened 2018-11-20
Affix grad fee calc spec vcr cassettes

jsugarman:
#### What
RSpec bug fix - prevent rerecording of cassettes

#### Why
Ensure LGFS fee calc specs use same values
Default factory was creating rep order date
based on current date which meant when specs
are run the day after it will change the
fee calc API query and therefore record a new
cassette.

This affixes the rep order date to that of the
start of scheme 8, meaning repeated test runs
on different days will NOT record new cassettes

TODO: really need to switch of recording cassettes
so new recordings are forbidden and will raise an
error.

#### How
Use transient attribute/flag in litigator claim factory
to set rep order date to start fo LGFS scheme 8.


colinbruce (approved):
:+1: nice

PR #2571, opened 2018-11-20
Fix price calculated for grad fee calc bug

jsugarman:
#### What
Fix bug whereby graduated fee calculation was not persisting `price_calculated` flag

#### Ticket

relates to [CBO-299](https://dsdmoj.atlassian.net/browse/CBO-299)

#### Why
bug fix, the flag was i believe primarily to prevent form navigation and saving as
draft resulting in a form state in which the fee cannot be calculated but is readonly
but need to check as the `pageLoad` js function may render this unnecessary now.

Is useful for stats in any event to determine how many fees have been "price calculated".


colinbruce (approved):
👍

PR #2572, opened 2018-11-21
Use unpresented fee rate for LGFS fixed fee rate

jsugarman:
#### What
Change LGFS fixed fee rate input value to use unpresented fee object value

#### Why
This field was using a presented fee rate that
resulted in, for example, £20.00 in the
value attribute of the tag. Does not
have an impact on value saved as it is
recalculated but would have impact
if we need too deactivate fee calc, as the
presented value is ignored

#### TODO:
Note, quantity field uses presented value to apply "not a decimal"
`number_with_precision` logic but this leads to inconsistency and 
confusion.


colinbruce (approved):
👍

PR #2573, opened 2018-11-21
Amend graduated fee calc help text

jsugarman:
#### What 
screen shot 2018-11-21 at 12 30 40

#### Why
Was wrong


simonwhatley (approved):
👍

PR #2574, opened 2018-11-21
Rename the Claims::FeeCalculator::Calculate superclass

jsugarman:
#### What 
Refactor - rename CCCD Calculate service object

#### Why
As a super class of Graduated/UnitPrice it should 
be named similarly IMO.


colinbruce (approved):
:+1:

PR #2575, opened 2018-11-21
Update calculated grad fee amount on effector keyup

jsugarman:
#### What
Update calculated grad fee amount on keyboard press

#### Why
In line with fixed fee calc rate update.


cleargif (commented):
👍

jsugarman:
dp forgot to approve, but thumbed up so force merging

PR #2577, opened 2018-11-22
CBO-459: Fixing a colour leaking out to body copy

cleargif:
#### What
The error class applies colour to all nested body copy

#### Ticket
CBO-459

#### How
Adding an extra level of scope and override style


colinbruce (approved):
LGTM 👍

PR #2578, opened 2018-11-23
Improve indexes

colinbruce:
#### What
Add indexes as detailed in the original PRs #2323, #2394 and #2402 
but updating to make the last one run concurrently and without  a transaction

#### Why
Again, see commit messages.

#### How
This will be deployed out of hours to prevent any table locking to a minimum


jsugarman (approved):
Cool 👍 - just a query on whether it is worth using the concurrent algorithm for claims.deleted_at as well

PR #2579, opened 2018-11-27
CBO-476 Seed the AJAX request with the current selected value

cleargif:
#### What
On allocation the data set and the selected radio button does not match up when the page is loaded

#### Ticket

CBO-476

#### Why
Caseworkers doing allocation have to do unnecessary clicks to reset the state

#### How

The selected value of the radio buttons is now checked when the AJAX url is created.


colinbruce (approved):
👍 and 🎉 for tidy

PR #2580, opened 2018-11-27
CBO-482: Update caseworker map links

colinbruce:
#### What
Refine the handling of the View map link on the caseworker claim view

#### Ticket
[CCCD: "View map" link on caseworker view](https://dsdmoj.atlassian.net/browse/CBO-482)

#### Why
Users were reporting confusion about when they needed to click on the link and what they needed to review

#### How
Update the helpers and presenters to refine the circumstances under which the link should be shown, where it should point and what the text should display


jsugarman (approved):
LGTM 👍

PR #2581, opened 2018-11-28
Upgrade to rails 5.1

jsugarman:
#### What
Upgrade to rails 5.1

#### Ticket

support and maintenance

#### Why

upgrade to rails 5.1 as a step to 5.2 to handle deployment issues
encountered on previous attempt


spec/models/fee/interim_fee_spec.rb
```diff
@@ -21,155 +21,154 @@
...
+    it 'should be true' do
+      expect(fee.is_interim?).to be true
     end
+  end
 
-    describe '#is_disbursemen?t' do
-      it 'should be true for disbursements' do
-        expect(disbursement_fee.is_disbursement?).to be true
-      end
+  describe '#is_disbursemen?t' do
```
colinbruce: `?` not at end of method name

colinbruce (approved):
Mind a little 💥 , but generally LGTM.

One tiny, non-blocking, typo in a test name, but otherwise :+1:

PR #2583, opened 2018-11-28
Add fix for missing calculated distance

colinbruce:
#### What
Expenses without a calculated distance failed the presenter

#### Why
Examples were missed during testing

#### How
Add an override to compare `distance` with itself if the `calculated_distance` is missing


jsugarman:
U might get away with just a ‘calculated_distance.to_f’ or to_i but if it works

jsugarman (approved):
LGTM

PR #2584, opened 2018-11-29
CBO-483: enhance additional travel information

colinbruce:
#### What
Acceptance Criteria
* Remove "(optional)" from the additional information input label on the travel information page
* Add validation to the additional information, if a car travel distance has changed; or, a 45p rate has been chosen on any expense
  - Display an error message if additional information is not provided when a user has claimed mileage at 45p
    > tell  us why at least one of your expenses has been claimed at 45p
  - display an error message if additional information is not provided when a user has increased the automatically calculated distance
    > tell us why you have increased the distance travelled for at least one of your expenses
* Update the form hint: "Provide any information that will help in the assessment of your expenses. For example, if you increase mileage on travel expenses, we need to know why."

#### Ticket
[Update additional information feature on travel expenses page](https://dsdmoj.atlassian.net/browse/CBO-483)

#### Why
The provision of the additional information will, more frequently, allow case workers to assess the claim during the initial assessment.  If they have to message and ask for the data, the assessments are delayed and it takes longer for the providers to receive their payments.

#### How
* Update the hint text and labelling
* Increase the validation
* Update the API to allow submission of additional information (but not yet validate)

#### TODO (wip)
- [x] Update the label (remove optional)
- [x] Update the hint text
- [x] Validate presence of info if provider has claimed higher rate mileage (web)
- [X] Validate presence of info if provider has increased calculated distance (web)
- [x] Add `travel_expense_additional_information` to claim endpoints on the API
- [x] Update release notes to alert Software vendors

### DIFF
--- a/app/validators/claim/base_claim_validator.rb
+++ b/app/validators/claim/base_claim_validator.rb
@@ -358,10 +358,10 @@ class Claim::BaseClaimValidator < BaseValidator
   end
 
   def has_higher_rate_mileage?
-    @record.expenses.map { |s| s.mileage_rate_id.eql?(2) }.any?
+    @record.expenses.where(mileage_rate_id: 2).any?
   end
 
   def increased_travel?
-    @record.expenses.map { |x| x.calculated_distance && (x.distance > x.calculated_distance) }.any?
+    @record.expenses.where('calculated_distance IS NOT NULL AND (distance > calculated_distance)').any?
   end
 end
--- a/app/validators/claim/base_claim_validator.rb
+++ b/app/validators/claim/base_claim_validator.rb
@@ -362,6 +362,6 @@ class Claim::BaseClaimValidator < BaseValidator
   end
 
   def increased_travel?
-    @record.expenses.where('calculated_distance IS NOT NULL AND (distance > calculated_distance)').any?
+    @record.expenses.where('distance > calculated_distance').any?
   end
 end
--- a/app/validators/claim/base_claim_validator.rb
+++ b/app/validators/claim/base_claim_validator.rb
@@ -291,6 +291,13 @@ class Claim::BaseClaimValidator < BaseValidator
     add_error(:retrial_actual_length, 'too_long')
   end
 
+  def validate_travel_expense_additional_information
+    return if @record.from_api?
+    return unless @record.expenses.any?
+    validate_presence(:travel_expense_additional_information, 'higher_rate_travel_claimed') if has_higher_rate_mileage?
+    validate_presence(:travel_expense_additional_information, 'calculated_travel_increased') if increased_travel?
+  end
+
   def actual_length_consistent?(requires_dates, actual_length, start_date, end_date)
     requires_dates &&
       actual_length.present? &&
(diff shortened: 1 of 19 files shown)


app/validators/claim/base_claim_validator.rb
```diff
@@ -349,4 +356,12 @@ def validate_retrial_start_and_end(start_attribute, end_attribute, inverse = fal
   def validate_too_far_in_past(start_attribute)
     validate_on_or_after(Settings.earliest_permitted_date, start_attribute, 'check_not_too_far_in_past')
   end
+
+  def has_higher_rate_mileage?
+    @record.expenses.map { |s| s.mileage_rate_id.eql?(2) }.any?
+  end
+
+  def increased_travel?
+    @record.expenses.map { |x| x.calculated_distance && (x.distance > x.calculated_distance) }.any?
```
jsugarman: not a blocker - could use a `.find`  - automatically breaks when first `true` result found. 

This will return true or nil tho.
```
@record.expenses.find { |x| x.calculated_distance && (x.distance > x.calculated_distance) }
```

jsugarman: i have not this tested either

config/locales/en.yml
```diff
@@ -846,8 +846,9 @@ en:
           remove: "Remove disbursement"
       expenses:
         textarea_additional_info:
-          additional_info_heading: Additional information (optional)
-          additional_info_label: Provide any information that will help in the assessment of your expenses
+          additional_info_heading: Additional information
+          agfs_additional_info_label: 'Provide any information that will help in the assessment of your expenses. For example, if you claim the private mileage rate on travel expenses, we need to know why.'
```
jsugarman: not a blocker: for multi-line translations can use | (pipe) and single quotes are default:
```
   agfs_additional_info_label: |
     Provide any information that will help in the assessment of your expenses. For 
     example, if you claim the private mileage rate on travel expenses, we need to 
     know why.
```

config/locales/en.yml
```diff
@@ -846,8 +846,9 @@ en:
...
+          lgfs_additional_info_label: 'Provide any information that will help in the assessment of your expenses. For example, if you claim the private mileage rate or increase the mileage on travel expenses, we need to know why.'
```
jsugarman: as above

jsugarman (approved):
👍 Looks good, just a couple of minor suggestions

app/validators/claim/base_claim_validator.rb
colinbruce: All tests pass with the change.  Also `Benchmark.measure` makes it (a fraction of a second) faster, so I'll swap this in.

jsugarman: 🥇

colinbruce: Hmm, looking at the `has_higher_rate_mileage` above, that can be replaced with 
```ruby
claim.expenses.where(mileage_rate_id: 2).any?
```

Revisiting both of these...

colinbruce: Both significantly faster by removing `.map` and replacing with simple `.where` clauses

app/validators/claim/base_claim_validator.rb
```diff
@@ -358,10 +358,10 @@ def validate_too_far_in_past(start_attribute)
   end
 
   def has_higher_rate_mileage?
-    @record.expenses.map { |s| s.mileage_rate_id.eql?(2) }.any?
+    @record.expenses.where(mileage_rate_id: 2).any?
   end
 
   def increased_travel?
-    @record.expenses.map { |x| x.calculated_distance && (x.distance > x.calculated_distance) }.any?
+    @record.expenses.where('calculated_distance IS NOT NULL AND (distance > calculated_distance)').any?
```
jsugarman: For SQL you might be able (he says over-confidently)  to get away with
```
where('distance > calculated_distance').any?
```
...as any expression with values evaluating to null in should return null, and therefore false

but either way 👍

colinbruce: Good call, rebuilding, then I'll merge

PR #2587, opened 2018-12-05
Replace Travel information validation

colinbruce:
#### What
Make the new additional-travel-information validation work

#### Why
The map methods were swapped out for `where` queries that failed
because the objects hadn't been persisted to the database yet.

#### How
Restoring to find queries that are quicker than map, but work on the
objects in memory


jsugarman (approved):
👍 💥

PR #2588, opened 2018-12-05
Fix for travel automation fix

colinbruce:
#### What
Fix for the previous fix for validating additional travel information introduced a regression failure

#### Why
Users were creating a car_travel expense, duplicating it, and changing it to parking.

#### How
This requires checking for the presence of both `distance` and `calculated_distance`.  Using safe navigation operators with numeric comparisons is blocked, so 
```
x&.distance > x&.calculated_distance
``` 
will generate new and exciting errors :(


jsugarman (approved):
👍

PR #2589, opened 2018-12-06
CBO-515: travel automation status designs

colinbruce:
#### What
Provide better guidance for caseworkers on when to assess travel expenses

#### Ticket
[CCCD: Implement Travel automation design ticket](https://dsdmoj.atlassian.net/browse/CBO-515)

#### Why
There has been some confusion around the need to process all travel expenses, this implements the designs in ticket CBO-484 that have been designed to reduce the cognitive load of case workers.  It endeavours to highlight only those travel expenses that need caseworker assessment

#### TODO (wip)

 - [X] Implement the designs from [CCCD: Create design for travel item assessment status](https://dsdmoj.atlassian.net/browse/CBO-484)
 - [x] Check edge cases in testing to ensure coverage of expense duplication issues

### DIFF
--- a/app/presenters/expense_presenter.rb
+++ b/app/presenters/expense_presenter.rb
@@ -76,4 +76,22 @@ class ExpensePresenter < BasePresenter
       true
     end
   end
+
+  def state
+    if expense.mileage_rate_id&.eql?(1) &&
+       expense.calculated_distance.present? &&
+       (expense.distance <= expense.calculated_distance)
+      'Accepted'
+    else
+      'Unverified'
+    end
+  end
+
+  def distance_label
+    if expense.calculated_distance.present? && (expense.distance <= expense.calculated_distance)
+      '.distance'
+    else
+      '.distance_claimed'
+    end
+  end
 end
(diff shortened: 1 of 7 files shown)


app/presenters/expense_presenter.rb
```diff
@@ -76,4 +76,22 @@ def show_map_link?
       true
     end
   end
+
+  def state
+    if expense.mileage_rate_id&.eql?(1) &&
+       expense.calculated_distance.present? &&
+       (expense.distance <= expense.calculated_distance)
```
jsugarman: does this handle the previously encountered nil value comparison problems?

jsugarman: perhaps also a method just for `expense.calculated_distance.present? && expense.distance <= expense.calculated_distance`
like `distance_acceptable?` as it is used below too.

app/presenters/expense_presenter.rb
```diff
@@ -76,4 +76,22 @@ def show_map_link?
...
+      'Accepted'
+    else
+      'Unverified'
+    end
+  end
+
+  def distance_label
```
jsugarman, line 88: just naming but i thought this was a CCS class at first - perhaps a `distance_t`,`distance_text` or something to reflect it is a translation - 

do you not need to call I18n.t directly?

jsugarman (approved):
👍 cool

PR #2591, opened 2018-12-06
Update the MI transformation to export PPE correctly

colinbruce:
#### What
Improve the MI data export prior to deletion

#### Why
Previously the code only looked at FeeTypes `BAPPE` that is used by AGFS claims only.  Therefore all 

#### How
This enhancement now extracts the quantity from the appropriate LGFS fee by type as well as still reading `BAPPE` from AGFS


jsugarman (approved):
👍

PR #2592, opened 2018-12-07
Update InjectionResponseService

colinbruce:
#### What
Only send slack notification when injection fails

#### Ticket
[Remove success from the injection channels](https://dsdmoj.atlassian.net/browse/CBO-508)

#### Why
As a product manager, or any other team member with access to the Slack channels, I want:
1. to enter the CCR and CCLF data injection slack channels and see only Katy Perry.

#### How
Add an `unless successful` check to the slack notification call


jsugarman (approved):
👍

PR #2593, opened 2018-12-10
CBO-517: Enhance private milage travel validation

colinbruce:
#### What
When visiting magistrates courts and prisons, providers should not have to provide reasons for claiming higher rate mileage

#### Ticket
[CCCD: Additional travel information - iteration 2](https://dsdmoj.atlassian.net/browse/CBO-517)

#### Why
This is a known exemption to the higher rate milage rules that was missed from the original validation ticket

#### How
Add new methods to the `Expense` model to get the, optionally, linked `Establishment` and it's `Category` and allow the validation to check the values while looping through the, about to be created, expenses

### DIFF
--- a/app/models/expense.rb
+++ b/app/models/expense.rb
@@ -146,6 +146,15 @@ class Expense < ApplicationRecord
     distance > calculated_distance
   end
 
+  def establishment
+    return unless expense_type.car_travel?
+    Establishment.find_by(name: location)
+  end
+
+  def establishment_category
+    establishment&.category
+  end
+
   private
 
   # we only calculate VAT for AGFS claims for vatable providers.  On LGFS claims, the VAT amount is entered in the form.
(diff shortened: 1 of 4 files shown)


app/validators/claim/base_claim_validator.rb
```diff
@@ -358,7 +358,8 @@ def validate_too_far_in_past(start_attribute)
   end
 
   def has_higher_rate_mileage?
-    @record.expenses.find { |x| x.mileage_rate_id.eql?(2) }
+    exclude_types = %w[magistrates_court prison]
```
jsugarman: as discussed, `destinations.exclude?(blah)` is clearer

jsugarman (approved):
👍 Just the one suggestion as discussed

app/models/expense.rb
```diff
@@ -146,6 +146,15 @@ def distance_gt_calculated?
     distance > calculated_distance
   end
 
+  def establishment
+    return unless expense_type.car_travel?
+    Establishment.find_by(name: location)
+  end
+
+  def establishment_category
```
jsugarman: just thinking. do we need this as the establishment method mimic a has_one relationship so:

```
expense.establishment&.category
```

PR #2595, opened 2018-12-11
FIX: Restrict travel expense statuses to LGFS claims

colinbruce:
#### What
Restrict travel expense statuses to LGFS claims

#### Ticket
N/A

#### Why
The travel automation change only apply to LGFS currently.  Adding status flags to AGFS expenses is causing confusion for case workers

#### How
Restrict the display in the view by checking the claim type using `claim.lgfs?`


jsugarman (approved):
👍

PR #2596, opened 2018-12-11
Fix styling on transfer claim disbursements

jsugarman:
#### What
Fix styling on transfer claim disbursements - see screenshot

#### Ticket

none

#### Why
**before:** 
[
screen shot 2018-12-11 at 16 35 15
](url)
Noticed that following the FE design changes
for interim fee calc, including for disbursements
the transfer claims disbursements form step
was not using the new stylesheets, causing the above.

#### How
**after:** 
screen shot 2018-12-11 at 16 40 54

modify disbursement form step for transfer claims to use
the new stylesheets.

#### TODO:

shouls really be sharing this at a lower level but relative paths using a dynamic
current_step means this would make this fix more involved


colinbruce (approved):
LGTM :+1:

PR #2597, opened 2018-12-12
Confirm Scheme 11 dates

colinbruce:
#### What
Update scheme 11 dates to reflect the SI

#### Why
This will allow us to deploy to api-sandbox and let vendors test

### DIFF
--- a/app/models/fee_scheme.rb
+++ b/app/models/fee_scheme.rb
@@ -16,7 +16,7 @@ class FeeScheme < ApplicationRecord
     where('(:now BETWEEN start_date AND end_date) OR (start_date < :now AND end_date IS NULL)', now: Time.zone.now)
   }
   scope :for, lambda { |check_date|
-    where('(:date BETWEEN start_date AND end_date) OR (start_date < :date AND end_date IS NULL)', date: check_date)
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍 had just done this myself - the sandbox date is what i suggested in slack channel too so good to go

app/models/fee_scheme.rb
```diff
@@ -16,7 +16,7 @@ class FeeScheme < ApplicationRecord
     where('(:now BETWEEN start_date AND end_date) OR (start_date < :now AND end_date IS NULL)', now: Time.zone.now)
   }
   scope :for, lambda { |check_date|
-    where('(:date BETWEEN start_date AND end_date) OR (start_date < :date AND end_date IS NULL)', date: check_date)
+    where('(:date BETWEEN start_date AND end_date) OR (start_date <= :date AND end_date IS NULL)', date: check_date)
```
jsugarman, line 19: 👍 makes sense - could use coalesce on the between but I would not for now as have not tested.

```
scope :for, lambda { |check_date|
    where('(:date BETWEEN start_date AND COALESCE(end_date, now())', date: check_date)
```

PR #2601, opened 2018-12-14
Fix count of claims in claim list

jsugarman:
#### What
Fix count of claims in claim list

#### Ticket
support and maintenance

#### Why
Seems to have been caused by rails 5.1 upgrade
where the alias used in the order by clause is
stripped when applying a count/size activerecord
operation on the statement.

#### How
Still looking for reason it has failed now but
by unscoping the order (which is irrelevant) on
the count operation it fixes it.


colinbruce (approved):
LGTM, as long as the staging test works

PR #2602, opened 2018-12-14
Fix case assessment reasons

jsugarman:
#### What
Fix case assessment refusal reason text feature 

#### Ticket

support and maintenance

#### Why
Following upgrade to rails 5.1 reports were received of 
case workers being unable to provide reasons for when
specifying 'other' refusal reasons.

It appears rails' collection_check_boxes
form helpers have changed to not put an underscore
in front of check box ids generated for
nil objects. This means the dynamically
generated ids:
```
'#_state_reason_other',
'#_state_reason_other_refuse
```

have become
```
'#state_reason_other'
'#state_reason_other_refuse'
```
and the javascript that relies on these
broke as a result.

#### How
change the ids used in the JS and write a regression feature test
to cover the 'other' reason text caseworker assessment flow.


colinbruce (approved):
LGTM :+1:

PR #2606, opened 2018-12-20
Change fee scheme specs to use seed helper

jsugarman:
#### What
Attempting to resolve flakey tests relating
to fee schemes.

#### Why
Specs that rely on fee schemes directly or indirectly
are occasionally failing.

#### How
Using seed helper only to create fee schemes rather than rely
on factories. Also remove any before(:all) hooks as these may 
not get cleaned up successfully and result in fee schemes persisting
between specs.


jsugarman:
cherry picked this commit into https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2615

PR #2608, opened 2018-12-20
Partially fix DB export

colinbruce:
#### What
Exclude attributes from export

#### Why
It was broken

#### How
Exclude arel output if the attribute is not a column


jsugarman (approved):
Cool 👍

PR #2612, opened 2018-12-21
Remove reason_text from expenses

colinbruce:
#### What
Add a clear reason text method and call it before validation

#### Why
This reduces the errors caused by duplicating expenses with other
text reasons that don't get removed when the type of the duplicated
expense is changed

#### How
The method removes data from the `reason_text` field after being called from the `before_validation` block

### DIFF
--- a/spec/validators/expense_validator_spec.rb
+++ b/spec/validators/expense_validator_spec.rb
@@ -311,20 +311,23 @@ RSpec.describe 'ExpenseValidator', type: :validator do
     end
 
     context 'validates absence when reason ID is other than 5 regardless of the reason set' do
+      subject { expense.valid? }
+
       before do
         expense.reason_id = 3
       end
 
-      it 'reason text is present' do
-        expense.reason_text = 'blah'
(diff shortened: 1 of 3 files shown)


spec/validators/expense_validator_spec.rb
```diff
@@ -311,20 +311,23 @@
...
-      it 'reason text is present' do
-        expense.reason_text = 'blah'
-        expense.valid?
-        expect(expense).not_to be_valid
-        expect(expense.errors[:reason_text]).to include('invalid')
+      context 'reason text is present' do
+        before { expense.reason_text = 'blah' }
+
+        it 'removes the reason text before validation' do
+          expect(subject).to be true
```
jsugarman, line 324: not a blocker - could use `is_expected.to`, as you have done below

jsugarman (approved):
LGTM 👍 - minor rspec tweak. do we have an expense duplication feature. would be good to have a scenario that duplicates this to exercise the JS IMO

PR #2614, opened 2018-12-27
CBO-352: fix archiving of purged claims

colinbruce:
#### What
Update the BaseClaim to destroy injection_attempts, to allow the transitioner to remove injection_attempts when running the overnight purge of archived claims

#### Why
The overnight archive job was failing on claims that had been injected

#### How
Set up a referential integrity link on the `BaseClaim` and `InjectionAttempt` by adding a destroy dependency


jsugarman (approved):
👍

PR #2615, opened 2019-01-02
Attempt to fix fee scheme spec flickers

jsugarman:
#### What
Amend factories and specs to tidy fee scheme logic

#### Why
Flickering specs related to fee schemes, particularly
`fee_scheme_spec` and `dropdown_data_spec`. plus
factories should reflect actual fee scheme dates
etc as much as possible.

#### How
- synchronise factory and seed data for fee scheme
- using seed helper instead of factories

### DIFF
--- a/spec/views/case_workers/claims/show_spec.rb
+++ b/spec/views/case_workers/claims/show_spec.rb
@@ -1,10 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe 'case_workers/claims/show.html.haml', type: :view do
-  let!(:lgfs_scheme_nine) { FeeScheme.find_by(name: 'LGFS', version: 9) || create(:fee_scheme, :lgfs) }
-  let!(:agfs_scheme_nine) { FeeScheme.find_by(name: 'AGFS', version: 9) || create(:fee_scheme, :agfs_nine) }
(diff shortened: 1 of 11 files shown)


spec/views/case_workers/claims/show_spec.rb
```diff
@@ -1,10 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe 'case_workers/claims/show.html.haml', type: :view do
-  let!(:lgfs_scheme_nine) { FeeScheme.find_by(name: 'LGFS', version: 9) || create(:fee_scheme, :lgfs_nine) }
-  let!(:agfs_scheme_nine) { FeeScheme.find_by(name: 'AGFS', version: 9) || create(:fee_scheme, :agfs_nine) }
-  let!(:agfs_scheme_ten) { FeeScheme.find_by(name: 'AGFS', version: 10) || create(:fee_scheme) }
-
   before do
```
colinbruce, line 4: You have deleted the manual creation but not re-seeded the FeeSchemes. 😕

colinbruce (approved):
One observation. but generally LGTM :+1:

spec/views/case_workers/claims/show_spec.rb
jsugarman: yeah not needed at all for the spec

PR #2616, opened 2019-01-03
Cbo 471 lgfs blank fields

kmahern:
#### What
As an LGFS provider
I do not want to be displayed 'first day of trial' and 'trial concluded on' on the case details summary
So that I am not shown unnecessary blank fields

#### Ticket
https://dsdmoj.atlassian.net/projects/CBO/issues/CBO-471?filter=allopenissues&orderby=priority%20DESC

#### Why
To avoid confusion?

#### How
Added a function hide_blank_litigator_fields? in presenter classes

### DIFF
--- a/features/claims/litigator/litigator_trial_claim_draft_submit.feature
+++ b/features/claims/litigator/litigator_trial_claim_draft_submit.feature
@@ -93,6 +93,13 @@ Feature: Litigator partially fills out a draft final fee claim, then later edits
 
     And I click Submit to LAA
     Then I should be on the check your claim page
+    And I should see the field 'Crown court' with value 'Blackfriars'
+    And I should see the field 'Case type' with value 'Trial'
+    And I should see the field 'Date case concluded' with value '01/04/2016'
+    And I should see the field 'Actual trial length' with value '1'
+    And I should not see 'First day of trial'
+    And I should not see 'Estimated trial length'
+    And I should not see 'Trial concluded on'
 
     When I click "Continue"
     Then I should be on the certification page
--- a/features/claims/litigator/litigator_trial_claim_draft_submit.feature
+++ b/features/claims/litigator/litigator_trial_claim_draft_submit.feature
@@ -93,10 +93,14 @@ Feature: Litigator partially fills out a draft final fee claim, then later edits
 
     And I click Submit to LAA
     Then I should be on the check your claim page
-    And I should see the field 'Crown court' with value 'Blackfriars'
-    And I should see the field 'Case type' with value 'Trial'
-    And I should see the field 'Date case concluded' with value '01/04/2016'
-    And I should see the field 'Actual trial length' with value '1'
+    And I should see the field 'Crown court' with value 'Blackfriars' in 'Case details'
+    And I should see the field 'Case type' with value 'Trial' in 'Case details'
+    And I should see the field 'Date case concluded' with value '01/04/2016' in 'Case details'
+    And I should see the field 'Type of fee' with value 'Trial' in 'Graduated fee'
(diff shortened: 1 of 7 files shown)


jsugarman (changes requested):
Unfortunately this ticket involves an area with a lot of tech debt and could do with a alot of refactoring. As a minimal solution i would instead suggest just using existing model methods in the view rather than creating generically named but actually quite specific methods, which could lead to confusion in the future. 

It would be good to extend the feature tests to cover this logic in the view too.

e.g. taken from `features/claims/litigator/interim_trial_claim_draft_submit.feature`

```
  Then I should be on the check your claim page
    And I should see 'Blackfriars'
    And I should see 'A20161234'
    And I should see 'Trial'

    And I should see 'Handling stolen goods'
    And I should see 'G: Other offences of dishonesty between £30,001 and £100,000'

    And I should see 'First day of hearing'

```

And, as discussed, we could apply requires_[re]trial_dates? false, on the presenter to handle this:
1. feature test to cover litigator final CYC claim
2. feature test to cover litigator fixed CYC page
3. presenter change and spec
4. remove the redundant view conditional logic

features/step_definitions/common_steps.rb
```diff
@@ -158,6 +158,10 @@
   expect(page).to have_content(arg1)
 end
 
+And(/^I should see the field '(.*?)' with value '(.*?)'$/) do |field, value|
```
jsugarman: 👍 nice

jsugarman (commented):
Looks great. 

Would you be able to extend the `Claim-for-Crown-Court-Defence/features/claims/litigator/litigator_trial_claim_draft_submit.feature` to cover the Check your claim page too as you have done with the contempt features. The reason is a litigator trial claim is also  a litigator claim, but  a "graduated" fee variety. It therefore shares a lot of the same logic but the CYC page is/should be slightly different.

features/claims/litigator/litigator_trial_claim_draft_submit.feature
```diff
@@ -93,6 +93,13 @@ Feature: Litigator partially fills out a draft final fee claim, then later edits
 
     And I click Submit to LAA
     Then I should be on the check your claim page
+    And I should see the field 'Crown court' with value 'Blackfriars'
+    And I should see the field 'Case type' with value 'Trial'
+    And I should see the field 'Date case concluded' with value '01/04/2016'
+    And I should see the field 'Actual trial length' with value '1'
+    And I should not see 'First day of trial'
+    And I should not see 'Estimated trial length'
+    And I should not see 'Trial concluded on'
```
jsugarman, line 106: In addition, to test the cyc page logic for graduated fees you could add

```
And I should see the field 'Type of fee' with value 'Trial'
And I should see the field 'First day of hearing' with value '01/01/2018'
And I should see the field 'Actual trial length' with value '1 day'
And I should see the field 'Total pages of evidence' with value '51'
And I should see the field 'Net amount' with value '£437.89'
```

jsugarman: You could even write a graduated fee section step test, specifically to test it appears in the correct section

have a look at the `within` capybara helper

features/step_definitions/common_steps.rb
```diff
@@ -158,6 +158,12 @@
   expect(page).to have_content(arg1)
 end
 
+And(/^I should see the field '(.*?)' with value '(.*?)' in '(.*?)'$/) do |field, value, section|
```
jsugarman, line 161: ace 🥇

jsugarman (approved):

PR #2618, opened 2019-01-03
CBO-381: DAC tables semantic review & translations

cleargif:
#### What
Tweaking tables, captions and summaries

#### Ticket

[CBO-381](https://dsdmoj.atlassian.net/browse/CBO-381)

#### Why
DAC Report


colinbruce (commented):
I'm a bit unsure why sometimes you are removing `aria` tags and sometimes adding them but, otherwise, looking good so far.

:+1:

cleargif:
Because the tables are simple, `th, td` - Voice over reads it out correctly. The additional `aria-label` is read out as well. `"Case number, Case number Txxxx..."`

colinbruce (approved):
Looking good :+1:

PR #2620, opened 2019-01-14
CBO-463: Improve sandbox provider creation

colinbruce:
#### What
Update the create_user regex

#### Ticket
[raise error for non-unique provider reference - api-sandbox](https://dsdmoj.atlassian.net/browse/CBO-463)

#### Why
The previous version was failing because of the caret(^)

#### How
This explicitly looks for the 3 digit numbers used by software vendors
on the api sandbox.  As the next line expects an integer, this seems
a safe proposal


jsugarman (approved):
looks good 👍

PR #2582, opened 2018-11-28
Agfs fixed fee checkboxes

jsugarman:
#### What
AGFS fixed fee UI improvements

#### Ticket

[board ticket description](link)

#### Why
Follow new designs

#### How
Instatiate eligible fixed fees in controller and present each one under its checkbox fee type

--------

#### TODO (wip)

 - [x] persistence problems
    - [x] destroy when nil value
    - [x] destroy when not applicable to case type?
 - [x] fee calc rewiring
 - [x] error message rewiring
 - [x] change case type journey

PR #2625, opened 2019-01-17
Agfs fixed fee checkboxes tidy

jsugarman:
#### What 

tidy of [agfs fixed fee checkboxes](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2582)


app/models/claim/advocate_claim.rb
```diff
@@ -72,7 +72,7 @@ class AdvocateClaim < BaseClaim
              dependent: :destroy,
              inverse_of: :claim,
              validate: proc { |claim| claim.step_validation_required?(:basic_fees) }
-    has_many :fixed_fees, -> { order(:created_at) },
```
jsugarman, line 75: **not a blocker** but @colinbruce this is the other one i meant to talk to you about. There is a `position` attribute on `fee_types` that could be used to control this in a more controlled/descriptive way.

app/models/claim/advocate_claim.rb
```diff
@@ -272,12 +272,24 @@ def instantiate_basic_fees
...
     end
 
+    def clear_basic_fees
+      basic_fees.map(&:clear) unless basic_fees.empty?
+    end
+
+    def destroy_ineligbile_fixed_fees
+      eligbile_fees = Claims::FetchEligibleFixedFeeTypes.new(self).call
+      fixed_fees.each do |fee|
+        fixed_fees.delete(fee) unless eligbile_fees.map(&:code).include?(fee.fee_type_code)
```
jsugarman, line 289:
```
fee.delete unless....
```
also: are we wanting to perform a delete instead of a destroy deliberately? [delete versus destroy](https://stackoverflow.com/questions/22757450/difference-between-destroy-and-delete)

```
# ensures callbacks are handled (dates_attended?)
fee.destroy unless...
```

colinbruce: It wasn't a conscious decision on my part, but I think the version without callbacks is preferable... happy to discuss...

app/models/claim/advocate_claim.rb
colinbruce: Agreed.  I think this change is a hangover of an attempt to sequence in-memory objects.  As long as the `position` field doesn't affect the rendering on the page it should be fine.

app/models/claim/advocate_claim.rb
jsugarman: it looks like `association.delete(child)` delegates its strategy to the associations `:dependant` option (destroy in our case) . so `fixed_fees.delete(fee)` will call :destroy on the fee and therefore destroy on its children
[rails association delete](https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete)

so `fixed_fees.delete(fee)` is the same as `fixed_fee.destroy(fee)`. `fee.delete` deletes the fee but not its children, `fee.destoy` deletes its children but cached results are not removed so   calling `claim.fixed_fees.flat_map(&:dates_attended).size` fails but `claim.fixed_fees.flat_map(&:dates_attended).count` does not

so better to use `association.delete(child)` or if we want to be explicit and not delegate strategy to the relation then `association.destroy(child)`. delegation magic is probably the "rails" way to go. 

I found it interesting anyway, and have added specs for it.

colinbruce (approved):
👍

app/validators/claim/advocate_claim_validator.rb
```diff
@@ -81,7 +81,7 @@ def validate_defendant_uplifts_basic_fees
   def validate_defendant_uplifts_fixed_fees
     return if @record.from_api? || !@record.fixed_fee_case?
     return unless defendant_uplifts_greater_than?(defendant_uplifts_fixed_fees_counts, number_of_defendants)
-    add_error(:base, 'defendant_uplifts_fixed_fees_mismatch')
+    add_error(:fixed_fee_4_quantity, 'defendant_uplifts_fixed_fees_mismatch')
```
jsugarman, line 84: just noting that there is a bug with this behaviour since it links to the 4th fixed fee regardless of whether the fee is the defendant uplift or not. That fee is not always the fourth

screen shot 2019-01-18 at 16 23 16

will revert this change out after the fact.

PR #2627, opened 2019-01-18
Improve post-test-run tidy-up

colinbruce:
#### What
There were a large number of zip files in my tmp folder

#### Why
It seems that they were not being deleted after the test run

#### How
Now they are.


jsugarman (approved):
👍 are these from simplecov

colinbruce:
No, these are from the spec and feature tests of the case workers generating and downloading zip files!

PR #2628, opened 2019-01-21
Attach def uplift error to correct fee

jsugarman:
#### What
Fix bug related to error linking

#### Why
Came out of change from the fixed fee checkbox
enhancement

#### How
determine position of fee at runtime. Also
remove reload from instantiation to allow
errorneous fees to be displayed.


kmahern:
Code changes look good to me (test failing currently but Joel is investigating).

kmahern (approved):

cleargif:
👍

cleargif (approved):

PR #2629, opened 2019-01-21
Amend wording and grouping of feature step call

jsugarman:
#### What
Amend wording and grouping of feature step call

#### Why
For plainer english


kmahern (approved):
Looks good to me.

PR #2630, opened 2019-01-21
Fix page title sentry error

jsugarman:
#### What
Fix page title sentry error

#### Why
- translation mappings incorrect for lit final claim
- LGFS claims with earliest rep order **before** scheme 8 start date
    have no fee scheme, making page title translation throw an error.

#### Fixes
 [example sentry error](https://sentry.service.dsd.io/mojds/private-beta/issues/34365/)


cleargif (approved):
👍

PR #2631, opened 2019-01-23
Validate retrial start dates being on or after trial end dates

jsugarman:
#### What
Validate retrial start dates being on or after trial end dates

#### Ticket

[CBO-547](https://dsdmoj.atlassian.net/browse/CBO-547)

#### Why
As part of retrial fee calculation it was found that
alot of `retrial_started_at` dates were being entered with
a date earlier than the `trial_concluded_at`. This leads
to a negative retrial interval, which in turn, causes the
fee calculator to return a basic fee amount ("Advocate fee")
equivalent to the trial (as if the user was not claiming
a reduction on the retrial).

- the "negative retrial interval" could be a result of
  user error, or attempting to claim no reduction.
- in any event the data should be accurate and it should
  not be possible for a retrial to start before a trial ends

This validation has been put in place to enforce the
business/legislative rules AND to enable accurate
calculation of retrial reductions based on retrial
intervals

TODO: 
- write use an RSpec custom matcher for the validation spec(s) [tech debt]

### DIFF
--- a/app/validators/claim/base_claim_validator.rb
+++ b/app/validators/claim/base_claim_validator.rb
@@ -234,6 +234,7 @@ class Claim::BaseClaimValidator < BaseValidator
   # cannot be before earliest rep order date
   # cannot be more than 5 years in the past
   def validate_retrial_started_at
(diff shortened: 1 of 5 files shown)


colinbruce (approved):
:+1:

PR #2632, opened 2019-01-24
Performance platform tweaks

colinbruce:
#### What
After deploying, and testing, the Performance Platform code a number of issues were discovered

#### Why
The live PP datasets have stricter validation than the test ones, therefore we needed to be clearer with the `period` type and capitalisation of the channels


jsugarman (approved):
👍

PR #2634, opened 2019-01-28
Move logic around displaying days and Actual trial length to presenter

kmahern:
Created display_days? to determine whether Days or Actual trial length is displayed in caseworker view.

#### What

Added display_days? to presenter classes to determine whether Actual trial length or days is displayed in claim summary caseworker view for lgfs claims

#### Ticket

n/a (residual tech debt from 542)


jsugarman (changes requested):
you said you wanted to amend the specs for the transfer claim too. so will wait for that, but otherwise looks good

jsugarman (approved):
👍

PR #2635, opened 2019-01-28
Fix invalid rep order user flow

jsugarman:
#### What
Handle invalid combinations of offence and rep order dates (i.e. fee schemes)

#### Ticket

[CBO-540](https://dsdmoj.atlassian.net/browse/CBO-540)

#### Why
Can result in conflicting offence for fee scheme

- [X] refactor to user cleaner service
- [x] only apply validation to API submitted claims
- [x] propagate validation error only for API submissions


colinbruce (approved):
👍 LGTM

PR #2636, opened 2019-01-29
Remove duplicate method

jsugarman:
#### What
Remove duplicate `earliest_representation_order` method

#### Why
there were two but only the latter was ever being called


app/models/claim/base_claim.rb
```diff
@@ -382,10 +382,18 @@ def representation_orders
       defendants.reduce([]) { |a, e| a.concat(e.representation_orders) }
     end
 
+    # def earliest_representation_order
+    #   representation_orders.sort do |a, b|
+    #     (a.representation_order_date || 100.years.from_now) <=> (b.representation_order_date || 100.years.from_now)
+    #   end.first
+    # end
+
```
colinbruce: Can the commented out one be deleted?

colinbruce (approved):
One observation, otherwise looks good 👍

PR #2637, opened 2019-01-31
Modify database housekeeping

jsugarman:
#### What
Modify database housekeeping

#### Why
To prevent the need to remember to
to update this method when new tables
are added.

To reeduce likelihood of flickering
test resulting from cross spec contamination

To speed up the cleaning
of test data by after(:all) hooks, using
truncation strategy.


colinbruce (approved):
:+1:

PR #2641, opened 2019-02-06
Remove retrial/trial start/end validation temporarily

jsugarman:
#### What
Remove retrial/trial start/end validation temporarily

#### Why
To give vendors time to "fix" their systems
inline.

#### How
Disable the validation for claims submitted
from api

#### TODO (wip)

 - [X] conditionally remove validation
 - [ ] update API docs


jsugarman:
PM does not want to remove this for specific vendor api consumer bugs. Closing...

PR #2642, opened 2019-02-06
Add api release notes for retrial validation

jsugarman:
#### What
Add api release notes for retrial validation

#### Why
Missed from previous worked. One vendor has
discovered their app has a bug/feature which
falls foul of this validation.


colinbruce (approved):
👍

PR #2646, opened 2019-02-11
Update validation for Litigator claim

willmcb:
#### What
Validation on LGFS supplier number field on Claim for litigator fees page, meaning that providers were able to continue with their application without a supplier number being collected. 

#### Ticket
[CBO-569](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&projectKey=CBO&modal=detail&selectedIssue=CBO-569)

#### Why
Caseworkers were unable to assess claims 

#### How
Changed validation restriction so that it was called when a claim was in `draft?` rather than just when `LitigatorClaimsController#create` was called, as was the previous trigger.

### DIFF
--- a/features/claims/litigator/litigator_trial_claim_draft_submit.feature
+++ b/features/claims/litigator/litigator_trial_claim_draft_submit.feature
@@ -17,7 +17,7 @@ Feature: Litigator partially fills out a draft final fee claim, then later edits
     And I select the court 'Blackfriars'
     And I enter a case number of 'A20161234'
     And I enter the case concluded date '2016-04-01'
-    Then I click "Continue" in the claim form and move to the 'Case details' form page
+    When I click "Continue" I should see a "Choose a supplier number" error
 
     When I choose the supplier number '1A222Z'
     Then I click "Continue" in the claim form and move to the 'Defendant details' form page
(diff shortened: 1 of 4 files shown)


features/claims/litigator/litigator_trial_claim_draft_submit.feature
```diff
@@ -13,11 +13,13 @@ Feature: Litigator partially fills out a draft final fee claim, then later edits
 
     And I should see 3 supplier number radios
 
-    When I choose the supplier number '1A222Z'
     And I select a case type of 'Trial'
     And I select the court 'Blackfriars'
     And I enter a case number of 'A20161234'
     And I enter the case concluded date '2016-04-01'
+    Then I click "Continue" in the claim form and move to the 'Case details' form page
```
jsugarman: is this expecting an error? if so you could test for that. also should it be `move to the 'Defendants details' form page`?

willmcb: Yep, we're expecting an error and we will add in a test for it.

willmcb: @jsugarman see subsequent commit

jsugarman (approved):
👍 that step and the original really needs looking at, but the change is good

PR #2648, opened 2019-02-12
Update quarterly volume generation of stats

colinbruce:
#### What
Switch param conversion to Time
#### Why
Previously the code added and removed hours for BST/GMT conversion
#### How
This strips out the timezone conversion when switching to beginning_of_quarter

### DIFF
--- a/app/interfaces/api/v2/performance_platform/quarterly_volume.rb
+++ b/app/interfaces/api/v2/performance_platform/quarterly_volume.rb
@@ -57,7 +57,7 @@ module API
 
             desc 'Submit calculated totals for quarterly volume performance platform report'
             post do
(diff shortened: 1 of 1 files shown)


app/interfaces/api/v2/performance_platform/quarterly_volume.rb
```diff
@@ -57,7 +57,7 @@ def filename(start_date)
 
             desc 'Submit calculated totals for quarterly volume performance platform report'
             post do
-              start_of_quarter = Date.parse(params[:start_date]).beginning_of_month
+              start_of_quarter = Time.parse(params[:start_date]).beginning_of_quarter
```
adrianclay, line 60: Today I learnt a new ActiveSupport method!

jsugarman (approved):
👍 LGTM

PR #2650, opened 2019-02-13
TEST - Modify the way claim state is handled in factories

jsugarman:
Modify the way claim state is handled in factories

Advocate claim factories for claims in a particular
state are simply altering the state without ensuring the
claim is valid. This can result in invalid claims in
a (post) submitted - which should never be possible and
can result in specs and features throwing invalid transition
errors unexpectedly.


jsugarman:
Closing...to do at another time as quite a few other specs need rejigging

PR #2654, opened 2019-02-20
Prevent supplmenetary claims with no miscellaneous fees

jsugarman:
#### What
Prevent supplementary claims with no miscellaneous fees
and calculate totals for such fees on the backend too

#### Ticket

[CBO-595](https://dsdmoj.atlassian.net/browse/CBO-595)

#### Why
Backend totals calculation required for API submissions in any
event. Plus a supplementary claim should(??) have a miscellaneous
fee value


willmcb:
Checked as a user would, and ran specs - everything works as expected 👍

willmcb (approved):

PR #2656, opened 2019-02-20
Change javascript driver to selenium headless chrome

jsugarman:
#### What
Change default javascript driver from poltergeist/phantomJS to selenium/headless-chrome

#### Why
PhantomJS (the headless browser used by poltergeist) is no longer supported
and is not as accurate a reflection of browser behaviour in any event.

[phantomJS desupport announcement](https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE)

#### How
- Register driver and implement as default
- implement a save as screenshot alternative
- configure travis to use google-chrome
- fix tests that were not reflecting behaviour


colinbruce (approved):
Nice :+1:

PR #2657, opened 2019-02-22
Update vcr chromedriver handling

colinbruce:
#### What
Add additional ports to the vcr ignore range

#### Why
If the chromedriver is cancelled or crashes, it can be left running as a background thread.  The next time features are started it increments the port number by one.

#### How
Check for a 10 sequential ports and ignore them.  After this users should probably be killing the zombie processes anyway!


jsugarman (approved):
🥇 cool. i was looking at fixing the port or finding it but i think this is simpler and clearer

PR #2659, opened 2019-02-22
Update ruby to 2.5.3

colinbruce:
#### What
Bump the ruby version form 2.5.1 to 2.5.3

#### Why
This is because the CircleCI standard build environment for 2.5.1 includes an outdated chrome version, and we need to update to CircleCI to take advantage of parallel test runs and kubernetes integration


jsugarman (approved):
👍 great!

PR #2663, opened 2019-02-27
Update JSON uploader claim endpoint use

jsugarman:
#### What
Change the JSON uploader to use the new advocates/final API claim endpoint
and remove a couple of uploader code files that were empty.

#### Why
The api/external_user/claims endpoint has been deprecated in
favour of api/external_users/advocates/final


colinbruce (approved):
Simples! :+1:

PR #2665, opened 2019-02-28
Add or_one method to NilClass

jsugarman:
#### What
Fixes sentry error related to calling an integer extension on nil

#### Why
Some actual trial lengths and other similar field values
are getting away with nil values which are then raising
NilClass errors on when CCR/CCLF calls the api endpoint
for DI.

[example sentry error](https://sentry.service.dsd.io/mojds/private-beta/issues/34897/)

#### How
default nil to 1 as well using an extension of nilClass itself

### DIFF
--- a/config/initializers/00_extensions.rb
+++ b/config/initializers/00_extensions.rb
@@ -45,3 +45,7 @@ end
 class Integer
   include IntegerExtension
 end
+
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
:+1:

PR #2669, opened 2019-03-04
Add missing agfs reform offences

jsugarman:
#### What
Add driving offences for AGFS scheme 10 and 11. Investigate
and add 'Other offences' for all (or just 17.1??) offence
bands.

#### Ticket

[CBO-597](https://dsdmoj.atlassian.net/browse/CBO-597)

#### Why
"Aiding, abetting, causing or permitting dangerous driving"
is missing for scheme 10 and 11 and is a common offence claimed
for.

In addition, catchall offences are needed for those not
explicitly covered. CCR has 'Other offences' for all bands

#### How

Add helper library class for adding offences generally and
use to create a rake task for adding new individual offences.

Add a rake task to add the missing 'Other offences' offences.

--------

#### TODO (wip)

 - [X] helper class for adding individual offence
 - [X] rake task to add individual offences
 - [X] rake task to wrap adding all 'Other offences'
 - [X] amend api to enable adding offences using offences.unique_code - [separate PR](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2672)


lib/data_migrator/offence_adder.rb
```diff
@@ -0,0 +1,106 @@
...
+      @fee_scheme = fee_scheme
+      @description = description
+      @offence_band = offence_band
+      @contrary = contrary
+      @year_chapter = year_chapter
+    end
+
+    def up
+      offence = find_or_create_offence!(attributes)
+      puts "-- found or created offence (#{offence.attributes})"
```
willmcb, line 31: Why have this as one method? If you wanted to check that an offence existed I think it would be hard with this as it would create it or find it. 

Could it return a message like `offence already exists`?

lib/data_migrator/offence_adder.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    def down
```
willmcb, line 34: Why `up` and `down`? Is this a traditional thing. Why not `add` and `remove`?

willmcb (approved):
Comments made for my benefit really. Rather than blocking the code.

kmahern:
Just a high level comment from me, but I am happy to approve. It would be great if we had some kind of readme/wiki/confluence page to describe the process of adding an offence to CCCD, anything to be aware of and any potential pitfalls.

kmahern (approved):
Looks good 👍

lib/data_migrator/offence_adder.rb
jsugarman: yeah, stand rails migration convention for schema changes

PR #2670, opened 2019-03-05
CBO-599: Autocomplete polyfill

cleargif:
#### What
A prototype to see if it will resolve the edge case issues with the auto complete

Branch also contains a `sleep` in the flickering feature

#### Ticket
CBO-599


cleargif:
This was a prototype. Closing.

PR #2672, opened 2019-03-06
Add offence unique code filter to API

jsugarman:
#### What
To expose and enable filtering offences using unique_code

#### Ticket

[CBO-624](https://dsdmoj.atlassian.net/browse/CBO-624)

#### Why
First step toward ensuring API consumers (software vendors) use 
offence.unique_code to specify the offences on a claim endpoint,
rather than the offence ID. We should not be exposing internal IDs
as they cannot be guaranteed to remain the same.

#### How
- expose unique code on `api/offences`
- add unique code filtering option on `api/offences`

#### Separate ticket
- add offence_unique_code param to all claim endpoints
- deprecate and eventually remove offence_id param on claim endpoints

### DIFF
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -80,7 +80,8 @@ module API
                      desc: 'OPTIONAL: Filter by offence description'
             optional :rep_order_date,
                      type: String,
-                     desc: 'OPTIONAL: Filter by scheme date based on representation order - format YYYY-MM-DD',
+                     default: '2016-04-01',
+                     desc: I18n.t('api.v1.dropdown_data.offences.params.rep_order_date'),
                      standard_json_format: true
             optional :unique_code,
                      type: String,
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -75,17 +75,24 @@ module API
         resource :offences do
           desc 'Return all Offence-ids to be used in advocate claims (see OffenceClasses for Litigator claims).'
           params do
-            optional :offence_description, type: String, desc: 'Offences matching description'
+            optional :offence_description,
+                     type: String,
+                     desc: 'OPTIONAL: Filter by offence description'
             optional :rep_order_date,
                      type: String,
-                     desc: 'OPTIONAL: Date of representation order in YYYY-MM-DD',
+                     desc: 'OPTIONAL: Filter by scheme date based on representation order - format YYYY-MM-DD',
                      standard_json_format: true
+            optional :unique_code,
+                     type: String,
+                     desc: 'OPTIONAL: Filter by offence unique code'
           end
           get do
-            scheme_date = params[:rep_order_date] || '2018-01-01'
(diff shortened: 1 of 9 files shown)


app/views/pages/api_release_notes.html.haml
```diff
@@ -9,6 +9,43 @@
...
+      To facilitate the lookup of offences the unique identifiers that
+      can be relied on not to change (i.e.
+      %code.code='unique_code'
+      ) have been exposed.
+      %br
+      %br
+      .panel
+        Note that an offence's
+        %code.code='id'
+        should not be persisted in consumers and relied on to add a specific offence to a claim. The reason for this is that internal id's could change unexpectedly. We will, however, endeavour to not change these for historical reasons. Instead an offence should be queried on this endpoint using the
```
kmahern: A cosmetic thing, but this line is very long!

spec/api/v1/dropdown_data_spec.rb
```diff
@@ -81,6 +81,7 @@
 
     it "should return a JSON formatted list of the required information" do
       results.each do |endpoint, json|
+        ap "#{endpoint}"
```
willmcb: Is this due to a flickering test?

kmahern (approved):
lgtm 👍

app/interfaces/api/v1/dropdown_data.rb
```diff
@@ -75,17 +75,25 @@ def unique_code
         resource :offences do
           desc 'Return all Offence-ids to be used in advocate claims (see OffenceClasses for Litigator claims).'
           params do
```
willmcb, line 77: This all looks fine to me, I guess this is the way that you define a new resource using Grape. Would you be able to go through this bit of code with me though so I understand what is going on? Thanks.

willmcb (changes requested):
Just that I think you may have left a `ap` print statement in.

See `dropdown_data_spec.rb` line 84

spec/api/v1/dropdown_data_spec.rb
jsugarman: oops, good catch - debug statement left in 👍

app/interfaces/api/v1/dropdown_data.rb
jsugarman: yes, just standard grape DSL - happy to talk on slack or you can review the [grape docs here](https://github.com/ruby-grape/grape) and [grape entity here](https://github.com/ruby-grape/grape-entity)

app/views/pages/api_release_notes.html.haml
jsugarman: no worries, will break up. has no impact on the HTML but you are right it makes it easier to read in github and locally (where text editor wrapping normally handles it anyway)

app/interfaces/api/v1/dropdown_data.rb
willmcb: Cool I'll have a look at the docs and ask if I have questions.

willmcb (approved):
👍

colinbruce (approved):
LGTM :+1:

PR #2674, opened 2019-03-07
Add release notes following addition of missing offences

jsugarman:
#### What
Add release notes section to inform vendors
of the addition of more offences.

#### Ticket

extension of [CBO-597](https://dsdmoj.atlassian.net/browse/CBO-597)

#### Why
To inform vendors they may need to update
their stores of offences.

Also importing api-documentation css in
v1/manifest as was not before. This has
some documentation specifc styling.


cleargif (approved):
👍

PR #2679, opened 2019-03-13
Update the Cost per Transaction Cron definition

colinbruce:
#### What
Fix cron timing of new task
#### Why
The previous configuration, despite being generated by specialist website(!)
was not recognised by the service
#### How
Remove the trailing star, the cron line still validates online, and hopefully will pass in app too


willmcb (approved):
👍

PR #2680, opened 2019-03-13
Correct path to new advocate final claim endpoint

jsugarman:
#### What
Fix the name of the endpoint used by the JSON uploader

#### Why
Following addition of new advocate final claim endpoint
and deprecating of old, the json uploader was pointed
to use the new endpoint, but namespacing was missed.


willmcb (approved):
👍

PR #2681, opened 2019-03-13
Fix Rails vulnerability

colinbruce:
#### What
Address [rails CVE ](https://hakiri.io/technologies/railties/issues/5a334b180fb3df)
#### Why
A vulnerability has been raised and we should address them ASAP
#### How
Update the rails version to 5.2.2.1


jsugarman (approved):
cool 👍

PR #2684, opened 2019-03-21
Removes old feature specs

willmcb:
These feature files are no longer in use - it is therefore not really
worthwhile having them in the repo.

#### What

Removes old features specs that were no longer being run. 

#### Ticket

[CBO-648/9](https://dsdmoj.atlassian.net/browse/CBO-648)

#### Why

Was dead unused code - makes sense to delete them.

#### How

`rm -rf /old-features`


jsugarman (approved):
👍 At last! :)

PR #2687, opened 2019-03-26
CBO-639: Numbering for multi item lists

cleargif:
#### What
Only add item numbers when there is more than one item
ie. Misc fee or Tavel expence

#### Ticket

CBO-639

#### How
- Condition to update the view if array has more than 1 item


colinbruce (approved):
LGTM :+1:

PR #2688, opened 2019-03-27
CBO-640:  Case details page review

cleargif:
#### What
Moving the page header to the correct position

#### Ticket

CBO-640

#### Why
DAC

#### How
Applying the pattern across all the views


colinbruce (approved):
LGTM 👍 
Do you want to add a `tech-debt` jira ticket around re-factoring the repeated code?

PR #2690, opened 2019-04-01
Cost Per Transaction report fix

colinbruce:
#### What
The Cost per transaction run on the first of April had a couple of issues, this seeks to address both

#### Why
1)The report did not run at the expected time.  As the scheduler cannot be queried for next run, the only way to test this is to update the cron definition and try again.

2) When the report was run manually, it miscalculated the count of transactions, leading to an elevated Cost Per Transaction figure

#### How
1) Change to cron format.  Each language/host OS seems to validate cron differently which makes it hard to 'validate' what will work.  The new version has been validated at [crontab.guru](https://crontab.guru/#15_4_2,3_1,4,7,10_*) and doesn't return any validation issues.  This can run for the next two days to see if it works, and then reverted to only run on the first of the month after that.
2) Change the count to run agains the whole quarter, not just the first week! 🤦‍♂️

### DIFF
--- a/app/services/reports/performance_platform/quarterly_volume.rb
+++ b/app/services/reports/performance_platform/quarterly_volume.rb
@@ -76,7 +76,7 @@ module Reports
       end
 
       def count_digital_claims
-        @count_digital_claims ||= Claims::Count.week(@start_date)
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍 LGTM

PR #2705, opened 2019-04-03
CBO-662: update cron instruction

Rolodophone:
#### What
update cron instruction
#### Ticket

#### Why
The dates were changed for debugging but now it needs to only run on the first day of the month
#### How
Remove 2,3 and replace with 1


colinbruce (approved):
Useful, thanks 👍

PR #2710, opened 2019-04-04
Update the quarterly claim counter

colinbruce:
#### What
This has been amended to count claim_state_transitions between the
required dates instead of just the claim submissions
#### Why
This should be counting submissions of redeterminations as well as
original submissions.
#### How
Switched the `#quarter` method to count the different type

### DIFF
--- a/app/services/claims/count.rb
+++ b/app/services/claims/count.rb
@@ -2,7 +2,7 @@ module Claims
   class Count
     class << self
       def quarter(date)
-        new(date, :quarter).call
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍 is it only required on the quarter then? not month/week

colinbruce:
> 👍 is it only required on the quarter then? not month/week

Yes, this claims::count was only added as I kept re-using the code for different reports for week and month.  I added quarter and re-used the previous setup.

PR #2719, opened 2019-04-10
Rubocop todo fix

willmcb:
Fix some rubocop errors.

Also removed some duplication from `.rubocop_todo.yml`

### DIFF
--- a/app/services/reports/injection_errors.rb
+++ b/app/services/reports/injection_errors.rb
@@ -27,37 +27,37 @@ module Reports
     def aggregrate_query
       %{SELECT
       CASE
-        #{when_cond('%Cannot create a bill of type%', 'BILL_TYPE#CREATION_ERROR')}
-        #{when_cond('%Bill Sub Type%not found in%', 'BILL_TYPE#NOT_FOUND_ERROR')}
-        #{when_cond('%Invalid scenario%for Bill Sub Type%', 'BILL_TYPE#INVALID_SCENARIO_ERROR')}
-        #{when_cond('%At least one Bill object MUST be included on a claim%', 'CLAIM#MISSING_BILL_ERROR')}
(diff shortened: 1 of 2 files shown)


app/services/reports/injection_errors.rb
```diff
@@ -70,6 +70,15 @@ def where_conditions
       ['ia.succeeded = false', date_range_clause].compact.join(' AND ')
     end
 
+    def when_cond(error_string, error_type)
+      "WHEN obj->>'error' ILIKE '#{error_string}' THEN '#{error_type}'"
+    end
+
+    def no_error_description_catch
```
jsugarman: for consistency i'd name this `when_no_error_description`

app/services/reports/injection_errors.rb
```diff
@@ -70,6 +70,15 @@ def where_conditions
       ['ia.succeeded = false', date_range_clause].compact.join(' AND ')
     end
 
+    def when_cond(error_string, error_type)
```
jsugarman: as it is quite specific method, could name - `when_error_like`

app/services/reports/injection_errors.rb
```diff
@@ -70,6 +70,15 @@ def where_conditions
...
+      "WHEN obj->>'error' ILIKE '#{error_string}' THEN '#{error_type}'"
```
jsugarman, line 74: could use substitution:
```
template = "WHEN obj->>'error' ILIKE %{error_message} THEN %{error_type}"
template % { error_message: error_string. error_type: error_type }
```
Other examples:
https://blog.revathskumar.com/2013/01/ruby-multiple-string-substitution-in-string-template.html

jsugarman (approved):
👍 looks good, couple of suggestions

PR #2723, opened 2019-04-16
Prevent fee calculation firing too often

jsugarman:
#### What
Debounce graduated price

#### Why
$.debounce enables a delay to be added to
an event to prevent firing fee calc bound
events too often.

e.g. when typing 1000, we only need to
calculate for 1000, not 1,10,100 and 1000.

#### How
use JQuery $.debounce method


cleargif (commented):
👍

cleargif (approved):

PR #2741, opened 2019-04-25
CBO-697: External user pages v2 styles upgrade

cleargif:
#### What
Upgrading the last few pages to use the V2 style path

#### Ticket
CBO-697


cleargif:
See #2748

PR #2742, opened 2019-04-25
CBO-694: Caseworker v2 styles upgrade

cleargif:
#### What
Update  pages to use the v2 path of the stylesheets

#### Ticket
CBO-694


cleargif:
See #2748

PR #2746, opened 2019-04-26
Rubocop fixes

colinbruce:
#### What
Address the `Style/IfInsideElse` cops

#### Why
Reduce the tech debt in the rubocop_todo file

#### How
Remove instances of nested if statements

#### Also
* extend tests of the ApplicationHelper to cover the `cp` method
* re-namespace a cop to remove a warning when running rubocop

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -42,12 +42,28 @@ module ApplicationHelper
   # Returns a "current" css class if the path = current_page
   # TODO: this will not work on those routes that are also rooted to for the namespace or which have js that interferes
   def cp(path)
-    tab = extract_uri_param(path, 'tab')
-    if tab.present?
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -42,12 +42,28 @@ def presenter_for(model)
...
-    tab = extract_uri_param(path, 'tab')
-    if tab.present?
-      'current' if request.path == strip_params(path) && request.GET[:tab] == tab
-    else
-      'current' if request.path == strip_params(path)
-    end
+    'current' if path_matches(path) && tab_check_passes?(path)
+  end
+
+  def path_matches(path)
```
jsugarman: might be worth making these private?

willmcb: 👍

willmcb (approved):

PR #2747, opened 2019-04-26
Fix Net::ReadTimeout at the end of all feature scenarios

jsugarman:
#### What
Fix Net::Timeout errors on master running cukes locally

#### Why
Following a local ruby and various dependecy updates cucumber no longer
appears to have been shutting down the chromedriver automatically.
This explicit close patches the isse but hopefully a chromedriver
or selenium webdrivers update will make this unnecessary in the near
future.


willmcb (approved):
👍

PR #2750, opened 2019-05-01
Remove phantomJS

jsugarman:
#### What
remove cask for phantomJS from brewfile, used for dev setup

#### Why
Using chromedriver alone now. phantomJS
is EOL.


colinbruce (approved):
:+1:

PR #2752, opened 2019-05-02
Add bullet gem to identify optimizations

jsugarman:
#### What
Add  [bullet gem](https://github.com/flyerhzm/bullet)

#### Why
The bullet gem has been added and configured to run against the
test suite only. It will log errors to `log/bullet.log` for manual 
review by a developer.

This can be used, like rubocop, to optimize the codebase
and address technical debt when we have time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,6 +105,7 @@ group :development, :devunicorn, :test do
 end
 
 group :test do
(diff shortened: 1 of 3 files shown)


kmahern (approved):
lgtm 👍 (does it need a rebase?)

PR #2753, opened 2019-05-02
Fix seed data bug with evidence checklist

willmcb:
`ERROR: StateMachines::InvalidTransition :: Cannot transition state via :submit from :draft (Reason(s): Evidence checklist ids Evidence checklist id 2 is invalid, please use valid evidence checklist ids)`

Change id of evidence to `2` to avoid this. This is because the DocType with this id was deleted from `/app/models/doc_type.rb` in the last PR as it was an old evidence type.


jsugarman (approved):
👍

PR #2754, opened 2019-05-02
Revert "Add bullet gem to identify optimizations"

jsugarman:
#### What
This reverts commit a83c4a5d79235fb5ce17e3a93e7c46888eede243.

#### Why
Bullet gem seems to break the build. reverting for
now rather than spending time working this out.


willmcb (approved):
👍

PR #2766, opened 2019-05-07
Amend factory default rep order date

jsugarman:
#### What
Amend factory default rep order date

#### Why
The default `400.days.ago` was hitting
conflicts with tests as of today.


willmcb (commented):
👍

willmcb (approved):

PR #2770, opened 2019-05-09
Remove parallel_tests

jsugarman:
#### What
Remove parallel_tests gem

#### Why
This test suite dependency is not being
used and is not configured to play well
with current test suite in any event
due to before hooks in both rspec
and cucumber.


colinbruce (approved):
LGTM (I removed the readme instructions) :+1:

PR #2771, opened 2019-05-09
Cbo 660 duplicate defendant

kmahern:
#### What
Show rep order details when a 2nd defendant is added

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-660

#### How
Create a new rep order when a new defendant is added (that doesn't already have any).

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


colinbruce (approved):
:+1:

PR #2775, opened 2019-05-10
Cbo 626 fix dump script

willmcb:
#### What
Fix database dump script

#### Ticket

[CBO-626 fix dump script](https://dsdmoj.atlassian.net/browse/CBO-626)

#### Why
It broke when we upgraded rails to 5.2.2

Issue was that the method `arel_attributes_with_values_for_create` had become deprecated

#### How
New class that should make it easier to maintain.

--------

#### TODO (wip)

- [] test it actually works for prod
- [] possible further refactors

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -275,7 +275,7 @@ namespace :db do
     open(file_name, 'a') do |file|
       yield ->(model) do
         type_caster = model.class.arel_table.send(:type_caster)
-        values = model.send(:arel_attributes_with_values_for_create, model.class.column_names)
+        values = model.send(:attributes_for_create, model.class.column_names)
         values.each do |attribute, value|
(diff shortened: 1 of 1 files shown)


jsugarman:
@willmcb looks good, have you tried running this against disaster env. It has a ton of old live data

lib/tasks/db.rake
```diff
@@ -1,3 +1,5 @@
+require 'ostruct'
```
jsugarman: is this being used?

willmcb: No - I was using it as an interim thing I'll remove.

jsugarman: cool

jsugarman (approved):
looks good 👍 give a go on disaster, we can merge now if you like

PR #2778, opened 2019-05-13
Update Docker to use non-root user

colinbruce:
#### What
Update `Dockerfile` to run all services as a non-root user

#### Ticket
[CCCD: Migrate Docker configuration](https://dsdmoj.atlassian.net/browse/CBO-707)

#### Why
The new Kubernetes cluster needs all pods to run as a non-root user.  As a first step then, we need to move away from using root to run the services inside our container.

#### How 
Create a new, alpine based, docker image for CCCD.  Switch all of the services to run as non-root.

--------

#### TO DISCUSS

 - [ ] should we use port 3000, or is it too rails standard?
 - [ ] best method for switching over.  This requires a change to the deploy repo to change the port decided above


docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
-FROM registry.service.dsd.io/ruby:2.6.3
+FROM ruby:2.6.3-alpine3.8
```
jsugarman, line 1: reminder: ruby:2.6.3-alpine3.9 available now

docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
...
 
-# add official nodejs repo
-RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
-        echo 'deb https://deb.nodesource.com/node jessie main' > /etc/apt/sources.list.d/nodesource.list
+# Install pre-requisites for building nokogiri & pg gems, as well
+# as the packages for converting and generating PDF's
+RUN apk --update add --virtual build_deps \
```
jsugarman, line 5: For discussion, be good  (for me if no one else, to understand what these packages are for. In my dummy app i have:
```
# Additional packages/libraries
# - install regular stuff build-base
# - install libxml2-dev libxslt-dev required for nokogiri
# - install postgresql-dev required for postgres
# - install nodejs needed for ExecJS functionality
# - install yarn needed for asset compilation
#
```

I know what pdftk, redis and libreoffice are for, but not clear what ruby-dev, libc-dev, linux-headers, runit and git are for

jsugarman: The `--virtual <name>` flag seems to be for when you want to clear up dependences later. [found this explanation](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#virtual-packages)  - since we are not cleaning `build_deps` up we may not need it although it is possible we could clear up the `-dev` packages as they may only bre required for the build

docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
...
+    build-base ruby-dev libc-dev linux-headers \
+    libxml2-dev libxslt-dev \
+    postgresql-dev pdftk runit nodejs libreoffice redis git
```
jsugarman, line 9: could add 
```
# tidy up installation
RUN rm -rf /var/cache/apk/*
```

docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
...
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/* && rm -fr *Release* *Sources* *Packages* && \
-    truncate -s 0 /var/log/*log
+RUN addgroup -g 1000 -S appgroup && \
+    adduser -u 1000 -S appuser -G appgroup
 
-# SSH proxy settings
-ENV SSH_AUTH_SOCK /tmp/ssh-auth
-ENV SSH_AUTH_PROXY_PORT 1234
+WORKDIR /app
```
jsugarman, line 13: I have gone with using `/usr/src/app` in the dummy app as it is a folder that already exists and seems to be a [linux convention](http://forums.devshed.com/linux-help-33/difference-usr-src-usr-local-src-60913.html) for such things

jsugarman: so 
```
RUN mkdir -p /usr/src/app
WORKDIR /usr/src/app
RUN mkdir -p /usr/src/app/tmp
```

docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
...
 RUN bundle config --global without test:development
-
-WORKDIR /usr/src/app
-
-COPY Gemfile /usr/src/app/
-COPY Gemfile.lock /usr/src/app/
-
-# Hack to install private gems
-#RUN socat UNIX-LISTEN:$SSH_AUTH_SOCK,fork TCP4:$(ip route|awk '/default/ {print $3}'):$SSH_AUTH_PROXY_PORT & bundle install
+RUN gem install bundler -v 1.17.3
```
jsugarman, line 18: to speed up builds it is, apparently, a good idea to only copy dependencies over - found this [9-common-dockerfile-mistakes](https://runnable.com/blog/9-common-dockerfile-mistakes)

Also we should probably be bundle installing as the user (i hit problems when i didn't). In total here is the way i did it in the dummy app

```
# set env vars needed for bundle install
# and asset precompilation 
#
ENV RAILS_ENV='production'
ENV RACK_ENV='production'

######################
# DEPENDENCIES START #
######################
# NOTE: all COPY commands set file permissions to root
COPY Gemfile Gemfile
COPY Gemfile.lock Gemfile.lock

# change to the added user
USER appuser
RUN gem install bundler -v 1.17.3

# remove test development gems from default bundle install
RUN bundle config --global without test:development

# nokogiri install problems
# https://github.com/gliderlabs/docker-alpine/issues/53
#
# Use libxml2-dev, libxslt-dev packages, installed above, for building nokogiri
RUN bundle config build.nokogiri --use-system-libraries

RUN bundle install

####################
# DEPENDENCIES END #
####################
```

docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
...
 RUN bundle install
 
-COPY . /usr/src/app
-RUN mkdir -p /usr/src/app/public/assets
-
-COPY ./docker/run.sh $APP_HOME/run.sh
-RUN chmod 777 $APP_HOME/run.sh
+RUN bundle exec rake assets:precompile RAILS_ENV=production SECRET_KEY_BASE=a-real-secret-key-is-not-needed-here
```
jsugarman, line 21: don't need `RAILS_ENV` here if we `ENV RAILS_ENV='production'`

docker/Dockerfile
```diff
@@ -1,41 +1,29 @@
...
 
-EXPOSE 80
+RUN chown -R appuser:appgroup /app/
```
jsugarman, line 23: I found that I only had to do this at this point?!
```
# copy app contents from host to container (owner will be root)
COPY . .

# take ownership of all app files
USER root
RUN chown -R appuser:appgroup /usr/src/
```

docker/non_root_run.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+# last modified 10-05-2019
+set -ex
```
jsugarman, line 3: This 'immediate fail and output' can go higher up near the top of the `Dockerfile` as `RUN set -ex` - to help catch errors early on - i imagine the setting will then persist while running this script?!

PR #2783, opened 2019-05-14
Update targeted-ruby-version in .rubocop.yml

colinbruce:
#### What
Updated targeted-ruby-version in .rubocop.yml

#### Why
It had got out of sync with the actual ruby version used in the application

#### How
Changed the value in .rubocop.yml, this doesn't need any cop fixes at this time


jsugarman (approved):
👍

PR #2787, opened 2019-05-15
Non root docker v2

jsugarman:
#### What
optimize docker image

#### Ticket
[CBO-707](https://dsdmoj.atlassian.net/browse/CBO-707)

#### Why
As a step to meeting platforms K8s requirements and delivering the
non-functional (performance and security) value provided by using a 
thinner open-source-maintained linux image and non-root user.


docker/Dockerfile
```diff
@@ -1,41 +1,75 @@
...
+# tidy up installation
+RUN apk del build-dependencies
 
-#SECRET_TOKEN set here because otherwise devise blows up during the precompile.
-RUN bundle exec rake assets:precompile RAILS_ENV=production SECRET_KEY_BASE=a-real-secret-key-is-not-needed-here
+# non-root/appuser should own only what they need to
+RUN chown -R appuser:appgroup log tmp db
 
-CMD "$APP_HOME/run.sh"
+USER appuser
```
exonian: I was talking to @colinbruce about this (`USER <name>` vs `USER <uid>`) yesterday, and today came across this in the cloud platforms guidance which clarifies it
https://user-guide.cloud-platform.service.justice.gov.uk/tasks.html#how-to-adapt-to-the-pod-security-policies
`You must specify the user by its numeric UID`
Colin said this branch was deploying to live-1 ok, so I'm not sure, but it looks like you may have to go uid like we have?

jsugarman: cool, cheers

docker/Dockerfile
```diff
@@ -1,41 +1,75 @@
-FROM registry.service.dsd.io/ruby:2.6.3
+FROM ruby:2.6.3-alpine3.8
+MAINTAINER Ministry of Justice <tools@digital.justice.gov.uk>
```
colinbruce: Should this not be crowncourtdefence@digital.justice.gov.uk?
The tools team do not maintain the CCCD Dockerfile.

colinbruce (approved):
LGTM :+1:

Couple of non-blocking observations

docker/Dockerfile
jsugarman: @colinbruce do we monitor that one, in which case yes 👍

exonian (approved):
👍

docker/Dockerfile
colinbruce: We do, or at least, _I_ do! :)

jsugarman: Cool, thats done

PR #2809, opened 2019-05-29
CBO-710: Implement provider manager role spike

colinbruce:
#### What
The spike was to investigate the work required to move the provider management function to role based.

#### Ticket
[SPIKE: CCCD: investigate moving the manage provider functionality to a role](https://dsdmoj.atlassian.net/browse/CBO-710)

#### Why
There is a single SuperAdmin user that has access to maintain the providers.  Adding an additional role that can be allocated, to case_workers only, means that users of the system can manage providers without logging out and in again. 
It also means that we no longer have to change the Super Admin password when a user with access leaves the department

#### How
Add an additional CaseWorker Role that can be implemented on a per-user basis.  This means that a user can have their permissions revoked by another admin and that, if they leave and their account is deleted we do not have to cycle the Super Admin password

#### Todo
- [x] Discuss the need for a single Super Admin login going forward
- [x] Agree on how to handle sidekiq monitoring if Super Admin is removed
- [x] Refactor the duplicated code if it is decided to retain both versions

These will all be discussed as part of this [jira ticket](https://dsdmoj.atlassian.net/browse/CBO-757)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -178,6 +178,17 @@ Rails.application.routes.draw do
       get 'management_information/download', to: 'management_information#download', as: :management_information_download
       get 'management_information/generate', to: 'management_information#generate', as: :management_information_generate
     end
+
+  end
+
+  namespace :provider_management do
+    root to: 'provider_management/providers#index'
+    resources :providers, except: [:destroy] do
+      resources :external_users, except: [:destroy] do
+        get 'change_password', on: :member
+        patch 'update_password', on: :member
+      end
+    end
   end
 
   namespace :geckoboard_api do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -71,7 +71,6 @@ Rails.application.routes.draw do
 
   # shared by super_admin and case worker provider managers
   concern :provider_management_routes do
-    root to: 'providers#index'
     get 'external_users/find', to: 'external_users#find'
     post 'external_users/find', to: 'external_users#search'
 
@@ -84,6 +83,7 @@ Rails.application.routes.draw do
   end
 
   namespace :super_admins do
+    root to: 'providers#index'
     concerns :provider_management_routes
 
     namespace :admin do
@@ -97,6 +97,7 @@ Rails.application.routes.draw do
   end
 
   namespace :provider_management do
+    root to: 'provider_management/providers#index'
     concerns :provider_management_routes
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -69,7 +69,8 @@ Rails.application.routes.draw do
     put :settings, on: :member, action: :update_settings, format: :js
   end
 
-  namespace :super_admins do
+  # shared by super_admin and case worker provider managers
+  concern :provider_management_routes do
     root to: 'providers#index'
     get 'external_users/find', to: 'external_users#find'
     post 'external_users/find', to: 'external_users#search'
@@ -80,6 +81,10 @@ Rails.application.routes.draw do
         patch 'update_password', on: :member
       end
     end
+  end
+
+  namespace :super_admins do
+    concerns :provider_management_routes
 
     namespace :admin do
       root to: 'providers#index'
@@ -89,7 +94,10 @@ Rails.application.routes.draw do
         patch 'update_password', on: :member
       end
     end
+  end
 
+  namespace :provider_management do
+    concerns :provider_management_routes
   end
 
   scope module: 'external_users' do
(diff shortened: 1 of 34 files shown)


colinbruce:
@jsugarman the plan is to remove this functionality from super admin. So there was no point in making the code shared as, hopefully, in the next sprint we'll be deleting the super admin version of it.
I think for now we just have to ignore the code-climate duplication warnings. :(

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -0,0 +1,245 @@
+require 'rails_helper'
+require 'json'
+
+RSpec.describe ProviderManagement::ExternalUsersController, type: :controller do
+  let(:case_worker_manager)   { create(:case_worker, :provider_manager) }
```
jsugarman: minor/non-blocking double whitespace

spec/controllers/provider_management/providers_controller_spec.rb
```diff
@@ -0,0 +1,192 @@
+require 'rails_helper'
+
+RSpec.describe ProviderManagement::ProvidersController, type: :controller do
+  let(:case_worker_manager)   { create(:case_worker, :provider_manager) }
```
jsugarman: non-blocking - double whitespace

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+    context 'when the email does not exist' do
+      let(:email) { 'vail.email@does.not.exist.com' }
+
+      it { is_expected.to redirect_to(provider_management_external_users_find_path) }
+    end
+  end
+
+  describe "GET #new" do
+    let(:external_user) do
+     a = ExternalUser.new(provider: provider)
```
jsugarman: possible refactor

```
ExternalUser.new(provider: provider).tap |eu|
 eu.build_user
end
```

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+    end
+
+    it 'builds user for @external_user' do
+      expect(assigns(:external_user).user.to_json).to eq(external_user.user.to_json)
+    end
+
+    it 'renders the new template' do
+      expect(response).to render_template(:new)
+    end
+
```
jsugarman: minor/non-blocking (copy an paste i imagine) - unnecessary linebreaks here and below

config/routes.rb
```diff
@@ -178,6 +178,19 @@
       get 'management_information/download', to: 'management_information#download', as: :management_information_download
       get 'management_information/generate', to: 'management_information#generate', as: :management_information_generate
     end
+
+  end
+
+  namespace :provider_management do
+    root to: 'provider_management/providers#index'
+    get 'external_users/find', to: 'external_users#find'
+    post 'external_users/find', to: 'external_users#search'
```
jsugarman: are these two routes used - they don't appear to be.

colinbruce: Yes, they are called from the provider_management index view... ```/provider_management/external_users/find``` it's the find provider by email page

jsugarman: nor even for superadmin?!

colinbruce: For superadmin too... try logging in to live and click the 'Find provider by email` link and see where you end up! :)

spec/controllers/provider_management/external_users_controller_spec.rb
colinbruce: image
👀 
🤣

jsugarman: oh, i have learnt so much. and my preferences changed. 🤣

jsugarman (approved):
Have made a few minor additions as discussed but LGTM 👍

Be good to have a feature test for the new provider management but if time is
short then thats fine.

config/routes.rb
jsugarman: oh yeah. had my code searching limited to specs. urrgh!

features/page_objects/provider_page.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'sections/checkbox_fee_section'
```
jsugarman: does not look like you are using this, no call to
```
CheckboxFeeSection
```

**Not a blocker**
You could, as a refactor, write a `CheckboxSection`, and have `CheckboxFeeSection` inherit from that and use the new CheckboxSection here, as below
```
section :fee_scheme_checkboxes, CheckboxSection, "div#fee-scheme-options"
```

jsugarman (approved):
LGTM 👍  
- one unused require
- a non-blocking siteprism section refactor option

jsugarman:
@colinbruce 🥇

PR #2814, opened 2019-05-30
Remove simplecov console

jsugarman:
#### What
Remove simplecov-console gem

#### Ticket
support and maintenance - depedency updates

#### Why
We do not use and it adds a lot of dependencies.
The latest bump also requires a json bump which
could have unintended side-effects.


colinbruce (approved):
👍 LGTM

PR #2819, opened 2019-06-03
Add rubocop rails

jsugarman:
#### What
Include separate gem for rubocop rails

#### Why
As of next rubocop bump (0.72) this will
need seperating out so introducing now peremtively
and to tidy up gemfile.


colinbruce (approved):
Generally LGTM :+1:

Not great that some of the offense counts are increasing...
Could we decrease the max and list exceptions to prevent more being added?

.rubocop_todo.yml
```diff
@@ -18,11 +18,11 @@ Lint/UnneededCopDisableDirective:
...
 
-# Offense count: 49
+# Offense count: 56
 Metrics/AbcSize:
   Max: 87
 
-# Offense count: 21
+# Offense count: 24
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
```
jsugarman, line 27: @colinbruce looks like this was the only cop that increased the max rather than excluding individual files so have readded the exclusions myself. The rest are new rails-specific cops.

PR #2823, opened 2019-06-04
Fix VAT on assessments and redeterminations

jsugarman:
#### What
Fixes the calculation and persisting of VAT amounts
on assessments and redeterminations for advocate
supplementary and interim claims.

#### Ticket

[CBO-752](https://dsdmoj.atlassian.net/browse/CBO-752)

#### Why
VAT was not being calculated and persisted on
assessments and redeterminations for advocate
supplementary and interim claims.

#### How
Use `.agfs?` method to filter those claims
requiring vat calculation, rather than check
the specific class type. This centralizes
the logic in the `agfs_claim_types` scope.

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -46,11 +46,15 @@ class Assessment < Determination
     save!
   end
 
+  # TODO: this appears to only be used by update_amount_assessed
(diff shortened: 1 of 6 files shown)


colinbruce (approved):
LGTM 👍

PR #2825, opened 2019-06-05
Increase batch size for nightly transition

jsugarman:
#### What
Increase batch size for nightly transition

#### Why
To catch up with achiving and deleting of
stale claims following failure over the last few
months.


colinbruce (approved):
:+1:

PR #2828, opened 2019-06-05
Add codeclimate-rubocop back using channel 0-71

jsugarman:
#### What
Add codeclimate-rubocop back using channel 0-71
instead of 0-70.

#### Why
previously removed due to problem
with `require: rubocop-performance`

Because codeclimate runs a diff against
the master branch and the master branches
codeclimate-rubocop was always failing due
to a problem with codeclimate-rubocop it had
be removed temporarily.


colinbruce (approved):
:+1:

PR #2833, opened 2019-06-10
Update messaging aws client

colinbruce:
#### What
Amend the way that the AWS SQS messaging client can be called

#### Why
As part of the Kubernetes work, queues will no longer have a statically defined name, but the queue_url will be generated and inserted into the environment variables

#### How
By adding an environment variable for the queue_url this can be loaded, if present.  If the value is not present the code falls back to the previous version and loads the queue_name and requests the URL from AWS.

A bonus of this is that a single call will be made to the AWS API now, `receive_message`, instead of two, `get_queue_url` *and* `receive_message`

### DIFF
--- a/scheduled_tasks/poll_injection_responses_task.rb
+++ b/scheduled_tasks/poll_injection_responses_task.rb
@@ -6,7 +6,7 @@ class PollInjectionResponsesTask < Scheduler::SchedulerTask
   every '1m'
 
   def run
-    queue = Settings.aws.response_queue
+    queue = Settings.aws.sqs.response_queue_url || Settings.aws.response_queue
     return unless queue
     log("Checking for messages on #{queue}")
     MessageQueue::AwsClient.new(queue).poll!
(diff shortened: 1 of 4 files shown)


scheduled_tasks/poll_injection_responses_task.rb
```diff
@@ -6,7 +6,7 @@ class PollInjectionResponsesTask < Scheduler::SchedulerTask
   every '1m'
 
   def run
-    queue = Settings.aws.response_queue
+    queue = Settings.aws.sqs.response_queue_url || Settings.aws.response_queue
```
jsugarman, line 9: just a query. is the `response_queue_url` fake value from settings going to be nilled, or should it not be set to an ENV var value so that this code falls back to  the `Settings.aws.response_queue` for environments that do not have the env var

colinbruce: It will be nil unless we explicitly set one.  The setting gem looks for an env var called SETTINGS__AWS__SQS__RESPONSE_QUEUE_URL and loads it from there... I'd expect that to be nil by default, but I want to check on demo before I fully deploy!

jsugarman: always forget about that settings gem. another magic abstraction.

jsugarman (dismissed):
LGTM

jsugarman (approved):
LGTM....again 👍

PR #2835, opened 2019-06-11
Afix chromdriver version feature tests

jsugarman:
#### What
Pin chromdriver version for webdrivers gem

#### Why
Latest version (75) is not running headless
locally. Possible chromedriver bug?!

Version 74 latest (see https://chromedriver.storage.googleapis.com/index.html)
is available and does run headless but have decided to use 2.46
which covers versions 71 to 73 in discussion with FE dev.


naseberry (approved):
Looks good.

PR #2839, opened 2019-06-14
Modify build and deploy scripts

jsugarman:
#### What
Seperate build and deploy scripts.

Also rename and enable running as bash
script.

i.e.
old:
```
$ builtit bash dev
```

new:
```
$ ./k8sbuild.sh
$ ./k8sdeploy.sh dev
```

#### Why
Seprate scripts so that you can, potentially, change deployments
(uncommitted) and deploy them without committing. Such as for
adding env vars and deploying the same build to multiple environments.

Could extend the deploy to take a shah to deploy an existing build but not the
current.


k8sdeploy.sh
```diff
@@ -0,0 +1,34 @@
...
+  version=$(git rev-parse $(git branch | grep \* | cut -d ' ' -f2))
+  docker_registry='754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-get-paid/cccd'
+  docker_tag=app-${version}
+  docker_image=${docker_registry}:${docker_tag}
+
+  echo "Deploying $docker_image to $environment on cluster $context"
+  kubectl set image -f kubectl_deploy/${environment}/deployment.yaml cccd-app=$docker_image --local -o yaml | kubectl apply -n cccd-${environment} --context ${context} -f -
+}
+
+_k8sdeploy $@
```
jsugarman: line break

k8sdeploy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
jsugarman: change to sh

k8sbuild.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+function _k8sbuild() {
+  usage="k8build -- build, tag and push image to ecr"
+
+  REGION='eu-west-2'
```
jsugarman: lower case for local/temp vars
https://stackoverflow.com/questions/673055/correct-bash-and-shell-script-variable-capitalization

k8sbuild.sh
```diff
@@ -0,0 +1,34 @@
...
+  CONTEXT='live-1'
```
jsugarman: not used

k8sbuild.sh
```diff
@@ -0,0 +1,34 @@
...
+  echo '------------------'
+  echo $docker_image
+  echo '------------------'
+
+  $(aws ecr --profile "$AWS_PROFILE" get-login --no-include-email --region "$REGION")
+  docker build \
+          --build-arg VERSION_NUMBER=$docker_image \
+          --build-arg BUILD_DATE=$(date +%Y-%m-%dT%H:%M:%S%z) \
+          --build-arg COMMIT_ID=$VERSION \
+          --build-arg BUILD_TAG=$CCCD_BUILD_TAG \
```
jsugarman: should be docker_tag

k8sdeploy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+function _k8sdeploy() {
+  usage="k8sdeploy -- build, tag and push image to ecr
```
colinbruce: `build, tag and deploy to k8s environment`?

colinbruce (approved):
Generally fine, one minor tweak to the description.

Also, is the `api-sandbox` deploy file supposed to be in here? It's not mentioned in the PR description.

👍

k8sdeploy.sh
jsugarman: copy paste 💥 - cheers

PR #2843, opened 2019-06-14
Refactor the S3 handling

colinbruce:
#### What
Improve the S3 header generation to allow it to work with both Kubernetes _and_ template deploy

#### Why
As part of the Kubernetes work, we needed to update the methods used to upload files to s3.  This has left the code slightly fragile in that it will work for K8s but not for the existing template deploy methods.  This refactor aims to allow the code to be deployed to both environments

#### How
 - [X] Extract the s3 headers into shared code to allow updates to be made in a single place
 - [x] extend the code to allow template deploy to work too

### DIFF
--- a/app/models/concerns/document_attachment.rb
+++ b/app/models/concerns/document_attachment.rb
@@ -1,24 +1,13 @@
 module DocumentAttachment
   extend ActiveSupport::Concern
+  include S3Headers
 
   included do
     attr_accessor :pdf_tmpfile
 
-    has_attached_file :converted_preview_document,
-                      { s3_headers: {
-                        'x-amz-meta-Cache-Control' => 'no-cache',
-                        'Expires' => 3.months.from_now.httpdate
-                      },
-                        s3_permissions: :private,
-                        s3_region: Settings.aws.s3.region }.merge(PAPERCLIP_STORAGE_OPTIONS)
+    has_attached_file :converted_preview_document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)
 
-    has_attached_file :document,
-                      { s3_headers: {
-                        'x-amz-meta-Cache-Control' => 'no-cache',
-                        'Expires' => 3.months.from_now.httpdate
-                      },
-                        s3_permissions: :private,
-                        s3_region: Settings.aws.s3.region }.merge(PAPERCLIP_STORAGE_OPTIONS)
+    has_attached_file :document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)
 
     validates_attachment_content_type :converted_preview_document, content_type: 'application/pdf'
   end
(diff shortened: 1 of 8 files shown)


app/models/concerns/document_attachment.rb
```diff
@@ -1,24 +1,13 @@
 module DocumentAttachment
   extend ActiveSupport::Concern
+  include S3Headers
```
jsugarman, line 3: nice tidy up! 

i wonder if you can use a kind of of `acts_as` like pattern
https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/app/services/simple_bill_typeable.rb
but might not work within a containing `concern`

app/models/concerns/s3_headers.rb
```diff
@@ -0,0 +1,22 @@
+module S3Headers
+  extend ActiveSupport::Concern
+
+  module ClassMethods
```
jsugarman: can just do 
```
class_methods do
...
end
```

app/models/concerns/s3_headers.rb
```diff
@@ -0,0 +1,22 @@
...
+        },
+        s3_permissions: :private,
+        s3_region: region
+      }
+    end
+
+    private
+
+    def region
+      Settings.aws.region || 'eu-west-1'
```
jsugarman, line 19: this will work fine on the kubernetes branch but merging the resulting kubernetes branch to master (which i think we are looking to do) would break current s3 by the looks of it.

I would think that setting the Settings.aws.region in settings to ENV['AWS_REGION'] then removing the || clause here would do it?!

TD envs have a `AWS_REGION` and looks like k8s secrets do  too (though why it is a secret is another matter and now has two REGION keys)
```
$ k exec -it claim-for-crown-court-defence-576d8b6cb-dqnwr sh
/usr/src/app $ env | grep REGION
AWS_REGION=eu-west-2
SETTINGS__AWS__S3__REGION=eu-west-2
```

spec/models/message_spec.rb
```diff
@@ -124,4 +124,23 @@
       expect(claim.state).to eq 'part_authorised'
     end
   end
+
+  describe '.s3_headers' do
```
jsugarman: **not a blocker**, but be good to have a shared example to use in the other models (document_attachment and stats_report)

app/models/concerns/s3_headers.rb
colinbruce: On TD at the moment...
```$ sudo docker exec -it advocatedefencepayments sh
/usr/src/app $ env | grep REGION
AWS_REGION=eu-west-1
```

spec/models/message_spec.rb
colinbruce: Noted, I'll work on that now

spec/models/document_spec.rb
```diff
@@ -41,6 +41,8 @@
   it { should have_attached_file(:converted_preview_document) }
   it { should validate_attachment_content_type(:converted_preview_document).allowing('application/pdf') }
 
+  it_behaves_like 'an s3 bucket'
```
jsugarman, line 44: 🏆 great

PR #2844, opened 2019-06-14
Kubernetes - update the CCCD queue response handling

colinbruce:
#### What
Update the CCCD response queue handling.  Allow for different actions if the K8s environment has set a response_queue URL, but not set a response_queue name

#### Why
This will allow us to track _all_ injection attempts while testing the Kubernetes stack, while still allowing the template deploy environments to function as they do now

#### How
- [x] replace the `aws.response_queue` env_var with an `aws.sqs.response_queue_url`
- [x] Allow an override to the slack channel if the `aws.sqs.response_queue_url` exists
- [x] Always send a slack notification if  the `aws.sqs.response_queue_url` exists
- [ ] Rename the `Settings.aws.sns.access` to be `Settings.aws.messgaging.access` etc  This has become necessary since the grouping of messaging objects in terraform


jsugarman:
closing in favour of a version rebased against the `kubernetes` "master" branch 
see [rebased version](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2852)

PR #2850, opened 2019-06-18
Remove unused active storage routes

jsugarman:
#### What
Remove unused active storage routes

#### Why
Still using paperclip for now so not needed

#### How
see [stackoverflow thread](https://stackoverflow.com/questions/52497044/how-to-disable-auto-generated-routes-by-active-storage)


willmcb (approved):
lgtm

PR #2851, opened 2019-06-18
Remove noisy warning for Hashdiff

jsugarman:
#### What
Remove noisy warning for Hashdiff

#### Why
The current version of Hashdiff
has a warning that pops up anytime
it gets loaded (tests, console, CI
jobs/tasks). noisy!

#### How
Update to version 1.0.0.beta1 which
gets rid of it. we only use this gem
in tests and seeding so no harm done
if the beta is imperfect.


willmcb:
👍

willmcb (approved):

PR #2852, opened 2019-06-18
Kubernetes - CCCD staging response queue (rebased version)

jsugarman:
#### What
Update the CCCD response queue handling.  Allow for different actions if the K8s environment has set a response_queue URL, but not set a response_queue name

#### Why
This will allow us to track _all_ injection attempts while testing the Kubernetes stack, while still allowing the template deploy environments to function as they do now

#### How
- [x] replace the `aws.response_queue` env_var with an `aws.sqs.response_queue_url`
- [x] Allow an override to the slack channel if the `aws.sqs.response_queue_url` exists
- [x] Always send a slack notification if  the `aws.sqs.response_queue_url` exists
- [ ] Rename the `Settings.aws.sns.access` to be `Settings.aws.messgaging.access` etc  This has become necessary since the grouping of messaging objects in terraform 
 ^^ I like the separation of SNS config from SQS. Even if the secrets are stored in the `cccd-messaging` secret they can referred to via settings in a more descriptive way

### DIFF
--- a/app/services/message_queue/aws_client.rb
+++ b/app/services/message_queue/aws_client.rb
@@ -1,7 +1,11 @@
 module MessageQueue
   class AwsClient
     def initialize(queue)
-      @sqs = Aws::SQS::Client.new(access_key_id: Settings.aws.access, secret_access_key: Settings.aws.secret)
+      @sqs = Aws::SQS::Client.new(
+        access_key_id: Settings.aws.sns.access,
+        secret_access_key: Settings.aws.sns.secret,
+        region: Settings.aws.region
+      )
       begin
         @queue_url = queue_url(queue)
       rescue Aws::SQS::Errors::NonExistentQueue
--- a/kubectl_deploy/staging/deployment.yaml
+++ b/kubectl_deploy/staging/deployment.yaml
@@ -24,6 +24,8 @@ spec:
 #              path: /healthcheck
 #              port: 4567
           env:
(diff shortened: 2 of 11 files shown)


kubectl_deploy/staging/deployment.yaml
```diff
@@ -152,6 +152,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: cccd-messaging
+                  key: SETTINGS__AWS__SUBMITTED_TOPIC_ARN
```
jsugarman: remove

app/services/message_queue/aws_client.rb
```diff
@@ -1,7 +1,11 @@
 module MessageQueue
   class AwsClient
     def initialize(queue)
-      @sqs = Aws::SQS::Client.new(access_key_id: Settings.aws.access, secret_access_key: Settings.aws.secret)
+      @sqs = Aws::SQS::Client.new(
+        access_key_id: Settings.aws.sns.access,
```
jsugarman, line 5: change to `messaging` form `sns`

app/services/message_queue/aws_client.rb
```diff
@@ -1,7 +1,11 @@
...
+        secret_access_key: Settings.aws.sns.secret,
```
jsugarman, line 6: ditto ^^

PR #2854, opened 2019-06-19
Ignore kubernetes `yml` and `yaml` secret file

jsugarman:
#### What
Ignore kubernetes `yml` and `yaml` secret file

#### Why
Putting this in just to make sure cross
branch work does not leak secrets.


kmahern (approved):
lgtm 👍

PR #2857, opened 2019-06-21
K8s/add latest tags

jsugarman:
Add and use `latest` tag to build and deploy

#### What
Add additional tag to builds
and have deploy use the latest
branch tag rather than the shah.

#### Ticket

kubernetes migration epic

#### Why
This prevents accidental deploy of an image for
the latet tag/commit-sha that has not yet been 
built. However it does not deal with situation when no
build of the branch has been done.

#### How
Add an additional tag to each build to mark as the latest for that branch.
i.e.
`app-<branch-name>-latest`
or for master branch
`app-latest`

Have deploy script use the latest branch tag during deploy.

Extension - provide a warning if no image with that tag exists
and abort.
Extension - have rolling update to prevent termination of pod
if replacement container creation fails

### DIFF
--- a/bin/k8secret
+++ b/bin/k8secret
@@ -12,12 +12,12 @@
 #
 # Examples:
 # most basic example if current-context is the one you want
-#  ``
+# `bin/k8secret -e dev`
 #
-# output default secrets oject to STDOUT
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM 👍

PR #2858, opened 2019-06-21
Cbo 781 remove super admin

willmcb:
#### What
Removed provider / external_user management for super_admin users.

#### Ticket

[CCCD: Remove provider management from superadmin](https://dsdmoj.atlassian.net/browse/CBO-781)

#### Why
Functionality duplicated elsewhere.


app/models/ability.rb
```diff
@@ -10,8 +10,7 @@ def initialize(user)
     persona = user.persona
 
     if persona.is_a? SuperAdmin
-      can %i[show index new create edit update], Provider
-      can %i[show index new create edit update change_password update_password find search], ExternalUser
+      can %i[index], Provider
```
jsugarman: as discussed this should go to, once a new landing page for superadmins can is indentified or added

app/controllers/super_admins/providers_controller.rb
```diff
@@ -1,42 +1,10 @@
...
-
-  def create
-    @provider = Provider.new(provider_params)
-    if @provider.save
-      redirect_to super_admins_root_path, notice: 'Provider successfully created'
-    else
-      render 'shared/providers/new'
-    end
-  end
+  def index; end
```
jsugarman: as discussed this whole file should go to and, if necessary, a new basic home page for superadmins created, i'd say

config/routes.rb
```diff
@@ -84,7 +84,6 @@
 
   namespace :super_admins do
     root to: 'providers#index'
-    concerns :provider_management_routes
```
jsugarman, line 87: **not a blocker** but i had added this "route concern" in only because we were sharing between case workers and super admins. now that super admins are not to have provider management i'd get rid of this concern and put the routes directly into the relevant resources namespace for case workers

spec/models/ability_spec.rb
```diff
@@ -454,19 +454,11 @@
...
     let(:other_super_admin) { create(:super_admin) }
     let(:provider)          { create(:provider) }
-    let(:other_provider)    { create(:provider) }
     let(:external_user)          { create(:external_user, provider: provider)}
-    let(:other_external_user)    { create(:external_user, provider: other_provider)}
 
     it { should be_able_to(:update_settings, user) }
     it { should_not be_able_to(:update_settings, another_user) }
 
-    context 'can manage any provider' do
```
jsugarman, line 464: probably worth having a negative test to replace this.

```
 context 'cannot manage any provider' do
      [:show, :index, :new, :create, :edit, :update].each do |action|	
        it { should_not be_able_to(action, provider) }	
      end	
    end
```

spec/models/ability_spec.rb
```diff
@@ -485,13 +477,6 @@
       end
     end
 
-    context 'can view, create and change any external_users details' do
```
jsugarman, line 488: ditto here on having a  negative test.

could use a [shared example](https://relishapp.com/rspec/rspec-core/v/3-8/docs/example-groups/shared-examples) with parameters to have positive and negative versions, but not a blocker

```
it_behaves_like 'a provider manager', false
it_behaves_like 'a provider manager', true
```

jsugarman (changes requested):
great. couple of things i think we should aim to get rid of here entirely:
-  `super_admin/providers _controller` 
- associated authorization/cancancan ability

have added some non-blocking refactoring suggestions too

app/controllers/super_admins/providers_controller.rb
willmcb: 👍

app/models/ability.rb
willmcb: 👍

spec/models/ability_spec.rb
willmcb: 👍

config/routes.rb
willmcb: 👍

config/routes.rb
```diff
@@ -97,7 +84,15 @@
 
   namespace :provider_management do
     root to: 'providers#index'
-    concerns :provider_management_routes
+    get 'external_users/find', to: 'external_users#find'
```
jsugarman, line 87: 👍

jsugarman (approved):

PR #2862, opened 2019-06-24
[CBO-771] Fix DAC report issues

naseberry:
#### What
Fix issues raised by DAC.

#### Ticket
[Investigate and fix issues raised in the DAC 2018 report](https://dsdmoj.atlassian.net/browse/CBO-771)

#### Why
To meet the governments accessibility standards.


app/views/external_users/claims/_summary_fees.html.haml
```diff
@@ -13,7 +13,7 @@
 
 - if local_assigns.has_key?(:fee)
   - if fee
-    %table{ class: 'summary', summary: t('.summary') }
+    %table.summary
```
colinbruce, line 16: The summary field had a text value drawn from `en.yml`, but I can't see that being added anywhere else. I know the summary field is deprecated in HTML5, but did the description have value for accessibility users?

This applies to the following tables too! :)

app/views/external_users/claims/supporting_documents/_summary.html.haml
```diff
@@ -44,9 +44,8 @@
           - if document.converted_preview_document.present?
             = link_to t('common.view'),
                       document_path(document),
-                      target: '_blank',
                       class: 'view',
-                      'aria-label': "View document: #{document.document_file_name}, Opens in a new window"
+                      'aria-label': "View document: #{document.document_file_name}"
```
colinbruce, line 48: What's the advantage of removing the `opens in a new window` prompt here?

I understand that indiscriminately opening new windows is bad, but this informs the user before doing so.

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
```diff
@@ -45,7 +45,7 @@
         ),
         %br
         please visit the
-        %a{ href: "https://www.gov.uk/government/publications/crown-court-fee-guidance", target: '_blank' } Crown Court fee guidance
+        %a{ href: "https://www.gov.uk/government/publications/crown-court-fee-guidance" } Crown Court fee guidance
```
colinbruce: Would it not be better to add an accessible warning and open in a new tab/window?

Leaving the current page to read the guidance could lead users to lose data inputted prior to leaving.

app/assets/javascripts/modules/Modules.CustomFileUpload.js
```diff
@@ -2,50 +2,50 @@ moj.Modules.CustomFileUpload = {
...
 
-  cacheEls: function() {
+  cacheEls: function () {
     this.$fileInputs = $(this.el);
     this.$forms = this.$fileInputs.closest('form');
     this.$chooseFileNameTmpl = $(this.tpml.text);
     this.$chooseFileButtonTmpl = $(this.tpml.button);
   }
-};
+};
```
colinbruce: Can you ensure files have an empty line at the end? This prevents warnings from github! 

We have set up rubocop to warn on ruby files, but have never found an auto linter for JS and view files.

colinbruce (changes requested):
Generally good (considering the number of changed files!) but a quick chat about some of the issues could be useful.

app/views/external_users/claims/_summary_fees.html.haml
naseberry: The summary attribute specifies a summary of the content of a table, whereby our text value was information the user how to navigate within the table.

The summary attribute will be removed and replaced by the caption element on all tables and it's translation key:value removed from the localization file in upcoming changes.

colinbruce (approved):

PR #2870, opened 2019-06-27
Use `Config` gems settings for s3

jsugarman:
#### What
Use `Config` gems settings for s3

#### Ticket

[CBO-794](https://dsdmoj.atlassian.net/browse/CBO-794)

#### Why
Tidy of secrets and prep for
migration of secrets to k8s


kmahern (approved):
lgtm

PR #2875, opened 2019-06-28
CBO 686 - Implement Google tag manager

colinbruce:
#### What
Implement Google Tag Manager in CCCD

#### Ticket
[CCCD – Implement Google Tag Manager](https://dsdmoj.atlassian.net/browse/CBO-686)

#### Why
GTM allows a more granular approach to GA event tracking 

#### How
* Update the Javascript to allow passing in GTM container IDs per environment
* Allow the container to start in GTM mode if the environment variable is set, otherwise fall back to the original, plain, Google Analytics version.
* Add a non-JS compatible GTM script to be called if invoking GTM and JS is turned off

--------

#### TODO (wip)

 - [x] create a GTM account and add a demo container
 - [x] test on the demo environment to ensure it works as expected

### DIFF
--- a/spec/models/google_analytics/data_tracking_spec.rb
+++ b/spec/models/google_analytics/data_tracking_spec.rb
@@ -39,5 +39,49 @@ module GoogleAnalytics
         end
       end
     end
+
+    context '#tag_manager?' do
+      subject(:tag_manager) { described_class.tag_manager? }
+
+      context 'when the adapter is google_tag_manager' do
+        before { allow(described_class).to receive(:adapter_name).and_return(:gtm) }
+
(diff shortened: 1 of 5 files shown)


spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -39,5 +39,56 @@ module GoogleAnalytics
         end
       end
     end
+
+    describe 'type methods' do
+
```
jsugarman: better without line-break under describe IMO.

jsugarman (approved):
👍 LGTM

PR #2879, opened 2019-07-02
Cbo 608 remove json uploader

willmcb:
#### What
Removes JSON import functionality for external users as this is hardly used since the introduction of the API.

#### Ticket

[CBO-608 Remove JSON Uploader](https://dsdmoj.atlassian.net/browse/CBO-608)


jsugarman:
some remaining clearup and documentation:

- `./app/views/external_users/json_document_importers/create.js.erb` looks removable
- add api/release_notes entry for 15th July
- check locales/en.yml for related translations (have remove 1 i found)

jsugarman:
this whole directory with both files should be removable `app/views/external_users/json_document_importers/format_error.js.erb`

jsugarman:
and this 
`app/assets/javascripts/modules/Modules.CustomFileUpload.js`

spec/models/json_document_importer_spec.rb
```diff
@@ -1,193 +0,0 @@
-require 'rails_helper'
-
-describe JsonDocumentImporter do
```
jsugarman, line 3: woohoo! hated this spec which was not really testing much anyway

jsugarman:
`config/schemas/basic_claim_schema.json` this too

plus amend `app/models/claim_json_schema_validator.rb` inline with removal

jsugarman (approved):
👍 Looks good @willmcb 
be good to have a single commit to revert or review from a historical point of view

PR #2884, opened 2019-07-03
Bugfix: allow discontinuance? to return false if no case_type set

colinbruce:
#### What: 

Add Safe Navigation Operator to case_type in the `discontinuance?` method of claims

#### Why: 
If a provider saved a draft before setting the case type, the `discontinuance?` check would fail as there was no case type to validate against


jsugarman (approved):
👍

PR #2891, opened 2019-07-09
Travel Automation report Fix

colinbruce:
What:
Update the Travel Automation Report mileage_rate calculation

Why:
The overnight report was failing because someone claimed a bike journey
and the rates are not included in the car_mileage_rates!

How:
Change the mileage_rate calculation to use the combined mileage rate


jsugarman (approved):
👍

PR #2892, opened 2019-07-11
[CBO-785][CB0-697] External_users updates

naseberry:
#### What
Apply accessibility, upgrade to v2 stylesheet and coding standards

#### Ticket
[CBO-787](https://dsdmoj.atlassian.net/browse/CBO-787)
[CBO-697](https://dsdmoj.atlassian.net/browse/CBO-697)

#### Why
To improve accessibility of all views within the `external_users` path

#### How
- Update code to fix identified accessibility errors
- Upgrade stylesheets from version 1 to version 2 of specific views
- Applying coding standard to touched files to improve the coding standards

### DIFF
--- a/app/views/external_users/claims/_main_claims_list.html.haml
+++ b/app/views/external_users/claims/_main_claims_list.html.haml
@@ -5,90 +5,92 @@
         = your_claims_header
       %span.claim-count
         - if @claims.none?
-          = t("shared.no_claims_found")
+          = t('shared.no_claims_found')
         - else
-          = "Showing "
-          #{pluralize(@claims.unscope(:order).size, 'claim')} of #{@claims.total_count}
+          = "Showing #{pluralize(@claims.unscope(:order).size, 'claim')} of #{@claims.total_count}"
           - if params[:search].present?
-            matching "#{params[:search]}"
+            = "matching #{params[:search]}"
     %thead
-
       - if user_requires_scheme_column?
-        %th{scope:'col'}
+        %th{ scope: 'col' }
           = t('.type')
 
-      %th{scope:'col'}
-        = sortable 'case_number', t(".case_number"), 'aria-label': t('.case_number_sort_label'), tabindex: 0
+      %th{ scope: 'col' }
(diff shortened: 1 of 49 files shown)


app/views/external_users/claims/_main_claims_list.html.haml
```diff
@@ -5,90 +5,93 @@
         = your_claims_header
       %span.claim-count
         - if @claims.none?
-          = t("shared.no_claims_found")
+          = t('shared.no_claims_found')
         - else
-          = "Showing "
+          = 'Showing '
           #{pluralize(@claims.unscope(:order).size, 'claim')} of #{@claims.total_count}
```
willmcb: I'm not sure this will work in terms of string interpolation - `"#{needs to be in double quotes}"`

features/step_definitions/common_claim_steps.rb
```diff
@@ -138,18 +138,18 @@
 end
 
 When(/^I click "Continue" in the claim form and move to the '(.*?)' form page$/) do |page_title|
-  original_title = page.title
+  original_header = page.first('h2').text
```
willmcb, line 141: Why move away from the page object model?

willmcb:
LGTM other than the single quote string issue.

Also the `page.title` to `page.first('h2').text` - if you prefer 'heading' to title you should be able to shange the page object model to reflect this.

app/views/external_users/claims/_main_claims_list.html.haml
naseberry: It works without double quotes, however for consistency, I have update it to use double quotes.

features/step_definitions/common_claim_steps.rb
naseberry: I am actually reverting to what it was before my prior commit and page object model will be revisited in future.

willmcb (approved):

colinbruce (approved):
LGTM :+1: 
As discussed, feel free to ignore the CodeClimate warning...

PR #2895, opened 2019-07-11
CBO-803: Add filter to the injection_response_service

colinbruce:
#### What
Reduce the noise in the slack injection channels

#### Ticket
[CCCD: SPIKE: Investigate methods to improve injection reporting](https://dsdmoj.atlassian.net/browse/CBO-803)

#### Why
After the CCLF injection reporting incident, it was determined that the injection channels in slack are very noisy

#### How
Apply some login in the injection response service to only send a slack message when the injection fail is not ignorable

#### TODO (wip)

 - [x] move the `ignorable` logic to the injection_attempt model
 - [X] Create some `ignorable` rules to prevent messages being sent

### DIFF
--- a/app/services/injection_response_service.rb
+++ b/app/services/injection_response_service.rb
@@ -10,11 +10,11 @@ class InjectionResponseService
     slack.build_injection_payload(@response)
     return failure(action: 'run!', uuid: @response['uuid']) unless @claim
 
-    injection_attempt = create_injection_attempt
+    injection_attempt
     if Settings.aws&.sqs&.response_queue_url
       slack.send_message!
     else
-      slack.send_message! unless injection_attempt.succeeded?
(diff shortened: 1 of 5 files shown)


app/services/injection_response_service.rb
```diff
@@ -10,11 +10,11 @@ def run!
     slack.build_injection_payload(@response)
     return failure(action: 'run!', uuid: @response['uuid']) unless @claim
 
-    injection_attempt = create_injection_attempt
+    injection_attempt
     if Settings.aws&.sqs&.response_queue_url
       slack.send_message!
```
colinbruce, line 15: This `if` block was temp code for debugging the K8s injection.
It can, and should, be removed once that has become more stable and this will revert to the single line 17

jsugarman (approved):
👍 LGTM

PR #2903, opened 2019-07-16
Remove fuubar reference from guard file

colinbruce:
After the Fuubar gem was removed, guard had stopped working

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -14,7 +14,7 @@ guard :jasmine, all_on_start: false do
   end
 end
 
-guard :rspec, cmd: "bundle exec rspec --format Fuubar --color", all_on_start: false do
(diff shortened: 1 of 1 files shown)


app/models/injection_attempt.rb
```diff
@@ -29,4 +29,10 @@ def error_messages
     messages = data&.fetch(:errors)&.map { |child| child[:error] }
     messages || []
   end
+
+  private
+
+  def ignorable?
```
jsugarman: do you want this in this PR?

colinbruce: No... that was _supposed_ to have been committed on a different branch 🤦‍♂ 

will force push a replacement

jsugarman: 👍

jsugarman (approved):
👍

PR #2905, opened 2019-07-17
Remove debug statement from archive task

jsugarman:
#### What
Remove debug statement from archive task

#### Why
Seems to be interfering with kibana ability to collect logs from the
cronjob that calls this.


kmahern (approved):
👍

PR #2906, opened 2019-07-17
Fix rubocop violations for bin/secret parser

jsugarman:
#### What
Fixes rubocop violations, or excludes cops, for bin/secret parser

#### Why
This is excluded from codeclimate anyway (don't
know how/why, perhaps because it is a "binary")
but fails running rubocop locally without these
fixes/exclusions.


colinbruce (approved):
:+1:

PR #2915, opened 2019-07-24
Set kubernetes image pull policy to always

jsugarman:
#### What
Always pull the specified docker image

#### Why
In order to honor the use of latest tags.

#### How
set kubernetes deployment config for app container images
to always pull - default is IfNotPresent, which only pulls
the image if does not already exist, which means `app-latest`
tags will never work as it does already exist.


willmcb (approved):
👍

PR #2916, opened 2019-07-24
Cbo 829 mi report generation error

willmcb:
#### What
Adds extra logging around the creation of MI Reports.

#### Ticket

[CBO-829](https://dsdmoj.atlassian.net/browse/CBO-829)

#### Why

To try and understand why we had a failed MI report generation.

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

### DIFF
--- a/app/services/stats/management_information_generator.rb
+++ b/app/services/stats/management_information_generator.rb
@@ -24,17 +24,38 @@ module Stats
     # TODO: separate data retrieval from exporting the data itself
     # keeping existent behaviour for now
     def generate_new_report
-      CSV.generate do |csv|
-        csv << Settings.claim_csv_headers.map { |header| header.to_s.humanize }
-        Claim::BaseClaim.active.non_draft.find_each do |claim|
-          ClaimCsvPresenter.new(claim, 'view').present! do |claim_journeys|
-            if claim_journeys.any?
-              claim_journeys.each do |claim_journey|
-                csv << claim_journey
+      begin
+        LogStuff.send(:info, "Report generation started...")
+        CSV.generate do |csv|
+          LogStuff.send(:info, "Adding CSV headers...")
+          csv << Settings.claim_csv_headers.map { |header| header.to_s.humanize }
+          LogStuff.send(:info, "Adding CSV headers complete")
+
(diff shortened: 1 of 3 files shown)


app/services/stats/management_information_generator.rb
```diff
@@ -21,21 +21,42 @@ def call
 
     attr_reader :format
 
+    def log_error(error)
+      LogStuff.send(:error, 'MI Report generation error has occured:')
```
jsugarman: This will create three error log entries for one error. I'd recommend having one error (or debug) entry that has attributes for the details you want:
e.g. 
```
LogStuff.send(
  :error, 
  self.class.name, 
  error_message: "#{error.class} - #{error.message}",
  error_backtrace: error.backtrace.inspect.to_s
) do
  'MI Report generation error'
end
```

app/services/stats/management_information_generator.rb
```diff
@@ -21,21 +21,42 @@ def call
...
-                csv << claim_journey
-              end
+              log_info("Adding journey for claim: #{claim.id}")
+              claim_journeys.each { |journey| csv << journey }
             end
           end
         end
       end
+    rescue StandardError => e
+      log_error(e)
```
jsugarman, line 62: You can right specs that raise an error and then "spy" on the behaviour to make sure `log_error` is called and even `LogStuff` is called with relevant arguments.

have a look at [this example](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/spec/models/timed_state_transitions/batch_transitioner_spec.rb#L60)

jsugarman (changes requested):
I'd recommend using one LogStuff call per error, and some specs to cover the logging. Have added comment with examples

app/services/stats/management_information_generator.rb
willmcb: 👍

app/services/stats/management_information_generator.rb
willmcb: Thanks @jsugarman I'll implement these changes

kmahern (approved):
looks good 👍

PR #2938, opened 2019-08-05
Update bundler version 2.0.1 --> 2.0.2 used by Dockerfile

jsugarman:
#### What
Update bundler version based on `Gemfile.lock`

#### Why
Updates 2.0.1 --> 2.0.2 and will keep inline
with latest `Gemfile.lock`. It will also prevent warnings
such as:
```
Warning: the running version of Bundler (2.0.1) is older than the version that created the lockfile (2.0.2). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,8 +33,9 @@ aliases:
     run:
       name: Install dependencies
       command: |
-        gem install bundler -v 2.0.1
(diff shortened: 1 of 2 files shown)


willmcb (commented):
👍

willmcb (approved):

willmcb (approved):
👍

PR #2939, opened 2019-08-05
Re-add accidently deleted binary

jsugarman:
#### What
Re-adds a binary that was accidentally deleted
in a previous commit

#### How
Copied it out of the version history


willmcb (approved):
👍

PR #2946, opened 2019-08-16
Amend live-1 DI to use named SQS queues

jsugarman:
#### What
Amend k8s/live-1 data injection to use new, named SQS queues

#### Ticket

[CBO-863](https://dsdmoj.atlassian.net/browse/CBO-863)

#### Why

Following cloud-platforms addition of the ability
to create named SQS queues, new queues have been
generated, with names. This enables the live-1
(kubernetes) version of CCCD to use the
same mechanism for receiving messages as current
(template-deploy) version. It also means that
in the event queues need to be recreated the
fixed name hides the new SQS queue URLs, making
the codebase more relient to infrastructure changes.

#### How
 - remove response_queue_url (previously used for k8s)
 - add response queue name secret to k8s config
 - consume named response queue from secret
 - adjust conditional slack message sending inline

### DIFF
--- a/app/services/claims/injection_channel.rb
+++ b/app/services/claims/injection_channel.rb
@@ -2,7 +2,7 @@ module Claims
   class InjectionChannel
     def self.for(claim)
       return 'cccd_development' if claim.nil?
-      return 'cccd-k8s-injection' if Settings.aws&.sqs&.response_queue_url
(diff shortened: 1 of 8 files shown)


naseberry:
LGTM

naseberry (approved):

PR #2961, opened 2019-09-02
CBO-384 Check your claim page

naseberry:
#### What
New design implementation and accessibility related tweaks for external_users ‘Check your claim’ page.

#### Ticket
[CBO-384](https://dsdmoj.atlassian.net/browse/CBO-384)

#### Why

#### How
- Update markup
- Implement accessible tables
- New functions to calculate VAT and Gross
- Spec for new functions

### DIFF
--- a/app/views/external_users/claims/_summary_claims_content.html.haml
+++ b/app/views/external_users/claims/_summary_claims_content.html.haml
@@ -1,19 +1,3 @@
-%p
-  = t('external_users.claims.check_your_claim.help_text')
-
-- unless claim.invalid_steps.empty?
-  .panel.panel-border-wide
-    %h3.heading-small
-      = t('external_users.claims.check_your_claim.missing_information.header')
-    %p
-      = t('external_users.claims.check_your_claim.missing_information.help_text')
-
-    %ul.list.list-bullet
-      - claim.summary_sections.each do |section, associated_step|
-        - if claim.invalid_steps.include?(associated_step)
-          %li
-            = t("external_users.claims.#{section}.summary.header")
-
 - claim.summary_sections.each do |section, associated_step|
-  - if claim.accessible_step?(associated_step)
+  - if claim.accessible_step?(associated_step) && section.to_s != 'additional_information'
     = render partial: "external_users/claims/#{section}/summary", locals: { claim: claim, step: associated_step, section: section, editable: claim.editable_step?(associated_step) }
--- a/app/views/external_users/claims/_summary_fees.html.haml
+++ b/app/views/external_users/claims/_summary_fees.html.haml
@@ -1,4 +1,4 @@
-%h3.heading-medium
+%h2.heading-large
   = local_assigns.has_key?(:header) ? header : t('common.fees')
   - if local_assigns[:editable]
     = link_to t('.change'), edit_polymorphic_path(claim, step: step, referrer: :summary), class: 'link-change', 'aria-label': t('.change_link_label')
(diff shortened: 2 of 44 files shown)


app/interfaces/api/v1/external_users/claims/advocate_claim.rb
```diff
@@ -27,15 +27,15 @@ def local_t(attribute)
       namespace '/' do
         desc 'DEPRECATED: Create an Advocate final claim. see advocates/final endpoint'
         post do
-          deprecate(datetime: Time.new(2019, 8, 30), link: "#{request.base_url}/api/release_notes")
+          deprecate(datetime: Time.new(2020, 8, 30), link: "#{request.base_url}/api/release_notes")
```
willmcb: Why have these dates changed?

app/views/external_users/claims/_summary_claims_content.html.haml
```diff
@@ -1,19 +1,3 @@
...
-
-    %ul.list.list-bullet
-      - claim.summary_sections.each do |section, associated_step|
-        - if claim.invalid_steps.include?(associated_step)
-          %li
-            = t("external_users.claims.#{section}.summary.header")
-
 - claim.summary_sections.each do |section, associated_step|
-  - if claim.accessible_step?(associated_step)
+  - if claim.accessible_step?(associated_step) && !section.to_s.eql?('additional_information')
```
willmcb: think we can do 

```ruby
section.to_s != 'additional_information'
```
here

app/views/external_users/claims/_summary_fees.html.haml
```diff
@@ -36,18 +36,53 @@
...
+          %td.numeric
+            = ''
+          %td.numeric
+            = ''
+          %td.numeric
+            = ''
+          %td.numeric
+            = claim.fixed_fees_total if section.to_s.eql?('fixed_fees')
+            = claim.misc_fees_total if section.to_s.eql?('misc_fees')
+          %td.numeric
```
willmcb, line 82: Could these be in a loop rather than declared literally? 

Also I think you could use a ternary operator given there are only two possibilities:

```ruby
= section.to_s.eql?('fixed_fees') ? claim.fix_fees_total : claim.misc_fees_total
```

app/views/external_users/claims/interim_fee/_summary.html.haml
```diff
@@ -13,83 +13,79 @@
...
+          %dd.govuk-summary-list__value
+            = fee.fee_type.description
 
-          - unless fee.is_interim_warrant?
-            %tr
-              %th{ class: 'bold' }
-                = t('shared.summary.ppe_total')
-              %td
-                = fee.quantity
+        - unless fee.is_interim_warrant?
```
willmcb, line 23: Maybe not for this PR but there seems to be some duplication in the views that could be added to partials with params to dry things out a bit.

willmcb (commented):
Nothing major but it would be good to talk about the changes with you 👍

app/interfaces/api/v1/external_users/claims/advocate_claim.rb
naseberry: `./spec/support/shared_examples_for_api.rb:325` expected the current time to be less than or equal to the above time.
As we are now in September, the spec expected time is now greater, therefore tests were failing.
I set a new time in the future in order to mock the test case and allow it to pass the given scenario.

app/views/external_users/claims/interim_fee/_summary.html.haml
naseberry: I agree, there are areas for improvement, however I am currently focused on the immediate tasks.

willmcb (approved):

PR #2965, opened 2019-09-05
Increase sunset date for deprecated api endpoint

jsugarman:
#### What
Increase sunset date for deprecated api endpoint

#### Why
Poor uptake of new endpoints means alot
of vendor software will cease to function
if we remove the `api/external_users/claims`
endpoint.


kmahern (approved):
lgtm 👍

PR #2998, opened 2019-09-26
Amend Dockerfile bundler config to exclude all but production gems

jsugarman:
#### What
Amend Dockerfile bundler config to exclude all but production gems

#### Why
Dockerfile was previously using faulty config setting resulting
in all gems being installed.


naseberry (approved):
LGTM

PR #2999, opened 2019-09-26
Replace current PDF generation mechanism

jsugarman:
#### What
Replace pdftk/pdf-forms gem with wkhtmltopdf/wicked_pdf
also bumps alpine on docker builds to 3.10 (latest)

#### Ticket

[CBO-916](https://dsdmoj.atlassian.net/browse/CBO-916)

#### Why

- PDFTK is old and increasing unsupported. Alpine 3.9+
  does not have an installable pdftk package.

- current disc evidence coversheet PDF is inaccessible and
  failed the DAC report. rendering from html gives
  us more control.

#### How

- Add model, controller and views for prefilled form
  To be used to generate the standard pdf
  based on html templates.

- Add `wkhtmltopdf` and `ttf-ubuntu-font-family` to alpine
  build - required by `wicked_pdf`

- Remove unneeded disc evidence coversheet builder,
  associated gems, config and binarys

--------

#### TODO (wip)

 - [X] generate PDF from html form
 - [X] add basic styling
 - [X] amend alpine build to work with new mechanism
 - [x] add unit tests/rspec
 - [x] add feature test or request spec
 - [x] improve accessibility
 - [x] handle button options - implement any design review changes


app/controllers/external_users/disc_evidence_coversheets_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  def create
+    @disc_evidence_coversheet = DiscEvidenceCoversheet.new(@claim, disc_evidence_coversheet_params)
+
+    respond_to do |format|
+      format.html do
+        render 'pdfs/templates/disc_evidence_coversheet.html.erb'
+      end
+
+      format.pdf do
+        # render(
```
jsugarman: remove

app/controllers/external_users/disc_evidence_coversheets_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+        # send_data(
+        #   pdf,
+        #   filename: "disc_evidence_coversheet_#{@disc_evidence_coversheet.case_number}.pdf",
+        #   type: 'application/pdf',
+        #   disposition: 'inline'
+        # )
+
+        render(
+          pdf: "disc_evidence_coversheet_#{@disc_evidence_coversheet.case_number}",
+          # save_to_file: Rails.root.join('tmp', "disc_evidence_coversheet_#{SecureRandom.uuid}.pdf"),
```
jsugarman: remove

app/models/disc_evidence_coversheet.rb
```diff
@@ -0,0 +1,76 @@
...
+              :data_storage_type
+
+  attr_reader :data_storage_type
+
+  gov_uk_date_accessor :current_date, :claim_submitted_at
+
+  def initialize(claim, params = {})
+    @claim = claim
+    @params = params
+    super(params) # use activemodel params writing behaviour
```
jsugarman: remove comment

app/models/disc_evidence_coversheet.rb
```diff
@@ -0,0 +1,76 @@
...
+    @provider_account_no || @claim.supplier_number
+  end
+
+  def provider_address
+    @provider_address&.strip
+  end
+
+  private
+
+  def date_from_parts(field)
```
jsugarman: handle this in date accessor elsewhere ideally

app/services/gov_uk/date_accessor.rb
```diff
@@ -0,0 +1,45 @@
+module GovUk
```
jsugarman: add comment on what used for - basically to enable use of gov_uk_date_field helper on virtual attributes

app/views/pdfs/templates/disc_evidence_coversheet.html.erb
```diff
@@ -0,0 +1,162 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+      <%# stylesheet_link_tag(wicked_pdf_asset_base64("pdf"), media: 'all') %>
```
jsugarman: remove unless base64'd version works n server - did not last time i tried

docker/Dockerfile
```diff
@@ -10,16 +10,20 @@ RUN set -ex
...
 # - nodejs: for ExecJS and asset compilation
 # - runit for process management (because we have multiple services)
-# - libreoffice/pdftk: for pdf conversion
-# - ttf-ubuntu-font-family: required for libreoffice pdf conversion
+# - libreoffice: for pdf conversion
+# - ttf-ubuntu-font-family: needed by wkhtmltopdf/wicked_pdf
+# - wkhtmltopdf: for pdf generation from html
 # - redis: for backend key-value store
 #
+# might ALSO need the following for wkhtmltopdf
```
jsugarman: remove

app/models/disc_evidence_coversheet.rb
```diff
@@ -0,0 +1,75 @@
+class DiscEvidenceCoversheet
+  include ActiveModel::Model
+  include GovUk::DateAccessor
+
+  attr_accessor :claim_id
+  attr_writer :fee_scheme,
```
willmcb, line 6: why not add `:data_storage_type` to the call to `attr_accessor`?

app/assets/stylesheets/pdf.css.scss
```diff
@@ -0,0 +1,111 @@
...
+    vertical-align: bottom;
+    width: 60%;
+  }
+}
+
+.pdf-form {
+  .pdf-field-row {
+    td {
+      padding-bottom: 20px;
+    }
```
naseberry: Duplication of line: 21

docker/Dockerfile
```diff
@@ -47,8 +48,8 @@ WORKDIR /usr/src/app
 ######################
 # DEPENDENCIES START #
 ######################
-# Env vars needed for dependency install and asset precompilation
-ENV RAILS_ENV='production'
+# Env vars needed for dependency install and asset precompilation?? maybe not
```
willmcb, line 51: Do we know this now, should we remove the comment if so?

spec/models/ability_spec.rb
```diff
@@ -19,6 +20,7 @@
     it { should_not be_able_to(:update, UserMessageStatus.new) }
     it { should_not be_able_to(:create, Document.new) }
     it { should_not be_able_to(:update_settings, User.new) }
+    it { should_not be_able_to(:create, DiscEvidenceCoversheet.new) }
```
willmcb, line 23: 👍

willmcb (approved):
lgtm - see the non-blocking comments I made 👍

docker/Dockerfile
jsugarman: Not sure so just added the comment as this was yak shaving issues coming up. But from my understanding a single docker command like this does not make the variable available to subsequent commands (i.e the RUN bundle install command in particular). We are handling production-only gem install via the bundle command directly now so probably not needed.

you could have a play and chck whether removing this line on a separate branch and then build changes the gems installed. I am pretty sure it won't.

app/assets/stylesheets/pdf.css.scss
jsugarman: 👍

app/controllers/external_users/disc_evidence_coversheets_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+        render 'pdfs/templates/disc_evidence_coversheet.html.erb'
+      end
+
+      format.pdf do
+        render(
+          pdf: "disc_evidence_coversheet_#{@disc_evidence_coversheet.case_number}",
+          margin: { top: 10, bottom: 10, left: 10, right: 10 },
+          template: 'pdfs/templates/disc_evidence_coversheet.html.erb',
+          disposition: 'inline'
+        )
```
naseberry: **Accessibility**: 
- Please add a document title using `title: 'Disc Evidence Coversheet'`
- Document language is missing, however, wicked_pdf doesn't seem to have an option to set this

app/models/disc_evidence_coversheet.rb
jsugarman: `super(@params)` handles setting it via activemodel mixin. i have only added accessors and writers needed by the controller new or create actions. readers are either defined individually here to specify defaults or additional processing required or, if no additional processing/defaults required then just a reader.

`data_storage_type` is the only attribute that does not require additional processing/defaults. 

i don't think it will break anything to make it an accessor so could simplify as you suggest but it is not required as creation/write is done via `super(@params)`

app/assets/stylesheets/pdf.css.scss
jsugarman: changed that @naseberry. if you could re-review would be appreciated

app/controllers/external_users/disc_evidence_coversheets_controller.rb
jsugarman: 👍 done

app/assets/stylesheets/pdf.css.scss
naseberry: :+1:  LGTM

PR #3003, opened 2019-09-27
Apply accessibility updates to static pages views

naseberry:
#### What
Accessibility and coding standard updates to pages views.

#### Ticket
[CBO-927](https://dsdmoj.atlassian.net/browse/CBO-927)

#### Why

#### How
- Minor accessibility updates
- Improve coding standard

#### NOT DONE
 - pages related to api documentation
 - localization of long form content, i.e. app/views/pages/tandcs.haml

### DIFF
--- a/app/views/pages/timed_retention.html.haml
+++ b/app/views/pages/timed_retention.html.haml
@@ -1,29 +1,29 @@
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
-
-- stale_weeks    = Settings.timed_transition_stale_weeks
-- archived_weeks = Settings.timed_transition_pending_weeks
+= render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
 
 %section.timed-retention
   %p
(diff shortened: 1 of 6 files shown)


app/views/pages/timed_retention.html.haml
```diff
@@ -1,29 +1,29 @@
...
+      = t('draft').capitalize
     %li
-      authorised
+      = t('authorised').capitalize
     %li
-      part-authorised
+      = t('part_authorised').capitalize
     %li
-      refused
+      = t('refused').capitalize
```
willmcb: I think this could be wrapped up into an array that builds the list dynamically something like:

```ruby
states = ['draft', 'authorised', 'part_authourised', 'refused', 'rejected']
states.each_with_index do |state, index|
  index.eql?(states.length - 1) ? '%li' : '%li.last_element'
  = t(state).capitalize 
end
```
Though this is not a big thing, and definitely not a blocker. More of something we could look into to further DRY out the views. I have seen similar hardcoded lists throughout the application, some form of view helper would be a better way of managing the repetition.

willmcb (approved):
👍  lgtm

PR #3008, opened 2019-09-30
Apply accessibility updates to provider management views

naseberry:
#### What
Accessibility and coding standard updates to provider management views.

#### Ticket
[CBO-788](https://dsdmoj.atlassian.net/browse/CBO-788)

#### Why

#### How
- Minor accessibility updates
- Improve coding standard

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -19,8 +19,11 @@
 # To learn more, please read the Rails Internationalization guide
 # available at http://guides.rubyonrails.org/i18n.html.
 en:
-  TrueClass: 'Yes'
-  FalseClass: 'No'
+  global_no: &global_no 'No'
+  global_yes: &global_yes 'Yes'
+
+  TrueClass: *global_yes
+  FalseClass: *global_no
   layouts:
     phase_banner_html: 'This is a new service – your <a href="%{feedback}"><span class="visuallyhidden">Give your </span>feedback</a> will help us to improve it. You can <a href="%{feedback_type}">report a fault here</a>.'
     primary_navigation:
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -2551,3 +2554,9 @@ en:
       comment_hint: "Don't include any personal or sensitive information."
       send: Send
       rating: 'Overall, how satisfied were you with this service?'
+
+
+
+  # global
+  global_no: "No"
```
jsugarman: i'd put these at the top under `en:`
you could also do this as further refactoring to anchor and reuse

```
en: 
  # anchor
  global_yes: &global_yes Yes
  global_no: &global_no No

  # and use
  my_view:
     anwser_yes: *global_yes
     anwser_no: *global_no
```
we should also probably have a standard/convention relating to use of quotes (single quotes are better IMO and are the default if no quotes specified, so no quotes actually needed except for interpolation and escaping chars)

jsugarman (commented):
looks good assuming the relevant views have all been visually checked.

I would take the opportunity to tidy up the Yes/No translations as they are used in various places. see my individual comment.

config/locales/en.yml
naseberry: 👍

jsugarman (approved):
👍 LGTM

separate conversation over whether we should rely on yaml's default of not requiring single quotes for strings as they default to being single quoted 
http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html
```
global_yes: 'Yes'
```
is the same as:
```
global_yes: Yes
```

naseberry:
> 👍 LGTM
> 
> separate conversation over whether we should rely on yaml's default of not requiring single quotes for strings as they default to being single quoted
> http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html
> 
> ```
> global_yes: 'Yes'
> ```
> 
> is the same as:
> 
> ```
> global_yes: Yes
> ```

I believe _Yes_ and _No_ are reserved, therefore I have to wrap them in quotes.

PR #3023, opened 2019-10-09
Add cccd-migrator config and deployment script

jsugarman:
#### What
Add cccd-migrator config and deployment script

#### Why
need to duplicate current s3 buckets
to live-1 environs

#### How
A separate repository - [cccd-migrator](https://github.com/ministryofjustice/cccd-migrator) -creates a simple
cli for migrating s3 buckets (and eventually
DBs)

This app provides a wrapper around `aws s3` tools
```
aws s3 sync --delete s3://$SOURCE_BUCKET_NAME s3://$DESTINATION_BUCKET_NAME
```

Note that:
`aws s3 sync` is used with the IAM user
of the live-1/destination s3 bucket and
via the source bucket's bucket policy (which
must grant relevant actions and resources to
that IAM user).


ddippolito (approved):
Lgtm 🚀

PR #3025, opened 2019-10-10
Delete the previous td migrator pod within the specified namespace

jsugarman:
#### What

Delete the previous template-deploy migrator pod within the
specified namespace.

#### Why

Prior to this it was not explicitly specifying the namespace
from which to delete the pod. This could resulting in a pod
in a different namespace being deleted.


kmahern (approved):
👍

PR #3029, opened 2019-10-14
Kubernetes fix master tag fail

kmahern:
#### What

The push_docker_image job in config.yml was failng when the branch was master was failing with the following error "No such image: app:latest". This was because it was looking for an image tagged app that did not exist.

#### Ticket

https://dsdmoj.atlassian.net/browse/CBO-900

#### Why

To fix CircleCI workflow

#### How

Tag ${ECR_ENDPOINT}/${GITHUB_TEAM_NAME_SLUG}/${REPO_NAME}:${CIRCLE_SHA1} with :latest instead of app

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


jsugarman (approved):
👍 **not a blocker** can we change this to `app-latest` in line with scripts and in the event we have "other" pods such as `migrator-latest` (which i am currently working on)

PR #3030, opened 2019-10-15
Kubernetes rails migrations

kmahern:
#### What
Enable rails migrations in Kubernetes

#### Ticket
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&modal=detail&selectedIssue=CBO-899

#### Why
In order to be able to update the db schema in Live1.

#### How
Created a new environment variable to determine whether or not a rails db migration is carried out in live1, LIVE1_DB_TASK. The is set in .circleci/config.yml. If it is set to 'migrate' any pending migrations will be applied in live1. Currently it is set to 'none' so that migrations are not applied by default in Template Deploy.

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


.circleci/config.yml
```diff
@@ -137,6 +137,7 @@ references:
           --build-arg BUILD_DATE=$(date +%Y-%m-%dT%H:%M:%S%z) \
           --build-arg COMMIT_ID=${CIRCLE_SHA1} \
           --build-arg BUILD_TAG=${CIRCLE_BRANCH} \
+          --build-arg LIVE1_DB_TASK=none \
```
jsugarman: you going to set this to migrate by default

jsugarman (approved):
Just change to `migrate` as discussed and we can merge this.

PR #3036, opened 2019-10-18
hotfix for disc evidence coversheet

naseberry:
#### What
Fix for disc evidence NoMethodError

#### Why
Sentry highlighted the issue, if claim is empty, then `name` method will be undefined

#### How
Check if method is defined.

### DIFF
--- a/app/models/disc_evidence_coversheet.rb
+++ b/app/models/disc_evidence_coversheet.rb
@@ -34,7 +34,7 @@ class DiscEvidenceCoversheet
   end
 
   def fee_scheme
-    @fee_scheme || claim.fee_scheme.name.upcase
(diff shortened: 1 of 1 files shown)


app/models/disc_evidence_coversheet.rb
```diff
@@ -34,7 +34,7 @@ def current_date
   end
 
   def fee_scheme
-    @fee_scheme || claim.fee_scheme.name.upcase
+    @fee_scheme || defined?(claim.fee_scheme.name)
```
jsugarman, line 37: you can just do a `claim&.fee_scheme&.name` here.

i saw this sentry error but I assumed it was an abberation as every claim should have a fee scheme and every fee scheme has a name.

but this would prevent that error in such circumstances

PR #3037, opened 2019-10-18
hotfix for disc evidence coversheet

naseberry:
#### What
Fix for disc evidence NoMethodError

#### Why
Sentry highlighted the issue, if claim is empty, then name method will be undefined

#### How
Check if method is defined.

### DIFF
--- a/app/models/disc_evidence_coversheet.rb
+++ b/app/models/disc_evidence_coversheet.rb
@@ -34,15 +34,15 @@ class DiscEvidenceCoversheet
   end
 
   def fee_scheme
-    @fee_scheme || claim.fee_scheme.name.upcase
+    @fee_scheme || claim&.fee_scheme&.name
   end
 
   def agfs?
-    fee_scheme.eql?('AGFS')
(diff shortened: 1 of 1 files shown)


app/models/disc_evidence_coversheet.rb
```diff
@@ -34,7 +34,7 @@ def current_date
   end
 
   def fee_scheme
-    @fee_scheme || claim.fee_scheme.name.upcase
+    @fee_scheme || claim&.fee_scheme&.name
```
jsugarman, line 37: I think you still need the upcase on this. As it is displayed like that in the PDF

jsugarman: Actually looks like fee schemes names are upper case in any event so the checks like `fee_scheme.eql?('AGFS')`on line37 and 41 in this file would work. Could make those checks case insensitive for safety e.g (`fee_scheme.match?(/AGFS/i)` or `fee_scheme.upcase.eql?('AGFS')`)

jsugarman (approved):
👍 just a **non-blocking** comment to consider

PR #3041, opened 2019-10-24
CBO-943: email notifications refused rejected

colinbruce:
#### What
We've had provider feedback to say even though they are signed up to receive an email notification on a case, they are not doing so for rejected or refused claims.
#### Ticket
[Email notifications for refused/rejected claims - fix issue](https://dsdmoj.atlassian.net/browse/CBO-943)

#### Why
Providers should be receiving email notifications when a caseworker sends them a message on any case.

#### How
MOve the logic for sending emails into the message itself.  As a message is created, if it was sent by a caseworker, and the provider has opted-in to receive emails and is still active, an email will be sent

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -66,10 +66,15 @@ class Message < ApplicationRecord
   private
 
   def send_email_if_required
-    return unless sender.persona.is_a?(CaseWorker)
-    return unless claim.creator.send_email_notification_of_message?
-    return if claim.creator.softly_deleted?
-    NotifyMailer.message_added_email(claim).deliver_later
+    NotifyMailer.message_added_email(claim).deliver_later if send_email?
+  end
+
+  def send_email?
+    [
+      sender.persona.is_a?(CaseWorker),
+      claim.creator.send_email_notification_of_message?,
+      claim.creator.active?
+    ].all?
   end
 
   def generate_statuses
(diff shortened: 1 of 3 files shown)


app/models/message.rb
```diff
@@ -65,6 +65,13 @@ def for(object)
 
   private
 
+  def send_email_if_required
+    return unless sender.persona.is_a?(CaseWorker)
```
jsugarman: **not-a-blocker** possible refactor while you are here

```
def email_required?
  [
    sender.persona.is_a?(CaseWorker),
    claim.creator.send_email_notification_of_message?,
    claim.creator.active?
  ].all?
end
```

and use:
```
return unless email_required?
NotifyMailer.message_added_email(claim).deliver_later
```
or
```
NotifyMailer.message_added_email(claim).deliver_later if email_required?
```

spec/models/message_spec.rb
```diff
@@ -126,4 +126,81 @@
       expect(claim.state).to eq 'part_authorised'
     end
   end
+
+  context 'send_message_if_required' do
```
jsugarman: **not-a-blocker** `send_email_if_required`

jsugarman (approved):
LGTM 👍 just a spec wording thing and a possible refactor

colinbruce:
Cancelled the dev deploy deliberately.  Testing on TD dev.

PR #3048, opened 2019-10-31
MIGRATION DAY: Amend env var inline with eventual live-1 domain name

jsugarman:
#### What
Amend env var inline with eventual live-1 domain name

IMPORTANT: needs merging after DNS changes applied and ingress
modifed for live-1 production

The GRAPE_SWAGGER_ROOT_URL is used for
1. internal api calls
2. use in email templates
3. api documenation


colinbruce (approved):
👍

PR #3049, opened 2019-10-31
CBO-973: transfer grad fee issue

colinbruce:
#### What
Prevent API users from submitting grad fees to a transfer claim

#### Ticket
[Provider able to claim 2 graduated fees on single bill](https://dsdmoj.atlassian.net/browse/CBO-973)

#### Why
Caseworkers reported this claim in CCCD which shows both a cracked trial and a transfer fee on the same bill.  this should not be possible for 2 reasons  - 1 it should not be possible to claim 2 graduated fees on the same bill in any scenario. Secondly, a transfer fee is an entirely separate bill type to a regular grad fee and it shouldn’t have been possible to select both a transfer and final fee LGFS bill within a single claim, transfer and final bills have different journeys within the app that cant be combined!
 
#### How
Traditionally we did not validate fee types against claim types, because the web app filters the correct fee pages to the user.  One of the CMS's was submitting grad fees to Transfer cases (at least Trial and Cracked Trial have been seen).  This validation will prevent that happening.

--------

#### TODO (wip)

 - [ ] Alert the software vendors via slack that this change will affect fee submission.  This will potentially break at least one CMS as it is, incorrectly, sending these fee types

### DIFF
--- a/app/validators/fee/graduated_fee_validator.rb
+++ b/app/validators/fee/graduated_fee_validator.rb
@@ -14,6 +14,7 @@ class Fee::GraduatedFeeValidator < Fee::BaseFeeValidator
 
   def validate_claim
     super
+    add_error(:claim, :incompatible_claim_type) if @record.claim&.transfer?
     return unless @record.claim&.final?
     add_error(:claim, :incompatible_case_type) if @record.claim.case_type&.is_fixed_fee?
   end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -253,6 +253,7 @@ en:
           attributes:
             claim:
               incompatible_case_type: Fixed fee invalid on non-fixed fee case types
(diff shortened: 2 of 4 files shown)


jsugarman:
@colinbruce probably worth putting this on the new sandbox and asking litigator api consumers to try it out sooner rather than later. PM may need to negogiate a timeframe for release. One alternative is we could also simply silently reject the inapplicable fee for now (and possibly tidy up historical claims).

jsugarman:
@kmahern to query some stats on existing submitted claims that this validation could break (prevent assessment of)

kmahern:
Stale issue and we have a ticket linked to this branch in the backlog.

jsugarman:
deleting as was fixed by https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4323

PR #3051, opened 2019-11-02
Remove migrator pod deploy script

jsugarman:
#### What
delete unneeded migrator deploy script

#### Why
This was missed and should have been remove
when the pod config and build script was
removed.

see c07b53d66f8c789e8a13cb5361557254aeb3ac12


kmahern (approved):
👍

PR #3056, opened 2019-11-04
Amend and synchronize local and circleCI build and deploy

jsugarman:
#### What
Amend and synchronize local and circleCI build and deploy

#### Why
so local build and deploy create/use same
docker tags and ping endpoint info

#### How
- Synchronize local build with current circleCI build
  regarding rails migration task.

- Refactor circleCI to use app- prefix for build tags

- Refactor build and image arguments in both to
  - remove version number (used by TD only)
  - amend build tag to be the build tag, not branch name
  - add branch name as separate build arg
  - use vars to reduce repetition


kmahern (commented):
👍

kmahern (approved):

PR #3057, opened 2019-11-04
MIGRATION DAY: servicedown page

jsugarman:
#### What
reroute all app traffic to servicedown page

#### Why
As an additional method for preventing
user access to the app (in addition to
DNS change for maintenance page).

#### Ticket

[CBO-984](https://dsdmoj.atlassian.net/browse/CBO-984)

#### Why
d-day migration to live-1

#### How
wildcard match on all routes including `'/'`

`'(*any)'` - matches all routes and none/root. the `any` is merely
needed for rails to bind url parameters to a variable.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -68,8 +68,8 @@ en:
       page_heading: API release notes
     servicedown:
       page_heading: "The service is unavailable"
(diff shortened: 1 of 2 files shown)


jsugarman:
build fails due to smoke test. Discussed with team and not deemed necessary - rely on cloudformation static website and DNS changes instead

PR #3058, opened 2019-11-04
Fix live-1 archive stale cronjob

jsugarman:
#### What
Fix live-1 archive stale cronjob

#### Ticket

[CBO-979](https://dsdmoj.atlassian.net/browse/CBO-979)

#### Why
live-1 cronjob failing on placeholder
prod environment with live data.

The error received was
```
undefined methods `match` for nil:NilClass
```

This is can be raised by paperclip. Could
also be caused by MI data problems.

Better to provide all env vars that are available
in any environment to be confident the job can work
in any environment. Note that zendesk, google api
and other keys are not available to all environments
and so are excluded.

#### How
Add s3 and other secrets to archive stale cronjob

### DIFF
--- a/app/services/transform/claim.rb
+++ b/app/services/transform/claim.rb
@@ -38,6 +38,7 @@ module Transform
     }.freeze
 
     CLAIM_TYPE_CONVERSIONS = {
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM :+1:

PR #3061, opened 2019-11-06
Correct docker tag naming

jsugarman:
#### What
correct tag name pushed and applied

#### Why
The tag from the build was not being used to
deploy. This meant deploys were not getting
deployed and instead the last good image
tag remained on the instances.

It would be nice if we could get this to
displayed as a deployment failure in circleCI.


colinbruce (approved):
LGTM 👍

PR #3064, opened 2019-11-08
Add region to footer

jsugarman:
#### What
output datacenter region in footer

#### Why
so DNS changes will be obvious in the app
for production switchover, both for GDS
delegation changes and for cutover to live-1.


naseberry (approved):

colinbruce (approved):
LGTM :+1:

PR #3066, opened 2019-11-11
Add region conditional http status code

jsugarman:
#### What
have servicedown respond with different http status by region

#### Why
kubernetes/live-1 intercepts 503s at nginx level,
redirecting to its own page. template-deploy does
not. So use 200 for live-1 for now as is not really
needed.

#### Also
- remove other links from servicedown page
  to mitigate users losing there last
  uri query.
- remove GA tracking so users on servivedown page
  eventually fall of GA
- correct GA tracking code for live-1 production
- use basic layout for statistics too to remove repetition


app/views/layouts/basic.html.haml
```diff
@@ -1,33 +1,35 @@
...
 
+- content_for :header_class do
+  = "with-proposition"
+
 - content_for :proposition_header do
   .header-proposition
     .content
       %nav#proposition-menu
-        %a#proposition-name{:href =>"/"}
+        %a#proposition-name{:href => "/"}
```
naseberry: Code consistency: You should most likely replace the hash rocker with a colon as you did on line 4

jsugarman: done

colinbruce (approved):
LGTM :+1:

PR #3068, opened 2019-11-11
Append to any settings references to gamma

jsugarman:
#### What
Check and append to any references to `gamma`
in the code base to add `production`

#### Why
`production` is the `ENV['ENV']` value in live-1
production, but is `gamma` in TD prod environments

- Ensure statistics error notification in
  live-1 production
- ensure GA datatracking enabled in live-1
  production (RailsHost is not actually
  used so this is simply and tidy in this
  respect)


colinbruce:
Do you have a plan for the dump/restore scripts?  There are gamma & adp_gamma references in there too, but the whole script is likely to need to re-writing post live.

colinbruce (approved):
LGTM 👍

PR #3070, opened 2019-11-11
Add app branch to ping output

jsugarman:
#### What
Add the name of the current branch
on which the build is based to ping output

#### Why
To easily identify which branch
is on what environment.

With the move to CirclCI for deployment
we are losing build version numbers and other
information that Jenkis build and deploy
provides for free. This is therefore to replace
some of that lost info.

#### How
docker build scripts already supply this info
and DockerFile already passes this information
and argument on to the environment. We just need
to output it.


colinbruce (approved):
LGTM :+1:

PR #3071, opened 2019-11-11
Put live-1 production into maintenance

jsugarman:
#### What
Put live-1 production into maintenance mode

#### Why
Prevents people using new production app
until we are ready for cutover.

More specifically this will help
to ensure any errors in DNS
delegation from GDS tomorrow do not
impact data - in the event of DNS delegation
going awry, if users are directed to the wrong
load balancer (unlikely), this will
direct them to a site in maintenance
mode and prevent any data being entered (which 
would subsequently be lost).

#### How
see [kubernetes-maintenance-mode](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence#kubernetes-maintenance-mode)

### DIFF
--- a/kubernetes_deploy/production/deployment.yaml
+++ b/kubernetes_deploy/production/deployment.yaml
@@ -56,7 +56,7 @@ spec:
             - name: GA_TRACKER_ID
               value: 'UA-37377084-37'
             - name: MAINTENANCE_MODE
-              value: 'false'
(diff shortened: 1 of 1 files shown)


CeeBeeUK (approved):
:+1: lgtm

Oops,  home account!

colinbruce (approved):
Looks good to work me too! 😊

PR #3075, opened 2019-11-13
CBO-994: Stylesheet refactoring, remove magic values

naseberry:
#### What
Refactor CCCD stylesheets by removing magic values.

#### Ticket
[Stylesheet refactoring, magic values](https://dsdmoj.atlassian.net/browse/CBO-994)

#### Why
Values were problematic and did not properly convey the intention, we have opted instead to use the preprocessor variables to add meaning.

### DIFF
--- a/app/assets/stylesheets/_claims_importer.scss
+++ b/app/assets/stylesheets/_claims_importer.scss
@@ -10,32 +10,21 @@
   label {
     @include core-16;
   }
+
   .js-enabled & {
+
     input[type="file"],
     .no-file-selected [type="submit"] {
       @include visually-hidden;
     }
   }
+
   button.button-secondary {
     @include core-19;
   }
+
   .errors {
-    margin-top: 5px;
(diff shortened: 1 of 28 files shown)


app/assets/stylesheets/_claims_importer.scss
```diff
@@ -10,28 +10,31 @@
...
+
   .errors {
-    margin-top: 5px;
-    line-height: 1.2;
+    margin-top: $gutter-one-sixth;
     color: $red;
   }
 }
 
 .importer-results {
```
jsugarman: i don't think this class is used. I think it was part of the json uploader/importer which has since been removed

jsugarman (approved):
👍 LGTM

app/assets/stylesheets/_claims_importer.scss
naseberry: Thanks, this has been removed.

PR #3079, opened 2019-11-18
Set explicit response js and all

jsugarman:
#### What
Respond to js with json and unhandled mime types with plain text

#### Why
A neatening up of response for maintenance mode.
Previous maintenance page trial run revealed errors on
sentry resulting from other mime type formats requests.


colinbruce (approved):
👍 LGTM

PR #3080, opened 2019-11-18
CBO-1024: Refine Zendesk handling of tickets

colinbruce:
#### What
Do not submit to zendesk unless comments included
#### Ticket
[Refine Zendesk handling of tickets](https://dsdmoj.atlassian.net/browse/CBO-1024)
#### Why
The feedback score is now handled by Google Analytics and the empty values simply clutter up the zendesk views
#### How
When creating a feedback, as opposed to bug_report, do not send to zendesk unless comments are included.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -62,6 +62,6 @@ class Feedback
   end
 
   def is_feedback_with_empty_comment?
-    feedback? && comment&.empty?
+    feedback? && comment.to_s.empty?
   end
 end
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -43,9 +43,7 @@ class Feedback
 
   def save
     return false unless valid?
-    # rubocop:disable Style/SafeNavigation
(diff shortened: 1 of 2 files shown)


app/models/feedback.rb
```diff
@@ -43,7 +43,7 @@ def is?(type)
 
   def save
     return false unless valid?
-    ZendeskSender.send!(self)
+    ZendeskSender.send!(self) unless is_feedback_with_empty_comment?
```
jsugarman, line 46: much clearer!

app/models/feedback.rb
```diff
@@ -60,4 +60,8 @@ def description
   def feedback_type_attributes
     FEEDBACK_TYPES[type.to_sym]
   end
+
+  def is_feedback_with_empty_comment?
+    feedback? && comment&.empty?
```
jsugarman: if nil is a problem you can do `comment&.to_s.empty?`

jsugarman (approved):
LGTM 👍 

just a suggestion if `nil.empty?` is a problem that needs covering

PR #3081, opened 2019-11-18
MIGRATION DAY: Turn of maintenance mode on live-1 production permanently

jsugarman:
#### What
Permanently set prod maintenance mode off in config

#### Why
Currently live-1 prod is in maintenance mode
to prevent access. This can be changed easily
locally by changing the env var `MAINTENANCE_MODE`
to false and applying directly.

This PR is to merge to master after live-1 prod
goes fully live on cutover day.


kmahern (approved):
lgtm 👍

PR #3082, opened 2019-11-18
Kubernetes admin workflow

kmahern:
#### What
Investigate possibility for admin user to deploy to kubernetes/live-1 with CircleCI without tests having to pass

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1022

#### Why
To spike the ability to have more flexibility on live-1

#### How
Added cccd-admins circleci context and workflow restricted to this role. NB Currently this context is not restricted to any particular group of github users. If we decide to take this ticket forward we would need to associate the context with a github group of CCCD admin users.

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -23,7 +23,6 @@ function _circleci_build() {
     return 1
   fi
 
-  # Cloud platforms circle ci solution does not handle hyphenated names
   case "$1" in
     app | admin-app)
       workflow=$1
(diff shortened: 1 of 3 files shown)


jsugarman:
@kmahern, good investigation but I don't think we are going to use this method for now as it it leads to a very complex workflow with little payoff. As discussed, we can do "admin" deploys from local machines if necessary for now.

PR #3083, opened 2019-11-19
MIGRATION DAY: Switch off downtime warning banner

jsugarman:
#### What
Switch off downtime warning banner

#### Why
Once cutover we can deactivate the warning
banner.

The banner deactivates the day after
the downtime_warning_date anyway, but this
can/should be deployed once we are done with
with cutover to live-1.


kmahern (approved):
👍

PR #3084, opened 2019-11-21
CBO-770: Add focus state for 'Add another' component

naseberry:
#### What
'Add another' component focus state.

#### Ticket
[DAC accessibility resolution](https://dsdmoj.atlassian.net/browse/CBO-770)

#### Why
The tabbing order and focus state was incorrect when the 'Add another' component is used, this PR sets the focus state on the first available input when a new component is added and thereby resolves the tabbing order.


jsugarman (approved):
👍 LGTM

PR #3085, opened 2019-11-22
Hotfix: Add the geckoboard api_key for travel automation

colinbruce:
#### What
Add the Geckoboard API widget key for travel automation 

#### Why
It got missed from the migration

#### How
Added to the secrets and deployment K8s Scripts


jsugarman (approved):
👍 entirely missed that one :(

PR #3087, opened 2019-11-22
Modify handling of DI slack messages

jsugarman:
#### What
Redirect DI slack messages to the
normal channels and filter out success
and ignorable errors

#### Why
To put it back to how it was working
in TD as injection appears to be working
fine via live-1

### DIFF
--- a/app/services/claims/injection_channel.rb
+++ b/app/services/claims/injection_channel.rb
@@ -2,7 +2,6 @@ module Claims
   class InjectionChannel
     def self.for(claim)
       return 'cccd_development' if claim.nil?
-      return 'cccd-k8s-injection' if Settings.aws&.response_queue&.match?('laa-get-paid')
       claim.agfs? ? 'cccd_ccr_injection' : 'cccd_cclf_injection'
     end
   end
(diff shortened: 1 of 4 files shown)


spec/services/injection_response_service_spec.rb
```diff
@@ -38,90 +38,49 @@
 
   describe '#run!' do
     subject(:run!) { irs.run! }
+    # before { allow(Settings.aws).to receive(:response_queue).and_return('laa-get-paid-test-responses-for-cccd') }
```
jsugarman: remove

app/services/claims/injection_channel.rb
```diff
@@ -2,7 +2,6 @@ module Claims
   class InjectionChannel
     def self.for(claim)
       return 'cccd_development' if claim.nil?
-      return 'cccd-k8s-injection' if Settings.aws&.response_queue&.match?('laa-get-paid')
       claim.agfs? ? 'cccd_ccr_injection' : 'cccd_cclf_injection'
```
colinbruce, line 5: Is it worth combining these into a single, `cccd-injection` channel now?  We are only getting 1 or 2 per in each channel.

colinbruce (approved):
One observation, otherwise :+1:

app/services/claims/injection_channel.rb
jsugarman: I rather like the separate channels as it makes searching for specific errors related to ccr or cclf clearer/easier. But could swap this out for single channel in the near future.

PR #3097, opened 2019-12-02
Remove references to old prod and TD deploy

jsugarman:
#### What
Remove vestiges of template-deploy version
of app from the codebase

#### Why
To remove obsolete code and thereby avoid
future confusion.

#### How

remove reference to `gamma` - old TD
name for production environment - and
DOCKER_STATE env var - used by TD to
apply fab(ric) tasks during deployment.


kmahern (approved):
Looks good to me

PR #3101, opened 2019-12-04
Remove superflous default live-1 cluster ingresses

jsugarman:
#### What
Removes the default live-1 cluster ingresses

Note: has already been applied on the fly using edit command

#### Ticket

[CBO-1075](https://dsdmoj.atlassian.net/browse/CBO-1075)

#### Why
To prevent connection using default live-1 cluster ingress and certificates. This is recommended by Cloud platform


kmahern (approved):
👍

PR #3108, opened 2019-12-06
Readme: update environment urls

naseberry:
#### What
Update environment urls within readme.

#### Why
Environment urls have changed since moving to k8.


jsugarman (approved):
🏅 brill

PR #3111, opened 2019-12-09
remove back-ticks as they are interpreted

jsugarman:
#### What
remove back-ticks as they are interpreted

#### Why
use of back-ticks to format text in
a slack attachment does not work and
results in the text being ignored.

remove until there is time to find a way
of escaping these (have tried the obvious
`\` escape char to no effect).


jsugarman:
closing in favour of #3113

PR #3116, opened 2019-12-10
Remove reference to env var in name of step

jsugarman:
#### What
Remove reference to env var in name of step

#### Why
Name keys do no resolve env var references


kmahern (approved):
👍

PR #3117, opened 2019-12-10
Use time zone helper for cronjobs

jsugarman:
#### What
Ensure BST is accounted for in cronjobs

#### Why
Cronjobs are in UTC which is an hour
out during British Summer Time (BST).

CP have created a simple way to handle
Time zone sensitive cronjobs.

see [time-zone-cronjobs guide](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/TZ-Cronjobs.html#time-zone-cronjobs)


kmahern (approved):
👍

jsugarman:
cannot apply this from my local machine so needs holding of on

jsugarman:
problems applying this from local machine. closing this for now

PR #3120, opened 2019-12-12
CBO-1085: Reduce cucumber test time

naseberry:
#### What
Reduce cucumber test times

#### Ticket
[Investigate cucumber performance issue](https://dsdmoj.atlassian.net/browse/CBO-1085)

#### Why
Cucumber runtime in CircleCI has rocketed to ~35mins, this is too long and should be reduced to an acceptable time frame <= 12mins.

#### How
Introducing AXE to Cucumber tests seems have dramatically increased the test time.
I had originally added axe steps at each page within each test scenario, most of the steps (i.e: `And I am on the 'Your claims' page`) were repeated across each feature.
This meant we were testing the same page multiple times, thereby increasing the time it took to run our tests, I have since remove such examples and reduce the amount of time axe is run on the same page.

--------

#### TODO (wip)

 - [ ] Research a more robust method of testing all variation within all views, while keeping test times <= 12mins

### DIFF
--- a/features/claims/advocate/advocate.feature
+++ b/features/claims/advocate/advocate.feature
@@ -1,34 +1,35 @@
 @javascript
-Feature: Advocate submits a claim for a Fixed fee (Appeal against conviction)
+Feature: Advocate submits a claim
 
-  @fee_calc_vcr
-  Scenario: I create an Appeal against conviction claim, then submit it
+  @accessibility
+  Scenario: I create a claim, then submit it
 
     Given I am a signed in advocate
     And I am on the 'Your claims' page
+    Then the page should be accessible within "#content"
     And I click 'Start a claim'
+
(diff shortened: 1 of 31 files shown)


features/claims/advocate/advocate.feature
```diff
@@ -0,0 +1,102 @@
+@javascript
```
jsugarman, line 1: Can we just call this file accessibility to  make it clear whats its purpose is? what do you think

**not a blocker** going forward we could then have a `litigators/accessibility.feature` and possibly a `caseworkers/accessibility.feature`

jsugarman (approved):
Great, LGTM 👍 I'd merge this but then just have a follow up PR to rename the feature test if you agree with my comment.

PR #3127, opened 2019-12-17
Delete ECR images that are tagged with branch tags

jsugarman:
#### What

A github action to delete any ECR image
with a tag named after a branch that has
just been deleted.

This follows current naming convention
of `app-<branch-name>-latest`. This is not
ideal but using this to test to start with.

eventual solution may be to rely on some sort
of BUILD_NUMBER image tagging.

NOTE: this does not, as yet, delete the related
image(s) as I won't to test this against master
by merging this and then creating and deleting
a branch.

#### Ticket

[CBO-1083](https://dsdmoj.atlassian.net/browse/CBO-1083)

#### Why
To remove ecr images of branches that have been deleted
since they cost money to store and are redundant.

#### How
Github action on delete to get the branch name and then
search for ecr images with a tag matching the convention
of `app-<branch-name>-latest`, to then delete.


kmahern (approved):
👍

PR #3131, opened 2019-12-17
Enable deletion of images via github actiion

jsugarman:
#### What
Uncomment the code for actual deletion
of ECR images for deleted branches.

#### Ticket

[CBO-1083](https://dsdmoj.atlassian.net/browse/CBO-1083)

#### Why
Remove ECR images for branches as part
of strategy to only retain images we need.


kmahern (approved):
lgtm

PR #3151, opened 2019-12-30
Correct stub for fee calc

jsugarman:
#### What
Amend the stub that matches fee calculator for failure

#### Why
This was missed and should have been changed
when fee calculator was migrated to live-1.


kmahern (approved):
lgtm

PR #3162, opened 2020-01-09
Fix new admin user password email failure

jsugarman:
#### What
Fix missing mailer template_id

#### Ticket

[CBO-1103](https://dsdmoj.atlassian.net/browse/CBO-1103)

#### Why

This secret was missing from staging
and production but not dev and api-sandbox.

The result is that a new external admin
user (only admin, without litigator or
advocate role) would not have received
an email.

#### How
add secret and amend deployment to expose
as expected env var.


kmahern (approved):
👍

PR #3166, opened 2020-01-13
Remove rails_12factor gem

jsugarman:
Not needed since rails 5

see https://github.com/heroku/rails_12factor#new-rails-5-apps

#### NOTE:
 will need testing on staging to ensure
 assets are still available

- [x] test on staging


kmahern (approved):
👍

PR #3191, opened 2020-01-31
extend advocate claim endpoint's deprecation date

jsugarman:
#### What
Extend deprecation date for old named API v1 endpoints

#### Why
There are still vendors using this API endpoint
and the team is currently focused on other work.

Supporting vendors to transition, or not,
needs discussing with product management.

In the meantime this prevents us from being
blocked while keeping the reminder to remove
the endpoint in place but bumping it to 31/03/2020
(which is the other Common platform works deadline).


naseberry (approved):
👍 LGTM

Solid decision to extend the date due to priorities.

PR #3227, opened 2020-03-17
Remove case number validation

kmahern:
#### What

Remove validation for case number pattern, to allow users to record common platform URNs in the case number field. Validate that these are alphanumeric and <= 20 characters.

#### Ticket

https://dsdmoj.atlassian.net/browse/CACP-180

#### Why

To allow users to record URNs when common platform goes live (CP cases will not have a case_number).

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

### DIFF
--- a/app/interfaces/api/v1/claim_params_helper.rb
+++ b/app/interfaces/api/v1/claim_params_helper.rb
@@ -10,7 +10,7 @@ module API
         optional :court_id, type: Integer, desc: 'REQUIRED: The unique identifier for this court'
         optional :case_type_id, type: Integer, desc: 'REQUIRED: The unique identifier of the case type'
         optional :offence_id, type: Integer, desc: 'REQUIRED: The unique identifier for this offence.'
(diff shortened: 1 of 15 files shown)


jsugarman (changes requested):
As discussed

1. reuse CASE_PATTERN_REGEX
2. use `/^[A-Za-z0-9]{9,20}$/i.freeze` to limit length to between 9 and 20
3. handle error message content via `invalid`
4. think about writing more tests (for API perhaps?)
5. add api release notes entry


worth checking how much existing data would violate the regex (on length >= 9 <=20)

kmahern:
👍 Requested changes made - please re-review? (I am going to look at validating could_be_case_number? as a separate branch)

jsugarman (commented):
this seems to be an older, less applicable version of https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3231. can you close it unless i am missing something

PR #3231, opened 2020-03-24
Add looks like a case number

kmahern:
#### What
As suggested, add looks_like_a_case_number? functionality to validate that when users enter a Case numbers or URN that looks similar to a current case number these are validated against the current pattern for case numbers. The format for the CN_PATTERN regex may need changing if it rejects too many valid URNs.

#### Ticket
https://dsdmoj.atlassian.net/browse/CACP-263

#### Why
To allow users to enter Common Platform URNs in the current case number field, and to attempt to ensure that in the short to medium term users do not enter too many invalid case numbers.

### DIFF
--- a/app/validators/base_validator.rb
+++ b/app/validators/base_validator.rb
@@ -1,5 +1,7 @@
 class BaseValidator < ActiveModel::Validator
   CASE_NUMBER_PATTERN ||= /^[BASTU](199|20\d)\d{5}$/i.freeze
+  CASE_URN_PATTERN ||= /^[A-Za-z0-9]{1,20}$/i.freeze
+  CASE_NUMBER_OR_URN_PATTERN ||= /^[A-Za-z](199|20\d)\d{4,6}$/i.freeze
 
   # Override this method in the derived class
   def validate_step_fields; end
@@ -202,5 +204,10 @@ class BaseValidator < ActiveModel::Validator
   def vat_exceeds_max?(vat:, net:)
     max_vat = VatRate.vat_amount(net, @record.claim.vat_date, calculate: true)
     vat.round(2) > max_vat.round(2)
+  end
+
+  def looks_like_a_case_number?(attribute)
+    return if attr_blank?(attribute)
+    @record.__send__(attribute).match?(CASE_NUMBER_OR_URN_PATTERN)
   end
 end
--- a/app/validators/claim/base_claim_validator.rb
+++ b/app/validators/claim/base_claim_validator.rb
@@ -74,6 +74,8 @@ class Claim::BaseClaimValidator < BaseValidator
   def validate_case_number
     @record.case_number&.upcase!
     validate_presence(:case_number, 'blank')
+    validate_pattern(:case_number, CASE_URN_PATTERN, 'invalid')
+    return unless looks_like_a_case_number?(:case_number)
     validate_pattern(:case_number, CASE_NUMBER_PATTERN, 'invalid')
   end
 
@@ -91,6 +93,8 @@ class Claim::BaseClaimValidator < BaseValidator
 
   def validate_transfer_case_number
     return if @record.errors[:transfer_case_number].present?
+    validate_pattern(:transfer_case_number, CASE_URN_PATTERN, 'invalid')
+    return unless looks_like_a_case_number?(:transfer_case_number)
     validate_pattern(:transfer_case_number, CASE_NUMBER_PATTERN, 'invalid')
   end
 
--- a/app/validators/fee/concerns/case_numbers_validator.rb
+++ b/app/validators/fee/concerns/case_numbers_validator.rb
@@ -8,8 +8,6 @@
 module Fee
   module Concerns
     module CaseNumbersValidator
-      CASE_NUMBER_PATTERN = BaseValidator::CASE_NUMBER_PATTERN
-
       private
 
       # TODO: At time of writing the AGFS case uplift fee types are
@@ -52,8 +50,13 @@ module Fee
       end
 
       def validate_case_number(case_number)
-        add_error(:case_numbers, 'invalid') unless case_number.match?(CASE_NUMBER_PATTERN)
+        add_error(:case_numbers, 'invalid') unless case_number.match?(BaseValidator::CASE_URN_PATTERN)
+        validate_case_number_pattern(case_number) if case_number.match?(BaseValidator::CASE_NUMBER_OR_URN_PATTERN)
         add_error(:case_numbers, 'eqls_claim_case_number') if case_number.casecmp?(claim.case_number)
+      end
+
+      def validate_case_number_pattern(case_number)
+        add_error(:case_numbers, 'invalid') unless case_number.match?(BaseValidator::CASE_NUMBER_PATTERN)
       end
     end
   end
--- a/app/views/pages/api_release_notes.html.haml
+++ b/app/views/pages/api_release_notes.html.haml
@@ -9,6 +9,18 @@
 
 %section.api-documentation
   %hr
+  %h2 1st September 2020
+  %ul.list-bullet
+    %li
+      Case Number modified to also accept URNs
+      %br/
+      %br/
+      These are <= 20 characters and alphanumeric
+      %br/
+      %br/
+
+%section.api-documentation
+  %hr
   %h2 12th May 2020
   %ul.list-bullet
     %li
@@ -34,8 +46,6 @@
       %table
         %tr
           %td GET /api/case_stages
-      %br/
-      %br/
 
 %section.api-documentation
   %hr
--- a/config/locales/error_messages.en.yml
+++ b/config/locales/error_messages.en.yml
@@ -170,13 +170,13 @@ court:
 case_number:
   _seq: 50
   blank:
-    long: Enter a case number for example A20161234
-    short: Enter a case number
-    api: Enter a case number for example A20161234
+    long: Enter a case number or URN
+    short: Enter a case number or URN
+    api: Enter a case number of URN
   invalid:
-    long: The case number must be in the format A20161234
+    long: The case number must be a case number (e.g. A20161234) or unique reference number (less than 21 letters and numbers)
     short: Invalid case number
-    api: The case number must be in the format A20161234
+    api: The case number must be a case number (e.g. A20161234) or unique reference number (less than 21 letters and numbers)
 
 transfer_court:
   _seq: 55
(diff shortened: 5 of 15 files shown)


app/validators/base_validator.rb
```diff
@@ -199,4 +201,9 @@ def vat_exceeds_max?(vat:, net:)
     max_vat = VatRate.vat_amount(net, @record.claim.vat_date, calculate: true)
     vat.round(2) > max_vat.round(2)
   end
+
+  def looks_like_a_case_number?(attribute)
+    return if attr_blank?(attribute)
+    @record.__send__(attribute).match?(CASE_NUMBER_OR_URN_PATTERN)
```
jsugarman, line 211: should this not be 
```
 @record.__send__(attribute).match?(CN_PATTERN)
```

kmahern: I don't think so. The CASE_NUMBER_OR_URN_PATTERN regex is a bit looser to try and pick up things that could be a case number. If looks_like_a_case_number? is true this then means the case number is validated against CN_PATTERN.

app/validators/base_validator.rb
```diff
@@ -1,5 +1,7 @@
 class BaseValidator < ActiveModel::Validator
-  CASE_NUMBER_PATTERN ||= /^[BASTU](199|20\d)\d{5}$/i.freeze
+  CN_PATTERN ||= /^[BASTU](199|20\d)\d{5}$/i.freeze
```
jsugarman: may as well use full names for clarity

`CASE_NUMBER_PATTERN ||= /^[BASTU](199|20\d)\d{5}$/i.freeze`

app/validators/base_validator.rb
```diff
@@ -1,5 +1,7 @@
...
+  URN_PATTERN ||= /^[A-Za-z0-9]{1,20}$/i.freeze
```
jsugarman: may as well use full names for clarity

`CASE_URN_PATTERN ||=  /^[A-Za-z0-9]{1,20}$/i.freeze`

app/validators/fee/concerns/case_numbers_validator.rb
```diff
@@ -52,9 +54,14 @@ def validate_each_case_number
       end
 
       def validate_case_number(case_number)
-        add_error(:case_numbers, 'invalid') unless case_number.match?(CASE_NUMBER_PATTERN)
+        add_error(:case_numbers, 'invalid') unless case_number.match?(URN_PATTERN)
```
jsugarman: can move this line `57` down to `validate_case_number_pattern` method

config/locales/error_messages.en.yml
```diff
@@ -144,10 +144,14 @@ court:
...
   _seq: 50
   blank:
-    long: Enter a case number for example A20161234
-    short: Enter a case number
-    api: Enter a case number for example A20161234
+    long: Enter a case number or URN
+    short: Enter a case number or URN
+    api: Enter a case number of URN
   invalid:
+    long: The case number should be less than 21 characters and alphanumeric
```
jsugarman: i would have one error message that covers both. long message can say something like
`The case number must be a case number (e.g. A20161234) or unique reference number ( less than 21 letters and numbers)`

app/validators/fee/concerns/case_numbers_validator.rb
```diff
@@ -8,7 +8,9 @@
 module Fee
   module Concerns
     module CaseNumbersValidator
-      CASE_NUMBER_PATTERN = BaseValidator::CASE_NUMBER_PATTERN
+      URN_PATTERN = BaseValidator::URN_PATTERN
```
jsugarman: **not-a-blocker* - i know you are merely extending what came before but seems better to me to either refer directly to constant in BaseValidator or define the constants here and refer to them directly where needed elsewhere.

app/validators/claim/base_claim_validator.rb
```diff
@@ -91,7 +92,8 @@ def validate_transfer_court(force: false)
 
   def validate_transfer_case_number
     return if @record.errors[:transfer_case_number].present?
-    validate_pattern(:transfer_case_number, CASE_NUMBER_PATTERN, 'invalid')
+    validate_pattern(:transfer_case_number, URN_PATTERN, 'invalid')
+    validate_pattern(:transfer_case_number, CN_PATTERN, 'format') if looks_like_a_case_number?(:transfer_case_number)
```
jsugarman: use the same message `invalid` and change that message as [i mention below](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3231/files#r397717110)

spec/validators/claim/base_claim_validator_spec.rb
```diff
@@ -224,10 +224,15 @@
     end
 
     it 'should error if invalid' do
-      claim.case_number = 'T87654321'
+      claim.case_number = 'T87654321?'
       should_error_with(claim, :case_number, 'invalid')
     end
 
+    it 'should error if case number has the incorrect format' do
```
jsugarman: given the cyclomatic complexity of the case number validation now I would look to extend this/these spec examples to more extensively unit test the regex

e.g.

```
context 'with unique reference numbers' do
  # test boundaries
  it 'invalid when less than 9 chars' do
  ...
 end
  it 'invalid when more than 20 chars' do
  ...
 end
  it 'invalid when includes symbol' do
  ...
 end
end

context 'with case number' do
...
end
```

jsugarman (changes requested):
Some name changing, refactoring and more extensive unit tests would be good

app/validators/base_validator.rb
kmahern: 👍

app/validators/base_validator.rb
kmahern: 👍

config/locales/error_messages.en.yml
kmahern: 👍

app/validators/fee/concerns/case_numbers_validator.rb
kmahern: 👍

app/validators/claim/base_claim_validator.rb
kmahern: 👍

spec/validators/claim/base_claim_validator_spec.rb
kmahern: 👍

app/validators/fee/concerns/case_numbers_validator.rb
kmahern: I don't think I can. I am only calling validate_case_number_pattern where looks_like_a_case_number is true. But all case numbers (not just URNs) need to be validated against URN_PATTERN.

kmahern:
@jsugarman I think I have addressed all the comments in your review?

app/validators/claim/base_claim_validator.rb
```diff
@@ -74,7 +74,8 @@ def validate_court
   def validate_case_number
     @record.case_number&.upcase!
     validate_presence(:case_number, 'blank')
-    validate_pattern(:case_number, CASE_NUMBER_PATTERN, 'invalid')
+    validate_pattern(:case_number, URN_PATTERN, 'invalid')
```
jsugarman: CASE_URN_PATTERN

app/validators/base_validator.rb
```diff
@@ -1,5 +1,7 @@
 class BaseValidator < ActiveModel::Validator
   CASE_NUMBER_PATTERN ||= /^[BASTU](199|20\d)\d{5}$/i.freeze
+  URN_PATTERN ||= /^[A-Za-z0-9]{1,20}$/i.freeze
```
jsugarman: CASE_URN_PATTERN

app/validators/claim/base_claim_validator.rb
```diff
@@ -91,6 +92,8 @@ def validate_transfer_court(force: false)
 
   def validate_transfer_case_number
     return if @record.errors[:transfer_case_number].present?
+    validate_pattern(:transfer_case_number, URN_PATTERN, 'invalid')
```
jsugarman: CASE_URN_PATTERN

app/validators/fee/concerns/case_numbers_validator.rb
```diff
@@ -52,9 +50,14 @@ def validate_each_case_number
       end
 
       def validate_case_number(case_number)
-        add_error(:case_numbers, 'invalid') unless case_number.match?(CASE_NUMBER_PATTERN)
+        add_error(:case_numbers, 'invalid') unless case_number.match?(BaseValidator::URN_PATTERN)
```
jsugarman: CASE_URN_PATTERN

spec/validators/claim/base_claim_validator_spec.rb
```diff
@@ -246,6 +256,25 @@
...
   end
 
+  context 'with unique reference numbers' do
+    it 'should not error if valid' do
+      claim.case_number = 'ABCDEFGHIJ1234567890'
+      expect(claim).to be_valid
+    end
+
+    it 'is invalid if contains non alphanumeric characters' do
+      %w(_ - * ? ,).each do |character|
```
jsugarman, line 266: 👍

app/validators/claim/base_claim_validator.rb
```diff
@@ -91,6 +92,8 @@ def validate_transfer_court(force: false)
...
+    return unless looks_like_a_case_number?(:transfer_case_number)
```
jsugarman, line 97: why not use [guard clause as you have here](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3231/files#diff-90d4e3266775fe3632d22743de6bc3f4R78)

kmahern: I couldn't because I kept getting into a circle of rubocop linelength/guard clause errors. Adding the return unless looks_like_a_case_number? was the suggested solution here https://github.com/rubocop-hq/rubocop/issues/7215.

app/validators/fee/concerns/case_numbers_validator.rb
kmahern: 👍

app/validators/claim/base_claim_validator.rb
kmahern: 👍

app/validators/base_validator.rb
kmahern: 👍

app/validators/claim/base_claim_validator.rb
kmahern: 👍

app/validators/claim/base_claim_validator.rb
jsugarman: its more about consistency as you are using guard clause at line `78` but not here

kmahern: 👍 As discussed, have amended line 78 so that the validation for case_number and transfer_case_number is consistent.

app/views/pages/api_release_notes.html.haml
```diff
@@ -7,6 +7,18 @@
   = link_to 'Github issues', 'https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues'
   for bugs or other problems.
 
+%section.api-documentation
+  %hr
+  %h2 1st April 2020
```
jsugarman: will need to change this when released

jsugarman (commented):
👍 LGTM. 

do not merge until ready with all the other changes and will need to amend the api release notes date. On that have you contacted the vendors regarding these changes to give them warning?

PR #3243, opened 2020-04-01
Extend advocate claim endpoint's deprecation date

kmahern:
#### What
Extend advocate claim endpoint's deprecation date

#### Ticket
N/A

#### Why
There are still vendors using this API endpoint
and the team is currently focused on other work.

Supporting vendors to transition, or not,
needs discussing with product management.

In the meantime this prevents us from being
blocked while keeping the reminder to remove
the endpoint in place.


jsugarman (approved):
great 👍

PR #3250, opened 2020-04-08
Refactor daily attendance validations

jsugarman:
#### What
abstract daily attendance validation out to separate methods

#### Why
Already confusing, these may need further
extension for hardship claims so considering
refactoring in now.

#### Ticket

[CBO-1174](https://dsdmoj.atlassian.net/browse/CBO-1174)


jsugarman:
daily attendances to be ignored on hardship claims for now

PR #3260, opened 2020-04-15
Extract common claim provider methods to mixin

jsugarman:
#### What
share claim provider delegation

#### Why
These are methods are duplicated between advocate
claim types and the advent of a new hardship claim
would benefit from sharing via composition.

#### How
extract rails concern and mixin to advocate claim classes


naseberry (approved):
LG

PR #3266, opened 2020-04-16
CBO-1171: retain hardship in archive

colinbruce:
#### What
Added a new state to claims that will allow hardship claims to be retained until a final bill is settled

#### Ticket

[Link to ticket](https://dsdmoj.atlassian.net/browse/CBO-1171)

#### How
Extended the state machine and transition models to allow different archive actions to take place depending on the type of claim being archived.
This model will work for HarshipClaims being placed into an `:archived_pending_review` state.

#### Next steps
A second ticket will be raised to allow caseworkers to manually archive claims from the `archived_pending_review` state when they have reviewed.  This should be separate as we don't know how that flow will work yet, but this flow allows an archived_pending_review to be transitioned into `archived_pending_delete` but does not attempt to implement a method of allowing a user to trigger it.

--------

#### TODO (wip)

 - [x] Create ticket for design of flow for caseworker to manually archive these in the future - [CBO-1183](https://dsdmoj.atlassian.net/browse/CBO-1183) created


jsugarman (approved):
LGTM 👍

PR #3268, opened 2020-04-17
Update the guard version

colinbruce:
#### What
Update the guard version

#### Why
Guard helps me run discrete tests, but the version had not been updated

#### How
```bundle update --source guard```


jsugarman (approved):
👍

PR #3269, opened 2020-04-17
Update rubocop_todo

colinbruce:
#### What
extend ABC limit in rubocop todo

#### Why
The hardship handling adds more complexity to the state machine, this excludes the check completely as it was skewing the numbers considerably


jsugarman (approved):
👍

PR #3271, opened 2020-04-17
Highlight optional form fields in workflow

naseberry:
#### What
Mark optional fields within the users workflow.

#### Ticket
[CBO-1184](https://dsdmoj.atlassian.net/browse/CBO-1184)

#### Why
Design / User Research has a need to inform users of optional inputs within the workflow.


config/locales/en.yml
```diff
@@ -1043,7 +1043,7 @@ en:
           trial_detail: Trial details
           actual_trial_length: *actual_trial_length
           actual_trial_length_hint: 'Number of days'
-          estimated_trial_length: 'Estimated trial length'
+          estimated_trial_length: 'Estimated trial length (optional)'
```
jsugarman: not sure it is optional?

jsugarman (commented):
not sure may of these are optional in particular scenarios. validations define whether or not they are required.

PR #3274, opened 2020-04-20
Remove duplicate advocate claim presenter spec

jsugarman:
#### What
remove duplicate presenter spec but move individual examples
as necessary.

#### Why
One spec existing correctly a `spec/presenters/claims/advocate_claim_presenter_spec.rb`. The other at a level
up. Tests were moved from the duplicate as necessary.


colinbruce (commented):
LGTM 👍

colinbruce (approved):

PR #3277, opened 2020-04-21
Add lgfs hardship allocation filter task

kmahern:
Enable caseworkers to filter on lgfs hardship claims

#### What
Allow caseworkers to allocate and filter on lgfs hardship claims

#### Ticket

https://dsdmoj.atlassian.net/browse/CBO-1185

#### Why
To allow caseworkers to process lgfs hardship claims

### DIFF
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -37,6 +37,7 @@ module API
         expose :injection_errored
         expose :cav_warning
         expose :supplementary
+        expose :lgfs_hardship
       end
 
       private
@@ -130,6 +131,10 @@ module API
 
       def supplementary
         object.case_type.eql?('Supplementary').to_i
+      end
+
+      def lgfs_hardship
+        object.scheme_type.eql?('LitigatorHardship').to_i
       end
     end
   end
--- a/spec/api/entities/search_result_spec.rb
+++ b/spec/api/entities/search_result_spec.rb
@@ -66,7 +66,8 @@ describe API::Entities::SearchResult do
           risk_based_bills: 0,
           injection_errored: 0,
           cav_warning: 0,
-          supplementary: 0
+          supplementary: 0,
+          lgfs_hardship: 0
         }
       end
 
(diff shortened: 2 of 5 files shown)


app/interfaces/api/entities/search_result.rb
```diff
@@ -37,6 +37,7 @@ class SearchResult < BaseEntity
         expose :injection_errored
         expose :cav_warning
         expose :supplementary
+        expose :lgfs_hardship
```
jsugarman, line 40: @kmahern once the agfs branch is in master with the agfs filter will it not be possible to simply use the `:hardship` filter you already added to that branch

app/interfaces/api/entities/search_result.rb
```diff
@@ -131,6 +132,10 @@ def cav_warning
       def supplementary
         object.case_type.eql?('Supplementary').to_i
       end
+
+      def lgfs_hardship
```
jsugarman, line 136: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3277/files#r412056455)

kmahern: I think this might be OK? I have called  the method def hardship in the agfs branch, so I don't think there should be a conflict (although possibly I should change this to def agfs_hardship in the agfs branch)? (There is a separate list of filter tasks for agfs and lgfs claims).

app/interfaces/api/entities/search_result.rb
kmahern: I think this might be OK? I have  added expose :hardship in the agfs branch, so I don't think there should be a conflict (although possibly I should change this to expose :agfs_hardship in the agfs branch)? (There is a separate list of filter tasks for agfs and lgfs claims).

jsugarman: if it can't be shared then yes the agfs equivalent should be named agfs_hardship (we do this for warrant filters by the looks of it). you could PR against my branch

spec/api/entities/search_result_spec.rb
```diff
@@ -66,7 +66,8 @@
           risk_based_bills: 0,
           injection_errored: 0,
           cav_warning: 0,
-          supplementary: 0
+          supplementary: 0,
+          lgfs_hardship: 0
```
jsugarman, line 70: @kmahern I think you need to  add a search_result_spec example to cover this filer type - see the examples in this file - agfs equivalent should too

app/interfaces/api/entities/search_result.rb
kmahern: 👍 see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3282

app/interfaces/api/entities/search_result.rb
kmahern: see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3282

spec/api/entities/search_result_spec.rb
kmahern: 👍 see also https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3282

colinbruce (approved):
:+1:

PR #3282, opened 2020-04-21
Rename hardship to agfs_hardship

kmahern:
Also add wrong instructed advocate refusal reason,
and search_result_spec test for agfs_hardship claims

#### What
Rename hardship filter for agfs claims from filter to agfs_hardship.
Also add Wrong Instructed Advocate refusal reason.
Also add search_result_spec test for agfs_hardship claims

#### Why
To be consistent with LGFS hardship claims.

### DIFF
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -37,7 +37,7 @@ module API
         expose :injection_errored
         expose :cav_warning
         expose :supplementary
-        expose :hardship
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
👍 

something looks fishy with the preceding tests (advocate interim/warrant and supplementary) in light of this and could do with tweaking against master if you want to PR

namely
- do they not need a `'scheme' => 'agfs'`?
- they are using case type to describe the claim_type/scheme_type

PR #3284, opened 2020-04-22
Add BillType to MI Report

colinbruce:
#### What
Add a new, Bill Type, column to the MI Report

#### Ticket

[Amend MI reporting to include the bill type](https://dsdmoj.atlassian.net/browse/CBO-1186)

#### Why
This is required for - case management to monitor bill volume intakes and use this information for resourcing. It is likely to also be used for policy and finance to determine how much providers are utilising this option (and given that we may not be able to identify the claims as hardship in CCR and CCLF it is important to have this in place.  It may also be useful for monitoring the number of claims we will end up having in the new state of "archived pending review" and allow caseworkers to match up the claims the final bills. 

#### How
Extend the MI report headers and add and a new generator for extracting the required values from the Claim::BaseClaim.type

--------

#### TODO (wip)

 - [ ] Extend this to include the AGFS case type once the two branches can be merged

### DIFF
--- a/app/presenters/claim_csv_presenter.rb
+++ b/app/presenters/claim_csv_presenter.rb
@@ -3,6 +3,17 @@ class ClaimCsvPresenter < BasePresenter
 
   COMPLETED_STATES = %w[rejected refused authorised part_authorised].freeze
   SUBMITTED_STATES = %w[submitted redetermination awaiting_written_reasons].freeze
+  AGFS_CLAIM_TYPES = %w[
+    Claim::AdvocateClaim
+    Claim::AdvocateInterimClaim
+    Claim::AdvocateSupplementaryClaim
+  ].freeze
+  LGFS_CLAIM_TYPES = %w[
+    Claim::LitigatorClaim
+    Claim::InterimClaim
+    Claim::TransferClaim
+    Claim::LitigatorHardshipClaim
+  ].freeze
 
   def present!
     yield parsed_journeys if block_given?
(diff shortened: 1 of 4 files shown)


app/presenters/claim_csv_presenter.rb
```diff
@@ -61,15 +72,27 @@ def case_type_name
   end
 
   def scheme
-    if %w[Claim::AdvocateClaim Claim::AdvocateInterimClaim Claim::AdvocateSupplementaryClaim].include? type
+    if AGFS_CLAIM_TYPES.include? type
```
jsugarman: so we don't have to repeat the claim types wold be good to have something like

```suggestion
    if  claim.class.agfs_claim_types.map(&:to_s).include? type
```

could tidy further with some kind of delegation - rails' delegate or forwardable
or manual as below
```
private
def agfs_claim_types
  claim.class.agfs_claim_types.map(&:to_s)
end
```

app/presenters/claim_csv_presenter.rb
```diff
@@ -61,15 +72,27 @@ def case_type_name
...
       'AGFS'
-    elsif %w[Claim::LitigatorClaim Claim::InterimClaim Claim::TransferClaim].include? type
+    elsif LGFS_CLAIM_TYPES.include? type
```
jsugarman: [as above](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3284/files#r413031287)

jsugarman (approved):
👍 looks good although not too familiar with the MI reports

be good to remove the hard-coding of claim types.

app/presenters/claim_csv_presenter.rb
jsugarman: i think you can actually do 
```
if claim.class.agfs?...
```

PR #3286, opened 2020-04-23
Add translations for all claim sidebars

jsugarman:
#### What
Add translations for sidebars

#### Why
missing translations

#### How
rely on separate file for sidebar translations and add to load path


naseberry (approved):
:+1:

PR #3290, opened 2020-04-24
test the smoke test

jsugarman:
want failure to not faile next step but still show failure


jsugarman:
closing as was just using to test

PR #3298, opened 2020-04-27
Add case stage table migration

jsugarman:
#### What
add new table and references for case stages

#### Ticket

[CBO-1187](https://dsdmoj.atlassian.net/browse/CBO-1187)

#### Why
Separate PR to push out to production out of
hours and so we can iterate on the details. separate
PR for the relevant seeds, model and other
objects

#### How
New table that belongs to case_types and is referenced
by claims. The idea is, in separate PR, to have hardship
claims delegate their case_type, to case_stage.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -2,21 +2,30 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -31,7 +40,7 @@
     t.uuid "uuid", default: -> { "uuid_generate_v4()" }
   end
 
-  create_table "case_worker_claims", force: :cascade do |t|
+  create_table "case_worker_claims", id: :serial, force: :cascade do |t|
```
colinbruce, line 43: Did you deliberately add the `id: :serial` or have you had a recent pg update? 🤔

colinbruce (approved):
Migration looks fine... weird update on all other tables though! :)

db/schema.rb
jsugarman: not deliberate no. its from an update to activerecord i believe - happened on the last migration i did too.

PR #3300, opened 2020-04-28
Prevent seeding of sample users on application load

jsugarman:
#### What
Bug fix to prevent seeding of sample users on application load

#### Why
unnecessary and slows console load

This has been happening for a long time and should
not have been.

It was caused by autoload of `lib` dir in combination
with the existence `lib/demo_data/demo_seeds.rb` file
which loads the seeds as soon as it is loaded (i.e.
on application boot, due to autoload of lib dir).

To explicitly load sample users use:
```
rake claims:sample_users
```

### DIFF
--- a/lib/demo_data/demo_seeds.rb
+++ b/lib/demo_data/demo_seeds.rb
@@ -1,8 +1,0 @@
-%w(
-  external_users
-  case_workers
-).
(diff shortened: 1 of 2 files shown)


kmahern (approved):
lgtm

PR #3302, opened 2020-04-28
Refactor transitioner spec

jsugarman:
#### What
redescribe some context using with/when standard
and mock in before blocks to keep examples small

#### Why
Very unclear from context descriptions
and too many lines in some examples, including
mocking that could go in before hooks.

Came out of hardship claim work

### DIFF
--- a/spec/models/timed_transitions/transitioner_spec.rb
+++ b/spec/models/timed_transitions/transitioner_spec.rb
@@ -65,22 +65,24 @@ RSpec.describe TimedTransitions::Transitioner do
   end
 
   describe '#run' do
-    context 'non dummy run' do
-      context 'transitioning to archived pending delete' do
-        context 'less than 60 days ago' do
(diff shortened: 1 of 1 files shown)


spec/models/timed_transitions/transitioner_spec.rb
```diff
@@ -106,7 +108,7 @@
               end
             end
 
-            it 'calls archive if last state change more than 16 weeks ago' do
+            it 'amends claim state to archives pending review' do
```
kmahern, line 111: Non-blocking minor typo, should it read 'amends claim state to archived pending review'

kmahern (approved):
lgtm other than minor possible typo.

spec/models/timed_transitions/transitioner_spec.rb
jsugarman: cheers, will fix in my next pr on top - test suite takes ages 😞

PR #3306, opened 2020-04-29
Set case_type to Hardship in allocation filter for AdvocateHardshipCl…

kmahern:
#### What
Set case_type to Hardship in allocation filter for AdvocateHardshipClaims.

#### Ticket
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&selectedIssue=CBO-1203

#### Why
This was previously being displayed as null

#### How
Amended query_helper.rb (this fix will need to be replicated for LitigatorHardshipClaims)

### DIFF
--- a/app/interfaces/api/v2/query_helper.rb
+++ b/app/interfaces/api/v2/query_helper.rb
@@ -8,7 +8,7 @@ module API::V2
           c.id,
           c.uuid,
           CASE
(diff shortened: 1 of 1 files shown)


jsugarman (approved):
👍 great

are they any existing tests that can be extended?

kmahern:
@jsugarman I couldn't see any existing tests that looked like I could extend them?

PR #3307, opened 2020-04-29
enable deploy of master to dev-agfs/dev-lgfs from local machine

jsugarman:
Just a quick fix, not necessarily worth changing
circleci master build and deploy at this late
stage - what do you think?


colinbruce (approved):
Looks fine... the CircleCI deploys okay so far :)

PR #3309, opened 2020-04-29
Conditionally show stage or case type

jsugarman:
#### What
show stage or case type based on whether
case stage exists

#### Ticket

[comment on CBO-1187](https://dsdmoj.atlassian.net/browse/CBO-1187?focusedCommentId=100069)

#### Why
Only need to show case stage for hardship claims.
And more generally for anything where the case stage
exists it should be preferred over case type.


colinbruce (approved):
👍

PR #3314, opened 2020-04-30
Cbo 1188 agfs hardship api

kmahern:
#### What

Added api endpoints for advocate hardship claims.

#### Ticket

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&modal=detail&selectedIssue=CBO-1188

#### Why

To allow external users (eventually) to submit advocate hardship claims with their case management systems via the the api endpoints.

#### How

Added POST /api/external_users/claims/advocates/hardship/validate and POST /api/external_users/claims/advocates/hardship endpoints.

### DIFF
--- a/app/interfaces/api/entities/case_stage.rb
+++ b/app/interfaces/api/entities/case_stage.rb
@@ -1,0 +1,10 @@
+module API
+  module Entities
+    class CaseStage < BaseEntity
+      expose :case_type_id
+      expose :unique_code
+      expose :description
+      expose :roles
+    end
+  end
+end
--- a/spec/api/v1/dropdown_data_spec.rb
+++ b/spec/api/v1/dropdown_data_spec.rb
@@ -16,6 +16,7 @@ RSpec.describe API::V1::DropdownData do
   DISBURSEMENT_TYPE_ENDPOINT  = "/api/disbursement_types"
   TRANSFER_STAGES_ENDPOINT    = "/api/transfer_stages"
   TRANSFER_CASE_CONCLUSIONS_ENDPOINT = "/api/transfer_case_conclusions"
+  CASE_STAGE_ENDPOINT         = "/api/case_stages"
 
   FORBIDDEN_DROPDOWN_VERBS = [:post, :put, :patch, :delete]
   ALL_DROPDOWN_ENDPOINTS = [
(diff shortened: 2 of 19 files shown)


spec/api/v1/dropdown_data_spec.rb
```diff
@@ -63,13 +65,15 @@
...
-        TRANSFER_CASE_CONCLUSIONS_ENDPOINT => API::Entities::SimpleKeyValueList.represent(Claim::TransferBrain::CASE_CONCLUSIONS.to_a).to_json
+        TRANSFER_CASE_CONCLUSIONS_ENDPOINT => API::Entities::SimpleKeyValueList.represent(Claim::TransferBrain::CASE_CONCLUSIONS.to_a).to_json,
+        CASE_STAGE_ENDPOINT => API::Entities::CaseStage.represent(CaseStage.all).to_json,
       }
     end
 
     before do
       seed_fee_schemes
       seed_case_types
+      seed_case_stage
```
jsugarman: my error, but this should be `seed_case_stages` (plural). you could add a commit to fix my typo

jsugarman (changes requested):
Can you go through the check llist in the ticket and make sure you have ticked them all off
https://dsdmoj.atlassian.net/browse/CBO-1188


you can exclude `seed_helper` from `simplecov` too, to improve coverage score. There should not be a reduction in coverage ideally.

app/interfaces/api/entities/case_stage.rb
```diff
@@ -0,0 +1,11 @@
+module API
+  module Entities
+    class CaseStage < BaseEntity
+      expose :id
```
jsugarman: we should not really be exposing `id` - vendors would then use it and rely on it. we should be enabling search by `unique_code` (which we can guarantee will not change)

kmahern: 👍 Have removed id from the case stage endpoint (also api users will enter unique code and not id when creating/validating a hardship claim).

spec/api/v1/dropdown_data_spec.rb
kmahern: 👍 have amended to seed_case_stages

kmahern:
@jsugarman I have been through the comments and the checklist on the ticket and I think I have made the changes requested.

kmahern:
Have removed db/seed_helper.rb from the scope of code climate to resolve issue with code coverage.

spec/api/v1/external_users/claim_spec.rb
```diff
@@ -16,6 +16,9 @@
     /api/external_users/claims/advocates/supplementary
     /api/external_users/claims/advocates/supplementary/validate
 
+    /api/external_users/claims/advocates/hardship
```
jsugarman, line 19: 👍 yes this is good for AGFS 
see LGFS equivalent https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3341/files/d71bea25dc2edb4144f9e7fd772c502d1bb515fc#r420650769

PR #3315, opened 2020-04-30
Add an LGFS hardship cleaner

colinbruce:
This needed a few aditions to factories and the
tests needed expanding


app/models/claim/litigator_hardship_claim_cleaner.rb
```diff
@@ -0,0 +1,39 @@
...
+      clear_inapplicable_fields
+    end
+
+    private
+
+    def clear_inapplicable_fields
+      clear_ppe if case_type.present? && ppe_not_required
+    end
+
+    def ppe_not_required
```
jsugarman: not a blocker 
```suggestion
    def ppe_not_required?
```

PR #3319, opened 2020-04-30
Enable hardship claims

jsugarman:
#### What
Switch hardship enabled feature flag on all environments

#### Why
Allow users to submit hardship claims


colinbruce (approved):
👍

PR #3323, opened 2020-05-01
Revert disbling of LGFS hardships

jsugarman:
Business would like to keep enabled
and handle via assessment feedback for now.

Could merge this for now and if business change their mind we can `git cherry-pick ae8762433` to add back in


colinbruce (approved):
:+1:

PR #3325, opened 2020-05-01
Fix lgfs case stages

colinbruce:
#### What
Add the correct handling for LGFS hardship fees

#### Why
The old way was broken

#### How
Add CaseStage for the new scenario with accurate description
Add new BillScenario handling that passes the correct scenario to the fee calculator with minimal impact on the views, minus removing an optional field that is now always required

--------

#### TODO (wip)

 - [ ] Make sure the feature tests pass
 - [ ] Make sure Rspec is good
 - [ ] User testing


db/seeds/case_stages.rb
```diff
@@ -84,19 +84,10 @@
...
-  id: 10,
-  description: 'Pre PTPH (no evidence served)',
-  unique_code: 'NOPTPHNOPPE',
-  position: 100,
-  case_type_id: CaseType.find_by(fee_type_code: 'GRDIS').id,
+  id: 12,
+  description: 'Pre PTPH or PTPH adjourned',
+  unique_code: 'PREPTPHADJ',
+  position: 110,
+  case_type_id: CaseType.find_by(fee_type_code: 'GRRAK').id,
```
jsugarman, line 118: treating as a cracked trial 'GRRAK' here - is that right?

app/services/cclf/case_type_adapter.rb
```diff
@@ -22,7 +22,8 @@ class CaseTypeAdapter
       INPCM: 'ST1TS0T0', # Interim Claim - Effective PCMH - Trial only
       INTDT: 'ST1TS1T0', # Interim Claim - Trial start - Trial only
       INRNS: 'ST1TS2T0', # Interim Claim - Retrial New solicitor - Retrial only
-      INRST: 'ST1TS3T0'  # Interim Claim - Retrial start - Retrial only
+      INRST: 'ST1TS3T0', # Interim Claim - Retrial start - Retrial only
+      HARDSHIP: 'ST2TS1T0' # Hardship Claim - up to and including PCMH transfer (org)
```
jsugarman, line 26: 💥 cool 🤞

app/views/external_users/claims/hardship_fee/_fields.html.haml
```diff
@@ -4,16 +4,15 @@
...
-        .js-fee-calculator-effector
-          = hf.adp_text_field :quantity,
-            label: t('.quantity'),
-            hint_text: t('.quantity_hint'),
-            input_classes:'quantity fee-quantity js-fee-calculator-ppe form-control-1-4',
-            input_type: 'number',
-            value: fee.quantity,
-            errors: @error_presenter,
-            error_key:'hardship_fee.quantity'
+      .js-fee-calculator-effector
```
jsugarman, line 7: 🎖️

db/seeds/case_stages.rb
jsugarman: saw that business regs snippet 👍

jsugarman: I think we should leave the old types/seeds in there because they exist in the database and their are claims that are using them so we can't get rid of just yet.

we should do it later. can chat about this as they will need to be excluded from the list of stages tho

jsugarman (approved):
👍 Looks good, assuming cukes pass

PR #3329, opened 2020-05-04
Add case stage event handling

naseberry:
#### What
Resolve case stage work-around

#### Ticket
[CBO-1209](https://dsdmoj.atlassian.net/browse/CBO-1209)

#### Why
Case stage input is "pretending" to be case type in order for the trial fields to work.

#### How
Bind event handling to case stage input


jsugarman (approved):
💯 LGTM

PR #3331, opened 2020-05-04
Fix evidence table long text wrapping

naseberry:
#### What
Long strings in Evidence table are breaking the page layout.

#### Ticket
[CBO-1216](https://dsdmoj.atlassian.net/browse/CBO-1216)

#### Why
Uploaded files with long names and no white space breaks the page layout

#### How
Wrap long string within table cell to stop the table breaking the page layout


jsugarman (approved):
👍 if it works!

PR #3332, opened 2020-05-04
Reduce time of rspec sweet

jsugarman:
#### What
RSpec suite has slowed, try to speed up

#### Ticket

[CBO-1212](https://dsdmoj.atlassian.net/browse/CBO-1212)

#### Why
Following introduction of hardship claims the code
base and test suite has grown significantly. The test
suite for rspec takes over 15 minutes to run.

#### How
Run rspec in 8 parrallel containers/executors, up
from 4. In addition split the specs across the containers
based on timings. This is the most efficient use
of containers but will require metadata storage on
circleci - which is analysed by circle to optimise
the runs across the "nodes", placing slowest tests with fewer
other tests...we shall see.

Already down from 15 minutes to sub 12 minutes with the
parrallel instances, although competition for containers
can delay the start potentially?!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,8 @@ references:
       command: |
         mkdir -p tmp/
         mkdir -p tmp/coverage/
+        mkdir -p /tmp/test-results/rspec
+        mkdir -p /tmp/test-results/cucumber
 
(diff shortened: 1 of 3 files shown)


kmahern (approved):
👍

naseberry (approved):

PR #3335, opened 2020-05-05
CBO-1208: Timestamp authorise on MI

colinbruce:
#### What
Add the timestamp of last authorise

#### Ticket
[CBO-1208](https://dsdmoj.atlassian.net/browse/CBO-1208)

#### Why
This was shown in the message view per claim, adding it
here makes it simpler for the MI users to manage
performance

#### How
Update the strftime format from `%d/%m/%Y` to `%d/%m/%Y %H:%M`


jsugarman (approved):
👍 LGTM

PR #3339, opened 2020-05-05
CBO-1221: Refactor the LGFS CaseStage view handling

colinbruce:
#### What
As a provider I do not want to have to make multiple clicks to show, and choose, a single item in a drop down list

#### Ticket
[CBO-1221](https://dsdmoj.atlassian.net/browse/CBO-1221)

#### How
* Add a panel to the top of the page explaining the LGFS
  case_stage limitations
* Replace the case_stage section drop down with a hidden
  field with the single LGFS case_stage id in it
* Update the LGFS hardship feature to remove the option
  to select the case stage, but still check the output
  of it on the summary page


jsugarman (approved):
👍 LGTM

PR #3341, opened 2020-05-06
Cbo 1189 lgfs hardship api

kmahern:
#### What
Create an api endpoint for lgfs hardship claims

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1189

#### Why
To allow users to submit lgfs hardship claims via an api endpoint using their CMS. See also https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3314 for agfs hardship claims.

#### How
Added POST /api/external_users/claims/hardship/validate and POST /api/external_users/claims/hardship endpoints.


spec/factories/case_stages.rb
```diff
@@ -93,5 +93,13 @@
...
       roles { %w(lgfs) }
     end
+
+    trait :pre_ptph_or_ptph_adjourned do
+      description { 'Pre PTPH or PTPH adjourned' }
+      unique_code { 'PREPTPHADJ' }
+      position { 110 }
+      case_type_id { create(:case_type, :cracked_trial).id }
+      roles { %w(lgfs) }
+    end
```
colinbruce, line 103: :trophy: Just adding this myself!

app/interfaces/api/v1/hardship_claim_params_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+        optional :additional_information, type: String, desc: 'OPTIONAL: Any additional information'
+        optional :apply_vat, type: Boolean, desc: 'OPTIONAL: Include VAT (JSON Boolean data type: true or false)'
+        optional :prosecution_evidence, type: Boolean, desc: 'OPTIONAL: Pages of prosecution evidence > 0?'
+        optional :api_key, type: String, desc: 'REQUIRED: The API authentication key of the provider'
+        optional :creator_email, type: String, desc: I18n.t('api.v1.common_params.creator_email')
+        optional :court_id, type: Integer, desc: 'REQUIRED: The unique identifier for this court'
+        optional :offence_id, type: Integer, desc: 'REQUIRED: The unique identifier for this offence.'
+        optional :case_number, type: String, desc: 'REQUIRED: The case number'
+        optional :providers_ref, type: String, desc: 'OPTIONAL: Providers reference number'
+        optional :cms_number, type: String, desc: 'OPTIONAL: The CMS number'
```
colinbruce: The main block seems to repeat twice (`api_key` to `prosecution_evidence`) seems to be duplicated

spec/support/api/claims/advocate_hardship_claim_test.rb
```diff
@@ -0,0 +1,89 @@
+require_relative 'base_claim_test'
```
colinbruce: This whole file seems to have leaked over from the AGFS branch, not a problem unless it changes before the merge

spec/api/v1/external_users/claim_spec.rb
```diff
@@ -24,6 +24,9 @@
 
     /api/external_users/claims/transfer
     /api/external_users/claims/transfer/validate
+
+    /api/external_users/claims/hardship
```
colinbruce: Should this be changed to`/api/external_users/claims/litigators/hardship`?

I know the plan was to move them over to the new namespace eventually, @jsugarman should probably weigh in on this one as it will affect you more than me!

colinbruce (changes requested):
The duplication needs removing but, otherwise, looking good.

A question that may be worth discussing now too.

spec/api/v1/external_users/claim_spec.rb
jsugarman: yeah :( the whole litigator side should really be name-spaced to:
```
/api/external_users/claims/litigators/[type]
/api/external_users/claims/litigators/[type]/validate
```

but we can't get rid of the old ones. we should probably only be introducing new endpoints in the new namespaces. same applies to AGFS branch for hardship - have not checked this on that branch.

jsugarman: actually looking at current API the whole LGFS side is not namespaced. so would be a bit of tangent to do now and sunset the old endpoints. Tech debt 😢 coming back to bite us

Screen Shot 2020-05-06 at 10 19 58

question is whether we should namespace just the lgfs hardship now?!

colinbruce: Hence why I asked! 😃 Do you want a single, 'correctly' namespaced litigator endpoint as a forerunner or add more tech debt?  🤔

jsugarman: could add the correct one for hardship and the others and it might prompt the vendors to correct the others while they are at it :dreamon:

app/interfaces/api/v1/hardship_claim_params_helper.rb
kmahern: 👍 have deleted the duplicated api_key

spec/support/api/claims/advocate_hardship_claim_test.rb
kmahern: 👍 have removed redundant file (it is in agfs branch).

spec/api/v1/external_users/claim_spec.rb
kmahern: Happy to create a namespaced endspoint for litigator hardship claims or not 😃 I originally added it to the non-namespaced endpoint as that is where all the other litigator endspoints were.

jsugarman: lets talk if over, but i am minded to add the namespaced version only for hardship and if @kmahern has time create a separate branch of off this branch that spikes swapping all the others over and deprecating the old versions.

app/views/pages/api_release_notes.html.haml
```diff
@@ -7,6 +7,32 @@
   = link_to 'Github issues', 'https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues'
   for bugs or other problems.
 
+%section.api-documentation
+  %hr
+  %h2 1st May 2020
```
jsugarman, line 12: will need to change, eventually

app/interfaces/api/v1/hardship_claim_params_helper.rb
kmahern: (as well as hopefully now also the other parameters I duplicated!)

spec/api/v1/external_users/claim_spec.rb
kmahern: 👍 I have created added the /api/external_users/claims/litigators/hardship and /api/external_users/claims/litigators/hardship/validate endpoints and a ticket for moving the other litigator claim endpoints (https://dsdmoj.atlassian.net/browse/CBO-1225)

colinbruce (approved):
LGTM :+1:

PR #3346, opened 2020-05-07
Remove hardship claims from redeterminations "queue"

jsugarman:
#### What
Remove hardship claims from redeterminations allocation filter

#### Ticket

[CBO-1200](https://dsdmoj.atlassian.net/browse/CBO-1200)

#### Why

Business requested to remove hardship redeterminations
from the redeteriminations "queue" (filter) for both
AGFS and LGFS hardships.

#### How
if the claim type is a hardship redetermination do not
treat as a redetermination in the allocaiton search
results object.

### DIFF
--- a/app/interfaces/api/entities/search_result.rb
+++ b/app/interfaces/api/entities/search_result.rb
@@ -75,7 +75,7 @@ module API
       end
 
       def redetermination
-        object.state.eql?('redetermination').to_i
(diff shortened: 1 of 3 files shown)


kmahern (approved):
👍

kmahern (approved):
👍 lgtm

PR #3353, opened 2020-05-11
CBO-1226: LGFS hardship fee page update

colinbruce:
#### What/Ticket
[remove unnecessary misc fee options from hardship page](https://dsdmoj.atlassian.net/browse/CBO-1226)

#### Why
The misc fee options in LGFS hardship are identical to the final fee options BUT some are incompatible with hardship claims. 
Both cost judge application and preparation cannot ever apply to a hardship claim and given these also can't data inject into CCLF onto a non-final fee claim its advisable to remove them entirely. 

#### How
Update Claims::FetchEligibleMiscFeeTypes  to add explicit handling for LGFS Hardship misc fee types, also updating the feature tests to ensure that only the correct fee types are displayed


kmahern (approved):
👍 lgtm

PR #3367, opened 2020-05-20
CBO-1236: allow archiving of hardship cases

colinbruce:
#### What
The caseworker user need was to retain hardship claims indefinitely, so we disabled a provider's ability to archive them. Providers weren't advised of this change in functionality and have been contacting via zendesk thinking there was a problem. I have responded advising of the rationale and asking for feedback about this approach as we didn't consider the impacts on the providers case view screens

#### Ticket
[Investigate hardship non-archiving strategy for provider view](https://dsdmoj.atlassian.net/browse/CBO-1236)

#### Why
Feedback contained in this ticket ( and will be added to as more comes in) indicates that the hardship claims may be cluttering up their claim views. So in order to improve the user experience we may need to come up with a method of moving the hardship claims away from the main claim view whilst still not archiving them - potentially creating another new state in the state machine

#### How
Add handlers for the new states required by hardship state changes

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -125,7 +125,7 @@ class ExternalUsers::ClaimsController < ExternalUsers::ApplicationController
 
   def unarchive
     claim_url = external_users_claim_url(@claim)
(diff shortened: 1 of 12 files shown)


jsugarman (approved):
👍 functionally looks ok.

PR #3373, opened 2020-05-29
Use basic executor from circleci where possible

jsugarman:
#### What
Use basic executor from circleci where possible

#### Why
To reduce run times. Thanks @colinbruce

certain jobs do not require anything other than a basic
docker container to run. circleci provides convenience 
images which are very fast to spin up
see [CircleCI Next-generation Convenience Images
](https://circleci.com/docs/2.0/circleci-images/#next-generation-convenience-images)
- slack (hold) notifications

#### Further improvements possible
- build and push docker images should be doable to BUT have to install python and awscli and this seems to negate the time savings gained from use the convenience image as executor. This can be tested further to ensure we are not losing potential time savings. one alternative is to extend the convenience image with python and awscli if this is not too slow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,14 +100,21 @@ references:
 
   _script-build-app-container: &script-build-app-container
     run:
-        name: Build and push cccd docker image
-        command: |
-          .circleci/build.sh
+      name: Build and push cccd docker image
+      command: |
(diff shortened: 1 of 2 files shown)


kmahern (approved):
lgtm 👍

PR #3381, opened 2020-06-03
Add correct id/object to edit user link

jsugarman:
#### What
Add correct id/object to edit user link

#### Ticket

[CBO-1259](https://dsdmoj.atlassian.net/browse/CBO-1259)

#### Why
Currently the link links based on an id
that is not the user in question. This could
result in a not found if no external_user with that id
exists or unauthorised if they exist but no in the same
chamber/firm or a working link to an entirely different
person, if `current_user` happens to have a access to that
persons account.

The route and controller expect the id of the
`external_user`, not the `user` itself.

#### TODO

needs some manual testing and possibly some automated testing
here.


kmahern (approved):
lgtm 👍

PR #3385, opened 2020-06-08
Increase server and cloning timeout

jsugarman:
#### What
increase the timeouts for all requests and cloning. Also
present total file sizes in log in human readable format.

#### Ticket

[relates to CBO-1233](https://dsdmoj.atlassian.net/browse/CBO-1233)

#### Why
At time of writing approximately 6.5% of
cloning attempts failed, mostly due
to timeouts. This is a quick fix to assess
impact and avoid the work needed for more rubost
appraoches such as backgrounding.

#### How
Increase unicorn web server request timeouts from
15 to 20 seconds generally, and cloning timeouts
inline with this.


kmahern (approved):
:lgtm 👍

PR #3396, opened 2020-06-10
Chore: Tidy Javascripts

naseberry:
#### What
Tidying of JS aspects of app

#### Why
To make `package.json` valid and to remove redundant Js modules.

#### How
- Validate package.json by adding in the required key value pairs
- Remove redundant js modules that were required but were fully commented


package.json
```diff
@@ -1,4 +1,18 @@
...
+    "type": "git",
+    "url": "git+https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence.git"
+  },
+  "author": "LAA Developers",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues"
+  },
+  "homepage": "https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence#readme",
   "dependencies": {
```
mathan11480-moj, line 16: Can you also add "main:application.js" for completion?

naseberry: thanks @mathan11480-moj but we don't utilise an entry file in this app.

mathan11480-moj: 👍 do we even ship this as an npm package?

naseberry: no, we do not @mathan11480-moj

jsugarman (approved):
👍 assuming this has been checked on an environment, this is great!

PR #3404, opened 2020-06-24
add bootsnap for faster loading

jsugarman:
#### What
adds bootsnap gem and use on boot

#### Why
This can have a [large performance benefit](https://github.com/Shopify/bootsnap#performance)
particularly for large apps such as CCCD.

[Rails also adds this as a default since 5.2](https://blog.bigbinary.com/2018/01/01/rails-5-2-adds-bootsnap-to-the-app-to-speed-up-boot-time.html)


naseberry (approved):

naseberry:
@jsugarman remember to add this to `config/boot.rb`

PR #3409, opened 2020-06-24
expose actual_trial_length for api creation of transfer claims

jsugarman:
#### What
enable transfer claims to be submitted via api with
and actual trial length

#### Ticket

[CBO-1340](https://dsdmoj.atlassian.net/browse/CBO-1340)

#### Why
currently the api does not expose this in swagger docs and a vendor queried this

--------

#### TODO (wip)

 - [X] expose in swagger docs
 - [x] add tests for actual trial length

### DIFF
--- a/spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
+++ b/spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
@@ -355,7 +355,7 @@ RSpec.describe 'API claim creation for LGFS' do
       let(:representation_order_date) { Date.new(2018, 03, 31).as_json }
 
       specify 'Case management system creates a valid scheme 9 transfer fee claim' do
-        post ClaimApiEndpoints.for(:transfer).create, claim_params.merge(offence_id: offence.id, **transfer_detail_params), format: :json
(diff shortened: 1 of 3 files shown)


jsugarman:
@jsugarman fix linter and merge

kmahern (approved):
lgtm 👍

spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
```diff
@@ -389,6 +389,7 @@ def results(claim)
         post endpoint(:expenses), expense_params.merge(claim_id: claim.uuid, expense_type_id: expense_hotel.id), format: :json
         expect(last_response.status).to eql 201
         expect(claim.expenses.size).to eql 2
+        expect(claim.actual_trial_length).to eql 11
```
jrmhaig, line 392: This is comparing with `eql` (as is done throughout the rest of the tests in this file). This ignores the type, so `11.0` would be accepted. Is this what we want or should we be ensuring that the correct types are returned? Is it possible, for instance, that a client application could fail if a particular value starts being returned as a float?

jrmhaig (approved):
Looks OK. I have added one comment but it isn't really important.

PR #3411, opened 2020-06-25
Add panel component helpers using a lib and railtie

naseberry:
#### What
Add the Govuk Panel component and replace existing markup

#### Ticket
[Add component - Panel](https://dsdmoj.atlassian.net/browse/CBO-1322)

#### Why

#### How
This pattern should make moving to a rails extension
gem easier, and make it easy to add components
in a modular fashion.

The GovukComponent::Railtie class mixes in the
helpers at boot time. see [this good blog on the subject](https://blog.capsens.eu/rails-6-boot-sequence-d289b44d2e94)

To get around the fact that railties need to be
intialized early in the boot sequence the railtie
is `require`'d in the `config/application.rb` (just after
normal Gem `require`ing). Once moved to a gem this can
be removed and would simply work as part of the normal
`Bundler.require(*Rails.groups)` call in that file.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -3,9 +3,13 @@ require_relative 'boot'
 require 'rails/all'
 
 require 'susy'
+
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
+
+# Custom railties that are not gems can be required here
+require_relative '../lib/govuk_component'
 
 module AdvocateDefencePayments
   class Application < Rails::Application
--- a/Gemfile
+++ b/Gemfile
@@ -118,6 +118,7 @@ group :test do
   gem 'kaminari-rspec'
   gem 'launchy', '~> 2.5.0'
   gem 'rails-controller-testing'
+  gem 'rspec-html-matchers', '~> 0.9.2'
   gem 'rspec-mocks'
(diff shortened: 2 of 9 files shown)


config/application.rb
```diff
@@ -2,11 +2,16 @@
 
 require 'rails/all'
 
+# TODO: test if this is (still) required
```
jsugarman: @naseberry unrelated but noticed this and not sure if we are using [susy]( https://www.oddbird.net/susy/)

jsugarman:
@naseberry are you wanting to get this PR in?

config/application.rb
naseberry: @jsugarman  Susy is being used by our style sheet, however it is now loaded via npm.

I need to create a task to test and if required remove redundant rails dependencies and configuration related to asset management.

jsugarman: cool, can we remove the comment then

naseberry: @jsugarman yes please

naseberry:
> @naseberry are you wanting to get this PR in?
@jsugarman Yes, I will deploy it to an env to be tested

jsugarman (approved):
👍  once tested on environment satisfactorily its good to go

naseberry:
@jsugarman would you care to review again. I have made the following changes:

- Updated helper
- Added rspec-html-matchers gem
- Update panel spec

PR #3421, opened 2020-07-02
Fix pcm not applicable error for supplementary claims with ptph fee

kmahern:
#### What
Provider's are currently unable to claim for PTPH on a supplementary claim because they are getting an error 'PCM not applicable'. This fixes that error to allow a provider to claim a quantity of between 1 and 3 PTPH on a supplementary claim (as with final claims). It also fixes the missing validation error messages so that a meaningful error is displayed.

#### Ticket
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&projectKey=CBO&modal=detail&selectedIssue=CBO-1163

#### Why
Software vendors have complained about this bug.

#### How
Amended existing validation for PTPH (which is based on case type) to include supplementary claims.

### DIFF
--- a/app/validators/fee/base_fee_validator.rb
+++ b/app/validators/fee/base_fee_validator.rb
@@ -74,7 +74,7 @@ module Fee
     end
 
     def validate_pcm_quantity
-      if @record.claim.case_type.try(:allow_pcmh_fee_type?)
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
LGTM 👍 

its a shame the supplementary claims do not have a case type or we could find some other way to work around the unfortunate `allow_pcmh_fee_type` issue. will have to do for now.

PR #3422, opened 2020-07-02
Fix message-body overflow

naseberry:
#### What
Add word break to long strings in claims message in order to resolve an overflow issue.

#### Ticket
[Text overflow](https://dsdmoj.atlassian.net/browse/CBO-1315)


kmahern (approved):
lgtm 👍

PR #3424, opened 2020-07-03
Add details component

naseberry:
#### What
Add the Govuk deatils component and replace existing markup

#### Ticket
[Add details component](https://dsdmoj.atlassian.net/browse/CBO-1320)

#### Why

#### How
 - [X] write component spec
 - [X] write component helper
 - [X] remove redundant styles

### DIFF
--- a/app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
+++ b/app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
@@ -44,13 +44,8 @@
       = f.hidden_field :price_calculated, value: f.object.price_calculated?
 
     .fee-calc-help-wrapper.form-group.hidden
-      %details
-        %summary
-          %span.summary
-            = t('.help_how_we_calculate_rate_title')
-        .panel.panel-border-narrow
-          %p
-            = t('.help_how_we_calculate_rate_body')
+      = govuk_detail t('.help_how_we_calculate_rate_title') do
+        = t('.help_how_we_calculate_rate_body')
 
     - if f.object.fee_type.requires_dates_attended?
       .dates-wrapper.form-group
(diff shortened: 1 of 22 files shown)


app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
```diff
@@ -44,13 +44,8 @@
...
 
     .fee-calc-help-wrapper.form-group.hidden
-      %details
-        %summary
-          %span.summary
-            = t('.help_how_we_calculate_rate_title')
-        .panel.panel-border-narrow
-          %p
-            = t('.help_how_we_calculate_rate_body')
+      = govuk_detail t('.help_how_we_calculate_rate_title'), false do
```
jsugarman: `false` is default so i would do without the second arg here.

unless "openness" is dynamic (i.e the result of a method call) or non-default (i.e. true),  i would not add it - which it is not in this case and all cases below where it is hardcoded as `false`

In context, i only added the  `open` arg option (in laa-court-data-ui) for those situations where open needs to be decided by code (error handling in laa-court-data-ui examples). I defaulted it to false as that is the typical use. So i would only add a second arg at all if it references some kind of boolean method to determine openness or it needs to always be "open".

jsugarman: as and aisde, and not a blocker, you could do this in a one-liner i think - your preference entirely tho!
```
 = govuk_detail t('.help_how_we_calculate_rate_title')  { t('.help_how_we_calculate_rate_body') }
```

jsugarman:
apologies for the early review, but I like this stuff and its all awesome!

app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
naseberry: > as and aisde, and not a blocker, you could do this in a one-liner i think - your preference entirely tho!
> 
> ```
>  = govuk_detail t('.help_how_we_calculate_rate_title')  { t('.help_how_we_calculate_rate_body') }
> ```

One-liner fails, unable to resolve at present therefore continuing with the block approach.

jsugarman (approved):
👍  LGTM

PR #3433, opened 2020-07-13
Add tag component

naseberry:
#### What
Add the Govuk tag component and replace existing markup

#### Ticket
[Add tag component](https://dsdmoj.atlassian.net/browse/CBO-1325)

#### Why

#### How
 - [X] write component spec
 - [X] write component helper
 - [X] remove redundant styles

### DIFF
--- a/lib/govuk_component/tag_helpers.rb
+++ b/lib/govuk_component/tag_helpers.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module GovukComponent
+  module TagHelpers
+    def govuk_tag(body = nil, color = nil, **tag_options)
+      tag_options = prepend_classes('govuk-tag--' + color, tag_options) if color.present?
+      tag_options = prepend_classes('govuk-tag', tag_options)
+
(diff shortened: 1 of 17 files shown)


lib/govuk_component/tag_helpers.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module GovukComponent
+  module TagHelpers
+    def govuk_tag(body: 'nil', color: nil, **tag_options)
```
naseberry: What are your thoughts on using keyword arguments over options hash @jsugarman @kmahern

jsugarman: i like keyword arguments but rails tag helpers (and others like [ViewComponent](https://github.com/github/view_component)  don't use them I believe and so I would emulate that as far as possible - from a practical point of view i think keeping views as concise as possible is preferable, and these args are verbose.

jsugarman (approved):
👍  LGTM

PR #3439, opened 2020-07-15
Replace cleaner delegate lists with delegate_missing_to

jsugarman:
#### What
Replace cleaner delegate lists with delegate_missing_to

#### Why
Remove repetetive and fragile manual delegation
with new(ish) rails method.

see this blog for more
https://blog.bigbinary.com/2017/05/30/rails-5-1-adds-delegate-missing-to.html


naseberry (approved):
:+1:

PR #3441, opened 2020-07-20
Fix claim case type field

naseberry:
#### What
Values simply typed into Case Type and Court combo boxes are lost on "save and continue"

I've found if you simply type the desired value in the Case Type and Court combo boxed in CCCD then their values are cleared when "Save and Continue" is pressed. This happens for both AGFS and LGFS claims.

#### Ticket

[Court details not saving](https://dsdmoj.atlassian.net/browse/CBO-934)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1021,6 +1021,7 @@ en:
         case_type_fields:
           case_type: Case type
           case_type_hint: For example Trial
+          select_option: Choose a case type
           stage_type: What stage has the case reached?
           agfs_stage_type_hint: For example Trial started but not concluded
         cracked_trial_fields:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -1021,6 +1021,7 @@ en:
         case_type_fields:
           case_type: Case type
           case_type_hint: For example Trial
+          select_option: Please select
```
jsugarman: **not a blocker** content wise this might need to not use please. but not sure - i was told ages ago not to use please (hints generally say Choose an option or, more specifically "choose a case type"

jsugarman (approved):
👍 have not tested manually but LGTM

config/locales/en.yml
naseberry: > **not a blocker** content wise this might need to not use please. but not sure - i was told ages ago not to use please (hints generally say Choose an option or, more specifically "choose a case type"

Cheers, currently being reviewed by Alan and Dan.

PR #3445, opened 2020-07-23
Fix API integration specs

jsugarman:
#### What
Runs the specs as part of test suite and fix LGFS hardship claim

#### Ticket

N/A

#### Why
The specs were not running as part of test suite :( due to naming.
Plus the LGFS hardship claim has probably never been passing as
it is calling the wring endpoint and not providing the correct
offence :(


kmahern (approved):
👍 lgtm. Apologies for missing the failing LGFS hardship claim test.

PR #3447, opened 2020-07-24
Add warning text component

naseberry:
#### What
Add the Govuk Warning text component and replace existing markup

#### Ticket
[Add warning text component](https://dsdmoj.atlassian.net/browse/CBO-1388)

#### Why

#### How
 - [X] write component spec
 - [X] write component helper
 - [x] remove redundant styles


app/views/external_users/claims/_api_promo_banner.html.haml
```diff
@@ -1,17 +1,12 @@
 - if claim.new_record? && show_api_promo_to_user?
 
   .js-callout-banner{'data-setting': 'api_promo_seen'}
-    .notice
-      %i.icon.icon-important
-        %span.visuallyhidden
-          = "Note"
+    = govuk_warning_text 'Do you use a case management system?', 'Note'
```
jsugarman, line 4: **not a blocker** - translation

app/views/external_users/claims/_api_promo_banner.html.haml
```diff
@@ -1,17 +1,12 @@
...
-        = "Do you use a case management system?"
     %p
       = "You may be able to import claims without re-typing. "
       = link_to 'Read more information.', api_landing_page_path
 
     %ul.list
       %li.hide-link
-        = link_to 'Do not show this message again', settings_user_path(current_user.id, api_promo_seen: 1), method: :put, remote: true
\ No newline at end of file
+        = link_to 'Do not show this message again', settings_user_path(current_user.id, api_promo_seen: 1), method: :put, remote: true
```
jsugarman, line 12: **not a blocker** - translation

lib/govuk_component/warning_text_helpers.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module GovukComponent
+  module WarningTextHelpers
```
jsugarman, line 4: **not a blocker** - just in terms of naming - have you considered a "higher" level name that could be used to add other similar helpers - such as a for notice or alert type messages

jsugarman (approved):
👍 looks fine, but is the name too specific - is there a higher level of grouping here, e.g. GovukNoticeHelpers. On the other hand I see GDS have [isolated warning text ](https://design-system.service.gov.uk/components/warning-text/) as a "group" so not a blocker

lib/govuk_component/warning_text_helpers.rb
naseberry: I have had a similar discussion with design (Dan) as CCCD currently uses this component for warning / note / information.
We have decided to keep it consistent with the Govuk Design System and have further discussions later for a "higher" level name.

PR #3448, opened 2020-07-27
Remove dropzone gem

naseberry:
#### What
Removal of the Dropzonejs gem

#### Why
Dropzonejs is now loaded via yarn


jsugarman:
have you tested this on an actual environment @naseberry  - the reason being i don't believe the testing round this area is not great and there was some custom js that may or may not be being used

naseberry:
> have you tested this on an actual environment @naseberry - the reason being i don't believe the testing round this area is not great and there was some custom js that may or may not be being used

Only tested locally, I shall test this in an environment

naseberry:
@jsugarman I have sanity checked user journeys for agfs, lgfs and caseworker on staging.
All references to 'dropzone' are in reference to the multiple file upload component, this dependency is now loaded via yarn.

jsugarman:
> @jsugarman I have sanity checked user journeys for agfs, lgfs and caseworker on staging.
> All references to 'dropzone' are in reference to the multiple file upload component, this dependency is now loaded via yarn.

great, approving

jsugarman (approved):
👍

PR #3455, opened 2020-08-06
Add table component

naseberry:
#### What
Spike to investigate a table helper

#### Ticket
[Spike to investigate table component](https://dsdmoj.atlassian.net/browse/CBO-1325)

#### How
 - [X] write component spec

### DIFF
--- a/app/views/external_users/claims/_final_claim_cost_summary.html.haml
+++ b/app/views/external_users/claims/_final_claim_cost_summary.html.haml
@@ -1,136 +1,85 @@
--# Cost summary
+= govuk_table do
+  = govuk_table_thead do
+    = govuk_table_th { t('common.description') }
+    = govuk_table_th(false, true) { t('shared.summary.amount') }
+    = govuk_table_th(false, true) { t('shared.summary.vat_amount') }
+    = govuk_table_th(false, true) { t('shared.summary.total_amount') }
 
-%table.summary
-  %caption
-  %thead
-    %tr
-      %th{ scope: 'col' }
-        = t('common.description')
-      %th.numeric{ scope: 'col' }
-        = t('shared.summary.amount')
-      %th.numeric{ scope: 'col' }
-        = t('shared.summary.vat_amount')
-      %th.numeric{ scope: 'col' }
-        = t('shared.summary.total_amount')
 
-  %tbody
+  = govuk_table_tbody do
     - if claim.fixed_fee_case?
-      %tr.fx-seed-fixedFees{ data: { autovat: @claim.apply_vat? ? 'true' : 'false', seed: claim.raw_fixed_fees_total } }
-        %td
-          = t('shared.summary.fixed_fees')
-        %td.numeric
-          = claim.fixed_fees_total
-        %td.numeric
-          = claim.fixed_fees_vat
-        %td.numeric
-          = claim.fixed_fees_gross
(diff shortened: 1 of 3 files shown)


lib/govuk_component/table_helpers.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    def govuk_table_row(**tag_options, &block)
+      tag_options = prepend_classes('govuk-table__row', tag_options)
+      tr_block = capture(&block)
+
+      tag.tr(tr_block, tag_options)
+    end
+
+    def govuk_table_th(row = false, numeric = false, width = nil, **tag_options, &block)
```
naseberry, line 55: This is a case for keyword arguments if I ever saw one. I have identified three arguments (scope, numeric and width) for Govuk table header cell so far.
Keyword arguments makes it much more readable, `govuk_table_th(row: true, numeric: true, width: 'one-third')` instead of `govuk_table_th(true, true, 'one-third')`.

https://thoughtbot.com/blog/ruby-2-keyword-arguments

cc @jsugarman @kmahern

app/views/external_users/claims/_final_claim_cost_summary.html.haml
```diff
@@ -1,136 +1,84 @@
--# Cost summary
+= govuk_table do
+  = govuk_table_thead do
+    = govuk_table_th { t('common.description') }
+    = govuk_table_th(false, true) { t('shared.summary.amount') }
```
naseberry, line 4: See keyword arguments comment below.

This cell needs to be numeric (2nd arg), however as this helper uses positional arguments, I had to call pass the row arg (1st arg) in order to use it, plus this is not easily read.

If keyword arguments were used, then we could call the helper using `= govuk_table_th(numeric: true) { t('shared.summary.amount') }`

lib/govuk_component/table_helpers.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
```
naseberry, line 1: Helper needs drying out.

lib/govuk_component/table_helpers.rb
jsugarman: yes in the sense of clarity but the number of arguments is very large. handling via `tag_options` is another way that is often used for view helpers in particular when args get unwieldy. i am not sure what is the "better" way

```
 def govuk_table_th(**tag_options, &block)
      klass = 'govuk-table__header'
      klass += ' govuk-table__header--numeric' if tag_options[:numeric]
      klass += " govuk-!-width-#{options[:width]}" if tag_options[:width]
      tag_options = prepend_classes(klass, tag_options)
      tag_options[:scope] = 'row' if tag_options[:row]
      ...
```

```
- tag_options = { row: '1|true|whatever'}
= govuk_table_th(tag_options) { t('shared.summary.expenses.travel_expenses') }
= govuk_table_td(tag_options) { claim.expenses_total }
```

there may be a case of having a different method for `row` or `col` if this is a common usage pattern - have been reading clean code and use of booleans to do different things in a method is discouraged generally, as is arguments exceeding 3

lib/govuk_component/table_helpers.rb
jsugarman: Hi, @naseberry can this PR be closed or are you still intending to get it in?

PR #3473, opened 2020-08-20
Refactor the APIs lookup data endpoints

jsugarman:
Refactor API's dropdown data endpoints and specs

campsite refactor coming out of looking
at adding a litigator categorie and
inside/outside london question


kmahern (approved):
👍 code looks good possibly the pr description above could be made a little clearer?

PR #3475, opened 2020-08-20
Sentence case all dropdown data endpoint descriptions

jsugarman:
Applies the standard set for other content
by content designers for GDS.


kmahern (approved):
👍

PR #3477, opened 2020-08-21
refactor base_claim_spec

jsugarman:
refactor base_claim_spec

Correct nesting and use mock
fr vat_registered? test.


jsugarman:
merging as is just a spec refactor and no other devs available on the team today

PR #3479, opened 2020-08-25
Prevent deletion of succesful jobs for approx. 3 days

jsugarman:
#### What
Prevent deletion of succesful jobs for just under 3 days

#### Why
Cloud platforms has introduced a pipeline
that destroys successful jobs regardless
of `successfulJobsHistoryLimit` value.

#### How
This can be circumnavigated by using
`ttlSecondsAfterFinished`

see https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/Cronjobs.html#deploying-a-cronjob-to-your-namespace
for more information.


kmahern (approved):
👍 lgtm

PR #3485, opened 2020-08-28
Move ttlSecondsAfterFinished config to correct level

jsugarman:
#### What
reposition ttlSecondsAfterFinished config to correct level

#### Why
This is currently wrong and having
no impact - jobs are being cleaned
up by pipeline too quickly.

see [user-guide](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/Cronjobs.html#kubernetes-cronjobs) for more


naseberry (approved):
👍

PR #3489, opened 2020-09-01
add CLAR warning test for LGFS claim

jsugarman:
Display caseworker-side warning of LGFS claim uninjected CLAR fees

#### What
provide caseworker warning to case workers for submitted claims that have CLAR fees on them

#### Why
because the CLAR fees are not injected into CCLF

#### Ticket
[CBO-1420](https://dsdmoj.atlassian.net/browse/CBO-1420)


kmahern (approved):
👍 lgtm

PR #3495, opened 2020-09-02
Cbo 1354 scheme 12 amendments to mi report

kmahern:
#### What
Add two new columns to MI report; Rep order issued date and Misc fees

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1354

#### Why
To enable caseworkers to identify CLAR claims

#### How

Rep order issued date is the rep order date associated with the earliest rep order on the claim; Misc fees is a space separated list of misc fee type descriptions listing the misc fee types associated with the claim.

--------

#### TODO (wip)

 - [ ] Check with Angela/someone from the business that they are happy with the new report format? (Are we happy with a space separated list of misc fees? I have attached a copy of the conversation with Angela, as well as a screen shot and copy of the new report generated locally, to the ticket)


app/presenters/concerns/claim_csv_helpers.rb
```diff
@@ -0,0 +1,33 @@
+module ClaimCSVHelpers
```
jsugarman: whats the reason for breaking out this concern. you generally don't use a concern/mixin unless it needs its logic shared with another class, or, perhaps if its super long  - see [this article ](https://rollout.io/blog/when-to-be-concerned-about-concerns/). This does not seem that long and the abstraction does not aid readability IMO

kmahern: Code climate was complaining, it has a max 24 methods per class, I added 2 which took it to 26.

spec/services/stats/management_information_generator_spec.rb
```diff
@@ -22,8 +22,8 @@
       expect(contents.size).to eq(valid_claims.size + 1)
     end
     
-    it 'has 19 columns' do
-      expect(contents.first.split(',').count).to eql 19
+    it 'has 21 columns' do
```
jsugarman: Not you, i know but it might be clearer and more useful to have something like

```
let(:report_columns) { [whatever whatever1] }

it 'has expected columns' do
  expect(contents.first.split(',').to match_array(report_columns)
end
```

app/presenters/concerns/claim_csv_helpers.rb
jsugarman: understandable but the stuff you have removed seems to be the core functionality of this class - i would either amend rubocop.yml to exclude this file from that cop and not use a concern at all, or create a concern called Something like `ClaimCsvColumns` and put all the "column" methods in there, and include that in here

app/presenters/concerns/claim_csv_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+  extend ActiveSupport::Concern
+
+  COMPLETED_STATES = %w[rejected refused authorised part_authorised].freeze
+  SUBMITTED_STATES = %w[submitted redetermination awaiting_written_reasons].freeze
+
```
jsugarman: also don't you need a `included do` block to mixin the methods? in any event

app/presenters/claim_csv_presenter.rb
```diff
@@ -1,40 +1,12 @@
 class ClaimCsvPresenter < BasePresenter
```
jsugarman: Also, again, not you, but this class seems to only be used by the ManagementInformationGenerator for generating the MI report. So its name would be clearer as ManagementInformationPresenter

app/presenters/concerns/claim_csv_helpers.rb
kmahern: 👍 will create a ClaimCsvColumns concern

app/presenters/management_information_presenter.rb
```diff
@@ -30,4 +33,17 @@ def clean_deallocations(journey)
     end
     journey
   end
+
+  def claim_state
```
jsugarman: is this not another column value method name that can go in the concern?

app/presenters/management_information_presenter.rb
```diff
@@ -30,4 +33,17 @@ def clean_deallocations(journey)
...
+    if state == 'archived_pending_delete'
+      claim_state_transitions.claim_state_transitions.sort.last.from
+    else
+      state
+    end
+  end
+
+  def previous(next_step)
```
jsugarman, line 37: and this method looks like it is only used by the `previous_redetermination_steps` method, so would be better being below that (in the concern), for clarity. unless i am missing something

jsugarman: forget that, probably better in here coz of dependence on `sorted_and_filtered_state_transitions`

app/presenters/management_information_presenter.rb
jsugarman: the idea being that any method that could be applied to a claim generally could go in the concern

app/presenters/concerns/claim_csv_helpers.rb
kmahern: Instead as suggested I have added a concern management_information_reportable (to follow 'able' naming convention) containing all the column methods. Code climate has also made me refactor the af1_lf1_processed_by method (not sure why, as I haven't changed it) as it was complaining about cognitive complexity.

spec/services/stats/management_information_generator_spec.rb
kmahern: 👍 have added it 'has expected columns' test to the spec.

app/presenters/concerns/claim_csv_helpers.rb
kmahern: 👍 have added included do block to the  (although everything seemed to work without?)

app/presenters/claim_csv_presenter.rb
kmahern: 👍 Have renamed ClaimCsvPresenter to ManagementInformationPresenter

jsugarman (changes requested):

jsugarman (approved):
even better 🔥

app/presenters/management_information_presenter.rb
kmahern: There is actually no column claim_state (see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/config/settings.yml) and I think this code is redundant - I have removed it.

PR #3498, opened 2020-09-04
Add CLAR Fee type validations

jsugarman:
#### What
Introduce fee type specific validations for the new
CLAR fee types across AGF and LGFS

#### Ticket

[CBO-1421](https://dsdmoj.atlassian.net/browse/CBO-1421)

#### Why
The fee types introduce some new validations
that are specific to them (and some more widely
applicable).

#### How
Extend the fee type rules to be able to
validate based on relations and their attributes.

--------

#### TODO (wip)

 - [x] Unused material (both types) not valid for fixed-fee case types
 - [x] Paper heavy case not valid for offence bands 1, 6 and 9
 - [ ] All new fees not valid for less than scheme 12 - move to separate PR if needed
 - [ ] Unused material (both types) are mutually exclusive - move to separate PR if needed


kmahern (approved):
lgtm 👍

PR #3506, opened 2020-09-09
Correct unused material (over 3 hours) boundaries

jsugarman:
Unused materials over 3 hours should be at least 3.01 hours to be valid

3 hours is not valid. currently it is accepting 3.00
hours incorrectly.


kmahern (approved):
Looks good 👍 .

PR #3507, opened 2020-09-10
Rename uppercase fee validator namespace folders to lowercase

jsugarman:
Inline with app dir naming - left over from
CLAR lgfs changes PR


kmahern (approved):
👍

PR #3508, opened 2020-09-10
Remove `agfs_scheme_11` fee_type dropdown data filter

jsugarman:
#### What
Remove `agfs_scheme_11` fee_type dropdown data filter
from API swagger doc endpoint

#### Why
There is no agfs_scheme_11 role and therefore we cannot
filter by this. Currently causing error when this filter
is selected

We may want to map role filter behavior differently so
that, but am asking vendors:

 - agfs_scheme_11 returns agfs_scheme_10 fee types
 - agfs_scheme_12 returns agfs_scheme_10 + agfs_scheme_12 fee types

### DIFF
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -14,7 +14,7 @@ module API
           optional :role,
                    type: String,
                    desc: I18n.t('api.v1.dropdown_data.params.role_filter'),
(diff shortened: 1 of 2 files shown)


kmahern (approved):
👍

kmahern:
Oops reviewed before I saw it looks like it is actually causing a spec failure API::V1::DropdownData GET api/advocate_categories[:category] when role is agfs_scheme_11 returns agfs scheme 10+ advocate categories - spec.api.v1.dropdown_data_spec

PR #3509, opened 2020-09-10
Remove Unused material (over 3 hours) quantity validation

jsugarman:
#### What
Remove Unused material (over 3 hours) quantity validation

#### Ticket

TODO

#### Why
We have become aware of a logic issue with the current
implementation of the CLAR “Unused material (over 3 hours)”
fee type for AGFS. The regulations stipulate that the first 3
hours are claimable at the fixed rate in addition to any hours
over 3 at an hourly rate. This means we will be allowing any
quantity to be claimed for this fee type where each unit/hour
should be in excess of the first 3 hours.

example:
  3 and half hours of work for “Unused material” would be claimable
  as two fees: `quantity: 1` of type “Unused material (upto 3 hours)”
  and `quantity: 0.5` of type “Unused material (over 3 hours)”.

### DIFF
--- a/app/views/pages/api_release_notes.html.haml
+++ b/app/views/pages/api_release_notes.html.haml
@@ -99,7 +99,9 @@
               Decimal
               %code
                 quantity
-              of over 3 (3.01+) only (not required for LGFS)
+              of over 0 (not required for LGFS)
+              %sup{ style: "font-size: 10px"}
+                3
               %br
               %br
               Only claimable on claims with a Graduated case type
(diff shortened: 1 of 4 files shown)


app/views/pages/api_release_notes.html.haml
```diff
@@ -99,7 +99,9 @@
               Decimal
               %code
                 quantity
-              of over 3 (3.01+) only (not required for LGFS)
+              of over 0 (not required for LGFS)
+              %sup{ style: "font-size: 10px"}
```
naseberry, line 103: minor: instead of using inline css, you could use one of the font override classes. see: https://design-system.service.gov.uk/styles/typography/#font-override-classes

spec/validators/fee/base_fee_validator_spec.rb
```diff
@@ -39,7 +39,7 @@
     it { should_error_if_not_present(fee, :claim, 'blank') }
   end
 
-  describe '#validate_fee_type' do
+  fdescribe '#validate_fee_type' do
```
naseberry: typo.

jsugarman: thats a focus command to run a specific spec rather than a typo, but it should not be there regardless 👍

naseberry (approved):
👍

PR #3511, opened 2020-09-11
Remove non-existent role reference

jsugarman:
#### What
Remove agfs scheme 11 role reference from api docs

#### Why
Should have been removed with[ previous PR](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3508) that
removed it from the role filter list

This role does not exist. Note, that it
should exist and a seperate piece of work
will be needed to apply it consistently (i.e. feature debt)

### DIFF
--- a/app/interfaces/api/v1/dropdown_data.rb
+++ b/app/interfaces/api/v1/dropdown_data.rb
@@ -58,7 +58,7 @@ module API
             case scheme_role
             when :lgfs
               []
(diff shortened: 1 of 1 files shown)


jsugarman:
Am going to force merge this as should have gone in with [previous PR](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3508)

PR #3512, opened 2020-09-11
Consistent agfs_scheme_12 fee type roles behaviour

jsugarman:
#### What
Apply agfs_scheme_12 role consistently for fee types

#### Ticket

[CBO-1479](https://dsdmoj.atlassian.net/browse/CBO-1479)

#### Why
This was never done for AGFS scheme 11 and has led to complexity/confusion
and lack of consistency in the codebase. Scheme 12 (CLAR accelerated measures)
has, so far, merely added the new fee types as having the agfs scheme 12 "role"
but really should give this role to all fee types applicable to scheme 12,
enabling more consistent handling and removal of complexity.

#### How

A patch is simply to promote the agfs_scheme_12s scope to the base_fee_type.

Instead we should
  - [x] amend seed files to add agfs_scheme_12 to all agfs_scheme_10 fee types
  - [x] run seed file as part of CLAR seeding
  - [x] amend rollback process to explicitly update any fee with agfs_scheme_10 to remove
  the agfs_scheme_12 role
  - [x] amend scopes in line with seed changes


kmahern (approved):
Tested locally and lgtm :

PR #3514, opened 2020-09-11
Ensure CLAR rollback task deletes ONLY new fee types

jsugarman:
#### What
Ensure CLAR rollback task deletes ONLY new CLAR fee types

#### Why
The change of fee scope in previous commit
means scheme 10 fee types are also scheme
12 fee types. meaning a rollback would delete
alot of fee types. Amend to use specific
unique_codes for assurance


naseberry (approved):
👍

PR #3516, opened 2020-09-11
Unused materials warning text

jsugarman:
#### What
Provider warning text on unused material (over 3 hours) fee types

#### Ticket

[CBO-1480](https://dsdmoj.atlassian.net/browse/CBO-1480)

#### Why
Mitigate user confusion

From a provider's perspective, if we're splitting 'up to 3 hours/over 3 hours' there's a real possibility for confusion. If a provider needs to claim for 5 hours of unused prep, they're going to be thinking "I enter 5 hours". Given the fee rules though, the system needs the provider to enter 2 hours (the extra hours).

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -29,6 +29,7 @@ Metrics/ClassLength:
   Exclude:
     - 'app/models/claim/base_claim.rb'
     - 'app/controllers/external_users/claims_controller.rb'
+    - 'app/presenters/claim/base_claim_presenter.rb'
 
 # Offense count: 68
 # Configuration parameters: CountComments, ExcludedMethods.
--- a/app/views/external_users/claims/misc_fees/advocates/_misc_fee_fields.html.haml
+++ b/app/views/external_users/claims/misc_fees/advocates/_misc_fee_fields.html.haml
@@ -1,4 +1,6 @@
 - fee = present(f.object)
+- claim = present(f.object.claim)
+
 .form-section.misc-fee-group.nested-fields.js-block.fx-do-init.fx-fee-group.fx-numberedList-item{ data: { 'type': 'miscFees', autovat: @claim.apply_vat? ? 'true' : 'false', 'block-type': 'FeeBlockCalculator' } }
   %span.bold-medium
     = t('.misc_fee')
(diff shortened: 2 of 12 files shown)


jsugarman:
@naseberry have started this off. The component helper makes it far neater

app/webpack/stylesheets/_shame.scss
```diff
@@ -869,3 +869,8 @@ table#download-files-report {
 .error {
   color: $error-colour;
 }
+
+#unused-materials-over-3-hours .js-hidden {
```
jsugarman: i reckon its better to just not use that `js-hidden` class when it is not needed - this could just lead to confusion

jsugarman: so i would not have it in the shared partial. if you need class for styling purposes i would create a separate one for clarity, no?!

app/views/external_users/claims/misc_fees/advocates/_misc_fee_fields.html.haml
```diff
@@ -13,13 +15,15 @@
...
-      = f.collection_select :fee_type_id,
-       @claim.eligible_misc_fee_types, :id, :description,
-       { include_blank: ''.html_safe },
-       { class: 'form-control form-control-full js-misc-fee-type js-fee-type js-fee-calculator-fee-type typeahead fx-misc-fee-calculation',
-       id: "misc_fee_#{@misc_fee_count}_fee_type",
-       'aria-label': t('.fee_type') }
+      = f.select :fee_type_id,
+        options_for_select(claim.eligible_misc_fee_type_options_for_select, f.object&.fee_type&.id),
+        { include_blank: ''.html_safe },
+        { class: 'new-fee-type form-control form-control-full js-misc-fee-type js-fee-type js-fee-calculator-fee-type typeahead fx-misc-fee-calculation',
```
jsugarman: remove new-fee-type class - debug leftover

app/webpack/javascripts/modules/external_users/claims/FeeTypeCtrl.js
```diff
@@ -0,0 +1,69 @@
+moj.Modules.FeeTypeCtrl = {
+  activate: function () {
+    return $('#claim_form_step').val() === 'miscellaneous_fees';
```
jsugarman, line 3: great 👍

.rubocop_todo.yml
```diff
@@ -29,6 +29,7 @@ Metrics/ClassLength:
   Exclude:
     - 'app/models/claim/base_claim.rb'
     - 'app/controllers/external_users/claims_controller.rb'
+    - 'app/presenters/claim/base_claim_presenter.rb'
```
kmahern, line 32: non-blocking refactor, possibly using concern?

app/webpack/javascripts/modules/external_users/claims/FeeTypeCtrl.js
```diff
@@ -0,0 +1,69 @@
+moj.Modules.FeeTypeCtrl = {
```
kmahern, line 1: Probably need some tests around this javascript?, but I think this is what @naseberry is already working on.

.rubocop_todo.yml
jsugarman: not really enough to justify a concern in itself, especially as it is not shared (relying oon inheritance instead in the presenters generally). Is Tech debt generally from the size of the class. 

Reminds me that we can/should have tech debt tickets around fixing certain rubocop TODOs as well, including this one - will add to the tech debt document for now

kmahern (approved):
lgtm from a Ruby/Rails perspective 👍

app/webpack/javascripts/modules/external_users/claims/FeeTypeCtrl.js
jsugarman: yes, he is

PR #3517, opened 2020-09-15
Remove email notification enabled flag

kmahern:
#### What
Remove email_notification_enabled? feature flag

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1490

#### Why


jsugarman (approved):
LGTM 👍  - i would have hoped there was a feature to test this minimally but hey ho

PR #3518, opened 2020-09-15
Remove timeout retention enabled flag

kmahern:
What
Remove timeout_retention_enabled? feature flag

Ticket
https://dsdmoj.atlassian.net/browse/CBO-1492

Why


jsugarman (approved):
LGTM 👍

PR #3519, opened 2020-09-15
Remove api_promo_enabled feature flag

kmahern:
#### What
Remove api_promo_enabled? feature flag

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1494

#### Why

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -25,7 +25,7 @@ module ClaimsHelper
   end
 
   def show_api_promo_to_user?
-    Settings.api_promo_enabled? && current_user.setting?(:api_promo_seen).nil?
+    current_user.setting?(:api_promo_seen).nil?
   end
 
   def show_claim_list_scheme_filters?(available_schemes)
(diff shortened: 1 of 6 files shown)


jsugarman (approved):
Looks good 👍 

i do wonder whether we need the promo at all anymore but i guess it could be useful for new accounts

jsugarman:
The wording of the promo should probably be changed as it refers to "importing" which we remove last year or so

kmahern:
@jsugarman  Do you have a suggestion what we should change the wording to? (Maybe "You may be able to enter claims without re-typing"? Or would it be better just to get rid of it altogether? (there are feature tests that run to test it so it is probably costing us money for circleci)

jsugarman:
> @jsugarman Do you have a suggestion what we should change the wording to? (Maybe "You may be able to enter claims without re-typing"? Or would it be better just to get rid of it altogether? (there are feature tests that run to test it so it is probably costing us money for circleci)

for now i reckon we could get away with just changing the word "import" to "create"
```
= "You may be able to import claims without re-typing. "
```

kmahern:
> > @jsugarman Do you have a suggestion what we should change the wording to? (Maybe "You may be able to enter claims without re-typing"? Or would it be better just to get rid of it altogether? (there are feature tests that run to test it so it is probably costing us money for circleci)
> 
> for now i reckon we could get away with just changing the word "import" to "create"
> 
> ```
> = "You may be able to import claims without re-typing. "
> ```

👍 Have made suggested change to use "create"

PR #3520, opened 2020-09-15
Enable clar on non prod envs ready for seeding

jsugarman:
#### What
Enable clar on non prod envs ready for seeding

#### Why
before seeding the new fee scheme we must set
this env var/flag. Note the flag is
also used directly to show link to the
published (hopefully) unused materials form


kmahern (approved):
👍 lgtm

PR #3524, opened 2020-09-16
Validate unused material against "trial" case types only

jsugarman:
#### What
Prevent unused material against claims with case types
that are not "trial" case types

#### Ticket

[CBO-1486](https://dsdmoj.atlassian.net/browse/CBO-1486)

#### Why
Only aplicable to the type of claims across LGFS and AGFS

### DIFF
--- a/app/models/case_type.rb
+++ b/app/models/case_type.rb
@@ -30,6 +30,7 @@ class CaseType < ApplicationRecord
   scope :fixed_fee,               -> { where(is_fixed_fee: true) }
   scope :not_fixed_fee,           -> { where(is_fixed_fee: false) }
   scope :graduated_fees,          -> { where(fee_type_code: Fee::GraduatedFeeType.pluck(:unique_code)) }
+  scope :trial_fees,              -> { where(fee_type_code: %w[GRCBR GRRAK GRRTR GRTRL]) }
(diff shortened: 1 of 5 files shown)


spec/services/claims/fetch_eligible_misc_fee_types_spec.rb
```diff
@@ -25,10 +25,14 @@
   context 'scheme 12 claim' do
     let(:claim) { create(claim_factory, :agfs_scheme_12) }
 
-    it 'returns misc fee types for AGFS scheme 10+ plus 12 without supplementary-only fee types' do
+    it 'returns misc fee types for AGFS scheme 12 without supplementary-only fee types' do
       is_expected.to have_at_least(1).items
       is_expected.to match_array Fee::MiscFeeType.agfs_scheme_12s.without_supplementary_only.map(&:unique_code)
     end
+
+    context 'when claim is a fixed fee type' do
```
kmahern: something missing here?

jsugarman: yep, left in branch inadvertently. remove now

kmahern (approved):
lgtm 👍

PR #3525, opened 2020-09-16
todos - remove update_amount_assessed

kmahern:
#### What
Remove update_amound_assessed, update_values and update_values! methods

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1495

#### Why
Remove code that is only used by specs (and fix code climate issue).

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -46,23 +46,6 @@ class Assessment < Determination
     save!
   end
 
-  # TODO: this appears to only be used by update_amount_assessed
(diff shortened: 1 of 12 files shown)


jsugarman (approved):
certainly looks ok to delete. I have checked for any obvious metaprogramming calls that could match and can't find any.

An interesting gem, [coverband](https://github.com/danmayer/coverband) i was just reading about in a medium article this morning could help to verify this and find other dead code.

PR #3529, opened 2020-09-17
AGFS - only display the Unused materials CLAR fee types for applicable case types

jsugarman:
#### What
AGFS - only display the Unused materials CLAR fee types for applicable case types

#### Ticket

[CBO-1463](https://dsdmoj.atlassian.net/browse/CBO-1463)

#### Why
AGFS, only display Unused materials CLAR fee types for applicable AGFS claims.

Final, hardship and supplementary claims with a case type or case stage
of trial, cracked trial, retrial and cracked before retrial


kmahern (approved):
lgtm 👍

PR #3533, opened 2020-09-18
Add accessibility statement

naseberry:
#### What
Publication of an Accessibility statement for the service

#### Ticket
[Add accessibility statement](https://dsdmoj.atlassian.net/browse/CBO-1467)

#### Why
The need to meet the legal requirements.

### DIFF
--- a/app/views/layouts/_footer.html.haml
+++ b/app/views/layouts/_footer.html.haml
@@ -4,6 +4,8 @@
   %li
     = govuk_footer_link_to t('contact'), contact_us_page_path
   %li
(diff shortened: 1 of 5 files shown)


jsugarman (approved):
LGTM and no blockers but from a maintenance and ease of change point of view would it be easier
to have one, or just a few, "html sections" in the translation files instead of the detailed breakdown for list items etc..

naseberry:
> LGTM and no blockers but from a maintenance and ease of change point of view would it be easier
> to have one, or just a few, "html sections" in the translation files instead of the detailed breakdown for list items etc..

updated @jsugarman

PR #3536, opened 2020-09-22
todos - share common defendant uplift fees aggregation validation

kmahern:
#### What
share share common defendant uplift fees aggregation validation as suggested by TODO

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1495

#### Why
DRY specs


spec/validators/claim/advocate_hardship_claim_validator_spec.rb
```diff
@@ -274,181 +281,30 @@
...
-            }
-          )
-          expect(claim).to be_valid
-        end
+      it 'should not error' do
+        should_not_error(claim, :base)
       end
 
-      context 'when defendants marked for destruction' do
+      context 'and form step is basic fees' do
```
jsugarman: not a blocker general rule of thumb use `when`, `with`, `for` rather than `and` here

spec/validators/claim/advocate_claim_validator_spec.rb
```diff
@@ -7,7 +7,14 @@
 
   let(:claim) { create(:advocate_claim) }
 
-  before { seed_fee_schemes }
+  before(:all) do
+    seed_fee_schemes
+    seed_fee_types
```
jsugarman: 🤔 quite a heavy operation to seed all fee type - will increase test suite runtime a fair bit

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -97,12 +97,163 @@
...
+      should_error_with(claim, :supplier_number, 'invalid')
+    end
+  end
+end
+
+RSpec.shared_examples 'common defendant uplift fees aggregation validation' do |options|
+  let(:fee_type) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][0]) }
+  let(:fee_type_2) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][2]) }
+  let(:uplift_fee_type) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][1]) }
+  let(:uplift_fee_type_2) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][3]) }
```
jsugarman: 🤔 while this is briefer and reusable it sacrifices clarity

spec/validators/claim/advocate_hardship_claim_validator_spec.rb
```diff
@@ -7,7 +7,14 @@
...
 
-  before { seed_fee_schemes }
+  before(:all) do
+    seed_fee_schemes
+    seed_fee_types
+  end
+
+  after(:all) do
+    clean_database
+  end
```
jsugarman: This is used in a few places so implies we can either use a [shared context](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-context) or make part of the shared example. This would also mean we could change this mechanism in one place if we wanted to change it for performance reasons

spec/validators/claim/advocate_claim_validator_spec.rb
```diff
@@ -7,7 +7,14 @@
...
+  end
+
+  after(:all) do
+    clean_database
+  end
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3536/files#r492591528)

spec/validators/claim/advocate_hardship_claim_validator_spec.rb
kmahern: 👍  have amended to 'when form step is basic fees'

spec/validators/claim/advocate_claim_validator_spec.rb
kmahern: 👍 I have refactored to not seed all fee types, and use the factories instead (as per the way advocate_claim_validator _spec.rb and advocate_hardship_claim_validator_spec.rb worked before). I am also now using this approach for supplementary_claim_validator_spec.rb which previously seeded all fee types - might improve runtime??

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
kmahern: I have refactored to attempt to make this clearer, and similar to the existing solution. 

Existing spec functionality:-
    let(:miaph) { create(:misc_fee_type, :miaph) }
    let(:miahu) { create(:misc_fee_type, :miahu) }
    let(:midtw) { create(:misc_fee_type, :midtw) }
    let(:midwu) { create(:misc_fee_type, :midwu) }

proposed change:-
  when 'advocate_claim', 'advocate_hardship_claim'
    let(:fee_type_1) { create(:misc_fee_type, :miaph) }
    let(:fee_type_2) { create(:misc_fee_type, :midtw) }
    let(:uplift_fee_type_1) { create(:misc_fee_type, :miahu) }
    let(:uplift_fee_type_2) { create(:misc_fee_type, :midwu) }

spec/validators/claim/advocate_hardship_claim_validator_spec.rb
kmahern: 👍 have amended to not seed fee_types or use the before/after(:all) blocks

spec/validators/claim/advocate_claim_validator_spec.rb
kmahern: 👍 have amended to not seed fee_types or use the before/after(:all) blocks

kmahern: 👍 have amended to not seed fee_types or use the before/after(:all) blocks

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -164,6 +172,33 @@
       end
     end
 
+    if ['advocate_claim', 'advocate_hardship_claim'].include? options[:claim_type]
```
jsugarman: this conditional logic implies this should not be in the shared set at all i'd say

spec/validators/claim/advocate_supplementary_claim_validator_spec.rb
```diff
@@ -7,14 +7,7 @@
...
 
-  before(:all) do
-    seed_fee_schemes
-    seed_fee_types
-  end
-
-  after(:all) do
-    clean_database
-  end
+  before { seed_fee_schemes }
```
jsugarman: i'd would still be tempted to use the `before(:all)` hook - depending on how many tests are running. a before each hook would need to seed for every example.

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -108,30 +108,38 @@
 end
 
 RSpec.shared_examples 'common defendant uplift fees aggregation validation' do |options|
-  let(:fee_type) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][0]) }
-  let(:fee_type_2) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][2]) }
-  let(:uplift_fee_type) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][1]) }
-  let(:uplift_fee_type_2) { Fee::MiscFeeType.find_by(unique_code: options[:unique_codes][3]) }
+  case options[:claim_type]
```
jsugarman: This is a complex area but all this conditional logic implies it should not be shared  (should not need it if behaviour is supposed to be same).

However, I think we can get rid of most if not all conditional logic and just use fee types that are valid/can-exist on any of the three claim types as all the test it trying to ensure is that uplifts are valid and invalid appropriately.

We can talk about this tomorrow but i reckon we should:

1. remove behaviour that is not shared between all three claims types to there or contexts or shared examples  - see my other comment below.
2. only use fee types here that are claimable on all three claim types, such as these:
```
Confiscation hearings (whole day),DTW,MIDTW
Confiscation hearings (whole day uplift),DWU,MIDWU
Deferred sentence hearings,DSE,MIDSE,
Deferred sentence hearings uplift,DSU,MIDSU
```
and get rid of the case statement and options entirely

3. For clarity rename `uplift_fee_type_1` to `uplift_for_fee_type_1` & rename `uplift_fee_type_2` to `uplift_for_fee_type_2`

spec/validators/claim/advocate_supplementary_claim_validator_spec.rb
kmahern: 👍 Have amended to use the before(:all) hook in a shared_context 'seeded fee schemes', as suggested above.

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
kmahern: 👍 Have removed conditional logic and refactored into shared example 'common defendant basic fees aggregation validation'.

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
kmahern: 1. 👍 Have created 3 new shared examples contexts; shared_context 'seeded fee schemes', shared_examples 'common defendant uplift fees aggregation validation' and shared_examples 'common defendant basic fees aggregation validation' .

2. 👍 Have amended to use suggested fee types that are claimable on all three claim types.

3. I have reverted to the original fee_type naming (midtw, midse...) for clarity.

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -97,12 +97,227 @@
...
+            }
+          }
+        )
+        expect(claim).to be_invalid
       end
     end
   end
+end
+
+RSpec.shared_examples 'common defendant basic fees aggregation validation' do
```
jsugarman, line 298: 👍  great, makes sense

spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -97,12 +97,227 @@
...
+    let(:advocate) { build(:external_user, :advocate, supplier_number: '9G606X') }
+
+    it 'should error when the supplier number does not match pattern' do
+      claim.external_user = advocate
+      should_error_with(claim, :supplier_number, 'invalid')
+    end
+  end
+end
+
+RSpec.shared_context 'seeded fee schemes' do
```
jsugarman: cool. **not a blocker** but for contexts i would use a command type statement - seed-fee-schemes - without  or without the hyphens

kmahern: 👍 madre suggested change "seed-fee-schemes"

jsugarman (approved):
👍

PR #3538, opened 2020-09-22
Add unused fee quantity logic

naseberry:
#### What
Change the state of the quantity field to be readonly with a default value when user chooses a misc. fee type of 'unused material upto 3 hours'

#### Ticket
[Change quantity attribute for unused material upto 3 hours](https://dsdmoj.atlassian.net/browse/CBO-1422)

#### Why
Validation based on the user flow and calculation of selected fee types.


kmahern (approved):
have pulled code and tested and lgtm 👍

PR #3539, opened 2020-09-22
Revert changes to feetype quantity

naseberry:
Quantity logic applied to all supplementary fees, this is incorrect.


kmahern (approved):
👍 Tested creating a supplementary claim, with various misc fees (including Unused materials upto 3 hours) and lgtm.

PR #3544, opened 2020-09-24
Update details component

naseberry:
#### What
Remove an argument from the govuk_details helper as it should be an html options.
Remove conflicting JS that has made the details component inaccessible for IE11 users.

#### Tickets
- [Update govuk_details helper](https://dsdmoj.atlassian.net/browse/CBO-1536)
- [Fix IE11 details component bug](https://dsdmoj.atlassian.net/browse/CBO-1502)

#### Why
To make the component helper cleaner and to resolve a bug which makes the component inaccessible within IE11.


jsugarman (approved):
👍

PR #3545, opened 2020-09-24
Run background worker (sidekiq) in its own pod(s)

jsugarman:
#### What
Move background worker process (sidekiq) to its own pod(s)

#### Ticket

[CBO-1487](https://dsdmoj.atlassian.net/browse/CBO-1487)

#### Why
So as not to drain web server processes and cpu
and prevent need for daemonization (which 
introduces other requirements for server OS - dedicated
server process manager such as Supervisor for sidekiq).

This will also unblock our ability to upgrade to
sidekiq 6+


jsugarman:
@naseberry @kmahern  - this PR contains commits in which i am amending the config for each environment we have. It would be good if you guys could check these commits in particular to really check that i have not missed any env var reference or namespace reference. Its the kind of thing where I could miss something small and we may not find out about it til something breaks on the given environment unfortunately.

I am getting code-blindness from checking it myself so would be good to have fresh eyes double/triple check my changes.

You can just look at the commits I have named  `Add cccd-[env-name] worker k8s config and optimize`

kubernetes_deploy/dev/app-config.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
naseberry: circleci/deploy.sh `kubectl apply -f kubernetes_deploy/${environment}/app-config.yaml` is expecting extension 'yaml' instead of 'yml'. Ditto to all the other app-config.yml files.

jsugarman: brill, good spot. Thats the effect of my blindness. 👍

jsugarman: corrected those files names now `yml --> yaml`

app/controllers/concerns/password_helpers.rb
```diff
@@ -20,11 +20,10 @@ def update_password
 
   def deliver_reset_password_instructions(user)
     token, enc = Devise.token_generator.generate(user.class, :reset_password_token)
-    user.reset_password_token   = enc
+    user.reset_password_token = enc
     user.reset_password_sent_at = Time.now.utc
     user.save(validate: false)
-    message = DeviseMailer.reset_password_instructions(user, token, current_user.name)
-    message.deliver_later
+    DeviseMailer.reset_password_instructions(user, token, current_user.name)
```
kmahern, line 27: Not sure why we are no longer using deliver_later here?

jsugarman: because it has no effect and is actually misleading. Devise ends up overriding the `deliver_later` and actually doing a `deliver_now`, as mentioned in the commit message - these messages ended up not running in the worker/via-sidekiq as a result (for years)

jsugarman: This commit is actually atomic and could go in on its own to correct the existing bug

kmahern: 👍

jsugarman: I think i will add a comment re devise here actually as it need to be clear to other devs

kmahern (approved):
lgtm 👍

naseberry:
lgtm 👍

PR #3547, opened 2020-09-28
Attempt to fix urn display issue for long urns

kmahern:
#### What
Fix issue identified by @alanmaydwell where long urns were being partially covered by the status field

#### Ticket
n/a

#### Why
Fix display issue

#### How
With thanks to suggested solution from @naseberry , allow casenumber/urn field to split into multiple lines.


jsugarman:
@kmahern @naseberry what are we doing with this one?

naseberry:
> @kmahern @naseberry what are we doing with this one?

I shall leave the decision up to @kmahern

naseberry (approved):

PR #3549, opened 2020-09-28
Add scheme 12 seeds file

kmahern:
#### What
allow developer to run db:seed  and/or db:reload to see fee scheme 12. 

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1489

#### Why
follow conventions/allow developers to setup their local dev environment


jsugarman (approved):
👍

PR #3559, opened 2020-09-30
Attempt to remove aws-sdk gem

kmahern:
#### What
Remove aws-sdk gem

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1550

#### Why
This aws-sdk gem is very large; it is recommended to use it only as a quick way to migrate from V2 or if you depend on many AWS services. see https://docs.aws.amazon.com/sdk-for-ruby/v3/api/

The whole gem is currently being loaded as part of the build-application-container step in our CircleCI builds - this takes time and ultimately costs money.

By only loading the gems we used in CCCD hopefully this will be improved.

#### How
load only gems we are using (S3, SNS and SQS)

#### TODO (wip)

 - [X] Tested successful CCR DI on staging
 - [X] Tested intentional failed CCR DI attempt on staging (invalid rep order number) - failed message appears in Slack channel
 - [X] Tested successful CCLF DI on staging
 - [X] Tested intentional failed CCLF DI attempt on staging (invalid rep order number) - failed message appears in Slack channel
 - [X] Tested uploading a pdf as an advocate and viewing/downloading document as caseworker.
 - [X] Tested attaching a pdf to a message as a caseworker and downloading as an advocate.


jsugarman (approved):
great 👍, 

I assume your testing would have included creating and updating DB records so I guess RDS just works as part of another gem like PG or rails.

PR #3561, opened 2020-10-01
Amend circleci resource_classes

jsugarman:
#### What
use circleci resource_class to provide extra RAM and
CPU for rspec job

#### Why
Improve speed. ucrrenlty can take 10 to 15 minutes.

see https://circleci.com/docs/2.0/configuration-reference/#resource_class

#### Detail
- [x] add large resource_class to rspec. 
      10000+ tests taking 15 minutes. is a bottleneck!

- [x] add small resource for basic-executor and cloud-platform-executor
   These typically only need a small amount of resource but some jobs
   may require more (like build and push build). Specific jobs using
   one of these executors can have their resource overridden at the
   job level. 

I note that, possibly just a coincidence, that changing the
app container build job to `large` from `medium(default)` 
saw no improvement on build time, but that making it a
`small` did cause an improvement.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,6 +74,7 @@ references:
 # ------------------
 executors:
   basic-executor:
+    resource_class: small
     docker:
       - image: cimg/base:2020.01
         environment:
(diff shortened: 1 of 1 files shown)


naseberry (approved):
👍

PR #3563, opened 2020-10-02
Change sample user scripts to modules

jsugarman:
Change sample user scripts to be modules

#### What
amend sample users scripts to be modules with explicit call required

#### Why
So they are only run when explicitly called

Because these files exist in the lib folder, and are scripts, 
any eager loading of that folder will run them. 

This is not desirable as we only need to run these seeds 
on request, such as for the smoke test, or calling the
rake task.


naseberry (approved):
LGTM, however don't like some much commented lines of code in `lib/demo_data/external_user_seeder.rb`

PR #3564, opened 2020-10-02
Show error in CI pipeline when deployment fails

jsugarman:
#### What
Fail deployment overtly if "pods" do not reach "readiness"

#### Ticket

[CBO-1334](https://dsdmoj.atlassian.net/browse/CBO-1334)

#### Why
Currently deployment is green as long as the Kubernetes apply commands are successful, but the pods may still fail to start the application up and various other scenarios could prevent the deployment actually succeeding.

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -10,6 +10,10 @@ function _circleci_deploy() {
     deploy.sh dev
     "
 
+  # exit when any command fails
+  set -e
+  trap 'echo command at lineno $LINENO completed with exit code $?.' EXIT
+
   if [[ -z "${ECR_ENDPOINT}" ]] || \
       [[ -z "${GIT_CRYPT_KEY}" ]] || \
       [[ -z "${AWS_DEFAULT_REGION}" ]] || \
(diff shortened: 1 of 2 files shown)


.circleci/deploy.sh
```diff
@@ -10,6 +10,12 @@ function _circleci_deploy() {
     deploy.sh dev
     "
 
+  # exit when any command fails, keep track of the last
+  # executed command and output error message before exiting
+  # https://intoli.com/blog/exit-on-errors-in-bash-scripts/
```
jsugarman: remove

.circleci/deploy.sh
```diff
@@ -55,6 +61,8 @@ function _circleci_deploy() {
   echo "${GIT_CRYPT_KEY}" | base64 -d > git-crypt.key
   git-crypt unlock git-crypt.key
 
+  echo $(kubectl version)
```
jsugarman: remove

.circleci/deploy.sh
```diff
@@ -85,6 +89,10 @@ function _circleci_deploy() {
 
   kubectl annotate deployments/claim-for-crown-court-defence kubernetes.io/change-cause="$(date +%Y-%m-%dT%H:%M:%S%z) - deploying: $docker_image_tag via CircleCI"
   kubectl annotate deployments/claim-for-crown-court-defence-worker kubernetes.io/change-cause="$(date +%Y-%m-%dT%H:%M:%S%z) - deploying: $docker_image_tag via CircleCI"
+
+  # wait for rollout to succeed or fail/timeout
+  kubectl rollout status deployments/claim-for-crown-court-defence
```
kmahern, line 94: Not sure when this will timeout?

jsugarman: default is 10 minutes. to have more control we need to upgrade kubectl 1.12+ client or set `spec.progressDeadlineSeconds`

jsugarman: these are other issues that i will need to address in another PR.

kmahern (approved):
lgtm 👍

PR #3565, opened 2020-10-05
Adjust message control timing

naseberry:
#### What
We have removed the timeout on `.message-status` on the claim history page as this failed to meet WCAG 2.2.1.
We have also added a `role` attribute to the element to describe the role of the element to screen readers.
On a side note, we have also applied the GOV.UK grid systems to help with the layout of the message status.

#### Ticket
[Timing adjustable](https://dsdmoj.atlassian.net/browse/CBO-1501)

#### Why
Status message had a set timeout, this would problematic for cognitive impaired users who may not be able to decipher the text quickly enough before it disappears.

### DIFF
--- a/app/views/shared/_claim_history.html.haml
+++ b/app/views/shared/_claim_history.html.haml
@@ -8,12 +8,13 @@
 .messages-container.no-print
   .messages-list{ 'aria-label': t('shared.claim_accordion.h2_messages'), role: 'region', tabindex: 0 }
     = render partial: 'shared/history_items_list', locals: { claim: claim }
+
   .messages-print-link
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
👍 looks good to me from a code point of view

Have you checked with dan, angela regarding the preferences of case workers. I assume this `toggleStatusBar` feature had been requested at some point.

naseberry:
> 👍 looks good to me from a code point of view
> 
> Have you checked with dan, angela regarding the preferences of case workers. I assume this `toggleStatusBar` feature had been requested at some point.

Thanks, I am in discussion with Dan, I will speak with Angela.

PR #3566, opened 2020-10-05
Set travel expense input focus

naseberry:
#### What
The select element, onchange, uses jQuery's `.detach()` method to remove elements, in doing so the tab order position is lost, we reapply its position by setting the focus state on the select element. 

#### Ticket
[Element initiated change of context with no warning](https://dsdmoj.atlassian.net/browse/CBO-1504)

#### Why
When users attempt to interact with the 'Type of expense' select element using the arrow keys, additional select and input fields appear on the page, however, focus does not remain in the select element.


kmahern (approved):
👍 lgtm

PR #3568, opened 2020-10-05
Change logging for MI and stats jobs generally

jsugarman:
#### What
Amend MI logging

#### Why
Excessive in some cases and not working in other.
For MI I have added some to help identify when
the error occurs.


kmahern (approved):
lgtm 👍

PR #3570, opened 2020-10-06
Revert ingress annotations

jsugarman:
#### What
Revert ingress annotations

services' ingresses back to the default ingress controller
(we'll publish full details of why this is necessary in our
post-incident report).

> If you have a dedicated ingress controller (i.e. you have a
custom kubernetes.io/ingress.class annotation), please remove this
annotation from your ingress definitions before redeploying your service.

> If you redeploy your service and change back to your dedicated ingress
controller, after we have moved your ingress to the default ingress
controller, your service will be unavailable.


naseberry (approved):
👍

PR #3576, opened 2020-10-08
Attempt to fix caseworker allocation filter

kmahern:
#### What
Try to fix delays in claims appearing in caseworkers allocated claims list reported on Zendesk.

"We’ve had multiple reports over the last couple of weeks that instead of this being instantaneous its taking anything upto 15 minutes for the claims to disappear from the allocation page.  Similar delays have been reported in the transitions once a claim has been assessed by a caseworker. "

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1449

#### Why
The delays appeared to start following a Grape/Grape-Swagger/Rackdependency update (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3466/files). Investigation appeared to confirm that the delays didn't happen on a branch without this update.

The difference seems to be a result of a change to response.headers; (testing locally) before dependency update:-
```
:content_type=>"application/json",
 :vary=>"Accept-Encoding",
 :etag=>"W/\"2ca6dcecf50860fb0ca8e75b40a7a46c\"",
 :cache_control=>"max-age=0, private, must-revalidate",
 :x_meta_request_version=>"0.7.2",
 :x_request_id=>"3bb45ab8-6c96-480b-aac1-3c663486f00f",
 :x_runtime=>"0.098327",
 :transfer_encoding=>"chunked"
```
after update:-
```
:cache_control=>"no-cache",
 :content_type=>"application/json",
 :vary=>"Accept-Encoding",
 :etag=>"W/\"a1bb929082d714a3ce835f3d3d8555be\"",
 :x_meta_request_version=>"0.7.2",
 :x_request_id=>"e92abfbe-00e5-4df9-9786-be3f7ebeefb3",
 :x_runtime=>"1.338278",
 :transfer_encoding=>"chunked"
```
This resulted in response.ttl being evaluated to nil and api-request.rb instead using default_ttl (900 seconds).

#### How
Set ttl = 0 if cache_control == 'no-cache'

--------

#### TODO (wip)

 - [X] Tested creating a claim/allocating claims/reallocating claims on CCCD-dev
 - [ ] additional testing required?

### DIFF
--- a/lib/caching/response.rb
+++ b/lib/caching/response.rb
@@ -16,10 +16,14 @@ module Caching
     end
 
     def ttl
-      cache_control[/max-age=([0-9]+)/, 1] || 0 if cache_control == 'no-cache'
+      max_age.to_i
     end
 
     private
+
+    def max_age
+      cache_control[/max-age=([0-9]+)/, 1] || 0
+    end
 
     def cache_control
       headers[:cache_control] || ''
--- a/lib/caching/response.rb
+++ b/lib/caching/response.rb
@@ -16,7 +16,7 @@ module Caching
     end
 
     def ttl
-      cache_control[/max-age=([0-9]+)/, 1]
+      cache_control[/max-age=([0-9]+)/, 1] || 0 if cache_control == 'no-cache'
     end
 
     private
--- a/lib/caching/response.rb
+++ b/lib/caching/response.rb
@@ -16,14 +16,11 @@ module Caching
     end
 
     def ttl
-      max_age.to_i
+      return 0 if cache_control.match?('no-cache')
+      cache_control[/max-age=([0-9]+)/, 1]
     end
 
     private
-
-    def max_age
-      cache_control[/max-age=([0-9]+)/, 1] || 0
-    end
 
     def cache_control
       headers[:cache_control] || ''
--- a/lib/caching/api_request.rb
+++ b/lib/caching/api_request.rb
@@ -71,7 +71,7 @@ module Caching
     end
 
     def save!(response)
-      ttl = (response.ttl || default_ttl).to_i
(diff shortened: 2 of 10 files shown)


jsugarman:
So it looks like this will fix the 15 minute problem. 

However, while testing (i added specs) it looks like it is also overrides the max-age Cache-Control if it does exist on the (internal) api request - meaning we would never be able to cache by setting headers. I have therefore amended and refactored so that it does use `max-age` for `ttl`. 

Essentially what we are doing though is disabling caching for all api calls which would mean that page could then suffer from timeouts more. so we may need to tweak this specific endpoint to have some level of caching (we could assess and ask case workers what is acceptable)

NOTE: Digging into it it looks like this the header change to `no-cache` is an issue with grape in version 1.4 which we are using. Apparently this was **[fixed in version 1.5](https://github.com/ruby-grape/grape/pull/2088)** . so will try bumping as well.

lib/caching/response.rb
```diff
@@ -16,11 +16,16 @@ def headers
     end
 
     def ttl
-      cache_control[/max-age=([0-9]+)/, 1]
+      max_age
```
jsugarman: pointless abstraction. merge methods

kmahern:
👍 Yes I see, I had a bug with the operator precedence in my fix (I had `cache_control[/max-age=([0-9]+)/, 1] || 0 if cache_control == 'no-cache'` I think `cache_control[/max-age=([0-9]+)/, 1] || (0 if cache_control == 'no-cache')` would have worked) - however fix by @jsugarman is clearer.

kmahern:
👍 new specs

jsugarman (approved):
👍

PR #3581, opened 2020-10-14
CBO-1505 Group related items

naseberry:
#### What
We improved the information, structure, and relationships.
We also updated the markup of the partial to match the GOVUK Design System and removed redundant styles.

#### Ticket
[Content to be marked up in a semantic list](https://dsdmoj.atlassian.net/browse/CBO-1505)

#### Why
Screen reader users are unable to understand the layout and cannot relate to the content.
The claim summary page fails to meet WCAG 1.3.1


jsugarman (approved):
Excellent. 

Feature tests are pretty light on the summary page (they only check for existence of content i believe) so would require a visual confirmation - do you want me to check this?

Otherwise ✅

naseberry:
> Excellent.
> 
> Feature tests are pretty light on the summary page (they only check for existence of content i believe) so would require a visual confirmation - do you want me to check this?
> 
> Otherwise ✅

Thank you, visually confirmed with Dan locally.

PR #3582, opened 2020-10-14
CBO-1510 Non-descriptive labelling

naseberry:
#### What
We have added visually hidden text to duplicate label text to provide context to differentiate between the form labels.

#### Ticket
[Non-descriptive labelling](https://dsdmoj.atlassian.net/browse/CBO-1510)

#### Why
There are multiple form labels on a page with the same text, these are ambiguous for screen reader users that browse out of context.

### DIFF
--- a/app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
+++ b/app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
@@ -23,7 +23,7 @@
     .form-group
       - fee_type_scope = fee.fee_type_code.downcase
       = f.adp_text_field :quantity,
-        label: t(".#{fee_type_scope}.quantity"),
+        label: t('.quantity'),
         hint_text: t(".#{fee_type_scope}.quantity_hint"),
         input_classes: 'quantity js-fee-quantity js-fee-calculator-quantity form-control-1-4',
         input_type: 'number',
@@ -33,7 +33,7 @@
     .form-group.js-unit-price-effectee
       .calculated-unit-fee
       = f.adp_text_field :rate,
-        label: t('.rate'),
+        label: t('.rate_html', context: f.object.description),
         input_classes: 'rate fee-rate js-fee-calculator-rate form-input-denote__input form-control-1-4',
         input_type: 'currency',
         value: number_with_precision(f.object.rate, precision: 2),
--- a/app/views/external_users/claims/fixed_fees/advocates/_fixed_fee_fields.html.haml
+++ b/app/views/external_users/claims/fixed_fees/advocates/_fixed_fee_fields.html.haml
@@ -18,14 +18,14 @@
     - if f.object.fee_type.case_uplift?
       .form-group
         = f.adp_text_field :case_numbers,
-          label: t('.case_numbers'),
+          label: t('.case_numbers_html', context: f.object.description),
           input_classes:'js-fixed-fee-case-numbers',
           hint_text:'Separate by commas',
           errors: @error_presenter
 
     .form-group
       = f.adp_text_field :quantity,
-        label: t('.quantity'),
+        label: t('.quantity_html', context: f.object.description),
         hint_text: t('.quantity_hint'),
         hide_hint: true,
         input_classes:"quantity fee-quantity form-control-1-8 js-fee-quantity js-fee-calculator-quantity",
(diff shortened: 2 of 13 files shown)


config/locales/en.yml
```diff
@@ -945,14 +946,15 @@ en:
...
+          net_amount_html: Net amount <span class="govuk-visually-hidden">in pounds for %{context}</span>
           help_how_we_calculate_rate_title: *how_is_this_calculated_help
           help_how_we_calculate_rate_body: We calculate the amount based upon the fee scheme, case type, offence, advocate category and fee type.
           npw:
             heading_html: Prosecution witnesses
             quantity: Total number of prosecution witnesses
             quantity_hint: *empty_input
           ppe:
-            quantity: Total number of pages
+            quantity: Total number of prosecution evidence pages
```
jsugarman: not-a-blocker: "Total number of pages of prosecution evidence" (i.e. PPE) might be a more typical rendering of this term

config/locales/en.yml
```diff
@@ -969,22 +971,23 @@ en:
...
+            quantity_hint: Number of daily attendance fee (2+)
           saf:
-            quantity_hint: *empty_input
+            quantity_hint: Number of standard appearance fee
           pcm:
-            quantity_hint: *empty_input
+            quantity_hint: Number of plea and trial preparation hearing
           cav:
-            quantity_hint: *empty_input
+            quantity_hint: Number of conferences and views
```
jsugarman: Have you tried this out. It may be pointless because the fee calculator JS will overwrite this with output from the api anyway. The "Number of days" hint is [set here I believe](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/261e13a847c6d99cdd30e8c37439ea38ffa92b8f/app/webpack/javascripts/modules/external_users/claims/fee_calculator/FeeCalculator.UnitPrice.js#L147)

Screen Shot 2020-10-16 at 09 22 42

jsugarman: This behaviour is also feature tested - [such as the here](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/features/fee_calculator/advocate/graduated_fee_calculator.feature#L50:L59)

app/views/external_users/claims/basic_fees/_additional_fee_fields.html.haml
```diff
@@ -23,7 +23,7 @@
     .form-group
       - fee_type_scope = fee.fee_type_code.downcase
       = f.adp_text_field :quantity,
-        label: t(".#{fee_type_scope}.quantity"),
+        label: t('.quantity'),
```
jsugarman, line 26: +1 - good to see removal of dynamic translations as will enable used on i18n-task missing and unused checks.

app/views/external_users/claims/fixed_fees/advocates/_fixed_fee_fields.html.haml
```diff
@@ -18,14 +18,14 @@
...
-          label: t('.case_numbers'),
+          label: t('.case_numbers_html', context: f.object.description),
           input_classes:'js-fixed-fee-case-numbers',
           hint_text:'Separate by commas',
           errors: @error_presenter
 
     .form-group
       = f.adp_text_field :quantity,
-        label: t('.quantity'),
+        label: t('.quantity_html', context: f.object.description),
```
jsugarman, line 28: not a blocker, and i like the use of `context`, but since it is not using a translation it will always be english - not that we are or probably ever will add another language (welsh 😬 )

jsugarman (commented):
Alot of fiddly work. Two things strike me:

1. the use of `_html` translation for the adp_text_field label attribute implies we could/should be applying this at the `adp_text_field` form helper level in someway. Although this could have unintended widespread impact.

2. No tests have been added and only one low level change to feature page object has been made implying that some of these changes are not working (see `graduated_fee_calculator.feature`) while others are probably untested.

Happy to discuss in person as this is a gritty area to work in generally

naseberry:
> Alot of fiddly work. Two things strike me:
> 
> 1. the use of `_html` translation for the adp_text_field label attribute implies we could/should be applying this at the `adp_text_field` form helper level in someway. Although this could have unintended widespread impact.
> 2. No tests have been added and only one low level change to feature page object has been made implying that some of these changes are not working (see `graduated_fee_calculator.feature`) while others are probably untested.
> 
> Happy to discuss in person as this is a gritty area to work in generally

In response to your points:

1. I'm agreement in the use of `_html` translation for the adp_text_field, however, as we have been discussing the migration to the new formbuilder, I think it would be best to handle it then.
2. We need a SPIKE to review feature specs

jsugarman (approved):
👍 LGTM

PR #3583, opened 2020-10-14
Update claim history messaging feature

naseberry:
#### What
- We have removed the timeout function attached to the status message
- We have added a `role` attribute to provide feedback to screen reader users after they have completed an action.

#### Tickets
- [Adjust message control timing](https://dsdmoj.atlassian.net/browse/CBO-1501)
- [Status message audio feedback](https://dsdmoj.atlassian.net/browse/CBO-1508)

#### Why
- Status message has a set timeout and disappears after a few seconds, this is problematic for cognitive impaired users who may not be able to decipher the text quickly enough before it disappears.
- There is no audible feedback to alert screen reader users regarding the status of their action.


jsugarman (approved):
👍 LGTM

PR #3584, opened 2020-10-14
Enable anonymised database dump and restore

jsugarman:
#### What
Provide means to dump and restore database

#### Ticket

[CBO-1252](https://dsdmoj.atlassian.net/browse/CBO-1252)

#### Why
So we can debug production problems including
MI report failure

### DIFF
--- a/kubernetes_deploy/jobs/migrate.yaml
+++ b/kubernetes_deploy/jobs/migrate.yaml
@@ -9,7 +9,7 @@ spec:
     spec:
       restartPolicy: Never
       containers:
-        - name: cccd-app
+        - name: cccd-job
           image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-get-paid/cccd:app-latest
           command:
             - bundle
@@ -24,13 +24,10 @@ spec:
                 secretKeyRef:
                   name: cccd-rds
                   key: url
-            - name: SECRET_KEY_BASE
-              valueFrom:
-                secretKeyRef:
-                  name: cccd-secrets
-                  key: SECRET_KEY_BASE
             - name: AWS_REGION
               valueFrom:
                 secretKeyRef:
                   name: cccd-secrets
                   key: AWS_REGION
+            - name: SECRET_KEY_BASE
+              value: just-be-present
--- a/kubernetes_deploy/jobs/seed.yaml
+++ b/kubernetes_deploy/jobs/seed.yaml
@@ -9,13 +9,20 @@ spec:
     spec:
       restartPolicy: Never
       containers:
-        - name: cccd-app
+        - name: cccd-job
           image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-get-paid/cccd:app-latest
           command:
             - bundle
             - exec
             - rake
             - db:seed
+
+          envFrom:
+            - configMapRef:
+                name: cccd-app-config
+            - secretRef:
+                name: cccd-secrets
+
           env:
             - name: RAILS_ENV
               value: 'production'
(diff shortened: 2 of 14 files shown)


kubernetes_deploy/jobs/migrate.yaml
```diff
@@ -24,13 +24,22 @@ spec:
...
+            - name: ADVOCATE_PASSWORD
+              value: just-be-present
+            - name: ADMIN_PASSWORD
+              value: just-be-present
+            - name: SUPERADMIN_USERNAME
+              value: just-be-present
+            - name: SUPERADMIN_PASSWORD
+              value: just-be-present
+            - name: TEST_CHAMBER_API_KEY
+              value: just-be-present
```
jsugarman: remove all but secret key base

kubernetes_deploy/jobs/seed.yaml
```diff
@@ -24,43 +24,22 @@ spec:
...
-                  key: SECRET_KEY_BASE
+                  key: AWS_REGION
+            - name: SECRET_KEY_BASE
+              value: just-be-present
             - name: CASE_WORKER_PASSWORD
-              valueFrom:
-                secretKeyRef:
-                  name: cccd-secrets
-                  key: CASE_WORKER_PASSWORD
+              value: just-be-present
```
jsugarman: Just use `ConfigMapRef` and `SecretRef` for simplicity

lib/tasks/dump.rake
```diff
@@ -0,0 +1,267 @@
...
+
+    def sensitive_table_exclusions
+      # make sure a db:dump task exists for each of the excluded tables (i.e. db:dump:providers)
+      excluded_tables.map { |table| "--exclude-table-data #{table}" }.join(' ')
+    end
+
+    def exec_with_output(cmd)
+      success = system(cmd)
+      success ? puts('done'.green) : puts('failed'.red)
+    end
```
jsugarman: not used - remove

lib/tasks/rake_helpers/rake_utils.rb
```diff
@@ -0,0 +1,32 @@
...
+  def decompress_file(filename)
+    shell_working "decompressing file #{filename}" do
+      system "gunzip -3 -f #{filename}"
+    end
+  end
+
+  def production_protected
+    raise 'This operation was aborted because the result might destroy production data' if Rails.host.production?
+  end
+end
```
jsugarman: add blank line

spec/rails_helper.rb
```diff
@@ -28,6 +28,7 @@
   # exclude patterns from test coverage results
   add_filter %r{^/lib\/rack/} # only used to prevent feature test flickering
   add_filter %r{^/lib\/demo_data/} # only used for generation of demo data
+  add_filter %r{^/lib\/tasks/} # only used for generation of demo data
```
jsugarman: amend or remove comment

kubernetes_deploy/scripts/job.sh
```diff
@@ -3,8 +3,8 @@ function _job() {
   usage="job -- run job in the specified environment
```
kmahern, line 3: context='live-1' (line 45) fails for me (not something introduced by this change though) - should be context='live-1.cloud-platform.service.justice.gov.uk' to make it generic for all devs?

kmahern (commented):
Looks great, but current (documented) solution only really works for people with access to the CCCD secrets. Do we either need to commit to setting up all devs investigating incidents in production with access to secrets (with time and risk this involves) or workaround solution?

kubernetes_deploy/scripts/job.sh
jsugarman: @kmahern I have modified to use the users `--current` context and just changes `--namespace` - can you give the job a go again and let me know if that fixes the issues

jsugarman:
> Looks great, but current (documented) solution only really works for people with access to the CCCD secrets. Do we either need to commit to setting up all devs investigating incidents in production with access to secrets (with time and risk this involves) or workaround solution?

So, for your purposes we need to add you to git-crypt  - can you send me your key. For devs without secrets access a dev with secrets access could restore the database for them on an environment, which we should do anyway once this is in. This needs testing but i believe it should be possible to go onto the environment and run the `db:dump:copy_s3_dump['s3_key_for_file']` (but there may be a need to provide credentials for the bucket actually) and then `db:restore['downloaded-dump-file']`.

This could be a separate PR and bit of work which could also do with separate documentation

kubernetes_deploy/scripts/job.sh
kmahern: 👍 yes this worked.

kmahern (approved):
great 👍

PR #3585, opened 2020-10-15
[CHORE] node, npm and yarn

naseberry:
#### What
- Set node version to use the LTS.
- Move 'webpack' from package dependency to resolution
- Clean yarn.lock


jsugarman (approved):
👍 LGTM. will we need to update node locally then?

naseberry:
> 👍 LGTM. will we need to update node locally then?

Yes, run `$ nvm use`

PR #3586, opened 2020-10-16
Enable use of k8s current context in "local" scripts

jsugarman:
#### What
Fix a reuseability issue in the k8s scripts

#### Why
k8s scripts had context hardcoded to 'live-1'
which is not the default context name. Some devs
used 'live-1' as the name. Devs can set any name
they like in the `~/kube/config` to the
https://api.live-1.cloud-platform.service.justice.gov.uk
server so it should not be hardcoded.

#### Also

 - tidy up unneeded/simpler secrets config


naseberry (approved):
LGTM

PR #3590, opened 2020-10-21
Enable acceptance of URN in case number field

jsugarman:
#### What
Enable acceptance of URN in case number field

#### Ticket

[CBO-1266](https://dsdmoj.atlassian.net/browse/CBO-1266)

#### Why

With the release of common platform cases will
start to be identifiable by URN or case number
so this field has been repurposed to accept them

The flag switches this on. CCR and CCLF have
been amended to handle URNs for DI now so
we can go ahead and accept them in CCCD and
they will be injected into CCR/CCLF as expected.


kmahern (approved):
👍

PR #3592, opened 2020-10-22
Use plain ruby script for worker healthcheck

jsugarman:
#### What
Use plain ruby script for worker healthcheck

Also, tweak `readinesProbe` period to reduce
interval.

#### Ticket

[CBO-1580](https://dsdmoj.atlassian.net/browse/CBO-1580)

#### Why
Previous healthcheck was a rake task that
was loading the rails app every 30 seconds.
This was using a lot of cpu and causing
the node to need cycling more often.

### DIFF
--- a/bin/worker_healthcheck
+++ b/bin/worker_healthcheck
@@ -1,0 +1,9 @@
+#!/usr/bin/env ruby
+
+module Worker
+  def self.healthcheck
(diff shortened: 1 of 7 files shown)


jsugarman:
FYI: @poornima-krishnasamy this should reduce the worker cpu usage to single/double digit millicores

kmahern (approved):
👍

PR #3594, opened 2020-10-23
Rename and add MI date fields

jsugarman:
#### What
Rename and add MI date fields

#### Ticket

[CBO-637](https://dsdmoj.atlassian.net/browse/CBO-637)

#### Why
Case workers are confused by the name "Submitted at".

They need to see either the last_submitted_at date or
the original_submission_date, but we are not sure what is
most applicable for their purposes.

So, we are adding the two new dates and they can then decide
what they want.

Note: "Submitted at" column in MI is the date the claim transitioned
to a state of submitted, awaiting_written_reasons or redetermination

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -58,12 +58,6 @@ Metrics/BlockLength:
     - state_machine
     - resource
 
-Metrics/ModuleLength:
-  Exclude:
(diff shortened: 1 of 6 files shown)


colinbruce (approved):
LGTM :+1:

kmahern (approved):

PR #3595, opened 2020-10-26
Disable messaging for lgfs interim claims post assessment

kmahern:
#### What
Disable external user messaging for lgfs interim claims post assessment

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1314

#### Why
Providers were adding messages to lgfs interim claims after they had been assessed. Because lgfs interim claims cannot be redetermined these messages were not being seen by caseworkers.

#### How
Fixed/amended messaging_permitted? in claims_helper.rb so that messaging is disabled for lgfs interim post assessment claims (previously this was relying on redeterminable? in base_claim.rb which did not work for lgfs interim claims, as these are not redeterminable). Added tests in claims_helper_spec.

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -39,7 +39,11 @@ module ClaimsHelper
   end
 
   def messaging_permitted?(message)
-    (current_user_is_external_user? && !message.claim.redeterminable?) || message.claim_action.present?
+    return true if message.claim_action.present?
+    if current_user_is_external_user?
+      return !Claims::StateMachine::VALID_STATES_FOR_REDETERMINATION.include?(message.claim.state)
+    end
+    false
   end
 
   def display_downtime_warning?
--- a/spec/helpers/claims_helper_spec.rb
+++ b/spec/helpers/claims_helper_spec.rb
@@ -212,4 +212,72 @@ RSpec.describe ClaimsHelper do
       it { is_expected. to be_falsey }
     end
   end
+
+  describe '#messaging_permitted' do
+    subject { messaging_permitted?(message) }
+
+    let(:claim) { build :claim, state: state }
+    let(:message) { build(:message, claim: claim, claim_action: claim_action) }
+    let(:claim_action) { nil }
+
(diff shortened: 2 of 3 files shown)


spec/helpers/claims_helper_spec.rb
```diff
@@ -212,4 +212,72 @@ def current_user
       it { is_expected. to be_falsey }
     end
   end
+
+  describe '#messaging_permitted' do
```
jsugarman:
```suggestion
  describe '#messaging_permitted?' do
```

app/helpers/claims_helper.rb
```diff
@@ -39,7 +39,11 @@ def show_message_controls?(claim)
   end
 
   def messaging_permitted?(message)
-    (current_user_is_external_user? && !message.claim.redeterminable?) || message.claim_action.present?
+    return true if message.claim_action.present?
+    if current_user_is_external_user?
```
jsugarman, line 43: hmm, if the [redeterminable? method](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/641d716be93a5b6c009b6764c11a712a2f9e24b6/app/models/claim/base_claim.rb#L379-L381) not working here then it should probably be simplified

spec/helpers/claims_helper_spec.rb
```diff
@@ -212,4 +212,72 @@ def current_user
...
+      end
+
+      context 'for redeterminable claim states' do
+        let(:claim) { build :claim, state: state }
+
+        %w[authorised part_authorised rejected refused].each do |state|
+          context "when claim state is #{state}" do
+            let(:state) { state }
+
+            it { is_expected. to be_falsey }
```
jsugarman, line 277: does this not mean that messaging will be disabled for any claim in one of these states?

jsugarman: The BA tested this so and said the functionality is as expected for non iterims too, so am going to merge

jsugarman (approved):
👍

PR #3596, opened 2020-10-27
Remove link from the legend

naseberry:
#### What
Remove the 'Remove' link from the `<legend>` so the hypertext does not interfere with the grouping of the inputs

#### Ticket
[Link within Legend](https://dsdmoj.atlassian.net/browse/CBO-1503)

#### Why
When navigating through the forms out of context, form elements that contain the word 'Remove' is confusing as users are entering information.

### DIFF
--- a/app/views/external_users/claims/_date_attended_fields.html.haml
+++ b/app/views/external_users/claims/_date_attended_fields.html.haml
@@ -8,7 +8,8 @@
     - error_prefix = "date_attended_0"
 
   - unless hide_remove
-    = link_to_remove_association t(".remove"), f, {class: "date-action"}
+    = link_to_remove_association f, class: 'govuk-link' do
+      = t('common.remove_html', context: t('.date_attended'))
 
   .fee-date-from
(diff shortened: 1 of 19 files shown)


jsugarman (commented):
Hard to see what has actually changed because it is just one commit. I assume most changes are just indentations in the views.

jsugarman (commented):
has anyone, other than yourself, manually reviewed this on an environment?

naseberry:
> Hard to see what has actually changed because it is just one commit. I assume most changes are just indentations in the views.

The cocoon remove link association was within the legend tag, when navigating through the forms out of context, form elements that contain the word 'Remove' is confusing as users are entering information.

This commit, moves the 'remove' link outside the `<legeng>` tag.

There were long standing markup errors where inputs were not properly grouped within the relevant `<fieldset>`, therefore the indentation changes.

remove-link

jsugarman (approved):
👍 reviewed and regression tested by designer so good to go.

PR #3600, opened 2020-11-09
Remove  non applicable litigator fields from summary screen

kmahern:
#### What
Remove dates, quantity and rates from summary/check your claim page for litigator claim misc fees.

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1475.

#### Why
These fields are not used.

#### How
Disabled the display of these fields in the relevant view(s).

### DIFF
--- a/app/views/external_users/claims/_summary_fee.html.haml
+++ b/app/views/external_users/claims/_summary_fee.html.haml
@@ -9,33 +9,34 @@
         %span
           = fee.case_numbers
 
-  %td
-    - if fee.date?
-      = fee.date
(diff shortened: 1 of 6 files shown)


jsugarman (changes requested):
i believe these partials are shared across both fee schemes, all claims types and all fee types so can we add/extend some feature tests or perhaps write view specs to try and pin this down and identify current of future regressions.

kmahern:
👍 @jsugarman I have extended the features to cover testing for the presence or absence of these fields

jsugarman (approved):
👍  LGTM 

I would get the BA to vet this if she has time

PR #3601, opened 2020-11-09
Non descriptive heading

naseberry:
#### What
Make headings descriptive by moving nested hyperlinks outside of the heading tags.

Secondary
- updated the heading classes
- replaced redundant class name `.form-section` with `.app-summary-section` in summary view and related partials

#### Ticket
[Non-descriptive headings](https://dsdmoj.atlassian.net/browse/CBO-1511)

#### Why
The “Change” links have been included as part of the heading structure. This is confusing for screen reader users as the heading is ambiguous with the hypertext included.

### DIFF
--- a/app/views/external_users/claims/_summary_fees.html.haml
+++ b/app/views/external_users/claims/_summary_fees.html.haml
@@ -1,7 +1,8 @@
-%h2.heading-large
+%h2.govuk-heading-l
   = local_assigns.has_key?(:header) ? header : t('common.fees')
-  - if local_assigns[:editable]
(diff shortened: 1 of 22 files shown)


jsugarman (approved):
👍  LGTM

**Not a blocker**: Given the repetition of `t('common.change_html', context: t('external_users.claims.<path-to-specific-partials-summary-header>')` I was wondering if it would have been possible to rely on a relative path to reduce repetition OR some other way to shorten the translation call.

naseberry:
> 👍 LGTM
> 
> **Not a blocker**: Given the repetition of `t('common.change_html', context: t('external_users.claims.<path-to-specific-partials-summary-header>')` I was wondering if it would have been possible to rely on a relative path to reduce repetition OR some other way to shorten the translation call.

👍 I agree, to be actioned at another time.

PR #3606, opened 2020-11-16
Some minor tweaks to the db dump mechanism

jsugarman:
#### What
Some minor tweaks to the db dump mechanism

#### Why
  - increase timeout waiting for job pod to start as was
    seeing it take upto 3 and half minutes.

  - improve formatting for list of s3 dump files for
    clarity

  - add arg to enable delete of All s3 dump files as
    we should not keep them as it has cost and security
    implications.


kmahern (approved):
👍

PR #3608, opened 2020-11-17
Remove URN_ENABLED flag and all conditional logic

jsugarman:
#### What
remove all reliance on the urn_enabled? flag

#### Why
System wil always accept URNs going forward
so the conditional logic is dead code.


kmahern (approved):
lgtm 👍

PR #3613, opened 2020-11-19
Replace content_tag with tag

jsugarman:
#### What
Replace content_tag with tag

#### Why
content_tag is legacy syntax

see https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag
for details.


naseberry (approved):
Code changes seem solid.

PR #3626, opened 2020-12-04
Standardise formatting options across supported editors

jsugarman:
#### What
Apply some standard formattting across supported editors

VS Code is supported and sublimetext has a plugin

see https://editorconfig.org/ for list of
editors with built in support or others
that require a plugin.

#### Why
To apply consistent standards across diffeerent
editors used by developers.

These default settings are the recommended
by the sublimetext plugin and allow for indent
being specified by tab width. editors should
set their tab width to 2 spaces.

https://packagecontrol.io/packages/EditorConfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# see https://editorconfig.org/
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,7 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
-indent_style = tab
+indent_size = 2
+indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true


.editorconfig
```diff
@@ -0,0 +1,9 @@
+# see https://editorconfig.org/
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
```
jrmhaig: Perhaps I'm misunderstanding. Doesn't this enforce tabs for indents (instead of spaces)? As far as I can tell we always using two spaces, it this is being enforced by Rubocop.

jsugarman: Thanks Joe, these are just the [defaults recommended by the sublime text plugin](https://github.com/sindresorhus/editorconfig-sublime#example-file) (that I use). We can change them, which is why i wanted everyone to review but I am not sure what effect it will have on ST.

The ST plugin does not support `tab_width` "intentionally" either (does not say why). 

In ST you typically define your tab width (and can change it easily, but is two spaces generally and should probably stay that way for `.rb` files at least) and then indenting is done using the tab key.

How do you guys indent?

jsugarman: @naseberry you could add a commit for `.js` and `.css` files specifically if you wanted. These are currently the fallbacks.

jsugarman: on tab issues it is worth looking at the [definition of indent_style](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_style) and indent size

```
If indent_size equals to tab, the indent_size will be set to the tab size, which should be tab_width if tab_width is specified, or the tab size set by editor if tab_width is not specified.
```

^^ but perhaps this is all rather too confusing and we could specify size 2 and style space for clarity

naseberry: @jsugarman suggest we change `indent_style` to space and define the `indent_size` as 2

This would be the preferred formatting for JS, SCSS and most other file type.

My experience in the past with an indent style of tab is that some IDE/Editors have a default size of 4, thereby causing formatting issues between engineers and their preferred IDE/Editor.

jsugarman: 👍 we can do that and see how it plays.

jsugarman: thats been applied now

naseberry (approved):
👍

PR #3627, opened 2020-12-04
Remove the Timecop gem

jrmhaig:
#### What

Replace Timecop with time helpers.

#### Ticket

N/A

#### Why

The time helpers that are built into Rails provide the same functionality as the Timecop gem so the gem could be removed.

#### How

I have updated many references to Timecop in the tests to use the time helpers instead. In some cases I have also cleaned up the tests.

--------

#### TODO (wip)

 - [ ] Timecop is still used in three factories; `spec/factories/claim/advocate_claims.rb`, `spec/factories/claim/base_claims.rb' and`spec/factories/claim/deterministic_claim.rb`. I am still trying to work out the correct way to use time helpers inside FactoryBot.

### DIFF
--- a/spec/api/v1/external_users/representation_order_spec.rb
+++ b/spec/api/v1/external_users/representation_order_spec.rb
@@ -35,15 +35,17 @@ RSpec.describe API::V1::ExternalUsers::RepresentationOrder do
   end
 
   describe "POST #{endpoint(:representation_orders)}" do
-    def post_to_create_endpoint(submission_date = Date.new(2017, 7, 1))
-      Timecop.freeze(submission_date) { post endpoint(:representation_orders), valid_params, format: :json }
+    def post_to_create_endpoint
+      post endpoint(:representation_orders), valid_params, format: :json
     end
+
+    subject { post endpoint(:representation_orders), valid_params, format: :json }
 
     include_examples "should NOT be able to amend a non-draft claim"
 
     context 'when representation_order params are valid' do
       it "should create fee, return 201 and expense JSON output including UUID" do
-        post_to_create_endpoint
+        subject
         expect(last_response.status).to eq 201
         json = JSON.parse(last_response.body)
         expect(json['id']).not_to be_nil
@@ -114,7 +116,7 @@ RSpec.describe API::V1::ExternalUsers::RepresentationOrder do
       describe 'and the rep_order_date pre-dates the start of the scheme' do
         let(:representation_order_date) { Date.new(2018, 3, 1) }
 
-        before { post_to_create_endpoint(Date.new(2018, 5, 1)) }
+        before { travel_to(Date.new(2018, 5, 1)) { post_to_create_endpoint } }
 
         specify { expect_error_response('Check the combination of the representation order date and offence') }
       end
@@ -122,7 +124,9 @@ RSpec.describe API::V1::ExternalUsers::RepresentationOrder do
       describe 'and the rep_order_date post-dates the start of the scheme' do
         let(:representation_order_date) { Date.new(2018, 4, 1) }
 
-        specify { expect{ post_to_create_endpoint(Date.new(2018, 5, 1)) }.to change { RepresentationOrder.count }.by(1) }
+        before { travel_to(Date.new(2018, 5, 1)) }
+
+        specify { expect{ post_to_create_endpoint }.to change { RepresentationOrder.count }.by(1) }
       end
     end
   end
--- a/spec/api/v1/external_users/representation_order_spec.rb
+++ b/spec/api/v1/external_users/representation_order_spec.rb
@@ -39,13 +39,11 @@ RSpec.describe API::V1::ExternalUsers::RepresentationOrder do
       post endpoint(:representation_orders), valid_params, format: :json
     end
 
-    subject { post endpoint(:representation_orders), valid_params, format: :json }
-
     include_examples "should NOT be able to amend a non-draft claim"
 
     context 'when representation_order params are valid' do
       it "should create fee, return 201 and expense JSON output including UUID" do
-        subject
+        post_to_create_endpoint
         expect(last_response.status).to eq 201
         json = JSON.parse(last_response.body)
         expect(json['id']).not_to be_nil
--- a/spec/controllers/external_users/claims_controller_spec.rb
+++ b/spec/controllers/external_users/claims_controller_spec.rb
@@ -829,7 +829,7 @@ end
 
 def build_sortable_claims_sample(advocate)
   [:draft, :submitted, :allocated, :authorised, :rejected].each_with_index do |state, i|
-    travel(-i.days) do
+    travel_to(i.days.ago) do
       n = i+1
       claim = create("#{state}_claim".to_sym, external_user: advocate, case_number: "A2016#{(n).to_s.rjust(4,'0')}")
       claim.fees.destroy_all
--- a/spec/controllers/external_users/claims_controller_spec.rb
+++ b/spec/controllers/external_users/claims_controller_spec.rb
@@ -829,7 +829,7 @@ end
 
 def build_sortable_claims_sample(advocate)
   [:draft, :submitted, :allocated, :authorised, :rejected].each_with_index do |state, i|
-    Timecop.freeze(i.days.ago) do
+    travel(-i.days) do
       n = i+1
       claim = create("#{state}_claim".to_sym, external_user: advocate, case_number: "A2016#{(n).to_s.rjust(4,'0')}")
       claim.fees.destroy_all
--- a/spec/controllers/provider_management/external_users_controller_spec.rb
+++ b/spec/controllers/provider_management/external_users_controller_spec.rb
@@ -205,17 +205,19 @@ RSpec.describe ProviderManagement::ExternalUsersController, type: :controller do
   describe "PUT #update_password" do
     let(:password) { 'password123' }
     let(:password_confirm) { password }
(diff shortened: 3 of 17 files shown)


spec/api/v1/external_users/representation_order_spec.rb
```diff
@@ -35,15 +35,17 @@
...
     end
 
+    subject { post endpoint(:representation_orders), valid_params, format: :json }
+
     include_examples "should NOT be able to amend a non-draft claim"
 
     context 'when representation_order params are valid' do
       it "should create fee, return 201 and expense JSON output including UUID" do
-        post_to_create_endpoint
+        subject
```
jsugarman: would prefer named `subject` at line 42 - for clarity and to avoid potential clashes. is also a rubocop preference (we are excluding specs from rubocop entirely at the moment i think due to the number of violations i think)

spec/controllers/external_users/claims_controller_spec.rb
```diff
@@ -829,7 +829,7 @@ def build_claim_in_state(state)
 
 def build_sortable_claims_sample(advocate)
   [:draft, :submitted, :allocated, :authorised, :rejected].each_with_index do |state, i|
-    Timecop.freeze(i.days.ago) do
+    travel(-i.days) do
```
jsugarman: was there a reason for not using `travel_to(i.days.ago)` here?

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -206,38 +203,34 @@ def post_to_create_external_user_action(options={})
...
+    let(:password_confirm) { password }
+    subject { put :update_password, params: { provider_id: provider, id: external_user, external_user: { user_attributes: { password: password, password_confirmation: password_confirm } } } }
 
-    context 'when valid' do
-
-      before(:each) do
-        put :update_password, params: { provider_id: provider, id: external_user, external_user: { user_attributes: { password: 'password123', password_confirmation: 'password123' } } }
-        external_user.reload
-      end
+    before(:each) { travel(-6.months) { external_user } }
```
jsugarman: It probably does not matter but to emulate frozen time that was used for external user creation prior to this you use `travel_to` (`travel` does not freeze time [from what i am reading](https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel))

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -206,38 +203,34 @@ def post_to_create_external_user_action(options={})
...
 
+    context 'when valid' do
       it 'does not require current password to be successful in updating the user record ' do
-        expect(external_user.user.updated_at).to_not eql frozen_time
+        expect { subject }.to change { external_user.reload.user.updated_at }
```
jsugarman: naming the subject above would make this clearer

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -206,38 +203,34 @@ def post_to_create_external_user_action(options={})
...
       end
 
       it 'redirects to external_user show action' do
+        subject
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3627/files#r538198950)

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -206,38 +203,34 @@ def post_to_create_external_user_action(options={})
...
     context 'when invalid' do
-
-      before(:each) do
-        put :update_password, params: { provider_id: provider, id: external_user, external_user: { user_attributes: { password: 'password123', password_confirmation: 'passwordxxx' } } }
-      end
+      let(:password_confirm) { 'passwordxxx' }
 
       it 'does not update the user record' do
-        expect(external_user.user.updated_at).to eql frozen_time
+        expect { subject }.not_to change { external_user.reload.user.updated_at }
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3627/files#r538198950)

spec/controllers/provider_management/external_users_controller_spec.rb
```diff
@@ -206,38 +203,34 @@ def post_to_create_external_user_action(options={})
...
       end
 
       it 'renders the change password template' do
+        subject
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3627/files#r538198950)

spec/models/claims/financial_summary_spec.rb
```diff
@@ -10,22 +10,20 @@
...
     let!(:old_part_authorised_claim) do
-      Timecop.freeze(Time.now - 2.week) do
-        claim = create(:part_authorised_claim)
-        Timecop.freeze(Time.now + 1.week) do
-          claim.determinations.first.update(fees: claim.fees_total/2, expenses: claim.expenses_total)
-          claim
-        end
+      travel_to(Time.now - 2.week)
+      create(:part_authorised_claim).tap do |claim|
+        travel_to(Time.now + 1.week)
```
jsugarman, line 15: can/should we be using a block here

spec/models/claims/financial_summary_spec.rb
```diff
@@ -10,22 +10,20 @@
...
     let!(:part_authorised_claim) do
-      Timecop.freeze(Time.now - 2.week) do
-        claim = create(:part_authorised_claim)
-        Timecop.freeze(Time.now + 2.week) do
-          claim.determinations.first.update(fees: claim.fees_total/2, expenses: claim.expenses_total)
-          claim
-        end
+      travel_to(Time.now - 2.week)
+      create(:part_authorised_claim).tap do |claim|
+        travel_to(Time.now + 2.week)
```
jsugarman, line 24: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3627/files#r538212486)

spec/models/claims/state_machine_spec.rb
```diff
@@ -297,24 +297,24 @@
...
 
       it 'set the last_submitted_at to the current time for awaiting_written_reasons' do
-        current_time = Time.now
+        current_time = Time.now.change(usec: 0)
         claim.submit!
         claim.allocate!
         claim.refuse!
 
-        Timecop.freeze 6.months.from_now do
+        travel_to current_time + 6.months do
```
jsugarman: **not a blocker** : braces preferred

```suggestion
        travel_to(current_time + 6.months) do
```

spec/models/claims/state_machine_spec.rb
```diff
@@ -297,24 +297,24 @@
...
     context 'update last_submitted_at on redetermination or await_written_reasons' do
       it 'set the last_submitted_at to the current time for redetermination' do
-        current_time = Time.now
+        current_time = Time.now.change(usec: 0)
         claim.submit!
         claim.allocate!
         claim.refuse!
 
-        Timecop.freeze 6.months.from_now do
+        travel_to current_time + 6.months do
```
jsugarman, line 305: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3627/files#r538245188)

spec/services/claim_reporter_spec.rb
```diff
@@ -25,7 +25,7 @@
 
   describe '#completion_rate' do
     before do
-      Timecop.freeze(5.weeks.ago) do
+      travel -5.weeks do
```
jsugarman, line 28: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3627/files#r538196976)

spec/api/v1/external_users/representation_order_spec.rb
jrmhaig: Sorry, I put `subject` in at one point and then thought I had taken it out again later, when I understood better what was going on. I'll change it back.

jsugarman (commented):
👍 looks good, no blockers but a few minor suggestions, style or otherwise

The use of named `subjects` is something i would advocate, for clarity

spec/models/claims/financial_summary_spec.rb
jrmhaig: I think that nesting `travel_to` can cause problems as the end of the block returns to the current time and not the previously frozen time. It is, in fact, calling `travel_back` after the `yield` but I thought that it would look clearer as:

```ruby
travel_to x
...
travel_to y
...
travel_back
```

although that was before I added `tap`.

I'm happy to add a block to the second `travel_to` (but nesting them, as they were before, is discouraged as it has complications).

spec/controllers/external_users/claims_controller_spec.rb
jrmhaig: I had an idea that it might be a slight optimisation, as it doesn't create an instance of `Time` but it turns out that `travel` just calls `travel_to Time.now + duration` anyway. That being the case, `travel_to` makes more sense (and in all the other places too).

spec/controllers/provider_management/external_users_controller_spec.rb
jrmhaig: The test was checking that the updated time is not exactly the same as the frozen time. There was a problem with this as the milliseconds part of the time is lost by `travel_to`.

I have changed it just to see if it has changed (or not changed as in the test further down).

spec/models/claims/financial_summary_spec.rb
jsugarman: 👍  makes sense, i'll leave it to you

PR #3629, opened 2020-12-08
Remove the Timecop gem

jrmhaig:
#### What

Replace Timecop with time helpers. (Reopening PR after changing branch name)

#### Ticket

N/A

#### Why

The time helpers that are built into Rails provide the same functionality as the Timecop gem so the gem could be removed.

#### How

I have updated many references to Timecop in the tests to use the time helpers instead. In some cases I have also cleaned up the tests.

--------

#### TODO (wip)

 - [ ] Timecop is still used in three factories; `spec/factories/claim/advocate_claims.rb`, `spec/factories/claim/base_claims.rb' and`spec/factories/claim/deterministic_claim.rb`. I am still trying to work out the correct way to use time helpers inside FactoryBot.


jsugarman (approved):
🏅 great stuff

PR #3635, opened 2020-12-09
Create report of archived claims

jrmhaig:
#### What

There are currently more than 82K claims in the CCCD archive. We know from zendesk reports and from looking into the archive in the past that there are issues with the archive and that it is not working optimally.  Unlike current/recent claims which can be extracted using the MI report there is no facility which enables us to extract information on the claims in the archive (other than a dev running an SQL query to obtain such data).

#### Ticket

[CBO-1611](https://dsdmoj.atlassian.net/browse/CBO-1611)

#### Why

In the short term I would like a one of extract of the archive data so that I can run some analysis on it, both for the purposes of investigating how widespread the issues with archiving are and also for more long term strategic considerations of or archiving timescales. This also ties into considerations of our MI generally, as well as data retention implications.

#### How

Create a rake task that can be executed as:

```bash
bundle exec rails 'claims:archive_report[test.csv]'
```

to create a CSV file containing the required data.

--------

#### TODO (wip)

 - [ ] Confirm the list of fields required.


lib/tasks/rake_helpers/archived_claims.rb
```diff
@@ -0,0 +1,48 @@
...
+    def self.query
+      Claim::BaseClaim.caseworker_dashboard_archived
+        .includes(
+          :case_type, :archived_claim_state_transitions,
+          defendants: :representation_orders,
+          external_user: :user
+        )
+    end
+  end
+end
```
jsugarman: newline

lib/tasks/rake_helpers/archived_claims.rb
```diff
@@ -0,0 +1,48 @@
...
+          'Main dependant',
+          'MAAT',
+          'Total (inc VAT)',
+          'Status',
+          'Case type',
+          'Submitted',
+          'Archived date'
+        ]
+
+        rows = query
```
jsugarman, line 18: not sure how many claims this would return in a production like environment but if this is a lot we should probably use a `find_each` to retrieve and write in batches - see [find_each](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) for details.

spec/models/claims/state_machine_scopes_spec.rb
```diff
@@ -40,4 +40,36 @@
       end
     end
   end
+
+  describe 'scoped associations' do
+    let(:claim) { create :advocate_claim }
+
+    describe '.archived_claim_state_transitions' do
+      it 'only returns transitions to archived state' do
```
jsugarman: as for the other PR would be good to have make use of contexts and have test setup separate from the actual test.

jsugarman (commented):
This looks good so far.

So i have tried running this against a local anonymised version of the database and get this.

```
Claim-for-Crown-Court-Defence (jh/archive_report) $ rake claims:archive_report['tmp/archived_claims_report.csv']
**************************************************
⛔️ WARNING: Sidekiq testing API enabled, but this is not the test environment.  Your jobs will not go to Redis.
**************************************************
Writing archived claims to tmp/archived_claims_report.csv
 - 79279 claims
rake aborted!
NoMethodError: undefined method `name' for nil:NilClass
.../Claim-for-Crown-Court-Defence/lib/tasks/rake_helpers/archived_claims.rb:30:in `block (2 levels) in write'
....
```

**couple of points:**
  - we need to get you an anonymised db dump- can pair on this
  - at a guess this would be `claim.case_type.name` needing a safe navigator as not all 
     claims have a case type - but there may be others

jsugarman (approved):
LGTM 👍

PR #3637, opened 2020-12-11
Delegate each to stages array

jsugarman:
#### What
Delegate each to stages array

#### Ticket

n/a

#### Why

This came out of a rubocop bump that
highlighted superfluos code in the
explict each method definition.

see this [PR comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3625#discussion_r534985240)

For brevity and simplicity we can delegate
each instead of defining our own explictly.

### DIFF
--- a/app/models/stage_collection.rb
+++ b/app/models/stage_collection.rb
@@ -4,17 +4,11 @@ class StageCollection
 
   attr_reader :stages
 
-  def_delegators :@stages, :second, :last, :size
+  def_delegators :@stages, :each, :second, :last, :size
 
   def initialize(stages, object)
     @object = object
     @stages = initialize_stages(stages)
-  end
-
-  def each(&block)
-    stages.each do |stage|
-      yield(stage) if block
-    end
   end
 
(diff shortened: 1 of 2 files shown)


jrmhaig:
👍 

I think that the tests for `#first`, `#second`, `#last` and `#size` can be deleted too. `#first` in particular comes automatically from including `Enumerable` and so the test was always unnecessary.

jrmhaig (approved):

jsugarman:
> 👍
> 
> I think that the tests for `#first`, `#second`, `#last` and `#size` can be deleted too. `#first` in particular comes automatically from including `Enumerable` and so the test was always unnecessary.

👍  yes, i did look at this. The  `#second`, `#last` and `#size` methods come from Array and require specific delegation so would keep the tests for them. `#first` implicitly falls out of including enumerable/defining-each but is relied on so will keep in the vien of testing the interface rather than the implemention, but will move the context to highlight its relation to enumerable.

PR #3641, opened 2020-12-11
Remove unused methods

jrmhaig:
#### What

Remove some unused methods from the `Claims::UserMessages` module.

#### Ticket

N/A

#### Why

While fixing #3632 I saw that some of the methods in the `Claims::UserMessages` module are not used anywhere, possibly since they were created in #429. Additionally, it looks as though the tests are broken.

#### How

* Delete any methods in `Claims::UserMessages` that are not used anywhere other than the tests.
* Move the one remaining method into `Claim::BaseClaim` and delete the `Claims::UserMessages` module.

I actually think that this the responsibility for finding unread messages belongs to the user rather than the claim, so instead of

```ruby
claim.unread_messages_for(user)
```

I would prefer

```ruby
user.unread_messages_for(claim)
```

However, this is a matter of opinion so I am not implementing it as part of this work.


spec/models/claim/base_claim_spec.rb
```diff
@@ -523,6 +523,39 @@ class MockSteppableClaim < Claim::BaseClaim
       specify { is_expected.to be_nil }
     end
   end
+
+  describe '#unread_messages_for' do
```
jsugarman, line 527: my personal preference is for tests to start with a subject and then use that subject

```
subject(:call) { claim.unread_messages_for(user) }
```

and then call
```
is_expected.to ...
```

spec/models/claim/base_claim_spec.rb
```diff
@@ -523,6 +523,39 @@ class MockSteppableClaim < Claim::BaseClaim
...
+    let(:claim) { create(:submitted_claim) }
+    let(:user) { claim.external_user.user }
+
+    it 'returns an empty array if there are no messages' do
```
jsugarman: I also think, mentioned in [betterspec.org](https://www.betterspecs.org/) we should make use of contexts and have setup for tests in a before block.

example
```
context 'when there are no messages for user' do
  it { is_expected.to eq([]) }
  # alternative composite spec
  it { is_expected.to be_an(Array).and be_empty }
end
```

jsugarman (commented):
Looks good, 

have added some spec suggestions

spec/models/claim/base_claim_spec.rb
```diff
@@ -528,32 +528,42 @@ class MockSteppableClaim < Claim::BaseClaim
     let(:claim) { create(:submitted_claim) }
     let(:user) { claim.external_user.user }
 
-    it 'returns an empty array if there are no messages' do
-      expect(claim.unread_messages_for(user)).to eq([])
+    subject(:call) { claim.unread_messages_for(user) }
```
jsugarman: **nit** subject should be first line of a block

spec/models/claim/base_claim_spec.rb
```diff
@@ -528,32 +528,42 @@ class MockSteppableClaim < Claim::BaseClaim
...
-      message4 = create(:message, claim: claim)
-      message2.user_message_statuses.where(user: user).update(read: true)
-      message3.user_message_statuses.where(user: user).update(read: true)
+    context 'with multiple messages' do
+      let!(:message1) { create(:message, claim: claim) }
+      let!(:message2) { create(:message, claim: claim) }
+      let!(:message3) { create(:message, claim: claim) }
+      let!(:message4) { create(:message, claim: claim) }
+
+      before :each do
```
jsugarman: **nit** `:each ` is the default

jsugarman (approved):
👍 looks great. 

couple of nits but no blockers.

PR #3648, opened 2020-12-16
Remove vat_registered? logging and use a fallback

jsugarman:
#### What
Remove vat_registered? logging and use a fallback of true

#### Ticket

n/a

#### Why
A very occasional sentry error exists whereby
the provider_delegator is nil. Rather than stop
user processing in such situations with a 500
we will swallow the error and default to true as
most providers are vat registered (**need to check this).

Fix swallow sentry error and fallback to default of true
for vat_registered? for a provider.

### DIFF
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -502,16 +502,9 @@ module Claim
     end
 
     def vat_registered?
-      if provider_delegator.nil?
-        LogStuff.send(:error, 'vat_registration',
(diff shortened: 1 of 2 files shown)


jsugarman:
small change but would like support to be available when this gets merged as historically there have been issues around this area and need to keep a eye on it zia zendesk reports once in

kmahern:
Closing as we have not encountered recent Sentry errors relating to this functionality

PR #3649, opened 2020-12-17
[CBO-1611] Ensure archived_claim_state_transitions is ordered by created_at

jrmhaig:
#### What

Ensure that the 'archive at' timestamp in the archived claim report is for the first archived state change.

#### Ticket

[CBO-1611](https://dsdmoj.atlassian.net/browse/CBO-1611)

#### Why

I had presumed that the list of create state transitions would be returned in order but this appears not to be the case when a scope is added.

#### How

Add an ordering to the `archived_claim_state_transition` scoped association.


app/models/claims/state_machine.rb
```diff
@@ -145,7 +145,7 @@ def self.included(klass)
       end
 
       klass.has_many :archived_claim_state_transitions, lambda {
-        where(to: CASEWORKER_DASHBOARD_ARCHIVED_STATES)
+        where(to: CASEWORKER_DASHBOARD_ARCHIVED_STATES).order(created_at: :asc)
```
jsugarman, line 148: I don't think you will need this but there may be a `default_scope` that needs unscoping `unscope(:order).order(created_at: :asc)`. Could not see any obvious one for claim_state_transitions though

jsugarman (approved):
👍

app/models/claims/state_machine.rb
jrmhaig: There is a scope that sets the order on the `claim_state_transitions` association in `Claim::BaseClaim` so I presume there isn't a default scope on the `ClaimStateTransition` model. I cannot see one either.

PR #3650, opened 2020-12-18
[CBO-1631] Apply Rubocop to RSpec files

jrmhaig:
#### What

Bring the RSpec tests into line with our Ruby coding standards.

#### Ticket

[CCCD: RSpecs rubocop adherence](https://dsdmoj.atlassian.net/browse/CBO-1631)

#### Why

A consistent coding standard makes it easier to understand a large code base.

#### How

1. ~Enable Rubocop on the `spec/` directory.~
2. ~Explicitly disable each cop for the `spec/` directory.~
3. ~Fix each cop that was disabled in step 2 in turn so that it can be reenabled.~

~**Note:** Most cops are disabled in a 'to do' list at the bottom of `.rubocop.yml`. There are, however, some that are already configured, either in `.rubocop.yml` or `.rubocop_todo.yml`. In these cases it was necessary to edit the existing configuration as configurations cannot be duplicated. My suggestion for a plan of attack is therefore;~

1. ~Resolve any issues relating to RSpec files in `.rubocop_todo.yml`. It will then be possible to auto-generate this file again.~
2. ~Resolve any issues for cops that were previously configured in `.rubocop.yml`.~
3. ~Tackle the 'to do' list at the bottom of `.rubocop.yml`.~

Following comments from @jsugarman, and seeing the length of the 'to do' list I added to the end of `.rubocop.yml`, I've had second thoughts about this. It makes more sense to let `.rubocop_todo.yml` handle this even though it makes it very large. I still think that the best idea is to tackle one cop at a time and concentrate on fixing just the offences and warnings that occur in the `spec/` directory.

However, for some reason running `rubocop --auto-gen-config` now doesn't create a `.rubocop_todo.yml` to correctly hide all errors. For this reason I have temporarily disabled the following cops in `.rubocop.yml`:

* `Layout/LineLength`
* `RSpec/StubbedMock`
* `Lint/ConstantDefinitionInBlock`
* `Style/IfUnlessModifier`
* `Lint/RedundantCopDisableDirective`

and these should be the first to be fixed.

# Fixes applied

## [Lint/FloatComparison](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/6e2d4ee42184e4a04c76bf866504e2986f4b80d9)

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/FloatComparison

Testing equality of floating point numbers can be dangerous due to their inherent inaccuracy. This fixed in two different ways;

* Testing instead for zero or non-zero
* Comparing as `BigDecimal` instead

## [Naming/AccessorMethodName](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/0c21ed2f353e034b7e3e04422d5629715d8c279e)

Until I work out how to compare between two arbitrary commits in a PR, these commits are also part of this fix:
* https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/da6326b88b5210460f7f42725bfda6cb9ba3cf0b
* https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/f24b9d1979948db75d6f3be0ca65b0e20ce37ea6

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AccessorMethodName

Getters and setters for attributes in Ruby are, by convention, named `attribute` and `attribute=`. Having methods called `get_attribute` and `set_attribute` cause confusion (particularly if they are not actually attribute getters and setters). I have therefore renamed instances if these.

## [Layout/SpaceInsideBlockBraces](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/628cebc1c475e7551a0e3cec866190e4be9632d0)

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceInsideBlockBraces

For readability, there is always a space between block braces (`{ }`) and their contents.

## [Layout/SpaceBeforeBlockBraces](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/176128416ff73876a94698d23c8c7a594f9216f9)

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceBeforeBlockBraces

For readability, there is always a space before block braces.

## [Layout/SpaceAroundEqualsInParameterDefault](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/6f38ad1aeb1a79c7aa17260af85b604b8b18de7c)

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceAroundEqualsInParameterDefault

For readability, there are always spaces between the `=` in a method's parameter list.

## [Layout/SpaceInsideHashLiteralBraces](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/9fbe3cfe79f9a794454603ee92f2f4b689d70fd8)

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceInsideHashLiteralBraces

For readability, hashes have a space after the opening brace and before the closing brace.

## [Style/MethodDefParentheses](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3650/commits/5fd65dc5c1459bd4f8d1d3debce8e7ce4ca680d9)

See https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodDefParentheses


spec/factories/claim/base_claims.rb
```diff
@@ -187,7 +187,7 @@ def random_case_number
   [%w(A S T).sample, rand(1990..2016), rand(1000..9999)].join
 end
 
-def set_amount_assessed(claim)
+def assign_fees_and_expenses_to(claim)
```
jsugarman: `assign_fees_and_expenses_to(claim)` is better described as `assess_fees_and_expenses_for(claim)` to convey the business logic

spec/support/factory_helpers.rb
```diff
@@ -85,7 +85,7 @@ def add_fee(factory, claim)
     claim.fees << build(factory, claim: claim)
   end
 
-  def set_creator(claim)
+  def assign_external_user_as_creator(claim)
```
jsugarman, line 88: 👍

.circleci/config.yml
```diff
@@ -21,7 +21,8 @@ references:
     run:
       name: Install Code Climate test-reporter
       command: |
-        curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > tmp/cc-test-reporter
+        # curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > tmp/cc-test-reporter
+        curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-0.7.0-linux-amd64 > tmp/cc-test-reporter
```
kmahern: Need to remember to revert this out at some point

app/models/expense.rb
```diff
@@ -135,7 +135,7 @@ def expense_type_unique_code=(code)
   end
 
   def vat_absent?
-    vat_amount.nil? || vat_amount == 0.0
+    vat_amount.nil? || vat_amount.zero?
```
kmahern: This is a large PR. IMO it would make sense to move the fixes for Lint/FloatComparison rubocop TODOs to a separate PR to keep it a little smaller as these are not related to specs.

app/services/vat_auditor.rb
```diff
@@ -43,7 +43,7 @@ def audit_lgfs_vat_fee
   end
 
   def audit_no_vat_on_fees
-    return unless @claim.fees_vat != 0.0
+    return unless @claim.fees_vat.nonzero?
```
kmahern: ditto

app/validators/fee/base_fee_validator.rb
```diff
@@ -94,7 +94,7 @@ def validate_any_quantity
 
     def validate_integer_decimal
       return if @record.fee_type.nil? || @record.quantity.nil? || @record.quantity_is_decimal?
-      add_error(:quantity, 'integer') unless @record.quantity.frac == 0.0
+      add_error(:quantity, 'integer') unless @record.quantity.frac.zero?
```
kmahern: ditto

lib/extensions/string_extension.rb
```diff
@@ -1,6 +1,6 @@
 module StringExtension
   def zero?
-    present? && to_f == 0.0
+    present? && to_f.zero?
```
kmahern: ditto

app/models/expense.rb
jrmhaig: I'm happy to change this (and the other 3 cases) back. I realised after doing the `Lint/FloatComparison` bit that I should restrict it to just the spec files so these aren't any changes in the actual app in any of the other commits.

.rubocop.yml
```diff
@@ -31,9 +30,11 @@ Layout/EmptyLineAfterGuardClause:
   Enabled: false
 
 Layout/LineLength:
-  Max: 120
-  IgnoredPatterns: ['(\A|\s)#','.*\s+ # .*']
-  AllowHeredoc: true
+  # Temporarily disable while fixing sped/* directory
```
kmahern: I am a little concerned about disabling these checks (and the ones below), albeit hopefully only temporarily and for a short time, did you discuss with @jsugarman/did he have any suggestions? Do we know why we can't autogenerate the .rubocop_todo.yml?

kmahern:
@jrmhaig the changes you have made to the specs lgtm as far as I can tell, I have some concerns about the strategy of temporarily disabling certain cops, but if there isn't another workable solution this isn't a blocker to me.

.rubocop.yml
jrmhaig: The trouble is that if I do not do it like this then for some reason (possibly a bug in Rubocop) it is not possible to auto-generate all the correct exceptions in `.rubocop_todo.yml` to get it to run without errors and warnings. I couldn't get to the bottom of what the issue was but I was trying to see if I could clean up enough to get it working by fixing the whitespace errors.

It should be simple to fix all the long lines but there are so many of them (nearly 1000) that I thought it would be better in a separate PR. Also, I think many of them could do with some more significant re-writing rather than simply breaking over multiple lines.

kmahern: 👍

app/models/expense.rb
kmahern: 👍

.circleci/config.yml
kmahern: @jrmhaig I think the issue with test-reporter has been resolved and this can be removed

kmahern (approved):
👍 lgtm

PR #3652, opened 2020-12-21
[CBO-1631] Rubocop: Fix FactoryBot/FactoryClassName

jrmhaig:
#### What

Bring the RSpec tests into line with our Ruby coding standards. In this case, fix the RSpec factory files so that they satisfy the [Fix FactoryBot/FactoryClassName](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/FactoryClassName) in Rubocop.

#### Ticket

[CCCD: RSpecs rubocop adherence](https://dsdmoj.atlassian.net/browse/CBO-1631)

#### Why

A consistent coding standard makes it easier to understand a large code base.

#### What

From the FactoryBot/FactoryClassName documentation:

> Use string value when setting the class attribute explicitly.
> 
> This cop would promote faster tests by lazy-loading of application files. Also, this could help you suppress potential bugs in combination with external libraries by avoiding a preload of application files from the factory files.

Whether or not the tests are faster as a result of lazy-loading can be seen by comparing the RSpec run time in Circle CI with the same run time of the parent PR, #3650 (There doesn't seem to be a significant improvement)

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -14,28 +14,6 @@ FactoryBot/CreateList:
   Exclude:
     - 'spec/controllers/external_users/admin/external_users_controller_spec.rb'
 
-# Offense count: 33
-# Cop supports --auto-correct.
-FactoryBot/FactoryClassName:
-  Exclude:
-    - 'spec/factories/claim/advocate_claims.rb'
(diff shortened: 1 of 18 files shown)


kmahern (approved):
👍 lgtm

PR #3666, opened 2021-01-04
Fix or ignore cop Metrics/AbcSize for bin/secret binary

jsugarman:
#### What
Fix or ignore Metrics/AbcSize cop for bin/secret.

#### Ticket

n/a

#### Why

This cop is only violated when run against an unencrypted
local version of the file - so does not fail on CircleCI.

#### How

- fix one method by breaking up
- ignore bigger method that necessarily has
  a large block to define OptionParser options


kmahern (approved):
lgtm 👍

PR #3667, opened 2021-01-04
Typeahead specs

naseberry:
#### What
The typeahead function duplicates the default element (`select`) classes to a dynamically created input.

If the type selector (`select`) is removed from the css, both elements become visually hidden, the test would still pass, as the spec could still interact with both elements, however users could not.

Now, the selector will be completely hidden, if the type selector (`select`) is removed, the spec cannot interact with it, the test will fail and engineers will be informed.

#### Ticket
[Test typeahead functionality](https://dsdmoj.atlassian.net/browse/CBO-1632)

### DIFF
--- a/app/webpack/stylesheets/_autocomplete.scss
+++ b/app/webpack/stylesheets/_autocomplete.scss
@@ -15,7 +15,7 @@
 
 // scss-lint:disable QualifyingElement, ImportantRule
 .js-enabled select.typeahead { // stylelint-disable-line selector-no-qualifying-type
-  @include visually-hidden;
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍 I Assume the visually-hidden is not needed in this context for accessibility purposes - i.e.  that screen readers and the like will still get a list of the select's values

naseberry:
> 👍 I Assume the visually-hidden is not needed in this context for accessibility purposes - i.e. that screen readers and the like will still get a list of the select's values

Tested using VoiceOver 👍

PR #3669, opened 2021-01-05
Add govuk_design_system_formbuilder

naseberry:
#### What
Install the govuk_design_system_formbuilder gem.

#### Ticket
[Add a GOVUK Frontend form builder](https://dsdmoj.atlassian.net/browse/CBO-1637)

#### Why
Part of a larger change to:

- reduce and remove redundant form builder throughout the codebase


Gemfile
```diff
@@ -18,6 +18,8 @@ gem 'factory_bot_rails',      '~> 6.1.0'
 gem 'faker',                  '~> 2.13.0'
 gem 'googlemaps-services',    '~> 1.5.0'
 gem 'gov_uk_date_fields',     '~> 3.0.0'
+# rails GDS design system form builder
```
jsugarman: don't really need this comment - gem name is pretty self-explanatory

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,5 @@
 class ApplicationController < ActionController::Base
+  default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
```
jsugarman: unclear on the impact of this. Seems like it could have wide reaching impacts  - is it not better to use the ad-hoc basis rather than the global basis as defined in the [gems setup](https://github.com/DFE-Digital/govuk_design_system_formbuilder#setup-) here

naseberry: I thought so as well, however, during manual testing, nothing happened unless another form builder tries to call the govuk_design_system_formbuilder elements, i.e `= form.govuk_text_field`.

I will remove this anyway and call it on an ad-hoc basis as your comment.

jrmhaig:
If I understand correctly, this will just install the gem but not actually have any change on the app. Is this the case? Or does including the gem have some sort of side-effect?

jsugarman (commented):
If you feel like putting this in I am ok with that but perhaps we could add it as the formbuilder for, say, the certification form, or one of the user management forms and give this PR more of a purpose beyond adding the gem alone. It does not, to my knowledge, doing anything just by adding it.

naseberry:
> If I understand correctly, this will just install the gem but not actually have any change on the app. Is this the case? Or does including the gem have some sort of side-effect?

This was just to install the gem in preparation for migrating all forms to this new formbuilder @jrmhaig

naseberry:
> If you feel like putting this in I am ok with that but perhaps we could add it as the formbuilder for, say, the certification form, or one of the user management forms and give this PR more of a purpose beyond adding the gem alone. It does not, to my knowledge, doing anything just by adding it.

The feedback / bug report forms have been added.
Deployed to dev-lgfs for testing, however fails due to no tokens on this env.
Deployed to staging and manually tested both forms 👍 

@jsugarman

config/locales/en/models/feedback.yml
```diff
@@ -0,0 +1,13 @@
+---
```
jsugarman, line 1: 🎖️

config/locales/en/views/feedback.yml
```diff
@@ -0,0 +1,18 @@
+---
```
jsugarman, line 1: 🎖️

app/views/feedback/bug_report.html.haml
```diff
@@ -1,39 +1,32 @@
...
       = f.hidden_field :type
       = f.hidden_field :referrer
 
-      .form-group
-        = f.label :case_number, class: 'form-label', id: :case_number do
-          = t('.case_number_label')
-          %span.form-hint
-            = t('.case_number_hint')
-        = f.text_field :case_number, class: 'form-control'
+      = f.govuk_error_summary
```
jsugarman, line 12: does this work with links to fields as expected? great if it does. I suspect it won't on the more complex forms

jsugarman: asking because I am not sure how much test coverage there is for the unhappy paths.

jsugarman (approved):
👍 LGTM barring question on errors and whether they just work with links to fields etc

naseberry:
Error handling works well on the feedback form, and will be manually tested against the more complex ones. 👍

PR #3672, opened 2021-01-05
Add default dependabot reviewers

jsugarman:
#### What
Add default dependabot reviewers for its PRs

#### Why
So we do not have to assign reviewers manually


kmahern (approved):
👍

PR #3681, opened 2021-01-06
[CBO-1631] Rubocop: FactoryBot/CreateList

jrmhaig:
#### What

Another Rubocop fix

#### Ticket

[RSpecs rubocop adherence](https://dsdmoj.atlassian.net/browse/CBO-1631)

#### Why

General housekeeping.

#### How

The `FactoryBot/CreateList` favours using `create_list` to create multiple factory instances for tests instead of creating them in a loop. In the one case that needs to be fixed, however, it looks as though the factory instances are not required and so their creation can be removed completely.

I was going to add this to a larger PR but it is probably useful to review in case I am missing something important.


jrmhaig:
This will be merged as part of #3682

PR #3682, opened 2021-01-06
[CBO-1631] More quick Rubocop fixes

jrmhaig:
#### What

More Rubocop fixes

#### Ticket

[RSpecs rubocop adherence](https://dsdmoj.atlassian.net/browse/CBO-1631)

#### Why

General housekeeping

#### How

* `FactoryBot/CreateList`: `create_list` is preferred to a loop when creating multiple instances. In the one case that needs to be fixed, however, it looks as though the factory instances are not required and so their creation can be removed completely.
* `Layout/CaseIndentation`: Align the `when` with the `case` in a case statement. Also remove two cases of duplicate code.
* `Style/MultilineWhenThen`: Remove unnecessary `then` from multi-line case statements.
* `Layout/ClosingParenthesisIndentation`: Closing parenthesis should match opening in a multi-line statement. In two cases this was fixed by putting the whole statement on a single line.
* `Layout/CommentIndentation`: Make sure that commented lines are aligned. In the process of this, an unnecessary `valid_params` method was removed.
* `Layout/DotPosition`: Multiline method calls should have dots at the start of lines instead of the end.
* `Layout/EmptyLineBetweenDefs`: I have set the option `AllowAdjacentOneLineDefs` to true to allow single line method definitions not to require empty lines between them.
* `Layout/EmptyLinesAroundClassBody`: Don't have empty lines around the body of a class.
* `Layout/EmptyLinesAroundModuleBody`: Don't have empty lines around the body of a module.
* `Layout/FirstArgumentIndentation`: First argument of a multi-line argument call is indented one step from the previous line.
* `Layout/FirstHashElementIndentation`: Check indentation of first element of a multi-line hash.
* `Layout/HeredocIndentation`: Consistent indentation for Heredocs.
* `Layout/IndentationWidth`: Consistent indentation width.


kmahern (approved):
lgtm 👍

PR #3684, opened 2021-01-06
Remove `can_have_expenses?` method

jrmhaig:
#### What

Remove the `can_have_expenses?` method from `Claim::BaseClaimPresenter`.

#### Ticket

N/A

#### Why

`Claim::BaseClaimPresenter#can_have_expenses?` is used to determine whether certain sections on claim views and forms are displayed to the user. However, it is set to true and has never been set to false since #2523

#### How

Remove the `can_have_expenses?` method together with all references. Then check the Cucumber tests to see if any features are affected.


jsugarman (approved):
👍 LGTM - did run the question past the BA - i.e. can any claim types not include expenses?

jrmhaig:
> 👍 LGTM - did run the question past the BA - i.e. can any claim types not include expenses?

Do you mean you have checked or asking if I checked?

jrmhaig:
I asked Angela and she responded;

> On any LGFS interim bill scenario other than those for a warrant fee expenses cannot be claimed.  If the expenses page is displaying where any other fee besides “warrant” is being selected by the user this would be an error.

jsugarman:
You might want to try this out, although i suspect if there is an error its already in master as I can't see this code change changing that behaviour

PR #3685, opened 2021-01-07
[CBO-1630] Enable Active Storage

jrmhaig:
#### What

Enable and configure Active Storage

#### Ticket

#### Why

This is just the first step to migrating the app to use Active Storage instead of Paperclip. The tables are set up and the configuration is created but nothing is being used yet.

#### How

Create the database migrations and the configuration in `config/storage.yml`.


jrmhaig:
Closing after accidentally merging jh/use-active-storage in. See #3702

PR #3695, opened 2021-01-14
[CBO-1664] Migrate axe-matchers to axe-core-cucumber

naseberry:
#### What
The package axe-matchers has been moved to axe-core-gems. We need to migrate to the new package and update our assertions.

#### Ticket
[Migrate axe-matchers to axe-core-cucumber](https://dsdmoj.atlassian.net/browse/CBO-1664)

#### Why
Because the package axe-matchers has been deprecated.

### DIFF
--- a/features/claims/hardship_claims_banner.feature
+++ b/features/claims/hardship_claims_banner.feature
@@ -9,7 +9,7 @@ Feature: A hardship claims banner will appear on the claims list, until the user
 
     And I click 'Start a claim'
     Then The hardship claims banner is visible
-    And the page should be accessible within "#content"
+    Then the page should be accessible within "#content"
 
     When I click the link 'Do not show hardship claims information again'
     Then The hardship claims banner is not visible
--- a/features/step_definitions/accessibility.rb
+++ b/features/step_definitions/accessibility.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+Then('the page should be accessible within {string}') do |selector|
+  steps %(Then the page should be axe clean within "#{selector}")
+end
--- a/Gemfile
+++ b/Gemfile
@@ -108,7 +108,7 @@ group :development, :devunicorn, :test do
 end
 
 group :test do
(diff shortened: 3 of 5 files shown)


jsugarman (commented):
If it is not to difficult to configure the gem to change the step name to use `accessible` or raise issue with the gem perhaps.

naseberry:
> If it is not to difficult to configure the gem to change the step name to use `accessible` or raise issue with the gem perhaps.

I've create our own step definition `the page should be accessible`.

The gem authors changed the method from `be_accessible` to `be_axe_clean` https://github.com/dequelabs/axe-core-gems/pull/138/files

@jsugarman

features/claims/hardship_claims_banner.feature
```diff
@@ -9,7 +9,7 @@ Feature: A hardship claims banner will appear on the claims list, until the user
 
     And I click 'Start a claim'
     Then The hardship claims banner is visible
-    And the page should be accessible within "#content"
+    Then the page should be accessible within "#content"
```
jrmhaig, line 12: Very unimportant question. Is there a reason for changing this 'And' to a 'Then'?

naseberry: > Very unimportant question. Is there a reason for changing this 'And' to a 'Then'?

To keep all our accessibility assertions consistent with the step definition

jsugarman: nit, not a blocker, the `Then` and `And` use don't need to match the step definition. Their use should be more inline with the "flow" of the language in the features scenario. but not important really

features/step_definitions/accessibility.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+Then('the page should be accessible within {string}') do |selector|
```
jsugarman, line 3: 👍

jsugarman (approved):
LGTM 👍

PR #3697, opened 2021-01-18
Get Jasmine gem from Rubygems

jrmhaig:
#### What

Use released versions of Jasmine instead of a version taken from the master branch on Github.

#### Ticket

N/A

#### Why

When CCCD was upgraded to Rails 6 (#3060) the latest released version of the Jasmine gem was not supported so the `Gemfile` was changed to pull the latest master from Github. [Rails 6 has subsequently been supported by Jasmine](https://github.com/jasmine/jasmine-gem/releases/tag/v3.5.1) so we can go back to using released versions. This will mean that Dependabot will be able to correctly suggest updates when they are released.

#### How

Update `Gemfile` to get the Jasmine gem from Rubygems.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,7 +87,7 @@ group :development, :devunicorn, :test do
   gem 'guard-rspec'
   gem 'guard-rubocop'
   gem 'guard-webpacker', '~> 0.2.1'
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍

PR #3701, opened 2021-01-20
Remove unused DocumentRecloner class

jrmhaig:
#### What

Remove the files app/rake_helpers/document_recloner.rb and spec/models/rake_helpers/document_recloner_spec.rb.

#### Ticket

N/A

#### Why

The `DocumentRecloner` class was created in #1453 but now it does not appear to be used anywhere. Presumably it was created for a fix at the time but is no longer needed.

#### How

```bash
git rm app/rake_helpers/document_recloner.rb spec/models/rake_helpers/document_recloner_spec.rb
```


jsugarman (approved):
👍  LGTM

PR #3702, opened 2021-01-20
[CBO-1690] [AS1] Configure Active Storage

jrmhaig:
#### What

Enable and configure Active Storage

This replaces #3685 because I accidentally merged the branch jh/use-active-storage into it.

#### Ticket

#### Why

This is just the first step to migrating the app to use Active Storage instead of Paperclip. The tables are set up and the configuration is created but nothing is being used yet.

#### How

Create the database migrations and the configuration in `config/storage.yml`.


config/environments/development.rb
```diff
@@ -13,6 +13,9 @@
     url: "tmp/dev/reports/:filename"
   }
 
+  config.active_storage.service = :local
+  # config.active_storage.service = :amazon
```
naseberry: Do we need this?

kmahern (approved):
👍 lgtm

naseberry (approved):
Looks good, best to get a second review

config/environments/development.rb
jrmhaig: Hmmm, probably not. I'll remove it.

db/schema.rb
```diff
@@ -10,12 +10,33 @@
...
+    t.string "key", null: false
+    t.string "filename", null: false
+    t.string "content_type"
+    t.text "metadata"
+    t.bigint "byte_size", null: false
+    t.string "checksum", null: false
+    t.datetime "created_at", null: false
+    t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
+  end
+
```
jsugarman, line 39: Are we missing an active_storage_variant_records table definition;

I was comparing to [laa-appy-for-legal-aid's limplementation in schema.rb](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/db/schema.rb#L19-L44) and can see a third table. 

Found this on a shallow investigation which implies that rails 6+ required [further active storage migrations](https://stackoverflow.com/questions/58373159/unknown-attribute-service-name-for-activestorageblob)

db/schema.rb
```diff
@@ -10,12 +10,33 @@
...
+  create_table "active_storage_blobs", force: :cascade do |t|
+    t.string "key", null: false
+    t.string "filename", null: false
+    t.string "content_type"
+    t.text "metadata"
+    t.bigint "byte_size", null: false
+    t.string "checksum", null: false
+    t.datetime "created_at", null: false
```
jsugarman, line 36: are we missing a `service_name` column as mentioned in the [comment below](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3702/files#r575083460)

jrmhaig: I was unaware of the `service_name` column but it appears to be introduced in Rails 6.1:

https://guides.rubyonrails.org/6_1_release_notes.html#active-storage-notable-changes

and will be added in the normal upgrade process.

db/schema.rb
jrmhaig: As with `service_name` this appears to be a change in Rails 6.1. It isn't mentioned in the change log but I found this in the PR where it was added: https://github.com/rails/rails/pull/37901#issuecomment-640766714

db/schema.rb
jrmhaig: I tried updating Rails to 6.1 to confirm this (and the other comment about the active_storage_variant_records  table). After running the `active_storage:update` task I get the following two migrations created:

`db/migrate/20210212095632_add_service_name_to_active_storage_blobs.active_storage.rb`:

```ruby
# This migration comes from active_storage (originally 20190112182829)
class AddServiceNameToActiveStorageBlobs < ActiveRecord::Migration[6.0]
  def up
    unless column_exists?(:active_storage_blobs, :service_name)
      add_column :active_storage_blobs, :service_name, :string

      if configured_service = ActiveStorage::Blob.service.name
        ActiveStorage::Blob.unscoped.update_all(service_name: configured_service)
      end

      change_column :active_storage_blobs, :service_name, :string, null: false
    end
  end

  def down
    remove_column :active_storage_blobs, :service_name
  end
end
```

`db/migrate/20210212095633_create_active_storage_variant_records.active_storage.rb`:

```ruby
# This migration comes from active_storage (originally 20191206030411)
class CreateActiveStorageVariantRecords < ActiveRecord::Migration[6.0]
  def change
    create_table :active_storage_variant_records do |t|
      t.belongs_to :blob, null: false, index: false
      t.string :variation_digest, null: false

      t.index %i[ blob_id variation_digest ], name: "index_active_storage_variant_records_uniqueness", unique: true
      t.foreign_key :active_storage_blobs, column: :blob_id
    end
  end
end
```

jsugarman: I thought we had bumped to rails 6.1, but that was VCD. I guess the question becomes more of should we upgrade to rails 6.1 first then? especially as it means more rework for actives storage if we do this now

jrmhaig: I would be open to upgrading to 6.1 first, especially as the Variant Records (which I did not know about) looks like something we could use for the PDF document previews (although it looks to be geared primarily at resizing images).

jsugarman: after having tried the 6.1 bump and it hitting various issues off the blocks. I am going to approve this for merging now. If you want to PR 6.1 before then I will leave it up to you. I am not going to have time this sprint.

jsugarman (approved):
👍  - can you squash commits up into logical groups before merging though. thanks

PR #3709, opened 2021-01-22
[CBO-1642] Migrate bill type form to GOVUK design system

naseberry:
#### What
Migrate bill type form to GOVUK design system

#### Ticket
[Migrate bill type form to GOVUK design system](https://dsdmoj.atlassian.net/browse/CBO-1642)

#### Why
Part of a larger change to:


kmahern (approved):
lgtm 👍

.rubocop.yml
```diff
@@ -62,6 +62,12 @@ Metrics/BlockLength:
     - resource
     - context
     - describe
+    - it
+    - specify
+    - fcontext
+    - fdescribe
+    - fit
+    - fspecify
```
jrmhaig, line 70: 1) I think that `it` and `specify` shouldn't be included in this list and, instead, we should be trying to reduce their size. There are already some flagged in the `.rubocop_todo.yml` file. I may be wrong - is there a standard convention?

2) The `f*` versions shouldn't really be checked in so they don't need to be listed here. At least, following (1), perhaps just `fcontext` and `fdescribe` are required.

app/views/external_users/claim_types/new.html.haml
```diff
@@ -0,0 +1,15 @@
+= content_for :page_title, flush: true do
+  = claim_type_page_header(current_user)
+
+= render partial: 'external_users/shared/hardship_claims_banner'
+
+= form_with model: @claim_type, url: external_users_claim_types_path, method: 'post', builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f|
```
jrmhaig, line 6: :+1: replacing `form_tag` with `form_with`.

jrmhaig (approved):
:+1: Looks good. Just one comment in `.rubocop.yml`, which you may or may not decide to change (not a blocker).

.rubocop.yml
jsugarman: if you see my comment in the commit the length of a specify or it block is already covered by RSpec/ExampleLength. This is just to prevent Metrics/BlockLength cop warnings unnecessarily cluttering the messages.

PR #3713, opened 2021-01-26
[CBO-1691] [AS2] Move assets from Paperclip to Active Storage

jrmhaig:
#### What

Migrate assets in Paperclip to Active Storage

#### Ticket

#### Why

This is the second step of migrating from the deprecated Paperclip to Active Storage. To retain access to assets in existing claims and reports their details need to be copied in the database from where Paperclip expects them to where they need to be for Active Storage.

#### How

Copy data from `*_file_name`, `*_content_type` and `*_file_size` in the `messages`, `documents` and `stats_reports` tables into the `active_storage_attachements` and `active_storage_blobs` tables.

There are three tasks:

* `storage:dummy_files[model]`: Create dummy files for assets in the database (for testing purposes) 
* `storage:calculate_checksums[model]`: Update the `messages`, `documents` or `stats_reports` table to set the `as_*_checksum` field with the checksum of the file, as required by Active Storage.
* `storage:migrate[model]`: Copy the details of files in the database from Paperclip to Active Storage.

The `model` should be `stats_reports`, `messages` or `documents`, as appropriate.

So, to perform the migration for message attachments:

```bash
# Create Active Storage checksums for all existing message attachments
bundle exec rails 'storage:calculate_checksums[messages]`
# Copy file details to Active Storage
bundle exec rails 'storage:migrate[messages]'
```

##### `storage:calculate_checksums`

The checksum calculation is the part of the process that takes the most time so we have added new columns in the database alongside the Paperclip fields so that the checksums can be prepared. This rake task will fill in the field for any attachments for which this is null. Additionally, any new attachments will have the checksum calculated on creation. Therefore, this task can be performed before any migration has started.

##### `storage:migrate`

The migration is done directly in SQL rather than using Active Record for performance reasons. It can be done before or after the models have been updated to use Active Storage, and it can be repeated to update any attachments that have been created subsequently.


jsugarman:
Following discussion between @jrmhaig @jsugarman these are some high level observations and questions:

1. Are there orphan or "unverified" `Document` and `Message` objects that we can delete, regardless? 

2. Do we need to checksum the file if we are just creating a new table record to the same s3 file?!
    - As the slowest part of the process so far, caused by need to download the file, it would have impact on need for or length of downtime, as well as Disk space limitations?!
    - if we need it for `active_storage_blobs.checksum` the can we get it from s3 directly or by some other means
    
3. Can we use a single SQL statement to insert the necessary records into the active storage tables?

4. Handle long file names issue, for anonymised dump and for migration

jrmhaig:
> 1. Are there orphan or "unverified" `Document` and `Message` objects that we can delete, regardless?

Based on an anonymised dump (taken a few weeks ago) there are only a small number of `Document` objects that are not attached to claims - 25 in total - out of a total of over 200,000.

The breakdown of claims in this version of the database is:

* Claim::AdvocateClaim: 39093
* Claim::InterimClaim: 10352
* Claim::LitigatorClaim: 38527
* Claim::TransferClaim: 1629
* Claim::AdvocateSupplementaryClaim: 1185
* Claim::AdvocateHardshipClaim: 1812
* Claim::LitigatorHardshipClaim: 319
* Claim::AdvocateInterimClaim: 25

Are there any of these claim types that do not require the documents to be kept?

> 2. Do we need to checksum the file if we are just creating a new table record to the same s3 file?!

See https://github.com/rails/rails/blob/d30dad4bc1039c80648f39c5171b146f8dd8abd6/activestorage/app/models/active_storage/blob.rb#L262-L278

So it looks as though something like this will raise an exception (`ActiveStorage::IntegrityError`) if the checksum recorded in the database is incorrect:

```ruby
document.document.blob.open do |file|
  # Do something with the file
end
```

Curiously, `.download` doesn't check the checksum. The difference between `.read` and `.download` is that `.read` saves the asset to a temporary local file while `.download` stores the file in memory (in chunks if a block is given).

We are not actually downloading files, either with `.read` or `.download` but it is important to keep this in mind in the future.

> 3. Can we use a single SQL statement to insert the necessary records into the active storage tables?

I have created three SQL scripts to do this:

* db/migrate_document_attachments.sql
* db/migrate_stats_report_attachments.sql
* db/migrate_message_attachments.sql

Each of these will;

a. Create records in `active_storage_blobs` for all the attachments in the relevant tables (documents, stats_reports or messages). The `key` field is used temporarily to record the attachment name (document, converted_preview_document or attachment) and record id so that they can be used in step (2).
b. Create records in `active_storage_attachments` for the `active_storage_blobs` records created in step (1).
c. Update the `key` fields in the `active_storage_blobs` records to the correct path to the files.

Issues with this are;

i. There is a problem with the stats reports script because there are cases where multiple reports have the same filename. The rake task deals with this by using the same blob for each of the attachments but I don't know how to do this in SQL.
ii. The checksums are not being calculated.

These scripts take a matter of seconds rather than hours to run.

> 4. Handle long file names issue, for anonymised dump and for migration

For the moment files with names that are too long are not having the checksum calculated correctly although their other details are being copied to Active Storage. This due to my attempt to fix the more serious problem, where the long file name would cause the task to crash.

jsugarman:
> Are there any of these claim types that do not require the documents to be kept?
nope, afraid not

I will take a closer look tomorrow if possible and then we can have a catchup. Particularly around checksum issues i think

app/models/document.rb
```diff
@@ -53,9 +53,30 @@ class Document < ApplicationRecord
 
   before_save :generate_pdf_tmpfile
   before_save :add_converted_preview_document
+  before_save :create_checksum
 
   validate :documents_count
 
+  # TODO: Remove after moving to Active Storage
```
jsugarman, line 61: Appreciate that this is an opimization for something that is temporary (famous last words) but we could create a helper/mixin like
```
module CheckSummable
    def add_checksum(checksum_field, io)
      checksum = Digest::MD5.new.tap do |checksum|
        while (chunk = io.read(5.megabytes))
          checksum << chunk
        end
        
        io.rewind
      end

      send("#{checksum_field}=", checksum.base64digest)

     rescue Errno::ENOENT
       send("#{checksum_field}=", nil)
    end
end
```

and eg. for use. For Document you could just loop over the the two documents i guess

```
class Message
  ....
  before_save :populate_checksum

  def polulate_checksum
    return if attachment_file_name.nil?
    io = Paperclip.io_adapters.for(attachment)

     add_checksum(:as_attachment_checksum, io)
  end

```

app/models/message.rb
```diff
@@ -49,6 +49,7 @@ class Message < ApplicationRecord
 
   scope :most_recent_last, -> { includes(:user_message_statuses).order(created_at: :asc) }
 
+  before_save :create_checksum
```
jsugarman: as [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3713/files#r575104297)

lib/tasks/rake_helpers/storage.rb
```diff
@@ -0,0 +1,138 @@
+module Storage
+  ATTACHMENTS = {
+    'stats_reports' => ['document'],
+    'messages' => ['attachment'],
+    'documents' => ['document', 'converted_preview_document']
+  }.freeze
+
+  def self.migrate model
```
jsugarman: nit - prefer braces around parameters

jsugarman (commented):
Great work first of all.

- I have added a suggested optimisation for creating checksums if it is easy enough to implement. 
- can we add tests to verify paperclip checksum creation by the hook

Also, seems like we should break this into two PRs:

1. paperclip checksum implementation
2. paperclip -> active storage data migration task

To my mind the migrate task should be run on a background worker, not on a web server instance. This is to mitigate impact of memory consumption and diskspace thresholds. This may be as simple as shelling into the worker pod to run the migrate task, but we may need to tweak memory and diskspace on the worker to compensate.

Some questions:
1. resilience: If the migrate task fails can it just be restarted and just pick up where it left off. Could partial checksums caused by failure cause rerun problems?! Could nil checks prevent a a successful restart?! may need to verify checksum is correct if it exists?!

2. verification: how can we tell if the checksums are correct? is there a task we can create to verify checksums and produce a report?

3. What is in place to handle the long name issue you identified previously?


Happy to talk through any of this.

lib/tasks/storage.rake
```diff
@@ -0,0 +1,52 @@
...
+      records = TempDocument.where(as_document_checksum: nil)
+    else
+      puts "Cannot calculate checksums for: #{args[:model]}"
+      exit
+    end
+
+    puts "Setting checksums for #{args[:model].green}"
+    Storage.set_checksums(records: records, model: args[:model])
+  end
+end
```
jsugarman: linebreak

lib/tasks/rake_helpers/storage.rb
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  def self.s3_path_pattern model
+    {
+      'stats_reports' => REPORTS_STORAGE_OPTIONS[:path],
+      'messages' => PAPERCLIP_STORAGE_OPTIONS[:path],
+      'documents' => PAPERCLIP_STORAGE_OPTIONS[:path]
+    }[model]
+  end
+end
```
jsugarman: linebreak

lib/tasks/rake_helpers/storage.rb
```diff
@@ -0,0 +1,138 @@
...
+      progress_mark: '#'.green,
+      remainder_mark: "\u{FF65}".yellow,
+      starting_at: 0,
+      total: total
+    )
+  end
+
+  private
+
+  def self.models model
```
jsugarman: prefer braces

lib/tasks/rake_helpers/storage.rb
```diff
@@ -0,0 +1,138 @@
...
+        UPDATE active_storage_blobs SET metadata='{}' WHERE metadata LIKE 'in-progress/%';
+      SQL
+
+      connection.exec_prepared('create_active_storage_blobs');
+      connection.exec_prepared('create_active_storage_attachments');
+      connection.exec_prepared('update_active_storage_blobs_metadata');
+    end
+  end
+
+  def self.make_dummy_files_for model
```
jsugarman: prefer braces

jsugarman:
also, @jrmhaig, once it is ready could you squash up commits

app/models/document.rb
jsugarman: do we also need to handle long name error
```
rescue Errno::ENAMETOOLONG
    log('FilenameTooLong')....blah
    send("#{checksum_field}=", nil)
 ```

Comment (jrmhaig) [reply app/models/document.rb:61]: I have moved this into a separate module now and have it logging (and not setting the checksum) when there are errors.

File: lib/tasks/storage.rake [hunk 8]
```diff
@@ -0,0 +1,114 @@
...
+
+namespace :storage do
+  desc 'Migrate assets from Paperclip to Active Storage'
+  task :migrate, [:model] => :environment do |_task, args|
+    Storage.migrate args[:model]
+  end
+
+  desc 'Create dummy assets files'
+  task :dummy_files, [:model] => :environment do |_task, args|
+    Storage.make_dummy_files_for args[:model]
```
jsugarman, line 11: put production protection error raising

app/models/concerns/check_summable.rb
```diff
@@ -0,0 +1,30 @@
...
+    checksum = Digest::MD5.new
+    while (chunk = io.read(5.megabytes))
+      checksum << chunk
+    end
+
+    io.rewind
+    checksum.base64digest
+  end
+
+  def checksum_log(message, error)
```
jsugarman, line 23: nit - could this be private

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+
+* Stats reports
+* Attachments on messages
+* Evidence documents
+
+and each of these is dealt with separately. The process is similar for each but
+with some small differences and can be summarised as:
+
+1) Generate checksums for all existing assets. The Checksum is required by
+   Active Storage and is time consuming. However, it can be done in advance of
```
jsugarman, line 24:
```suggestion
   Active Storage and is time consuming to generate. However, it can be done in advance of
```

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+# Number of stats reports with nil checksum
+rails> srs.where(as_attachment_checksum: nil).count
+# Should be zero
+
+# Number of stats reports with a valid checksum
+# A valid checksum ends with '==', such as '4+g5Nmjax5aCg3HtRmCO5Q=='
+rails> srs.select { |sr| sr.as_attachment_checksum&.match(/==$/) }.count
+# Should be the same as srs.count, above
+```
+
```
jsugarman, line 58: don't you have a rake task that outputs missing checksums. be good to document that here

jsugarman: and maybe extend that to perform the basic checksum validity check you have written here
```
srs.select { |sr| sr.as_attachment_checksum&.match(/==$/) }.count
```

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+2) Next, regenerate the report and download again. Confirm that the new
+   document is different from the previous.
+
+### Rollback
+
+**Note 1:** This rollback should not be attempted after PR #3722 has been deployed
+(final step, above) as this will delete assets from Active Storage that do not
+exist in Paperclip.
+
+**Note 2:** Using `destroy` or `destroy_all` on `ActiveStorage::Attachment` may
```
jsugarman, line 107: is there a way to have a `rails 'storage:rollback[stats_reports]'` task that encapsulates this logic in the event we need to do it.

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+```ruby
+rails> attachments.delete_all
+rails> blobs.delete_all
+```
+
+4) Delete checksums
+
+```ruby
+rails> Stats::StatsReport.update_all(as_document_checksum: nil)
+```
```
jsugarman, line 134: is there a way to have a `rails 'storage:rollback[stats_reports]'` task that encapsulates this logic in the event we need to do it.

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+6) (Final step) Activate Active Storage for stats reports by deploying PR #3735 (CBO-1692)
+
+### Testing
+
+1) Find an existing case with a message attachment. Confirm that the attachment
+   can be downloaded correctly.
+2) Add a new message to a case and add an attachment. Confirm that the
+   attachment can be downloaded correctly.
+
+### Rollback
```
jsugarman, line 201: [comment as for stats_reports](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3713/files#r581210705)

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+### Testing
+
+1) Find an case with a message attachment. Confirm that evidence documents can
+   be downloaded and viewed in browser.
+2) As a caseworker, confirm that all documents can be downloaded as a zip file.
+   can be downloaded correctly.
+3) Add a new case and confirm that documents can be accessed as in steps (1)
+   and (2)
+
+### Rollback
```
jsugarman, line 328: [comment as for stats report](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3713/files#r581210705)

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+rails> blobs.delete_all
+```
+
+4) Delete checksums
+
+```ruby
+rails> Document.update_all(as_document_checksum: nil, as_converted_preview_document_checksum: nil)
+```
+
+## Clean up
```
jsugarman, line 363: 🎖️

db/migrate/20210204164407_add_attachment_checksums.rb
```diff
@@ -0,0 +1,8 @@
+class AddAttachmentChecksums < ActiveRecord::Migration[6.0]
```
jsugarman, line 1: not a blocker - generally better, particularly if deploys are slow, to get migrations in as a separate deploy so long as nil values are considered valid/no-db-contraints. Not doing so could risk conditions where the code is updated just as someone is saving an attachment but migrations have not been applied - Not sure this is possible but we have had instances in the past where we have had errors related to code and migration changes going in together.

jsugarman: although i think it is more the reverse - i.e. if db migration gets applied first but code is still old code, which i do not believe will matter in this scenario?!

lib/tasks/rake_helpers/storage.rb
```diff
@@ -0,0 +1,128 @@
...
+        FileUtils.mkdir_p File.dirname(filename)
+        File.open(filename, 'wb') { |file| file.write(SecureRandom.random_bytes(record.send("#{name}_file_size"))) }
+      end
+    end
+  end
+
+  def self.set_checksums(records:, model:)
+    bar = self.progress_bar title: ATTACHMENTS[model].join(', '), total: records.count
+
+    records.each do |record|
```
jsugarman, line 85: could  be something to look at testing if this work is on a separate branch, but it might be possible to use batches to speed this up?!

jsugarman:
As discussed:

1. we need some specs for the generation of checksums for the new paperclip attachments
2. environment protection to prevent accidental running of destructive rake tasks on production (e.g. dummy document generation) - just put `production_protected` as first line of the task and include rake_utils.
3. ideally split into 2 PRs - one for "checksums for the new paperclip attachments" and the other for the rake tasks to "backdate" the checksum
4. Not in this PR but A rollback task to encapsulate whatever is necessary to rollback an actual active storage migration as mentioned in `docs/active_storage_migration.md`

docs/active_storage_migration.md
```diff
@@ -0,0 +1,387 @@
...
+    `document_file_size`, `document_updated_at`, `as_document_checksum`
+  * In `messages`; `attachment_file_name`, `attachment_content_type`,
+    `attachment_file_size`, `attachment_updated_at`, `as_attachment_checksum`
+  * In `documents`; `document_file_name`, `document_content_type`
+    `document_file_size`, `document_updated_at`,
+    `converted_preview_document_file_name`,
+    `converted_preview_document_content_type`
+    `converted_preview_document_file_size`,
+    `converted_preview_document_updated_at`,
+    `as_document_checksum`, `as_converted_preview_document_checksum`
```
jsugarman, line 387: Could add a note here to check if cloning redraft failures due to timeouts improve - Check kibana logs for timeout failures on redrafting/cloning

PR #3720, opened 2021-01-29
Anonymise entire filename

jsugarman:
#### What
Anonymise entire attachment filenames during db dump

#### Ticket
relates to active storage migration work

#### Why
Previously the extension (text after last dot)
was not anonymised. However, some files with extremely
long names are being uploaded and the names truncated.
this can result in an unanonymised part of the file name.

#### How

This solution renames the entire file and relies on the
mime content_type to then reattach the correct extension
to the name.


jrmhaig:
👍 Looks good.

jrmhaig (approved):

PR #3721, opened 2021-01-29
A few more Rubocop things

jrmhaig:
#### What

More Rubocop fixes.

* Before: `1161 files inspected, 14279 offenses detected, 4857 offenses auto-correctable`
* After: `1161 files inspected, 14232 offenses detected, 4812 offenses auto-correctable`

#### Ticket

#### Why

* Lint/RaiseException: `StandardException` should be used instead of `Exception`
* Lint/UnusedBlockArgument: Removal of unused block arguments. In some places this involved changing `each` to `each_key` or `each_value`.
* Naming/HeredocDelimiterCase: Heredoc delimiters should be all caps
* Naming/RescuedExceptionsVariableName: Rescued exception variable should be called `e`


spec/api/v1/external_users/claims/integration/advocate_claim_creation_spec.rb
```diff
@@ -2,15 +2,15 @@
 
 RSpec::Matchers.define :be_valid_api_agfs_claim do |expected|
   match do |claim|
-    @results = results(claim)
+    @results = results(claim, expected)
```
jrmhaig, line 5: The reason for this change (and on line 13 below) is because `expected` was getting changed to `_expected` on line 3 as it was unused. As it happens, this didn't cause an error in `results` so I wonder if it actually needs to be explicitly given. Does anyone know if `expected` is automatically defined inside `RSpec::Matchers.define` blocks?

jsugarman: This is just the conventional name for custom matchers arg yielding - `match do |claim|` could be changed to `match do |actual|` and `@results = results(actual, expected)` to be more conventional too. 

looking at the [definition](https://github.com/rspec/rspec-expectations/blob/955151da74fd6356e614e62addb3da26ea1ccaf6/lib/rspec/matchers/dsl.rb#L72) here (i think), it don't see it being anything other than a normal yielded parameter (not predined). You could try out one of its uses and see but probably better to sticked to convention in any event

jsugarman: it seems a bit odd that it come up as an unused param however as the results method is inside the block and clearly uses it 🤷 . rubocop oversite?!

jsugarman (approved):
👍 LGTM

PR #3722, opened 2021-02-01
[CBO-1683] [AS3a] Use Active Storage for stats reports

jrmhaig:
#### What

Switch over to using Active Storage for stats report.

#### Ticket

#### Why

We need to switch over to using Active Storage everywhere we have documents in CCCD. The simplest of these, and the place with the least risk, is for the stats reports and so this is being done first.

#### How


jsugarman (approved):
Looks ok. Am just querying the redirect mechanism with the apply team to see how if at all they are handling such things, but not a blocker.

Not directly related and not a blocker but also wondering whether we are handling authorization correctly for downloading anything (any of the models attachments). the `skip_load_and_authorize_resource` for `download` would appear to suggest not.

jsugarman:
FYI @jrmhaig  I found a thread i had since lost that had a third alternative to the `download` which involved [ActionController::Live streaming](https://api.rubyonrails.org/v6.0.0/classes/ActionController/Live.html)

jrmhaig:
Closing in favour of new version - #3783

PR #512, opened 2015-09-10
Date attended validation

stephenrichards:
Also validates associated models and fixes issues where factory is creating reporder dates out of sequence

PR #3723, opened 2021-02-01
Remove awesome_print gem

jrmhaig:
#### What

Remove the [Awesome Print](https://github.com/awesome-print/awesome_print) gem.

#### Ticket

N/A

#### Why

As far as I can see, Awesome Print is only used in one place and that is for debugging that should probably have been deleted. This is not a massive saving in disk space or memory but it does (slightly) clear the list of libraries that we are pulling in.

#### How

Remove `awesome_print` from `Gemfile` and replace any instances of `ap` in the code with `pp` (which probably didn't exist when the was added).

#### Important notes for reviewers

* Am I right in my assumption that this gem is not being used elsewhere? Are you aware of anywhere else it appears?
* Is that `rescue` block in spec/factories/claim/base_claims.rb actually adding any value or can it be deleted? See comments on #512

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gem 'aws-sdk-costexplorer',   '~> 1'
 gem 'aws-sdk-s3',             '~> 1'
 gem 'aws-sdk-sns',            '~> 1'
 gem 'aws-sdk-sqs',            '~> 1'
-gem 'awesome_print'
 gem 'bootsnap', require: false
 gem 'cancancan',              '~> 3.2'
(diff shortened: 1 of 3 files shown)


jsugarman (commented):
I'd like this gem kept in. I use his gem regularly as an alternative to pp, puts. i also use it in consoles on hosted envs.

jrmhaig:
> I'd like this gem kept in. I use his gem regularly as an alternative to pp, puts. i also use it in consoles on hosted envs.

OK. I just saw it being used in one place only and thought it had been added and then forgotten about.

PR #3363, opened 2020-05-15
Migrating from Sprockets to Webpacker

naseberry:
# Migrating from Sprockets to Webpacker

## What
To successfully migrate our existing assets from Sprockets to Webpacker.

## Ticket
[Migration from Sprockets to Webpacker](https://dsdmoj.atlassian.net/browse/CBO-1196)

## Why
With the upgrade to Rails 6, Webpacker is now the default asset compiler. Although Sprocket is still available, the benefits of Webpacker highlights a need to understand how we may successfully migrate from the Sprockets pipeline.

### Add Webpacker to the project
[Webpacker](https://github.com/rails/webpacker) is a gem which wraps Webpack. Webpacker  configuration differs slightly to Webpack and has a slight learning curve. Installation guide was well documented and easy to follow.

We have renamed the webpack source path from `app/javascript` to `app/webpack`. Why? It is confusing to nest stylesheets, images and javascripts within a "javascript" directory, plus it is a more descriptive directory name and it follows the naming consistency of the Webpacker default directory structures.

### Migrate Stylesheets
Stylesheets moved from `app/assets/stylesheets` to `app/webpack/stylesheets`
`vendor/assets/stylesheets` have also been moved to `app/webpack/stylesheets`

Sass compiler is now handled by node-sass that is installed as a dependency to Webpacker and will automatically process our SCSS files once we import it within our entrypoint.

As our stylesheet uses images, we need to specify our new image path within our entrypoint.

### Migrate Images
Images moved from `app/assets/images` to `app/webpack/images`
`vendor/assets/images` have also been moved to `app/webpack/images`

This migration was achieved by context requiring our images directory, Webpacker recursively checks this directory and adds all files to the manifest file, then we are able to use the view helper `image_pack_tag`.

### Migrate Javascripts
Javascripts moved from `app/assets/javascripts` to `app/webpack/javascripts`
`vendor/assets/javascripts` have also been moved to `app/webpack/javascripts`

The Javascript entrypoint changes from `app/assets/javascripts/application.js` to `app/javascript/packs/application.js`.

Webpack is different than Sprockets in the sense that it compiles modules whereas Sprockets concatenates, you should not try to use Webpack as you would use Sprockets!, however, I ended up using Webpack like sprockets, by including [webpack-concat-plugin](https://www.npmjs.com/package/webpack-concat-plugin) as there were compounding errors trying to compile our JS files as modules.

[...]

PR #3724, opened 2021-02-01
[CBO-1685] Remove Susy gem

naseberry:
#### What
Remove susy gem.

#### Ticket
[Remove susy gem](https://dsdmoj.atlassian.net/browse/CBO-1685)

#### Why
Leftover technical debt when we move asset management from [Sprockets to Webpacker ](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3363)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,6 @@ gem 'remotipart',             '~> 1.4'
 gem 'rest-client',            '~> 2.1' # needed for scheduled smoke testing plus others
 gem 'sass-rails',             '~> 6.0.0'
 gem 'scheduler_daemon',       git: 'https://github.com/jalkoby/scheduler_daemon.git'
-gem 'susy',                   '~> 2.2.14'
 gem 'sentry-raven',           '~> 2.13.0'
 gem 'simple_form',            '~> 5.0.3'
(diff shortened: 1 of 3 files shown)


jrmhaig:
I think there need to be changes on the Javascript side too. Susy is imported in app/webpack/stylesheets/application.scss, and do updated versions of package.json and yarn.lock need to be checked in?

When I remove susy from these files I then get errors around `@include container;` in various places. I guess there are still remnants of susy around, but I don't understand enough of the Javascript side to see what.

naseberry:
> I think there need to be changes on the Javascript side too. Susy is imported in app/webpack/stylesheets/application.scss, and do updated versions of package.json and yarn.lock need to be checked in?
> 
> When I remove susy from these files I then get errors around `@include container;` in various places. I guess there are still remnants of susy around, but I don't understand enough of the Javascript side to see what.

Susy is still required, just not as a gem, as we no longer use Sprockets for the asset pipeline.

Susy is now loaded via npm for use in Webpacker, we require the specific version due to the fact we are using certain mixins (deprecated in newer versions) throughout our stylesheets.
This will be actioned as we migrate and or remove deprecated dependencies throughout the codebase.

jrmhaig (approved):

PR #3726, opened 2021-02-02
[CBO-1658] Migrate MI report form to GOVUK design system

naseberry:
#### What
Migrate MI report form to GOVUK design system

#### Ticket
[Migrate MI report form to GOVUK design system](https://dsdmoj.atlassian.net/browse/CBO-1658)

#### Why
Part of a larger change to:


jrmhaig (approved):
:+1: I don't really understand all the GOVUK Frontend stuff too well but it looks much cleaner now and I've checked the form.

One issue (which applies to the old version too so is not strictly part of this PR); The 'Download' button looks as though it is disabled, although it does work. This screenshot shows the buttons for the rejections/refusals report and the provisional assessment by date form.

Screenshot 2021-02-02 at 14 26 37

(I have approved the PR as the problem with the button is an existing issue)

app/views/case_workers/admin/management_information/index.html.haml
```diff
@@ -20,68 +20,25 @@
...
 
-    -# GDS Date start
-    .form-group.fx-start-date
-      %fieldset
-        %legend
-          %span.form-label-bold
-            = t('.start_date')
-          %span.form-hint
-            = t('.example_date')
+  = form_with url: '', builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f|
```
jsugarman: is the `url: ''` needed? - default [form_with url is nil](https://api.rubyonrails.org/v6.1/classes/ActionView/Helpers/FormHelper.html#method-i-form_with) which [will POST to the current page](https://guides.rubyonrails.org/form_helpers.html#dealing-with-basic-forms) according to that guide - but have not tested

jsugarman (approved):
Just one suggestion related to `url: ''` but if it works then all good 👍

PR #3727, opened 2021-02-02
Attempt to quieten diskspace alerts on cccd production

kmahern:
#### What
Increase the threshold for prometheus diskspace alerts on CCCD production.

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1684

#### Why
This alert is too noisy/appears to be firing even when there is no issue with diskspace 

#### How
Increased the diskspace alert threshold on CCCD production from 150MB to 300MB.

--------

#### TODO (wip)

 - [ ] Investigate whether this alert is useful?


jsugarman (approved):
👍 makes sense to try this. this command `du -h /usr/src/app -d 1` shows that node_modules is taking up 148.4M on its own, 217.8Mb altogether for `/usr/src/app`.  Tmp upload files files might then be causing it to increase and thereby trigger a threshold alarm (which has already been breached). 

We shall see! 🎖️

PR #3735, opened 2021-02-03
[CBO-1692] [AS3b] Use Active Storage for message attachments

jrmhaig:
#### What

Switch over to using Active Storage for attachments of claim messages.

#### Ticket

#### Why

We need to switch over to using Active Storage everywhere we have documents in CCCD. Following on from the PR to turn this on for stats report (#3722) the next step is for attachments of messages.

#### How


jrmhaig:
Replaced by #3824

PR #3737, opened 2021-02-04
Use new version of fee-calculator-client gem

kmahern:
#### What
Use v1.2.0 of laa-fee-calculator-client gem.

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1671

#### Why
v1.1.0 of the laa-fee-calculator-client gem had become out of date and was preventing us from updating CCCD dependencies (e.g. sentry-rails)

#### How
Updated Gemfile


jsugarman (approved):
I assume the vcr cassette agent version is not causing problem then?!

jsugarman:
@kmahern for completeness you should problaby rerecord all fee-calculation feature cassettes. This will exercise the fee calculator for all the scenarios and record the updated responses

you can see that the cassettes are still using recordings of version 1.1.0. Which means they are basically stubbing any fee calc request with 1.1.0 responses and so the tests pass. Rerecording will ensure that they all still work against v1.2.0 and create new stubs in vcr. If you want to pair ping me, but you can read about it in the [testing readme](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/docs/testing.md#how-we-test-external-services)

```
      User-Agent:
      - laa-fee-calculator-client/1.1.0
```

jrmhaig (approved):
:+1:

PR #3738, opened 2021-02-04
Some more Rubocop

jrmhaig:
#### What

Some more Rubocop fixes.

After these fixes:

```
1161 files inspected, 13316 offenses detected, 3902 offenses auto-correctable
```

#### Ticket

#### How

* Naming/MethodParameterName
* Rubocop: RSpec/ExampleWording; Tests called `it 'should do something'` changed to `it 'does something'`. These were auto-corrected and sometimes the wording confused Rubocop (eg `should only do X` was auto-corrected to `onlies do X`) but I think I corrected them all.
* Rubocop: RSpec/ExpectChange; `expect { ... }.to change { thing.value }` changed to `expect { ... }.to change(thing, :value)`.
* Rubocop: RSpec/HookArgument; `:each` is the default on `before` and `after` do it doesn't need to be explicitly given.


kmahern (approved):
👍 lgtm, I have tried to check as many of the 'should' wording changes as possible

jsugarman (approved):
🎖️ looks good. I'd get it in quick as it will get stale quickly

PR #3740, opened 2021-02-09
Fix linting issues

naseberry:
#### What

StandardJS erred with exit code 1 due to an unexpected trailing comma.
The pre-commit hook failed to stop commits due to our last lint command passing.
This commit updates pre-commit hook to immediate return an exit code should it fail and stop the commit

This commit also updates the Stylelint glob that now validates files in nested directories.

#### Ticket

n/a


jrmhaig (approved):
👍 Looks OK

PR #3748, opened 2021-02-10
More Rubocop

jrmhaig:
#### What

Some more Rubocop fixes.

Before: `1161 files inspected, 13307 offenses detected, 3897 offenses auto-correctable`
After: `1161 files inspected, 12960 offenses detected, 3565 offenses auto-correctable`

#### Ticket

#### How

Most of these are small fixes:

* [Performance/InefficientHashSearch](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/ee2960456df37eaa42e1d9fdc84f620ef395b262)
* [Performance/RangeInclude](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/537b023987251e12de5d6c86a8937ab6a08d4571)
* [Performance/RedundantBlockCall](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/a393068e8cfb6ca6c5182293980b6699bb8170a2)
* [Performance/RedundantMerge](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/d3e5fa2a9cff76ead7fe983ebe0e3c42dd248055)
* [Performance/StringReplacement](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/8cbe1362dd18da01e00051bb79616efc10a61155)

The final one is much larger: [RSpec/BeEql](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/8cbe1362dd18da01e00051bb79616efc10a61155)

From the documentation:

> Check for expectations where `be(…)` can replace `eql(…)`.

After allowing Rubocop to auto-correct `eql` with `be` I have also made the following changes;

* `eql` instead of `eq`
  * `expect(x).to eq y` compares `x == y`, which will be true if `x` and `y` are of different types. So `1 == 1.0` is true
  * `expect(x).to eql y` compares `x.eql? y`, which will only be true if `x` and `y` are of the same type. So `1.eql? 1.0` is false
* `be_truthy` or `be_falsey` instead of `eql true` or `eql false`
* `be_nil` instead of `be nil`
* `expect(subject.flag?).to be_truthy` replaced by `expect(subject).to be_flag`, and similar for `be_falsey` (see [RSpec/PredicateMatcher](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/PredicateMatcher))
  * I expect there may be some discussion about this and I may need to make changes to `.rubocop.yml` if this is not the desired behaviour


spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -66,7 +66,7 @@
     it 'is valid when absent' do
       expect(claim.case_concluded_at).to be_nil
       claim.valid?
-      expect(claim.errors.key?(:case_concluded_at)).to be false
+      expect(claim.errors).not_to be_key(:case_concluded_at)
```
jsugarman:
```suggestion
      expect(claim.errors).not_to have_key(:case_concluded_at)
```

spec/api/entities/ccr/defendant_spec.rb
```diff
@@ -25,12 +25,12 @@
...
   it 'returns main defendant true for the defendant created first' do
     @created_at = 1.minute.ago
-    expect(response[:main_defendant]).to eql true
+    expect(response[:main_defendant]).to be_truthy
   end
 
   it 'returns main defendant false for defendant created after other defendants' do
     @created_at = 1.minute.from_now
-    expect(response[:main_defendant]).to eql false
+    expect(response[:main_defendant]).to be_falsey
```
kmahern: Was there a reason for preferring .to be_falsey over .to be false when replacing eql false? (and .to be_truthy over .to be true)? (is it possible to chose what rubocop autocorrects to?)

spec/models/case_type_spec.rb
```diff
@@ -31,19 +31,19 @@
 
     describe 'is_graduated_fee?' do
       it 'returns false if no fee_type_code' do
-        expect(nil_case_type.is_graduated_fee?).to eql false
+        expect(nil_case_type).not_to be_is_graduated_fee
```
kmahern: (Personal opinion) I am not keen on the expect(subject).to be_flag syntax, it seems convoluted to me, especially when it is not_to be_flag like here. Was the existing code (expect(nil_case_type.is_graduated_fee?).to eql false) raising an offence?

kmahern:
I had a couple of questions about the RSpec/BeEql cop, all the rest look fine.

spec/api/entities/ccr/defendant_spec.rb
jrmhaig: Rubocop changes `eql false` to `be false`, which will check that the value is exactly `false`. `be_falsey` checks that the value evaluates to `false`, so `nil` will also be accepted.

In most places I had been changing the `eql` to `eq`, which is slightly different but I think (maybe I'm wrong) this is what is actually intended. `eql` compares as `actual.eql? expected`, which requires that `actual` and `expected` have the same type as well as the same value, whereas `eq` compares as `actual == expected`, where type conversion is allowed. I believe that for `true` and `false` this makes `eq false` identical to `be_falsey`.

What do people think? Should we be checking for exactly boolean values in some cases or is it generally OK to check `truthy` and `falsey`?

spec/models/case_type_spec.rb
jrmhaig: This is the sort of thing that made me wonder whether this is the correct policy but yes Rubocop is raising offences over this (more or less).

The original `.to eql false` is picked up by `RSpec/BeEql`, which prefers `be` over `eql`.

I actually changed it to `.to be_falsey`, rather than `.to be false`, and it was then flagged by [`RSpec/PredicateMatcher`](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/PredicateMatcher), which prefers the `be_is_graduated_fee?`.

As it happens, the `is_graduated_fee?` flag fails the [`Naming/PredicateName`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/PredicateName) cop but is currently being ignored by the todo file. According to that cop it should be `graduated_fee?` in which case the test would read:

```ruby
expect(nil_case_type).not_to be_graduated_fee
```

and this makes a bit more sense.

Another issue is that I cannot find `is_graduated_fee?` used anywhere other than in the tests. Can it be deleted? @jsugarman

jsugarman: This whole section could do with a refactor as it is not using contexts and reusing memoized lets. ideally it would be something like these two separate describes

```
describe '#graduated_fee_type do
   subject  { case_type.graduated_fee_type }
   
   before { create :fixed_fee_type, unique_code: 'FIXED' }
   
   context 'when case type has no fee_type_code' do
      let(:case_type) { build :case_type, fee_type_code: 'FIXED' }
      
      it { is_expected.to be_nil }
   end
...
end
```
and
```
describe '#is_graduated_fee?' do
    subject  { case_type.is_graduated_fee? }
    
    before { create :fixed_fee_type, unique_code: 'FIXED' }
    
    context 'when case type is a fixed fee type' do
      let(:case_type) { build :case_type, fee_type_code: 'FIXED' }
      
      it { is_expected.to be false }
   end
   ...
end
```

Without more investigation into it use or refactoring of the actual method i would be minded to test the actual existing functionality which is to only ever return true or false
```
  def is_graduated_fee?
    graduated_fee_type.nil? ? false : true
  end
```
therefore `it { is_expected.to be false }` and `it { is_expected.to be true }`

>Another issue is that I cannot find is_graduated_fee? used anywhere other than in the tests. Can it be deleted? @jsugarman

I cannot either, nor any potential meta-programming calls to it. I would have a separate PR to remove though

jsugarman: BTW, i am not saying you need to refactor but i think `be false` is the best bet

spec/api/entities/ccr/defendant_spec.rb
jsugarman: As per other comment it is spec'ing this method 
```
  def main_defendant
    object.claim.defendants.unscope(:order).order(created_at: :asc)&.first == object || false
  end
  ```
so without investigating impact further i would say `be false` , `be true` should be used

File: spec/models/claim/transfer_claim_spec.rb [hunk 4]
```diff
@@ -124,14 +124,14 @@
       claim = Claim::TransferClaim.new(case_number: 'A20161234', litigator_type: 'new', elected_case: false, transfer_stage_id: 10, transfer_date: today, case_conclusion_id: 30)
       expect(claim.case_number).to eq 'A20161234'
       expect(claim.litigator_type).to eq 'new'
-      expect(claim.elected_case).to be false
+      expect(claim.elected_case).to be_falsey
```
kmahern: Can we stick with expect(claim.elected_case).to be false ?

spec/services/claims/fee_calculator/modifier_decorator_spec.rb
```diff
@@ -24,8 +24,8 @@
   context 'when called with arg of same class of object' do
     let(:other_decorated_modifier) { described_class.new(OpenStruct.new(fixed_percent: '-20.00', percent_per_unit: '0.00')) }
     it 'calls super' do
-      expect(decorated_modifier.send(method, decorated_modifier)).to eql true
-      expect(decorated_modifier.send(method, other_decorated_modifier)).to eql false
+      expect(decorated_modifier.send(method, decorated_modifier)).to be_truthy
```
kmahern: For consistency, can we use expect(decorated_modifier.send(method, other_decorated_modifier)).to be true ?

spec/models/claim/transfer_claim_spec.rb
```diff
@@ -124,14 +124,14 @@
...
       expect(claim.transfer_stage_id).to eq 10
       expect(claim.transfer_date).to eq today
       expect(claim.case_conclusion_id).to eq 30
 
       detail = claim.transfer_detail
       expect(detail.litigator_type).to eq 'new'
-      expect(detail.elected_case).to be false
+      expect(detail.elected_case).to be_falsey
```
kmahern: Can we stick with expect(detail.elected_case).to be false?

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -1479,7 +1479,7 @@
       }
       claim = Claim::AdvocateClaim.new(params['claim'])
       claim.creator = external_user
-      expect(claim.save).to be true
+      expect(claim.save).to be_truthy
```
kmahern: Can we stick with expect(claim.save).to be true ?

spec/services/claims/fee_calculator/modifier_decorator_spec.rb
```diff
@@ -24,8 +24,8 @@
...
+      expect(decorated_modifier.send(method, other_decorated_modifier)).to be_falsey
```
kmahern: For consistency, can we use expect(decorated_modifier.send(method, other_decorated_modifier)).to be false ?

kmahern:
Have flagged a few places where we are using be_truthy and I think we could be using be true, or be_falsey and I think we could be using be false otherwise looking good. Is it also possible to squash some of the commits/rebase?

kmahern (approved):
🏅 this is looking good to me @jrmhaig

PR #3749, opened 2021-02-11
Remove unused includes_state? method

jrmhaig:
#### What

Remove the `includes_state?` method from the `ClaimsHelper` module.

#### Ticket

N/A

#### Why

The method is no longer being used anywhere.

#### How

Remove the method together with its unit tests.


kmahern (approved):
👍 lgtm

PR #3698, opened 2021-01-18
[CBO-1659] Migrate devise forms to GOVUK design system

naseberry:
#### What
Migrate devise forms to the GOVUK design system formbuilder

#### Ticket
[Migrate devise forms to GOVUK design system](https://dsdmoj.atlassian.net/browse/CBO-1659)

#### Why
Part of a larger change to:

- continue migration to GOVUK Frontend
- resolve accessibility issues identified in the 2020 DAC report
- reduce then remove deprecated code

PR #3751, opened 2021-02-11
Remove simple_form gem

jrmhaig:
#### What

Remove dependency on the `simple_form` gem.

#### Ticket

N/A

#### Why

The [`simple_form`](https://github.com/heartcombo/simple_form) gem provides helpers to simplify the creation of forms. However, the only place we are using it is in the messages form on a claim where the helpers are used to create two fields; a hidden field containing the claim id and a checkbox for 'Written reasons submitted'. This strikes me as an unnecessary extravagance.

#### How

The form is defined using `form_with` instead of `simple_form_for`.

The hidden field is created using `f.hidden_field :claim_id, value: @claim.id` instead of `f.input :claim_id, value: @claim.id, as: :hidden`

The checkbox is created with:

```ruby
  = f.check_box :written_reasons_submitted
  = f.label :written_reasons_submitted
```

instead of:

```ruby
  = f.input :written_reasons_submitted, as: :boolean
```

**Question:** How should the label be styled? I tried adding `class: 'govuk-label'` but that put it on a new line.


app/views/shared/_message_controls.html.haml
```diff
@@ -1,5 +1,5 @@
-= simple_form_for message, remote: !@claim.redeterminable?, authenticity_token: true do |f|
-  = f.input :claim_id, value: @claim.id, as: :hidden
+= form_with model: message, remote: !@claim.redeterminable?, authenticity_token: true do |f|
```
naseberry: @jrmhaig Thanks.

I had already migrated other forms that use the `simple_form` gem to use the `govuk_design_system_formbuilder` gem. Due to the changes in prioritised tickets, I was unable to continue this migration.

Could you migrate the message control form and form elements to use `govuk_design_system_formbuilder`

See EPIC: CBO-1639

jrmhaig: I can have a go. Can you point me towards a similar PR that I can use as a reference? Thanks.

naseberry: See: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3698

It was mostly the Devise related forms that used simple_form.

@jrmhaig

naseberry (approved):

PR #3752, opened 2021-02-11
Migrate to sentry-rails

kmahern:
#### What
Migrate sentry-raven gem to sentry-ruby gem

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1662

#### Why
Sentry-raven is now in maintenance mode and it is recommended to migrate to sentry-ruby/sentry-rails.

#### How
Updated Gemfile, config and associated spec.


jrmhaig (approved):
👍 Looks good

PR #3753, opened 2021-02-12
Bump rails 6.0 to to 6.1

jsugarman:
#### What
Bump rails 6.0 to to 6.1

#### Why
Maintenance and support

#### TODO:

 - [X] all tests passing
 - [X] handle deprecation warnings in app
 - [x] handle deprecation warnings in gov_uk_date_fields gem
 - [x] run through [this upgrade guide](https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-6-0-to-6-1.html)
 - [x] check Gemfile.lock on [RailsBump](https://railsbump.org/lockfiles/df623b960c62c5f88f58f029d9dcd4c771cb770e)
 - [x] manual test of external user flows 
 - [x] manual test of case worker user flows


Gemfile
```diff
@@ -77,7 +77,7 @@ group :development, :devunicorn, :test do
   gem 'binding_of_caller'
   gem 'byebug'
   gem 'guard-cucumber'
-  gem 'guard-jasmine', '~> 3.1'
+  # gem 'guard-jasmine', '~> 3.1'
```
naseberry: Is this gem no longer required?

jrmhaig: The is the gem that prevents upgrading to Rails 6.1. See https://github.com/guard/guard-jasmine/issues/206

If it is required then it must be downgraded to version 1.19.2, and that is only because it is the last version that did not restrict the version of `thor`.

config/initializers/permissions_policy.rb
```diff
@@ -0,0 +1,11 @@
...
+# information see https://developers.google.com/web/updates/2018/06/feature-policy
+#
+# Rails.application.config.permissions_policy do |f|
+#   f.camera      :none
+#   f.gyroscope   :none
+#   f.microphone  :none
+#   f.usb         :none
+#   f.fullscreen  :self
+#   f.payment     :self, "https://secure.example.com"
+# end
```
naseberry, line 11: The entire file is commented, do we need to commit it?

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,28 @@
+# Be sure to restart your server when you modify this file.
```
naseberry, line 1: The entire file is commented, do we need to commit it?

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
+# Be sure to restart your server when you modify this file.
```
naseberry, line 1: The entire file is commented, do we need to commit it?

config/initializers/permissions_policy.rb
jsugarman: 🤷 placeholder file that came out of `rails app:update` - i'd be inclined to keep it! what do you think?

config/initializers/content_security_policy.rb
jsugarman: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3753#discussion_r655476724

config/initializers/new_framework_defaults_6_1.rb
jsugarman: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3753#discussion_r655476724

Gemfile
colinbruce: Does anyone actually use Guard on your team?  I know a dev prior to me used it and I got in the habit to reduce overhead on running the test suite locally... having said that, I'm not sure I ever ran the jasmine tests via guard 🤷‍♂️

jsugarman: yeah, we have agreed to remove for the time being, especially as looking to move to cypress to some extent anyway

config/initializers/new_framework_defaults_6_1.rb
jsugarman: I'd be inclined to keep  - to [quote this blog](http://joshfrankel.me/blog/a-guide-for-upgrading-to-rails-6/)
>This is a new convention from Rails 5 that lists out specific feature flags that can be opted into at a later date allowing for an easier time migrating to new Rails’ versions. Keep these as either commented out or false. Enabling these should be places as individual stories within the post-upgrade category.

jsugarman: This [SO thread](https://stackoverflow.com/questions/50102584/upgrading-rails-what-am-i-to-do-with-new-framework-defaults-file/51275459) has a good description. I think we need to have a post upgrade ticket or tickets to flip these options 

The file itself contains comments for some like
>This change is not backwards compatible with earlier Rails versions.
It's best enabled when your entire app is migrated and stable on 6.1.

Gemfile
naseberry: I (we) can still run jasmine test using bundle instead of guard, so i'm happy to remove it, especially consider no one uses it.

config/initializers/new_framework_defaults_6_1.rb
naseberry: Got it, I will mark these comments as resolved.

db/schema.rb
```diff
@@ -672,8 +679,8 @@
   end
 
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
+  add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
   add_foreign_key "claims", "case_stages", name: "fk_claims_case_stage_id"
-  add_foreign_key "injection_attempts", "claims"
```
jrmhaig, line 676: This was added in this #3940 and I do not know why.

jrmhaig (approved):
I cannot see any blockers to this. I will test it on dev-lgfs.

PR #3754, opened 2021-02-12
[CBO-1695] Migrate layouts to govuk-frontend

naseberry:
#### What

Migrate layout page(s) from `govuk_template` to `govuk-frontend`

#### Ticket

[Migrate layout page(s) to the GOVUK Design System](https://dsdmoj.atlassian.net/browse/CBO-1695)

#### Why

- A prerequisite to implementing a cookie banner that is Data Protection Act 2018 compliant.
- Improving accessibility
- Actioning deprecated code
- Removing dead code


jessicabrentnall (approved):
I've reviewed at a very high level as someone who isn't overly familiar with rails. I'd be interested to learn from @jrmhaig's suggestion for improvements.

app/controllers/application_controller.rb
```diff
@@ -98,4 +98,24 @@ def track_visit(*args)
...
+
+  def disable_analytics
+    @disable_analytics = true
+  end
+
+  def disable_phase_banner
+    @disable_phase_banner = true
+  end
+
+  def disable_navigation
```
jrmhaig: I cannot see where `disable_navigation` is being used. Is it necessary?

app/controllers/application_controller.rb
```diff
@@ -98,4 +98,24 @@ def track_visit(*args)
   def suppress_hotline_link
     @suppress_contact_us_message = true
   end
+
+  def disable_analytics
```
jrmhaig, line 102: The only places where this is used it is used as `!@disable_anaytics`. Could it be `enable_analytics` instead (and then used as a `before_action` in the `external_users` controllers)?

wordshaker: This sounds more readable to me. I am happy to make these changes for Pete while he is off and give it a test

jessicabrentnall: I could be reading this wrong, but isn't it used in `statistics_controller.rb` and `widgets_controller.rb` without the bang?

app/controllers/application_controller.rb
jessicabrentnall: I could be reading this wrong, but isn't it used in `statistics_controller.rb` and `widgets_controller.rb`?

jrmhaig: The `before_action` calls this methods and so sets the variable. However, unless I am missing it, the actual variable isn't being used.

app/controllers/application_controller.rb
jrmhaig: The `before_action` calls this method and so sets the variable. If it is changed to `enable_analytics` then the `before_action`s in each of the controllers will need to be changed so that it is set to `true` in the relevant cases.

jrmhaig: Actually, I'm having second thoughts about this. The default (as it stands) is for analytics to be enabled and this is the case for all controllers, not just the five that are being changed here, so an `enable_analytics` action would have to be applied to all. It makes more sense for the setting to be 'disable'.

I'm currently looking at the `GoogleAnalytics::DataTracking` module to see if that is where this lives.

app/controllers/application_controller.rb
jrmhaig: The place where `@disable_navigation` was used was removed in this commit: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3754/commits/ea07759a12b6518702e62914daae831a6fcdff95

I'm wondering if this was a mistake or if it has been replaced by `@enable_breadcrumb`, which would be the case if you either see navigation or breadcrumbs.

jessicabrentnall: Oh! Peter did say something along those lines in the dev catchup. From what I recall he mentioned that it had been replaced by enable breadcrumb because it was a poorly named variable before?

app/controllers/application_controller.rb
jrmhaig: I've created a branch where I've put this flag inside `GoogleAnalytics::DataTracker`:

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/compare/1695-layout-page-migration...1695-layout-page-migration_disable_analytics

I didn't push it straight into this PR as I'm not totally familiar with how `GoogleAnalytics::DataTracking`, so this PR can go as it is and then that change can be reviewed later.

app/controllers/application_controller.rb
jrmhaig: I remember that too. I have removed it.

jrmhaig (approved):
Approved as is. I have a small suggestion, which can either go into this or be reviewed later:

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/compare/1695-layout-page-migration...1695-layout-page-migration_disable_analytics

app/views/layouts/_govuk_logo.html.haml
```diff
@@ -0,0 +1,7 @@
...
...9 2.8c-.2-.3-.5-.6-.9-.9L72.4 0H59.6l3.5 11.2c-.3.3-.6.5-.9.9l-8.8-2.8v13.1l8.8-4.7c.3.3.6.7.9.9l-5 15.4v.1c-.2.8-.4 1.6-.4 2.4 0 4.1 3.1 7.5 7 8.1h.2c.3 0 .7.1 1 .1.4 0 .7 0 1-.1h.2c4-.6 7.1-4.1 7.1-8.1 0-.8-.1-1.7-.4-2.4V34l-5.1-15.4c.4-.2.7-.6 1-.9zM66 92.8c16.9 0 32.8 1.1 47.1 3.2 4-16.9 8.9-26.7 14-33.5l-9.6-3.4c1 4.9 1.1 7.2 0 10.2-1.5-1.4-3-4.3-4.2-8.7L108.6 76c2.8-2 5-3.2 7.5-3.3-4.4 9.4-10 11.9-13.6 11.2-4.3-.8-6.3-4.6-5.6-7.9 1-4.7 5.7-5.9 8-.5 4.3-8.7-3-11.4-7.6-8.8 7.1-7.2 7.9-13.5 2.1-21.1-8 6.1-8.1 12.3-4.5 20.8-4.7-5.4-12.1-2.5-9.5 6.2 3.4-5.2 7.9-2 7.2 3.1-.6 4.3-6.4 7.8-13.5 7.2-10.3-.9-10.9-8-11.2-13.8 2.5-.5 7.1 1.8 11 7.3L80.2 60c-4.1 4.4-8 5.3-12.3 5.4 1.4-4.4 8-11.6 8-11.6H55.5s6.4 7.2 7.9 11.6c-4.2-.1-8-1-12.3-5.4l1.4 16.4c3.9-5.5 8.5-7.7 10.9-7.3-.3 5.8-.9 12.8-11.1 13.8-7.2.6-12.9-2.9-13.5-7.2-.7-5 3.8-8.3 7.1-3.1 2.7-8.7-4.6-11.6-9.4-6.2 3.7-8.5 3.6-14.7-4.6-20.8-5.8 7.6-5 13.9 2.2 21.1-4.7-2.6-11.9.1-7.7 8.8 2.3-5.5 7.1-4.2 8.1.5.7 3.3-1.3 7.1-5.7 7.9-3.5.7-9-1.8-13.5-11.2 2.5.1 4.7 1.3 7.5 3.3l-4.7-15.4c-1.2 4.4-2.7 7.2-4.3 8.7-1.1-3-.9-5.3 0-10.2l-9.5 3.4c5 6.9 9.9 16.7 14 33.5 14.8-2.1 30.8-3.2 47.7-3.2z', fill: 'currentColor', 'fill-rule': 'evenodd' }
```
jsugarman, line 4: not a blocker; is there any way to store this as an image asset.

naseberry: Apologies, we are following GDS recommended use of inline SVG.

PR #3764, opened 2021-02-19
Remove unused build_claim_in_state method

jrmhaig:
#### What

Remove the `build_claim_in_state` method.

#### Ticket

N/A

#### Why

Dead code. The `build_claim_in_state` method is a helper for specs but all uses of it were removed in this PR: #913


kmahern (approved):
👍

PR #3765, opened 2021-02-22
Remove simple_form

jrmhaig:
#### What

Remove `simple_form` from `Gemfile`.

#### Ticket

N/A

#### Why

This gem was remove (#3751) but it was re-introduced into `Gemfile` accidentally as a result of a merge conflict.


kmahern (approved):
👍

PR #3769, opened 2021-02-24
[CBO-1691] [AS2a]Add checksum to Paperclip attachments

jrmhaig:
#### What

Add checksum fields for evidence documents, message attachments and stats reports.

#### Ticket

[CCCD: migrate paperclip to active storage - migration scripts](https://dsdmoj.atlassian.net/browse/CBO-1691)

#### Why

In preparation for migrating assets from Paperclip to Active Storage we are creating checksums, required by Active Storage, for all existing assets. This PR will generate these checksums for newly created assets and a later PR will create a task to retrospectively add checksums for old assets.

#### How

The `CheckSummable` mixin is added to `Message`, `Document` and `Stats::StatsReport` to create the check sum when assets are created.

This is based on #3713.

### DIFF
--- a/spec/models/concerns/check_summable_spec.rb
+++ b/spec/models/concerns/check_summable_spec.rb
@@ -1,0 +1,95 @@
+require 'rails_helper'
+
+RSpec.shared_examples 'add_checksum failure cases' do
+  before { stub_const('TestClass', test_class) }
+
+  it 'does not raise an error' do
+    expect { add_checksum }.not_to raise_error
+  end
+
+  it 'logs the error' do
+    allow(LogStuff).to receive(:warn)
+    add_checksum
+    expect(LogStuff).to have_received(:warn)
+      .with(hash_including(class: 'TestClass', action: 'add_checksum'))
+  end
+end
+
+RSpec.describe CheckSummable do
+  let(:test_object) { test_class.new }
+  let(:checksum) { Digest::MD5.new.tap { |digest| digest << 'Test document' }.base64digest }
+  let(:io) { StringIO.new('Test document') }
+  let(:test_class) do
+    Class.new do
+      include CheckSummable
+
+      attr_accessor :as_document_checksum
+
+      def document
+        @document ||= OpenStruct.new(path: 'example.pdf')
+      end
+    end
+  end
+
+  describe '#calculate_checksum' do
+    subject(:calculate_checksum) { test_object.calculate_checksum(io) }
+
+    context 'with a valid io stream' do
+      it { is_expected.to eq checksum }
+
+      it 'rewinds the io stream' do
+        expect { calculate_checksum }.not_to change(io, :readline)
+      end
+    end
+
+    context 'with a nil io stream' do
+      let(:io) { nil }
+
+      it 'is nil if the input stream is nil' do
+        expect(calculate_checksum).to be_nil
+      end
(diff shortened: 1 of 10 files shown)


spec/models/concerns/check_summable_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'rails_helper'
+
+RSpec.describe CheckSummable do
+  let(:test_object) { TestClass.new }
+  let(:checksum) { Digest::MD5.new.tap { |digest| digest << 'Test document' }.base64digest }
+  let(:io) { StringIO.new('Test document') }
+
+  before do
+    test_class = Class.new do
```
jsugarman: nit - you may be able to use a let for the class as well

spec/models/concerns/check_summable_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      it 'returns a base64digest checksum for an IO stream' do
+        expect(calculate_checksum).to eq checksum
+      end
+
+      it 'rewinds the io stream' do
+        expect { calculate_checksum }.not_to change(io, :readline)
+      end
+    end
+
+    context 'with a nil input stream' do
```
jsugarman: nit
```suggestion
    context 'with a nil io stream' do
```

spec/models/concerns/check_summable_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    stub_const('TestClass', test_class)
+  end
+
+  describe '#calculate_checksum' do
+    subject(:calculate_checksum) { test_object.calculate_checksum(io) }
+
+    context 'with a valid io stream' do
+      it 'returns a base64digest checksum for an IO stream' do
+        expect(calculate_checksum).to eq checksum
+      end
```
jsugarman:
```suggestion
      it { is_expected.to eq checksum }
```

jsugarman: happy for you to keep as is, more just example

spec/models/concerns/check_summable_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      before do
+        allow(registry).to receive(:for).with(test_object.document).and_return(io)
+      end
+
+      it 'sets the checksum' do
+        expect { add_checksum }.to change(test_object, :as_document_checksum).to checksum
+      end
+    end
+
+    shared_examples 'CheckSummable.add_checksum failure cases' do
```
jsugarman: nit - I prefer shared examples at top of

```ruby
    RSpec.shared_examples 'CheckSummable.add_checksum failure cases' do
    end
    
    RSpec.describe CheckSummable do
    ...
    end
```

jsugarman: personally don't think we need checksummable here 
```suggestion
    shared_examples 'add_checksum failure cases' do
```

spec/models/message_spec.rb
```diff
@@ -86,7 +86,7 @@
     end
   end
 
-  context 'automotic state change of claim on message creation' do
+  describe 'automatic state change of claim on message creation' do
```
jsugarman: possibly use `context 'when automatically...'`

jsugarman (approved):
Looks good 👍

PR #3775, opened 2021-03-01
Rename some rake tasks and their methods

jsugarman:
Rename some tasks and their methods for clarity as well as add some prompting
for input/verification (optional) and some out put.


jrmhaig (approved):
Looks good

PR #3783, opened 2021-03-03
[CBO-1683] [AS3a] Use Active Storage for stats reports

jrmhaig:
#### What

Switch over to using Active Storage for stats reports.

#### Ticket

#### Why

We need to switch over to using Active Storage everywhere we have documents in CCCD. The simplest of these, and the place with the least risk, is for the stats reports and so this is being done first.

#### How

| How to ... | Paperclip | Active Storage |
|---|---|---|
| Configure an attachment for reports | `has_attached_file :document` | `has_one_attached :document` |
| Attach a report file | `update(...)` | `document.attach(...)` |
| Download a report | Fetch file to the server and then stream with `send_data` | Redirect to `document.blob.service_url` |

This is based on #3722, which is now closed.

### DIFF
--- a/app/controllers/case_workers/admin/management_information_controller.rb
+++ b/app/controllers/case_workers/admin/management_information_controller.rb
@@ -1,6 +1,8 @@
 require 'csv'
 
 class CaseWorkers::Admin::ManagementInformationController < CaseWorkers::Admin::ApplicationController
+  include ActiveStorage::SetCurrent
+
   skip_load_and_authorize_resource only: %i[index download generate]
   before_action -> { authorize! :view, :management_information }, only: %i[index download generate]
   before_action :validate_report_type, only: %i[download generate]
@@ -14,14 +16,13 @@ class CaseWorkers::Admin::ManagementInformationController < CaseWorkers::Admin::
   def download
     log_download_start
     record = Stats::StatsReport.most_recent_by_type(params[:report_type])
-    if record.document?
-      data = open(record.document_url).read
-      content_type = record.document_content_type
+
+    if record.document.attached?
+      # TODO: Change service_url to url when moving to Rails 6.1
+      redirect_to record.document.blob.service_url(disposition: 'attachment')
     else
-      data = record.report
-      content_type = 'text/csv'
+      redirect_to case_workers_admin_management_information_url, alert: t('.missing_report')
     end
-    send_data data, filename: record.download_filename, type: content_type
   end
 
   def generate
--- a/app/models/stats/stats_report.rb
+++ b/app/models/stats/stats_report.rb
@@ -64,6 +64,7 @@ module Stats
         document_file_name: filename,
         document_file_size: document.byte_size,
         document_content_type: report_result.content_type,
+        document_updated_at: document.created_at,
         as_document_checksum: document.checksum
       )
     rescue StandardError => e
--- a/app/models/stats/stats_report.rb
+++ b/app/models/stats/stats_report.rb
@@ -51,15 +51,26 @@ module Stats
       create!(report_name: report_name, status: 'started', started_at: Time.now)
     end
 
+    # High ABC Size due to setting Paperclip fields for possible revert.
+    # This 'rubocop:disable' can be removed when the Paperclip fields are removed.
+    # rubocop:disable Metrics/AbcSize
     def write_report(report_result)
       filename = "#{report_name}_#{started_at.to_s(:number)}.#{report_result.format}"
       log(:info, :write_report, "Writing report #{report_name} to #{filename}")
       document.attach(io: report_result.io, filename: filename, content_type: report_result.content_type)
-      update(status: 'completed', completed_at: Time.zone.now)
+      update(
+        status: 'completed',
+        completed_at: Time.zone.now,
+        document_file_name: filename,
+        document_file_size: document.byte_size,
+        document_content_type: report_result.content_type,
+        as_document_checksum: document.checksum
+      )
     rescue StandardError => e
       log(:error, :write_report, "error writing report #{report_name}...", e)
       raise
     end
+    # rubocop:enable Metrics/AbcSize
 
     def write_error(report_contents)
       update(report: report_contents, status: 'error', completed_at: nil)
--- a/app/models/stats/stats_report.rb
+++ b/app/models/stats/stats_report.rb
@@ -28,9 +28,7 @@ module Stats
     scope :management_information, -> { not_errored.where(report_name: 'management_information') }
     scope :provisional_assessment, -> { not_errored.where(report_name: 'provisional_assessment') }
 
-    has_attached_file :document, s3_headers.merge(REPORTS_STORAGE_OPTIONS)
-
-    validates_attachment_content_type :document, content_type: ['text/csv']
+    has_one_attached :document
 
     def self.clean_up(report_name)
       destroy_reports_older_than(report_name, 1.month.ago)
@@ -54,16 +52,10 @@ module Stats
     end
 
     def write_report(report_result)
-      log(:info, :write_report, "Writing report #{report_name} to DB...")
-      io = StringIO.new(report_result.content)
-      update(
-        document: io,
-        document_file_name: "#{report_name}_#{started_at.to_s(:number)}.#{report_result.format}",
-        document_content_type: report_result.content_type,
-        as_document_checksum: calculate_checksum(io),
-        status: 'completed',
-        completed_at: Time.now
-      )
+      filename = "#{report_name}_#{started_at.to_s(:number)}.#{report_result.format}"
+      log(:info, :write_report, "Writing report #{report_name} to #{filename}")
+      document.attach(io: StringIO.new(report_result.content), filename: filename)
+      update(status: 'completed', completed_at: Time.zone.now)
     rescue StandardError => e
       log(:error, :write_report, "error writing report #{report_name}...", e)
       raise
--- a/app/models/stats/stats_report.rb
+++ b/app/models/stats/stats_report.rb
@@ -54,7 +54,7 @@ module Stats
     def write_report(report_result)
       filename = "#{report_name}_#{started_at.to_s(:number)}.#{report_result.format}"
       log(:info, :write_report, "Writing report #{report_name} to #{filename}")
-      document.attach(io: StringIO.new(report_result.content), filename: filename)
+      document.attach(io: report_result.io, filename: filename, content_type: report_result.content_type)
       update(status: 'completed', completed_at: Time.zone.now)
     rescue StandardError => e
       log(:error, :write_report, "error writing report #{report_name}...", e)
--- a/config/initializers/active_storage.rb
+++ b/config/initializers/active_storage.rb
@@ -1,0 +1,1 @@
+Rails.application.config.active_storage.service_urls_expire_in = 5.minutes # default
--- a/spec/controllers/case_workers/admin/management_information_controller_spec.rb
+++ b/spec/controllers/case_workers/admin/management_information_controller_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe CaseWorkers::Admin::ManagementInformationController, type: :contr
         end
 
         it 'redirects to the service url of the document' do
-          expect(response.location).to match %{#{disk_service_prefix}/.*/#{filename}}
+          expect(response.location).to match %(#{disk_service_prefix}/.*/#{filename})
         end
       end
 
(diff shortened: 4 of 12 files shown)


app/models/stats/stats_report.rb
```diff
@@ -28,9 +28,7 @@ class StatsReport < ApplicationRecord
     scope :management_information, -> { not_errored.where(report_name: 'management_information') }
     scope :provisional_assessment, -> { not_errored.where(report_name: 'provisional_assessment') }
 
-    has_attached_file :document, s3_headers.merge(REPORTS_STORAGE_OPTIONS)
-
-    validates_attachment_content_type :document, content_type: ['text/csv']
```
jsugarman, line 33: not so important for reports but for messages and [evidence] documents we would need to have a validation on the content-type.  Is that possible? and if so I'd say it is worth doing it for reports too, if for no other reason than to prove we can later for documents etcs

app/models/stats/stats_report.rb
```diff
@@ -54,16 +52,10 @@ def self.record_start(report_name)
...
-        document_file_name: "#{report_name}_#{started_at.to_s(:number)}.#{report_result.format}",
-        document_content_type: report_result.content_type,
-        as_document_checksum: calculate_checksum(io),
-        status: 'completed',
-        completed_at: Time.now
-      )
+      filename = "#{report_name}_#{started_at.to_s(:number)}.#{report_result.format}"
+      log(:info, :write_report, "Writing report #{report_name} to #{filename}")
+      document.attach(io: StringIO.new(report_result.content), filename: filename)
+      update(status: 'completed', completed_at: Time.zone.now)
```
jsugarman: 🎖️ `Time.zone.now`

app/models/stats/stats_report.rb
```diff
@@ -54,16 +52,10 @@ def self.record_start(report_name)
...
     rescue StandardError => e
       log(:error, :write_report, "error writing report #{report_name}...", e)
```
jsugarman, line 71: does log call LogStuff? Ideally we want any errors available in kibana in Log Stash format (which log stuff does)

spec/controllers/case_workers/admin/management_information_controller_spec.rb
```diff
@@ -26,65 +23,50 @@
...
-              document: document,
-              document_file_name: "#{report_type}_#{Time.now.to_s(:number)}.csv",
-              document_content_type: 'text/csv'
-            )
-          }
-
-          before do
-            get :download, params: { report_type: report_type }
-          end
+        it 'redirects to the service url of the document' do
```
jsugarman: Is it possible to do something more idiomatic like:

```
let(:stats_report) do
  create(:stats_report, report_name: report_type).tap do |report|
    report.document.attach(io: StringIO.new, filename: filename)
  end
end

before { download }
...
expect(response).to redirect_to(stats_report.document.blob.service_url)
```

spec/models/stats/stats_report_spec.rb
```diff
@@ -123,47 +123,6 @@
...
-
-      context 'when the document storage is S3' do
-        let(:options) { { storage: :s3 } }
-
-        before do
-          original_options = report.document.options
-          allow(report.document).to receive(:options).and_return(original_options.merge(options))
-        end
-
-        it 'returns the an expiring url for the document' do
```
jsugarman, line 177: Is there anyway to emulate this test for an expired URL? - I assume, perhaps wrongly, that it is

spec/services/stats/stats_report_generator_spec.rb
```diff
@@ -2,88 +2,85 @@
...
+          args = ['call_failed.stats_report', id: record.id, name: report_type, error: instance_of(StandardError)]
+          expect(ActiveSupport::Notifications).to have_received(:instrument).with(*args)
         end
+      end
 
-        context 'and when the error notifications are disabled' do
-          before do
-            allow(Settings).to receive(:notify_report_errors).and_return(false)
-          end
+      context 'when when the error notifications are disabled' do
```
jsugarman:
```suggestion
      context 'when the error notifications are disabled' do
```

spec/services/stats/stats_report_generator_spec.rb
```diff
@@ -2,88 +2,85 @@
...
+        before { allow(Settings).to receive(:notify_report_errors).and_return(false) }
 
-          it 'does not sends an error notification' do
-            expect(ActiveSupport::Notifications).not_to receive(:instrument)
-            described_class.call(report_type) rescue nil
-          end
+        it 'does not sends an error notification' do
+          expect(ActiveSupport::Notifications).not_to receive(:instrument)
+          call_report_generator rescue nil
```
jsugarman:
```suggestion
          call_report_generator rescue nil
          expect(ActiveSupport::Notifications).not_to have_received(:instrument)
```

app/controllers/case_workers/admin/management_information_controller.rb
```diff
@@ -14,14 +16,13 @@ def index
   def download
     log_download_start
     record = Stats::StatsReport.most_recent_by_type(params[:report_type])
-    if record.document?
-      data = open(record.document_url).read
-      content_type = record.document_content_type
+
+    if record.document.attached?
+      # TODO: Change service_url to url when moving to Rails 6.1
+      redirect_to record.document.blob.service_url(disposition: 'attachment')
```
jsugarman, line 22: are we not using an [expiring url like this perhaps](https://api.rubyonrails.org/v6.1.3/classes/ActiveStorage/Blob.html#method-i-service_url)

jsugarman (commented):
Some possible test amendments and extensions.

What are we doing about:

1. content type validation in AS
2. url expiry - was [looking at this](https://api.rubyonrails.org/v6.1.3/classes/ActiveStorage/Blob.html#method-i-service_url)

app/models/stats/stats_report.rb
jrmhaig: I removed the validation because the file is being created in the app. Potentially validation could fail if something goes wrong with the CSV creation, indicating a bug. I can put it back in but I'm not sure of the value.

There is (and will remain) validation for uploads and there are tests for this.

app/controllers/case_workers/admin/management_information_controller.rb
jrmhaig: URLs have a 5 minute expiry by default but perhaps it should be explicitly?

app/models/stats/stats_report.rb
jrmhaig: Yes. `log` is a private method defined at the bottom of the file.

spec/controllers/case_workers/admin/management_information_controller_spec.rb
jrmhaig: `service_url` creates a new expiring url each time so that test will fail.

I originally had something using `allow_any_instance_of(ActiveStorage::Blob).to receive(:service_url)` but I discovered that `allow_any_instance_of` is deprecated. I'll have another go using `allow(...).to receive(...)`.

spec/models/stats/stats_report_spec.rb
jrmhaig: This is what I was attempting at line 41 of spec/controllers/case_workers/admin/management_information_controller_spec.rb

app/controllers/case_workers/admin/management_information_controller.rb
jsugarman: I certainly prefer explicit myself. We can configure this at the app level.

```
# config/initializers/active_storage.rb
Rails.application.config.active_storage.service_urls_expire_in = 5.minutes # default
```

spec/controllers/case_workers/admin/management_information_controller_spec.rb
jsugarman: ok, makes sense.

could try 
```
allow(stats_report).to receive_message_chain(:document, :blob, :service_url).and_return('my-expiring-object-url')
```

but perhaps a regex matcher is better after all and perhaps with a separate test to match any expiry limit of on the url. maybe excessive though?! 
```
expect(response.location).to match(%r{expiry=blah})
```

jrmhaig: I found that `receive_message_chain` gets flagged by Rubocop (see [Cop/RSpec/MessageChain](https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/MessageChain)) and, searching around, it does seem to be discouraged (see [here](https://stackoverflow.com/questions/53648078/rspec-rubocop-why-receive-message-chain-is-a-code-smell)). I have achieved the same with:

```ruby
allow(Stats::StatsReport).to receive(:most_recent_by_type).and_return(stats_report)
allow(stats_report.document.blob).to receive(:service_url).and_return(test_url)
```

I'm not entirely convinced of this but it works.

jsugarman: as discussed, testing AS internals is probably overkill

spec/models/stats/stats_report_spec.rb
jsugarman: > as discussed, testing AS internals is probably overkill

config/initializers/active_storage.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.active_storage.service_urls_expire_in = 5.minutes # default
```
jsugarman, line 1: 💯

spec/services/stats/stats_report_generator_spec.rb
```diff
@@ -2,88 +2,86 @@
...
+        expect {
+          call_report_generator rescue nil
+        }.to change { Stats::StatsReport.where(report_name: report_type).errored.count }.from(0).to(1)
+      end
 
-        it 'creates a new report marked as errored' do
-          expect {
-            described_class.call(report_type) rescue nil
-          }.to change { Stats::StatsReport.where(report_name: report_type).errored.count }.from(0).to(1)
+      context 'when when the error notifications are enabled' do
```
jsugarman:
```suggestion
      context 'when the error notifications are enabled' do
```

spec/services/stats/stats_report_generator_spec.rb
```diff
@@ -2,88 +2,86 @@
...
-            allow(Settings).to receive(:notify_report_errors).and_return(false)
-          end
+      context 'when the error notifications are disabled' do
+        before { allow(Settings).to receive(:notify_report_errors).and_return(false) }
 
-          it 'does not sends an error notification' do
-            expect(ActiveSupport::Notifications).not_to receive(:instrument)
-            described_class.call(report_type) rescue nil
-          end
+        it 'does not sends an error notification' do
```
jsugarman:
```suggestion
        it 'does not send an error notification' do
```

spec/services/stats/stats_report_generator_spec.rb
```diff
@@ -2,88 +2,86 @@
...
+          allow(ActiveSupport::Notifications).to receive(:instrument)
+          call_report_generator rescue nil
```
jsugarman: nit - can put both these lines in the before/setup block i believe

jsugarman (commented):
Just minor nits from a code perspective, and a commit tidy prior to merge. 

**However**
This also needs to go in to master only after the manual testing of active storage migration of `stats_reports` on dev-lgfs and check that:
 - a report generated before AS migration is still accessible after AS migration
 - a report generated after AS migration is accessible.
 
 Once AS2b is in we/I/whoever can use dev-lgfs to test the above

jrmhaig:
> This also needs to go in to master only after the manual testing of active storage migration of `stats_reports` on dev-lgfs and check that:
> 
> * a report generated before AS migration is still accessible after AS migration
> * a report generated after AS migration is accessible.

I have rebased and deployed to dev-lgfs. After running the `storage:migrate[stats_reports]` task I confirmed that I was able to download reports that had previously been created as Paperclip attachments. I then created new reports for 'Provisional assessment' and 'Rejections/Refusals' and was able to download these too.

**Note:** I did not create a new 'Management information' report as I have found that the worker fails with this one due to running out of memory. The report that is created by a cron job at 3am appears to succeed, though, so I do not know what the difference is.

jsugarman:
> > This also needs to go in to master only after the manual testing of active storage migration of `stats_reports` on dev-lgfs and check that:
> > 
> > * a report generated before AS migration is still accessible after AS migration
> > * a report generated after AS migration is accessible.
> 
> I have rebased and deployed to dev-lgfs. After running the `storage:migrate[stats_reports]` task I confirmed that I was able to download reports that had previously been created as Paperclip attachments. I then created new reports for 'Provisional assessment' and 'Rejections/Refusals' and was able to download these too.
> 
> **Note:** I did not create a new 'Management information' report as I have found that the worker fails with this one due to running out of memory. The report that is created by a cron job at 3am appears to succeed, though, so I do not know what the difference is.

great, but oh dear on the report regeneration.

Did you try the rollback too?

jsugarman (approved):
👍  Looks good, maybe we can pair on squashing this up but no biggee

PR #3785, opened 2021-03-04
Enable/disable analytics in GoogleAnalytics::DataTracking

jrmhaig:
#### What

Use `GoogleAnalytics::DataTracking` to disable analytics.

#### Ticket

[Migrate layout page(s) to the GOVUK Design System](https://dsdmoj.atlassian.net/browse/CBO-1695)

#### Why

This is a small suggestion for the main PR #3754.

The `GoogleAnalytics::DataTracking` class already determines if analytics is enabled in some cases. If there needs to be a switch to turn it on or off at will then it makes sense that it should go here.

#### How

Create a new attribute `active` on the `GoogleAnalytics::DataTracking` class that allows analytics to be turned on or off at will. This attribute is set to `true` by default in `config/initializers/analytics.rb` and can be turned off with `ApplicationController#disable_analytics`.

### DIFF
--- a/spec/models/google_analytics/data_tracking_spec.rb
+++ b/spec/models/google_analytics/data_tracking_spec.rb
@@ -2,22 +2,36 @@ require 'rails_helper'
 
 module GoogleAnalytics
   describe DataTracking do
+    before do
+      allow(described_class).to receive(:adapter).and_return('Adapter')
+      allow(Rails).to receive(:env).and_return('production'.inquiry)
+      described_class.active = true
+    end
+
     describe '#enabled?' do
       subject { described_class.enabled? }
-      before do
-        allow(Rails).to receive(:env).and_return('production'.inquiry)
+
+      context 'when host is staging' do
+        before { allow(RailsHost).to receive(:env).and_return('staging') }
+
+        it { is_expected.to be_truthy }
+
+        context 'when active is false' do
+          before { described_class.active = false }
+
+          it { is_expected.to be_falsey }
+        end
       end
 
-      context 'with an adapter set' do
-        before do
-          allow(described_class).to receive(:adapter).and_return('Adapter')
-        end
+      context 'when host is production' do
+        before { allow(RailsHost).to receive(:env).and_return('production') }
 
-        %w(staging production).each do |host|
-          it "returns true when host is #{host}" do
-            allow(RailsHost).to receive(:env).and_return(host)
-            expect(described_class.enabled?).to be_truthy
-          end
+        it { is_expected.to be_truthy }
+
+        context 'when active is false' do
+          before { described_class.active = false }
+
+          it { is_expected.to be_falsey }
         end
       end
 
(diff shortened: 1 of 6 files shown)


jsugarman (changes requested):
Looks good @jrmhaig  but as @naseberry mentioned, in slack, we should extend the existing unit testing to cover the `enabled?` changes and the new `.active` method

jrmhaig:
> Looks good @jrmhaig but as @naseberry mentioned, in slack, we should extend the existing unit testing to cover the `enabled?` changes and the new `.active` method

I have added to `spec/models/google_analytics/data_tracking_spec.rb`

spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -2,22 +2,36 @@
 
 module GoogleAnalytics
   describe DataTracking do
+    before do
+      allow(described_class).to receive(:adapter).and_return('Adapter')
+      allow(Rails).to receive(:env).and_return('production'.inquiry)
```
jsugarman, line 7: 👍 ooh, did not know about `inquiry`.

spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -40,53 +54,47 @@ module GoogleAnalytics
       end
     end
 
-    describe 'type methods' do
-      before do
-        allow(described_class).to receive(:adapter).and_return('Adapter')
-        allow(Rails).to receive(:env).and_return('production'.inquiry)
-      end
+    describe '#tag_manager?' do
```
jsugarman, line 57: 🎖️ much better than inappropriate describe block description

spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -40,53 +54,47 @@ module GoogleAnalytics
...
+      subject(:tag_manager) { described_class.tag_manager? }
 
-      context '#tag_manager?' do
-        subject(:tag_manager) { described_class.tag_manager? }
-        context 'when the adapter is google_tag_manager' do
-          before { allow(described_class).to receive(:adapter_name).and_return(:gtm) }
+      context 'when the adapter is google_tag_manager' do
```
jsugarman, line 60: 🎖️

spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -2,22 +2,36 @@
...
       subject { described_class.enabled? }
-      before do
-        allow(Rails).to receive(:env).and_return('production'.inquiry)
-      end
 
-      context 'with an adapter set' do
-        before do
-          allow(described_class).to receive(:adapter).and_return('Adapter')
+      context 'when host is staging' do
+        before { allow(RailsHost).to receive(:env).and_return('staging') }
```
jsugarman, line 15: need to check this ENV['ENV'] is Rails.env....

spec/models/google_analytics/data_tracking_spec.rb
jrmhaig: I don't really know what it is doing and Rubocop doesn't like it. I just left it in - that line is copied from elsewhere.

spec/models/google_analytics/data_tracking_spec.rb
jrmhaig: This is from lines 16-21 in the old version. I just expanded the loop and put the test setup in a `before` to make it more readable but I can change the setup if that is desired.

spec/models/google_analytics/data_tracking_spec.rb
jrmhaig: I actually deleted this `describe` and changed the `context` on what was line 49 to a `describe`. Most of this section is just whitespace changes after removing one level of nesting.

spec/models/google_analytics/data_tracking_spec.rb
jsugarman: ok, this is just a refactor of the existing tests I see now, breaking the loop down. The diff is a bit confusing.

jsugarman (approved):
👍

jsugarman (approved):
👍

PR #3786, opened 2021-03-04
Enable writing to s3 for hosted envs

jsugarman:
To load test the adding of checksums we need
to create a production like amount of s3 hosted
files.

These files should also be deleted once we are done with
them to avoid costs.

### DIFF
--- a/app/models/dummy_document.rb
+++ b/app/models/dummy_document.rb
@@ -1,21 +1,21 @@
-  # DummyDocument: class used to shadow a real
-  # Document for dummy paperclip s3 file creation.
-  #
-  # The "Shadow" Document object does not include callbacks
-  # that would otherwise generate a pdf from the original file
-  # when it has changed and is saved. This would take a long time and
-  # not needed for dummy s3 file creation
-  #
-  # It also prevents validation of the file content and type so
-  # random bytes can be used to quickly create the dummy file content
-  #
-  class DummyDocument < ApplicationRecord
-    include S3Headers
-    include CheckSummable
+# DummyDocument: class used to shadow a real
+# Document for dummy paperclip s3 file creation.
+#
+# The "Shadow" Document object does not include callbacks
+# that would otherwise generate a pdf from the original file
+# when it has changed and is saved. This would take a long time and
+# not needed for dummy s3 file creation
+#
+# It also prevents validation of the file content and type so
+# random bytes can be used to quickly create the dummy file content
+#
+class DummyDocument < ApplicationRecord
+  include S3Headers
+  include CheckSummable
 
-    self.table_name = 'documents'
-    has_attached_file :converted_preview_document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)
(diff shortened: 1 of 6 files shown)


lib/tasks/rake_helpers/storage.rb
```diff
@@ -105,16 +105,74 @@ def self.create_dummy_paperclip_files_for(model)
...
+  # Deleting filesystem/s3 files
+  # doc.document.destroy will delete the doc and its meta data in the doc model (not want we want)
+  # doc.document.clear wil delete the paperclip attachment file, without changing the meta data, unless you save
+  # the model object immediatley afterwards.
+
+  # NOTE: for the Document model the `doc.document.clear` message removes the "folder" (tested locally)
+  # which includes the converted_preview_document object too. However, calling `.clear` on a non-existant
+  # attachment does not raise an error.
+  #
+  def self.clear_dummy_paperclip_files_for(model)
```
jsugarman, line 158: I have tried out and am able to delete the s3 objects directly using code similar to below. this is likely to be a lot quicker.

```
# delete all objects with reports as a prefix
s3_bucket = S3Bucket.new('does-not-matter-if-on-host')
s3_bucket.list('reports').map { |item| item.delete }
```

The disadvantage of this approach is that there is no simple way to distinguish message attachments from evidence document attachments as their s3 keys are both prefixed with `document/...`.

app/models/dummy_document.rb
```diff
@@ -0,0 +1,21 @@
...
+  class DummyDocument < ApplicationRecord
+    include S3Headers
+    include CheckSummable
+
+    self.table_name = 'documents'
+    has_attached_file :converted_preview_document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)
+    has_attached_file :document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)
+
+    do_not_validate_attachment_file_type :document
+  end
```
jrmhaig: Rubocop is complaining because the whole of this file is indented.

```suggestion
# DummyDocument: class used to shadow a real
# Document for dummy paperclip s3 file creation.
#
# The "Shadow" Document object does not include callbacks
# that would otherwise generate a pdf from the original file
# when it has changed and is saved. This would take a long time and
# not needed for dummy s3 file creation
#
# It also prevents validation of the file content and type so
# random bytes can be used to quickly create the dummy file content
#
class DummyDocument < ApplicationRecord
  include S3Headers
  include CheckSummable

  self.table_name = 'documents'
  has_attached_file :converted_preview_document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)
  has_attached_file :document, s3_headers.merge(PAPERCLIP_STORAGE_OPTIONS)

  do_not_validate_attachment_file_type :document
end
```

jrmhaig (changes requested):
Looks OK. Just one change required to fix for Rubocop.

app/models/dummy_document.rb
jsugarman: oops, ta

jrmhaig (approved):

PR #3790, opened 2021-03-08
Add RACK_ENV var to all environments

jsugarman:
#### What
Add RACK_ENV var to all environments

#### Why
This fixes an error whereby stacktraces
get displayed on production instances.

Longer term we need to understand if we
can remove this line from `application.rb`

```
config.exceptions_app = self.routes
```

see this SO thread for more:

[Rack::ShowExceptions enabled in production Rails 4.0.4](https://stackoverflow.com/questions/27113841/rackshowexceptions-enabled-in-production-rails-4-0-4)

### DIFF
--- a/kubernetes_deploy/api-sandbox/app-config.yaml
+++ b/kubernetes_deploy/api-sandbox/app-config.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: cccd-api-sandbox
 data:
   ENV: 'api-sandbox'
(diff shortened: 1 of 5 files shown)


jrmhaig (approved):
:+1:

jrmhaig:
The Jira ticket I created for this is CBO-1734

kmahern (approved):
👍

PR #3792, opened 2021-03-09
Set diskspace exceeded prom alert to 2Gb on CCCD prod

kmahern:
#### What
Set the prometheus diskspace threshold exceeded prometheus alert to 2GB (it is currently 300MB)

#### Ticket
N/A

#### Why
This alert is currently very noisy/therefore not useful. The current threshold of 300MB is being regularly exceeded and this not seem to indicate that there is a problem.

#### How
Once approved this will need to be applied with the following (until we move this to the Cloud Platforms environment repo)

kubectl apply -f kubernetes_deploy/production/prometheus-custom-rules.yaml -n cccd-production


jsugarman (approved):
👍 LGTM

PR #3793, opened 2021-03-09
Extract disabling methods for view-only layouts to mixin

jsugarman:
#### What
Extract disabling methods for view-only layouts to mixin

#### Why
Fixes a lot of [sentry errors such as this](https://sentry.io/organizations/ministryofjustice/issues/2261816305/?project=5436820&query=is%3Aunresolved) which result from the disable_analytics 
and other disable... methods not being in the inheritance tree for widgets controller.

The inheritance for the controllers in question
is odd and needs looking at, so have removed
to mixin and included as a quick fix.

Needs testing and automated tests.


naseberry (approved):
LGTM

jrmhaig (approved):
:+1:

PR #3801, opened 2021-03-10
Increase threshold for DiskThreshold prometheus alerts

jsugarman:
#### What
Increase threshold for DiskThreshold prometheus alerts
and add alerts to api-sandbox too.

#### Why
These are rather noisy and desensitising devs
to more important alerts. It is not a big
problem to be using some disk space since
file upload will create tmp files for
upload and download.

Adding prometheus alerting to api-sandbox as
this appears to have been missed, but as a user
facing app environment for CCCD (vendors use it)
it should have alerting on it as well.


kmahern (approved):
Looks good 👍

jsugarman:
These changes have been applied manually to each environment now

PR #3806, opened 2021-03-12
Unroll loops of tests

jrmhaig:
#### What

Some tests were defined by iterating over lists like:

```ruby
array.each do |param|
  it "Does something for #{param}" do
    ...
  end
end
```

This PR expands these arrays.

#### Ticket

N/A

#### Why

For clarity, the details of each test can now be seen in place. Also, there were previously no negative tests for advocate categories that should not be valid.

#### How

The body of the tests have been extracted to a shared example that can be called together with the relevant details as:

```ruby
it_behaves_like 'claim is valid for advocate category', 'QC', :with_scheme_nine_offence
```

--------

#### TODO (?)

 - [x] move shared examples into `spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb`
 - [x] expand shared examples so that it can also be used in the 'advocate category validations' examples


spec/support/shared_examples/validators/shared_examples_for_advocate_claim_validators.rb
```diff
@@ -1,53 +1,121 @@
...
+    subject(:claim) { create(options[:factory], options[:trait]) }
 
-  before do
-    claim.form_step = options[:form_step]
-  end
+    before { claim.form_step = options[:form_step] }
+
+    it 'is valid for QC' do
+      claim.advocate_category = 'QC'
+      should_not_error(claim, :advocate_category)
```
jrmhaig, line 9: The `should_not_error` method calls `claim.valid?` and then checks that `claim.errors[:advocate_category]` is empty (similarly in `should_error_with`). I would think that it is sufficient just to test `claim.valid?`. These could tests could then be replaced with:

```ruby
context 'with QC advocate category' do
  before { claim.advocate_category = 'QC' }
  it { is_expected.to be_valid }
end
(etc)
```

How do people feel about this?

jsugarman: The `should_not_error` method is expliclty testing if `claim.errors[:advocate_category]` is populated, regardless of whether the claim is valid or not. This is, from memory, because the claim may not be valid (due to factory issues and claim creation setup. It was done like this to enable testing the validity of a specific field, rather than the object itself. 

Even if all tests pass now, this change would mean that changes to factories or setup unrelated to these tests could cause them to break, which i believe is why it was done this way

kmahern: Consider closing if no-one has a chance to review in the next few days

PR #3807, opened 2021-03-12
Anonymise Active Storage blobs

jrmhaig:
#### What

Anonymise `ActiveStorage::Blob` when creating database dumps.

#### Ticket

N/A

#### Why

Active Storage blobs include the filename of messages and documents so they
are anonymised. Additionally, the keys of attachments that have been migrated
from Paperclip also includes the filename so this is modified too.

Stats reports do not need anonymising.

#### How

Add `active_storage_blobs` to the list of excluded tables and add the `db:dump:active_storage_blobs` task.

### DIFF
--- a/lib/tasks/dump.rake
+++ b/lib/tasks/dump.rake
@@ -203,6 +203,25 @@ namespace :db do
       end
     end
 
+    desc 'Export anonymised active storage blob data'
+    task :active_storage_blobs, [:file] => :environment do |task, args|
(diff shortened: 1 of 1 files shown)


jsugarman (approved):
LGTM  - i guess we can try it out after the fact for real or have you test locally with migrated data?

jrmhaig:
> LGTM - i guess we can try it out after the fact for real or have you test locally with migrated data?

I have tested it locally but I'll give it a go when it is live.

PR #3808, opened 2021-03-15
Add ability/docs to run specs in parallel

jsugarman:
#### What
Add ability/docs to run specs in parallel, locally

#### Why
The are 10k+ specs that take approx 1 hour
to run locally in a single process.

This enables them to be run in parallel in
as many processes as the machine has cores.
On an 8 core machine it took ~15 minutes.

#### How
To install, prepare and run:
```
bundle
rake parallel:create
rake parallel:prepare
rake parallel:migrate # needed after each migration
rake parallel:spec
```
Note, feature specs not working and alot of setup and customization
needed it looks like

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -59,10 +59,12 @@ ENV['ADP_API_PASS'] = 'api_password'
 require 'spec_helper'
 require File.expand_path('../../config/environment', __FILE__)
 require 'rspec/rails'
+
 # Add additional requires below this line. Rails is not loaded until this point!
 require 'shoulda/matchers'
 require 'paperclip/matchers'
 require 'webmock/rspec'
+require 'parallel_spec_helper'
 require 'vcr_helper'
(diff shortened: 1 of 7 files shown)


spec/rails_helper.rb
```diff
@@ -120,6 +122,9 @@
     end
   end
 
+  # mute noise for parallel tests
```
jsugarman: remove

kmahern (approved):
🥇 Looks good to me. I pulled the branch and tried it and ran the specs in around 16 minutes

naseberry:
Looks good, In CircleCI it seems to take a minute less to run the suite.

jsugarman:
> Looks good, In CircleCI it seems to take a minute less to run the suite.

That will probably be because it is optimises by spreading load more evenly across the processors by keep timings from previous runs. Can't do that with parallel_tests gem locally, at least not out of the box

PR #3809, opened 2021-03-15
Add mechanism for low impact secret rotation

jsugarman:
#### What
Adds deactivated configuration that enables
rotation of the `secret_key_base` without
logging users out.

#### Ticket

[CBO-1734](https://dsdmoj.atlassian.net/browse/CBO-1734)

#### Why
To prevent users being logged out when the `secret_key_base`
secret is rotated, which in turn would be an inconvenience
with some potetntial for input field data loss, as well
as edges cases for error being encountered by the app.

#### How
 See `docs/cookie_rotation.md` for details.

### DIFF
--- a/config/initializers/cookie_rotation.rb
+++ b/config/initializers/cookie_rotation.rb
@@ -1,0 +1,36 @@
+# See `docs/cookie_rotation.md` for details.
+#
+# This initializer can be kept for future
+# rotation or it can be deleted.
+#
+# It will only take action if there is an
+# `old_secret_key_base` secret in `config/secrets.yml`.
+#
+# NOTE: any changes to rails `action_dispatch` encryption config will
+# impact this functionality, rendering it in need of revisiting - manual
+# testing.
+#
+return unless Rails.application.secrets.old_secret_key_base.present?
+
+Rails.application.configure do
+  # Not technically necessary, as this is the current default,
+  # but it is to line up with the config below.
+  config.action_dispatch.use_authenticated_cookie_encryption = false
+
+  # From:
+  # https://www.gitmemory.com/issue/rails/rails/39964/668147345
+  # --------------------------------------------------------------------------------
+  config.action_dispatch.cookies_rotations.tap do |cookies|
+    salt = config.action_dispatch.encrypted_cookie_salt
+    signed_salt = config.action_dispatch.encrypted_signed_cookie_salt
+    cipher = config.action_dispatch.encrypted_cookie_cipher || 'aes-256-cbc'
+
+    old_secret_key_base = Rails.application.secrets.old_secret_key_base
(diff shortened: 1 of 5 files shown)


docs/cookie_rotation.md
```diff
@@ -0,0 +1,82 @@
...
+   , where `environment` can be dev, staging, production or
+   any other "namespace".
+
+4. Deploy the application, ensuring the new secrets
+   are applied first. The `kubernetes_deploy/scripts/deploy.sh`
+   or `.circleci/deploy.sh` scripts will apply the secrets
+   before the image, thereby ensuring this.
+
+5. Leave the rotation in place for x, where x is however
+   long (days?!) we think it will take the majority of
```
jrmhaig, line 50: Do users get automatically logged out if they haven't access the site in a certain period of time? If so, that could be you *x* days.

config/initializers/cookie_rotation.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+return unless Rails.application.secrets.old_secret_key_base.present?
+
+Rails.application.configure do
+  # Not technically necessary, as this is the current default,
+  # but it is to line up with the config below.
+  config.action_dispatch.use_authenticated_cookie_encryption = false
+
+  # From:
+  # https://www.gitmemory.com/issue/rails/rails/39964/668147345
```
jrmhaig, line 21: This is a link to details of an issue with this approach, where an extra parameter `value` is added to the cookie. The original article at https://blog.kmkonline.co.id/rotating-ruby-on-rails-secret-key-base-9a4dcdf0d817 suggests having `before_action { session.delete :value }` in the base controller as a work-around. Do we need this?

jrmhaig (approved):
I'm still getting my head around the details but the strategy appears to be sensible.

docs/cookie_rotation.md
jsugarman: Essentially yes,

if user `A` has visited site during the x days the old/new secret rotation config is active, but user `B` has not then when we remove the old_secret_key_base value (ENV var on hosted environments) after the x days:

1. user  `A` would not encounter any problem
2. user `B` would be logged out/suffer a redirect (i.e. encounter `InvalidAuthenticityToken` error) when they do, eventually visit the site again.

config/initializers/cookie_rotation.rb
jsugarman: I just linked to this because it refers to the original article but handles our specific issue which is that we are still using the older cipher (we should move to the new cipher at some point). I have not encountered the `value` issue it refers to

jsugarman: The original article solution does not work because it does not include the `signed_secret` element, in particular, amongst other problems

kmahern:
If I am understanding this correctly, this pr adds the mechanism for rotating the secret, but does not actually rotate the key itself?

PR #3811, opened 2021-03-15
Rotate secret_key_base using old/new cookie rotation

jsugarman:
#### What
Rotate secret_key_base using old/new cookie rotation

#### Ticket

[CBO-1734](https://dsdmoj.atlassian.net/browse/CBO-1734)

#### Why
App security

#### How
This moves the current SECRET_KEY_BASE env
var to a new one called OLD_SECRET_KEY_BASE.

Session cookies encypted with either of these
secrets will be valid, but new cookies will
be generated using the new secret.

This means we can leave in place for several
days to allow users that visit the site a chance
get a new session cookie encypted with the new secret.

We can then delete the OLD_SECRET_KEY_BASE secret
env var without affecting those users.

Am minded to leave it for 5 days.


jrmhaig (approved):
Looks OK.

PR #3816, opened 2021-03-16
Fetch attachment path from Active Storage

jrmhaig:
#### What

If an attachment has been created using Active Storage and then Paperclip has been turned back on then fetch the path to the asset from the `ActiveStorage::Blob` record.

#### Ticket

[Migrate paperclip to active storage - stats reports](https://dsdmoj.atlassian.net/browse/CBO-1683)

#### Why

In the event of having to roll back the switch to Active Storage we need to be able to access assets that have been created using Active Storage. However, the pattern for files saved by Active Storage does not match Paperclip.

#### How

A custom [Paperclip interpolation](https://github.com/thoughtbot/paperclip/wiki/Interpolations) is created that will fetch the path from Active Storage if it exists (and not previously migrated from Paperclip) and fall back to the Paperclip path otherwise.


jsugarman (changes requested):
Looks good but need
- manual testing of rollback
- unit spec of the mixin
- update the cleanup documentation to get rid of it

spec/models/message_spec.rb
```diff
@@ -218,4 +218,64 @@
...
+      it 'has a path based on the filename' do
+        expect(message.attachment.path).to eq "public/assets/test/images/#{id_partition}/#{filename}"
+      end
+    end
+
+    context 'with an Active Storage attachment in disk storage' do
+      require 'active_storage/service/disk_service'
+
+      before do
+        ActiveStorage::Attachment.connection.execute(<<~SQL)
```
jsugarman: can you extract these to a shared context as i did for documents

spec/models/message_spec.rb
```diff
@@ -218,4 +218,64 @@
...
+      it 'has the path for disk storage' do
+        expect(message.attachment.path).to eq '/root/te/st/test_key'
+      end
+    end
+
+    context 'with an Active Storage attachment in S3' do
+      require 'active_storage/service/s3_service'
+
+      before do
+        ActiveStorage::Attachment.connection.execute(<<~SQL)
```
jsugarman: [as above](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3816/files#r596672810)

spec/models/stats/stats_report_spec.rb
```diff
@@ -167,6 +167,64 @@
...
+      it 'has a path based on the filename' do
+        expect(report.document.path).to eq 'reports/test_file.csv'
+      end
+    end
+
+    context 'with an Active Storage attachment in disk storage' do
+      require 'active_storage/service/disk_service'
+
+      before do
+        ActiveStorage::Attachment.connection.execute(<<~SQL)
```
jsugarman: [as above](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3816/files#r596672810)

spec/models/stats/stats_report_spec.rb
```diff
@@ -167,6 +167,64 @@
...
+      it 'has the path for disk storage' do
+        expect(report.document.path).to eq '/root/te/st/test_key'
+      end
+    end
+
+    context 'with an Active Storage attachment in S3' do
+      require 'active_storage/service/s3_service'
+
+      before do
+        ActiveStorage::Attachment.connection.execute(<<~SQL)
```
jsugarman: [as above](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3816/files#r596672810)

jsugarman (changes requested):
👍 Looks ok

- needs some `shared_contexts` to DRY out the tests further before merge i'd say.
- **not a blocker** but could even abstract AS setup further to parameterise the differences but probably not worth it. Ideally the AS stuff would have factorybot factories to deal with it too/instead - e.g. `spec/factories/documents.rb`
- have black-box tested on staging for documents and message attachments ✅

jrmhaig:
> Ideally the AS stuff would have factorybot factories to deal with it too/instead - e.g. `spec/factories/documents.rb`

I don't know how Factory Bot would cope with this. The AS data has be to inserted directly into the database because it is not attached to real associations.

jsugarman (approved):
👍

jsugarman:
> > Ideally the AS stuff would have factorybot factories to deal with it too/instead - e.g. `spec/factories/documents.rb`
> 
> I don't know how Factory Bot would cope with this. The AS data has be to inserted directly into the database because it is not attached to real associations.

On a very quick/shallow search [i found this](https://blog.eq8.eu/til/factory-bot-trait-for-active-storange-has_attached.html) but something for another PR i'd say

jrmhaig:
> > > Ideally the AS stuff would have factorybot factories to deal with it too/instead - e.g. `spec/factories/documents.rb`
> > 
> > 
> > I don't know how Factory Bot would cope with this. The AS data has be to inserted directly into the database because it is not attached to real associations.
> 
> On a very quick/shallow search [i found this](https://blog.eq8.eu/til/factory-bot-trait-for-active-storange-has_attached.html) but something for another PR i'd say

The problem is that we need to create an Active Storage attachment without a `has_attached`. At this stage the `Document` model doesn't know anything about Active Storage so a simple:

```ruby
document { fixture_file_upload(Rails.root.join(...) }
```

is going to, at best, create something with Paperclip and

```ruby
document.document.attach(...)
```

will just fail.

PR #3820, opened 2021-03-17
Add cronjob to process document checksums

jsugarman:
#### What
Adds a kubernetes cronjob to incrementally process
the 400k+ document for checksum

#### Ticket

[CBO-1742](https://dsdmoj.atlassian.net/browse/CBO-1742)

#### Why
Running the checksum task for documents on a worker
pod manually was failing after x (about 15-20k) records
because the process exceeded the memory request limit - 2Gi.

Increasing the memory would just have delayed the failure
as the single process would continue to "leak" memory - not
sure where this was/is happening.

#### How
This handles the "leak" by processing a limited number of
records and schedules the task to repeat every 30 minutes.

The cronjob can be applied to an environment using
```
kubernetes_deploy/scripts/cronjob.sh add_evidence_document_checksums <environment> latest
```

#### Load test results
On a pre-prod environment set up with a prod-like load this task took
approximately 8 hours to complete, processing 10K document records
every 30 minutes

### DIFF
--- a/docs/active_storage_migration.md
+++ b/docs/active_storage_migration.md
@@ -338,6 +338,8 @@ Paperclip can be removed.
 * Remove `PAPERCLIP_STORAGE_OPTIONS`, `REPORTS_STORAGE_OPTIONS` and
   `REPORTS_STORAGE_OPTIONS` from the configuration files in
   `config/environments`.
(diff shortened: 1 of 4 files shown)


jrmhaig (approved):
:+1:

PR #3822, opened 2021-03-18
Exclude storage directory from Rubocop

jrmhaig:
#### What

Exclude the contents of the `storage/` directory when running Rubocop.

#### Ticket

N/A

#### Why

Rubocop will scan all files but there are none in `storage/` that need to be checked so these will unnecessarily slow down the check.

#### How

Add `storage/**/*` to the exclusion list in `.rubocop.yml`.


kmahern (approved):
👍

PR #3823, opened 2021-03-18
Exclude public folder from rubocop

jsugarman:
#### What
Exclude public folder from rubocop

#### Why
Unnecessary to lint this folder, plus
if using for disk storage if can get very
large and cause slow startup for rubocop.


kmahern (approved):
👍

jrmhaig (approved):
:+1:

PR #3824, opened 2021-03-18
[CBO-1692] [AS3b] Enable Active Storage for message attachments

jrmhaig:
#### What

Switch over to using Active Storage for attachments of claim messages.

#### Ticket

[Migrate paperclip to active storage - enable for messages](https://dsdmoj.atlassian.net/browse/CBO-1692)

#### Why

We need to switch over to using Active Storage everywhere we have documents in CCCD. Following on from the PR to turn this on for stats report the next step is for attachments of messages.

#### How

### DIFF
--- a/spec/factories/messages.rb
+++ b/spec/factories/messages.rb
@@ -31,7 +31,11 @@ FactoryBot.define do
   end
 
   trait :with_attachment do
-    attachment { File.open(Rails.root + 'features/examples/shorter_lorem.docx') }
-    attachment_content_type { 'application/msword' }
+    attachment do
+      Rack::Test::UploadedFile.new(
+        File.expand_path('features/examples/shorter_lorem.docx', Rails.root),
+        'application/msword'
+      )
+    end
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -66,6 +66,7 @@ gem 'geckoboard-ruby'
 gem 'posix-spawn', '~> 0.3.15'
 gem 'laa-fee-calculator-client', '~> 1.2'
 gem 'webpacker', '~> 5.2'
+gem 'active_storage_validations'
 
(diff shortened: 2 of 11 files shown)


spec/factories/messages.rb
```diff
@@ -31,7 +31,11 @@
   end
 
   trait :with_attachment do
-    attachment { File.open(Rails.root + 'features/examples/shorter_lorem.docx') }
-    attachment_content_type { 'application/msword' }
+    attachment do
+      Rack::Test::UploadedFile.new(
+        File.expand_path('features/examples/shorter_lorem.docx', Rails.root),
+        'application/msword'
+      )
```
jsugarman, line 38: Not sure what is preferable but we are using `fixture_file_upload` in other factories and specs. it is [included globally](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/spec/rails_helper.rb#L71). However i have seen older comments suggesting it is bad practice to use this.

[example discussion - see comments](https://collectiveidea.com/blog/archives/2017/01/16/testing-an-uploaded-file-in-rspec)

jsugarman: I have [PR'd a change to not use fixture_file_upload](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3835) to see how it goes

jsugarman (commented):
LGTM, just need to test migrate and rollback on dev-lgfs

spec/factories/messages.rb
jrmhaig: I hadn't specifically avoided using `fixture_file_upload` and I don't know about the suggestion of it being bad practice. If we are moving away from `fixture_file_upload` are you happy for this to stay as it is?

jsugarman (approved):
👍

PR #3825, opened 2021-03-18
Exclude versions table data from dump

jsugarman:
#### What
Exclude versions table data from dump

#### Why
paper_trial gem auditing generates alot of
data for changed objects which is typically
not required for dump and restore

### DIFF
--- a/lib/tasks/dump.rake
+++ b/lib/tasks/dump.rake
@@ -47,7 +47,7 @@ namespace :db do
       filename = File.join('tmp', "#{Time.now.strftime('%Y%m%d%H%M%S')}_dump.psql")
 
       shell_working "exporting unanonymised database data to #{filename}..." do
-        cmd = "pg_dump $DATABASE_URL --no-owner --no-privileges --no-password #{sensitive_table_exclusions} #{unneeded_table_exclusions} -f #{filename} "
+        cmd = "pg_dump $DATABASE_URL --no-owner --no-privileges --no-password #{sensitive_table_exclusions} #{unneeded_table_exclusions} -f #{filename}"
         system(cmd)
       end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/dump.rake
```diff
@@ -242,13 +242,21 @@ namespace :db do
...
-      excluded_tables.map { |table| "--exclude-table-data #{table}" }.join(' ')
+      sensitive_tables.map { |table| "--exclude-table-data #{table}" }.join(' ')
+    end
+
+    def unneeded_tables
+      %w(versions)
+    end
+
+    def unneeded_table_exclusions
+      unneeded_tables.map { |table| "--exclude-table-data #{table}" }.join(' ')
```
jrmhaig: You might consider replacing this and `sensitive_table_exlusions` with something like:

```ruby
    def exclusions_for(tables)
      tables.map { |table| "--exclude-table-data #{table}" }.join(' ')
    end
```

lib/tasks/dump.rake
```diff
@@ -47,12 +47,12 @@ namespace :db do
       filename = File.join('tmp', "#{Time.now.strftime('%Y%m%d%H%M%S')}_dump.psql")
 
       shell_working "exporting unanonymised database data to #{filename}..." do
-        cmd = "pg_dump $DATABASE_URL --no-owner --no-privileges --no-password #{sensitive_table_exclusions} -f #{filename}"
+        cmd = "pg_dump $DATABASE_URL --no-owner --no-privileges --no-password #{sensitive_table_exclusions} #{unneeded_table_exclusions} -f #{filename} "
```
jsugarman: remove extra whitespace

jrmhaig (approved):
:+1:

PR #3826, opened 2021-03-18
Remove the climate_control gem

jrmhaig:
#### What

Remove the `climate_control` gem.

#### Ticket

N/A

#### Why

This gem is no longer being used.

#### How

Remove from `Gemfile` and recreate `Gemfile.lock`.


jsugarman (approved):
👍

PR #3835, opened 2021-03-22
Do not use include ActionDispatch::TestProcess

jsugarman:
#### What
Do not use include ActionDispatch::TestProcess for file
fixtures

#### Why
Not sure if this is preferable but some commentators
have argued against using this lib.

https://collectiveidea.com/blog/archives/2017/01/16/testing-an-uploaded-file-in-rspec

An alternative may be to just include the FixtureFile aspect
```
include ActionDispatch::TestProcess::FixtureFile
```


jrmhaig (approved):
Looks OK. I don't know too much about the pros and cons of using `fixture_file_upload` but as we are only using it in a couple of places is probably doesn't make sense to globally include `ActionDispatch::TestProcess`.

PR #3836, opened 2021-03-22
Prefetch blob and record to avoid N+1 problem

jrmhaig:
#### What

Pre-fetch `ActiveStorage::Blob` and `Document`/`Message`/`Stats::StatsReport` when validating migration data.

#### Ticket

N/A

#### Why

The `storage:status` Rake task checks each Active Storage attachment to make sure that the values in the `ActiveStorage::Blob` records match the corresponding Paperclip data.

#### How

Add `includes(:blob, :record)`


jsugarman (approved):
👍

PR #3839, opened 2021-03-23
[CBO-1681] GDPR compliant cookies

mabely:
#### What
Change cookie banner and cookies handling to be GDPR compliant. 

The “usage_opt_in” cookie determines whether Google Analytics cookies and tracking is switched on or off. These constitute non-essential cookies and under GDPR, should be optional and disabled by default. The user will be prompted with a cookie banner to accept or reject these usage cookies when they first visit the page, if they do so a confirmation banner appears, which can be hidden. 

Once usage cookies are accepted or rejected, a "cookies_preference” cookie is stored to indicate cookie preferences are set. The user will be prompted again if that cookie expires or does not exist. 

There is a new:
1. 'Cookie settings' page, where the user can change their cookie preference at any time
2. 'Cookie details' page, where the user can learn about the cookies we use
   

#### Ticket
[Create compliant Cookie Banner DPA2018 (GDPR)](https://dsdmoj.atlassian.net/browse/CBO-1681)

#### Why
To ensure that the way we handle cookies is GDPR compliant.

#### How
- Update cookie banner
- Set usage cookies as disabled by default, prompt user to accept or reject them if they do not have the cookies_preference cookie
- Set new cookies_preference cookie when user accepts or rejects usage cookies
- Create page to allow users to change cookie preferences
- Add 'cookie details' page with all cookie details


naseberry:
Note: Add Google Analytics Annotations when feature is deployed so that the possible change in data is easier to understand.

spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -13,10 +13,23 @@ module GoogleAnalytics
...
+            it "returns true when host is #{host}" do
+              allow(RailsHost).to receive(:env).and_return(host)
+              allow(described_class).to receive(:usage_name).and_return('true')
+              expect(described_class.enabled?).to be_truthy
+            end
+          end
+        end
+
+        context 'with usage tracking rejected' do
+          %w(staging production).each do |host|
```
jsugarman: we were looking to unfurl these rspec loops. can use context blocks with before hooks to set these
```
context 'when on staging' do
  before { allow(RailsHost).to receive(:env).and_return('staging') }
  ...
end
```

spec/models/google_analytics/data_tracking_spec.rb
```diff
@@ -13,10 +13,23 @@ module GoogleAnalytics
           allow(described_class).to receive(:adapter).and_return('Adapter')
         end
 
-        %w(staging production).each do |host|
-          it "returns true when host is #{host}" do
-            allow(RailsHost).to receive(:env).and_return(host)
-            expect(described_class.enabled?).to be_truthy
+        context 'with usage tracking accepted' do
```
jsugarman, line 16: see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3839/files#r603201741

features/step_definitions/cookies_steps.rb
```diff
@@ -0,0 +1,24 @@
...
+
+When('I choose to turn cookies {string}') do |status|
+  option_label = "cookies.new.analytics_cookie_#{status}_label"
+  choose(I18n.t(option_label))
+end
+
+Then('the cookie preference is saved') do
+  expect(find('.govuk-notification-banner--success').visible?).to eq true
+  expect(page).to have_content I18n.t('cookies.new.cookie_notification')
+end
```
jsugarman: you can use site prism to define a cookie banner section and then use syntax like
```
  expect(cooke_banner).to be_visible
  expect(cooke_banner).to have_content "expected_text"
```

see similar setups for other pages

features/step_definitions/cookies_steps.rb
```diff
@@ -0,0 +1,24 @@
+When('I see the cookies banner') do
+  expect(page).to have_css '.govuk-cookie-banner'
+  expect(page.find('.govuk-cookie-banner').visible?).to eq true
+  expect(page).to have_content I18n.t('layouts.cookie_banner.accept_button')
```
jsugarman: prefer to copy the actual english text, rather then call the translation the view uses. This then tests the content.

config/locales/en/views/cookie.en.yml
```diff
@@ -0,0 +1,119 @@
...
+        </tr>
+        </tbody>
+        </table>'
+      cookie_details_heading: Details about cookies on Claim for crown court defence
+      cookie_settings_html:
+        '<h2 class="govuk-heading-l">Change your settings</h2>
+        <p class="govuk-body">You can <a class="govuk-link" href="%{link}">change which cookies you are happy for us to use</a>.</p>'
+      intro_html:
+        '<p class="govuk-body">Claim for crown court defence puts small files (known as ‘cookies’) onto your computer to collect information about how you browse the site. Find out more about the cookies we use, what they are for and when they expire.</p>'
+      strictly_necessary_cookies_html:
```
jsugarman, line 75: **not a blocker**: given the amount of html in here, is a partial not a better option

config/locales/en/views/cookie.en.yml
```diff
@@ -0,0 +1,119 @@
...
+      essential_cookies_html:
+        '<h2 class="govuk-heading-l">Essential cookies</h2>
+        <p class="govuk-body">These essential cookies do things like keep your information secure while you use this service. They always need to be on.</p>'
+      intro_html:
+        '<p class="govuk-body">Cookies are files saved on your phone, tablet or computer when you visit a website.</p>
+        <p class="govuk-body">We use cookies to store information about how you use the Claim for crown court defence website, such as the pages you visit.</p>'
+      save_changes: Save changes
+
+    cookie_details:
+      analytics_cookies_html:
```
jsugarman, line 31: see [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3839/files#r603204733)

jsugarman (approved):
Looks good, couple of suggestions from a feature test point of view mainly

app/models/cookies.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class Cookies
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations
+
+  attribute :analytics
```
jrmhaig, line 8: Is it possible for `analytics` to be a boolean rather than a string? It feels more natural to be using `true` and `false` rather than `'true'` and `'false'` in various places (eg, in `GoogleAnalytics::DataTracking#enabled?`)

I realise that it is being stored in a cookie. Is it possible to have a boolean stored in a cookie?

features/step_definitions/cookies_steps.rb
mabely: I could only make it work by creating a `CookiePage` and including the section rather than setting a standalone `CookieSection`, even though the documentation seems to show it is possible. It seems counterintuitive though since the cookie banner is just a section?

When I try `CookieSection.new`, I get `wrong number of arguments (given 0, expected 2) (ArgumentError)`.

config/locales/en/views/cookie.en.yml
mabely: @naseberry - did you have any thoughts on this?

Seeing as there's lots of snippets of text there, am I right to understand that we generally keep all wordings out of views?

naseberry: Following standards, yes, I wold keep all long form content out of views.
This is what I have been doing to other views that have long form content.

@mabely @jsugarman

features/step_definitions/cookies_steps.rb
mabely: Perhaps it'd make more sense for me to add the element to `ExternalUserHomePage` or `UserHomePage`? @jsugarman

app/models/cookies.rb
mabely: I would agree too, but after looking into it and it seems that it is stored as boolean until the page is refreshed, where it automatically converts the value to string. As you said the `analytics` value is assigned via the cookie, so it would have to involve converting the string to boolean. Open to opinions on whether that is worth doing?

@jrmhaig (tagging as I'm not sure whether you'd be notified of this reply)

features/step_definitions/cookies_steps.rb
jsugarman: Its not a blocker anyway, so if you want to do as a refactor after the fact that is fine. There is a hierarchy with `BasePage` as the superclass. You could put it somewhere in that hierarchy that makes sense for the cookie banner "section"

app/models/cookies.rb
jrmhaig: I guess that the values of cookies have to be strings. I don't know if it helps but you can get a boolean value from a parameter (or cookie) that is a string using:

```ruby
ActiveModel::Type::Boolean.new.cast(params[:something])
```

(See https://api.rubyonrails.org/v6.1.0/classes/ActiveModel/Type/Boolean.html)

You could possibly use this in the `CookieConcern#set_datatracking_usage` method and then `usage_name` `GoogleAnalytics::DataTracking` would be a boolean.

PR #3841, opened 2021-03-23
Add feature steps to excercise "Duplicate last expense"

jsugarman:
#### What
Add feature steps to excercise "Duplicate last expense"
expense numbering and expense removal.

#### Why
This feature was not being feature tested. The feature
itself broke due to a govuk_component change but the test
suite passed.

This also adds steps to black box test the removal
feature and its impact on expense numbering.


naseberry (approved):
lgtm

PR #3844, opened 2021-03-24
Extend and document intended use of `govuk_link_button`

jsugarman:
#### What
Extend and document intended use of `govuk_link_button`

#### Ticket
[CBO-1755](https://dsdmoj.atlassian.net/browse/CBO-1755)

#### Why
A bump to ruby 2.7.1 caused a breakage with
no tests covering it. A feature test has
been added to cover this previously from the app level.

Some unit specs should also be added cover and document the
intended happy path for this helper and those
that depend on it - the `govuk_link_button_secondary`
and `govuk_link_button_warning`.


naseberry (approved):
ty

PR #3845, opened 2021-03-24
Bump mimemagic dependency

jsugarman:
All previous versions yanked for licensing reasons

see [issue for more](https://github.com/rails/rails/issues/41750)


bitboxer:
Sadly, this is not possible legally because the library is under GPL and this project is using MIT. Both are incompatible.

PR #3846, opened 2021-03-25
Bump mimemagic to 0.3.6

jsugarman:
#### What
Bump mimemagic to 0.3.6

#### Why
Previous versions yanked causing build failures.

Until rails fixes its dependency and paperclip
is removed we need to use the only compatible version
avialable.


jrmhaig (approved):
:+1:

PR #3850, opened 2021-03-25
Bump mimemagic and install library

jsugarman:


jsugarman:
close in favour of #3851

PR #3851, opened 2021-03-26
Bump mimemagic to 0.3.9

jsugarman:
This latest version reintroduces overlays, should help with docx issue.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,5 +1,6 @@
 brew "postgresql", restart_service: true
 brew "redis", restart_service: true
+brew "shared-mime-info"
 brew "nvm"
 brew "yarn"
 cask "chromedriver"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -474,7 +474,9 @@ GEM
     mime-types (3.3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2020.1104)
-    mimemagic (0.3.6)
+    mimemagic (0.3.9)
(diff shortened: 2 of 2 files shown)


docker/Dockerfile
```diff
@@ -37,7 +38,8 @@ RUN apk --no-cache add --virtual build-dependencies \
                   ttf-ubuntu-font-family \
                   libreoffice \
                   redis \
-                  postgresql-client
+                  postgresql-client \
+                  shared-mime-info
```
jrmhaig: I think that `shared-mime-info` is already installed but it doesn't hurt to list it here, especially as it is now a dependency.

config/initializers/mimemagic.rb
```diff
@@ -0,0 +1,6 @@
+require 'mimemagic'
+# require 'mimemagic/overlay'
+# file = File.expand_path('features/examples/shorter_lorem.docx', Rails.root)
+# MimeMagic.by_magic(File.open(file))
+
+MimeMagic.add('application/vnd.openxmlformats-officedocument.wordprocessingml.document', { extension: 'docx' })
```
jrmhaig: The overlay file has been reintroduced for version 0.3.9 so this may not be required any more.

jrmhaig:
We may need to add `shared-mime-info` to `Brewfile` as it is not installed on Mac by default.

jrmhaig (approved):
:+1: Looks good.

PR #3852, opened 2021-03-26
[CBO-1693] [AS3c] Use Active Storage for documents

jrmhaig:
#### What

Switch over to using Active Storage for documents.

#### Ticket

[Migrate Paperclip to Active Storage](https://dsdmoj.atlassian.net/browse/CBO-1693)

#### Why

We need to switch over to using Active Storage everywhere we have documents in CCCD. The final part is to enable Active Storage for case evidence documents, in the `Document` model.

#### How

#### Notes

1. The conversion of documents to PDF for preview was previously handled by the `DocumentAttachment` concern that was included in the `Document` model. This has been moved into a service, `DocumentConverterService`. The concern `DocumentAttachment` also included some configuration and validation of attributes for `Document` and this has been moved into the model.
2. `spec/controllers/documents_controller_spec.rb` has been replaced by a requests spec, `spec/requests/documents_spec.rb`. The sections for the download and view routes, `GET /documents/:id` and `GET /documents/:id/download`, are rewritten while the other sections are only modified to change into requests specs and to fix for Rubocop.

--------

#### TODO (wip)

 - [x] Populate Paperclip fields, to allow rollback
 - [x] Download document
 - [x] Preview document online
 - [x] `copy_from` method, for copying claims
 - [x] Download documents as zip


app/models/document.rb
```diff
@@ -24,107 +24,65 @@
...
-
-  validate :documents_count
+  before_save :create_preview_document
+  before_save -> { populate_paperclip_for :document }
+  before_save -> { populate_paperclip_for :converted_preview_document }
 
-  def copy_from(original_doc, verify: false)
-    self.document = original_doc.document
-    verify ? save_and_verify : save
+  def copy_from(other)
```
jsugarman: **nit** `original` feels a better name than `other`

app/services/document_converter_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    if @original.content_type == 'application/pdf'
+      converted.attach(@original.blob)
+    else
+      convert_document_to_pdf from: @original, to: converted
+    end
+  end
+
+  private
+
+  def convert_document_to_pdf(from:, to:)
```
jsugarman: **nit** given context and use the name `convert` feels sufficient or `convert_document`

app/services/document_converter_service.rb
```diff
@@ -0,0 +1,42 @@
+class DocumentConverterService
```
jsugarman, line 1: a service typically just responds to `call` - this has not been adhered to very well previously either.  see [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3852/files#r603999262)

app/models/document.rb
```diff
@@ -133,9 +91,7 @@ def documents_count
     errors.add(:document, "Total documents exceed maximum of #{max_doc_count}. This document has not been uploaded.")
   end
 
-  def transform_cryptic_paperclip_error
-    return unless errors[:document].include?('has contents that are not what they are reported to be')
-    errors[:document].delete('has contents that are not what they are reported to be')
-    errors[:document] << 'The contents of the file do not match the file extension'
+  def create_preview_document
+    DocumentConverterService.new(document).to(converted_preview_document)
```
jsugarman: could this just be
```
DocumentConverterService.new(document, converted_preview_document).call
```
see [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3852/files#r603996875) below and [this blog example](https://www.toptal.com/ruby-on-rails/rails-service-objects-tutorial)

app/models/document.rb
```diff
@@ -24,107 +24,65 @@
...
 
   alias attachment document # to have a consistent interface to both Document and Message
   delegate :provider_id, to: :external_user
 
-  before_save :generate_pdf_tmpfile
-  before_save :add_converted_preview_document
-  before_save :populate_checksum
-
-  validate :documents_count
+  before_save :create_preview_document
```
jsugarman: could a `before_create` make this simpler.

app/services/s3_zip_downloader.rb
```diff
@@ -3,31 +3,29 @@ class S3ZipDownloader
...
     @claim = claim
-    @folder = Dir.mktmpdir("#{@claim.case_number}-", './tmp')
-    move_files
-    @files_to_zip = Dir.entries(@folder) - %w[. ..]
-    @zip_file = "./tmp/#{@claim.case_number}-#{SecureRandom.uuid}-documents.zip"
   end
 
   def generate!
-    zip_files
-    FileUtils.rm_rf(@folder)
```
jsugarman, line 14: Should we be removing this, or does block deal with it?

jsugarman: >If a block is given, it is yielded with the path of the directory. The directory and its contents are removed using FileUtils.remove_entry before ::mktmpdir returns. The value of the block is returned.

app/services/s3_zip_downloader.rb
```diff
@@ -3,31 +3,29 @@ class S3ZipDownloader
...
   end
 
   private
 
-  def zip_files
-    Zip::File.open(@zip_file, Zip::File::CREATE) do |zip_file|
-      @files_to_zip.each do |filename|
-        zip_file.add(filename, File.join(@folder, filename))
+  def build_zip_file(documents, bundle)
+    Dir.mktmpdir("#{@claim.case_number}-", './tmp') do |tmp_dir|
```
jsugarman: `Rails.root.join('tmp')` as  fallback instead of './tmp'. but not an issue for now.

spec/requests/documents_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+
+  describe 'DELETE /documents/:id' do
+    subject(:delete_document) { delete document_path(document, format: :json) }
+
+    let!(:document) { create(:document, external_user_id: external_user.id) }
+
+    it 'destroys the document' do
+      expect { delete_document }.to change(Document, :count).by(-1)
+    end
+
```
jsugarman, line 155: **not a blocker** can we possibly test that the document is deleted as well.

spec/requests/documents_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+    context 'when unable to destroy a document' do
```
jsugarman: feel free to remove, barring codeclimate scorings

spec/models/timed_transitions/transitioner_spec.rb
```diff
@@ -263,90 +263,110 @@
...
-            expect(RepresentationOrder.where(defendant_id: @defendant.id)).to be_empty
-            expect(Document.where(claim_id: @claim.id)).to be_empty
-            expect(File.exist?(@document.document.path)).to be false
-            expect(File.exist?(@document.converted_preview_document.path)).to be false
-            expect(Message.where(claim_id: @claim_id)).to be_empty
-            expect(ClaimStateTransition.where(claim_id: @claim.id)).to be_empty
-            expect(Determination.where(claim_id: @claim.id)).to be_empty
-            expect(Certification.where(claim_id: @claim.id)).to be_empty
-            expect(InjectionAttempt.where(claim_id: @claim.id)).to be_empty
+            expect { Claim::BaseClaim.find(claim.id) }.to raise_error ActiveRecord::RecordNotFound, "Couldn't find Claim::BaseClaim with 'id'=#{claim.id}"
```
jsugarman, line 359: put a `TODO:` to clear this up/rip it out. individual tests or remove if pointless. also avoid false positives

spec/models/timed_transitions/transitioner_spec.rb
```diff
@@ -263,90 +263,110 @@
...
+            end
+
+            it 'destroys all associated records', delete: true do
+              check_associations
+              described_class.new(claim).run
+              expect_claim_and_all_associations_to_be_gone
+            end
+          end
+
+          context 'with an associated document' do
```
jsugarman, line 306: 👍

spec/services/document_converter_service_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe DocumentConverterService do
+  subject(:convert_document) { described_class.new(document, new_document).call }
+
+  let(:document) { message.attachment }
```
jsugarman: within the context of the `app` the use of the term `message` here could be confusing. 

While this service is isolated such that it could be applied to any `ActiveStorage` "backed" object i think it would be clearer to use either a `document` object and affiliated factory OR a mocked object.

other generic terms that could be used might include `storage_record`,`doc_record`, `doc`,  `evidence`, `evidence_rec`

spec/services/document_converter_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+          File.open(to_file, 'wb') do |output_stream|
+            IO.copy_stream(input_stream, output_stream)
+          end
+        end
+      end
+    end
+  end
+
+  context 'when the source attachment is a PDF' do
+    let(:message) { build :message, :with_pdf_attachment }
```
jsugarman: see [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3852/files#r604083739)

spec/services/document_converter_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    before { convert_document }
+
+    it { expect(new_document.blob).to eq document.blob }
+  end
+
+  context 'when the source attachment has been saved' do
+    let(:message) { create :message, :with_docx_attachment }
```
jsugarman: see [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3852/files#r604083739)

spec/services/document_converter_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    before { convert_document }
+
+    it { expect(new_document.content_type).to eq 'application/pdf' }
+    it { expect(new_document.checksum).to eq checksum }
+  end
+
+  context 'when the source attachment is new (i.e., not saved)' do
+    let(:message) { build :message, :with_docx_attachment }
```
jsugarman: see [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3852/files#r604083739)

spec/factories/messages.rb
```diff
@@ -38,4 +38,22 @@
...
+  end
+
+  trait :with_pdf_attachment do
+    attachment do
+      Rack::Test::UploadedFile.new(
+        File.expand_path('features/examples/longer_lorem.pdf', Rails.root),
+        'application/pdf'
+      )
+    end
+  end
```
jsugarman: do we need this if we use documents factory -  related to [comment](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3852/files#r604083739)

spec/services/document_converter_service_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe DocumentConverterService do
+  subject(:convert_document) { described_class.new(document, new_document).call }
```
jsugarman: would this not be clear if it was
```suggestion
  subject(:convert_document) { described_class.new(document, converted_document).call }
```
or similar `converted_preview_document`, `pdf_document`, `preview_document`

spec/requests/documents_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+    expect(response.status).to eq(422)
+  end
+
+  it 'returns errors in response' do
+    create_document
+    expect(JSON.parse(response.body)).to have_key('error')
+  end
+end
+
+RSpec.describe '/documents', type: :request do
```
jsugarman:
```suggestion
RSpec.describe 'documents', type: :request do
```
OR
```suggestion
RSpec.describe 'Document management', type: :request do
```

I always check what it looks like when you use `rspec -fd` to see the verbose output

spec/requests/documents_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+  before do
+    allow(Document).to receive(:find).with(document.to_param).and_return(document)
+    allow(document.document.blob).to receive(:service_url).and_return(test_url)
+    allow(document.converted_preview_document.blob).to receive(:service_url) { |args|
+      "#{test_url}?response-content-disposition=#{args[:disposition] || 'inline'}"
+    }
+  end
+
+  context 'when the document is owned by the logged in user' do
+    it { is_expected.to redirect_to "#{test_url}?response-content-disposition=#{expected_disposition}" }
```
jsugarman, line 17: can `response_headers` be tested for content disposition? - I don't know myself cause it is a redirect with active storage

jrmhaig: It will be in the response headers that comes back from S3 so we cannot check that. The headers in the response is set using the parameters in the url, like this. See https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html

app/services/s3_zip_downloader.rb
jrmhaig: The second argument is the location to put the temporary directory, not a fallback. I do not think it is necessary so I am removing it.

jsugarman: where is the api reference for that?

jrmhaig: https://ruby-doc.org/stdlib-2.7.1/libdoc/tmpdir/rdoc/Dir.html

> The directory is created under ::tmpdir or the optional second argument tmpdir if non-nil value is given.

jsugarman (commented):
Findings from testing this:

1. rails storage:status is killed after an otherwise apparrently successful migration:
```
Documents
=========
Total records:      203985
Missing checksums
  Document:         0
  Preview:          0
AS records
  Document:         203984
Killed
```
2. the download all file link/functionality of caseworker results in the a zip with files using their uploaded name now. previously they had random uuid names. I am not sure if the previous behaviour was for a requirement for security/obfuscation reasons?!

3. when i deploy the branch, migrate, create a new claim and add some documents, then deploy the master branch (switchover rollback) rollback the data as well (deletion of AS record) seems to work fine but any records created by AS become no longer accessible/are empty. Is this a known problem we can do nothing about?

jrmhaig:
> Findings from testing this:
> 
> 1. rails storage:status is killed after an otherwise apparrently successful migration:

I have changed how the sanity check is performed and I believe this will solve the problem.
 
> 2. the download all file link/functionality of caseworker results in the a zip with files using their uploaded name now.

[This PR](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2110) is where the zip file was introduced and this line:

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/f16ab9ffad62c700717e3a3a7ed47e2ab3ae77fa/app/services/s3_zip_downloader.rb#L30

is where the temporary file, together with the temporary file name, is created. I don't see anything to suggest that there is any intentional obfuscation.

> 3. when i deploy the branch, migrate, create a new claim and add some documents, then deploy the master branch (switchover rollback) rollback the data as well (deletion of AS record) seems to work fine but any records created by AS become no longer accessible/are empty. Is this a known problem we can do nothing about?

These files should still be accessible. I'll have a look.

app/services/document_converter_service.rb
```diff
@@ -0,0 +1,43 @@
...
+  private
+
+  def convert_to_pdf(from:, to:)
+    with_attached_file(from) do |original|
+      pdf_tmpfile = Tempfile.new
+      Libreconv.convert(original, pdf_tmpfile)
+      to.attach(io: pdf_tmpfile, filename: "#{from.filename}.pdf", content_type: 'application/pdf')
+    end
+  rescue IOError
+    nil # raised if Libreoffice exe is not in PATH
```
jsugarman: rather see a log error than a comment, if necessary

jrmhaig: Thanks. This was copied verbatim from the existing code but the rescue can actually be removed as we want it to raise the exception so that the job will be retried.

**Edit:** Actually, this is not the background job (yet) so the rescue is required.

app/services/document_converter_service.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def with_attached_file(document)
+    # Currently, evidence documents are converted to PDF when the document is
+    # uploaded and before the instance of Document is saved. This is how it
+    # was done with Paperclip but with Active the file is not accessible at
+    # until it is saved. This is why the path is taken as the temporary file
+    # of the attachment.
```
jsugarman: Seems a bit of an excessive comment that will go out of date quite quickly. better that this stuff is in commit message history but I think the meat of it is "AS files are not readable/accessible until saved"

app/services/document_converter_service.rb
```diff
@@ -22,8 +22,6 @@ def convert_to_pdf(from:, to:)
       Libreconv.convert(original, pdf_tmpfile)
       to.attach(io: pdf_tmpfile, filename: "#{from.filename}.pdf", content_type: 'application/pdf')
     end
-  rescue IOError
```
jsugarman: You previously mentioned i does need rescuing?!

jsugarman (commented):
👍 LGTM

Some comment issues and just that issue around whether to rescue the IOError or not. sounds like it needs rescuing for now?! with a bit of logging to capture the problem

jsugarman (approved):
👍

PR #3858, opened 2021-03-31
Fix BigDecimal.new calls

jsugarman:
#### What
Fix BigDecimal.new calls

#### Why
This should fix [sentry errors we have been getting](https://sentry.io/organizations/ministryofjustice/issues/2303416769/?project=5436820&query=is%3Aunresolved)

#### How
Ruby 2.6 deprecated and ruby 2.7 broke
calls to `.new`

see the ruby 2.6 definition for which
also contains the fix.

```
def BigDecimal.new(*args, **kwargs)
  warn "BigDecimal.new is deprecated; use BigDecimal() method instead.", uplevel: 1
  BigDecimal(*args, **kwargs)
end
```

### DIFF
--- a/app/interfaces/api/v1/external_users/claims/advocate_claim.rb
+++ b/app/interfaces/api/v1/external_users/claims/advocate_claim.rb
@@ -27,7 +27,7 @@ module API::V1::ExternalUsers
       namespace '/' do
         desc 'DEPRECATED: Create an Advocate final claim. see advocates/final endpoint'
         post do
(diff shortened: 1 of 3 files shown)


jrmhaig:
Did you intend to check in the `Brewfile.lock.json` file?

jsugarman:
> Did you intend to check in the `Brewfile.lock.json` file?

no, removed.

although we probably could/should, but not here

jrmhaig (approved):
:+1:

PR #3861, opened 2021-04-01
[security] selectively bump ssri from 6.0.1 to 8.0.1

naseberry:
CVE-2021-27290
high severity
Vulnerable versions: >= 5.2.2, < 8.0.1
Patched version: 8.0.1
ssri 5.2.2-8.0.0, fixed in 8.0.1, processes SRIs using a regular expression which is vulnerable to a denial of service. Malicious SRIs could take an extremely long time to process, leading to denial of service. This issue only affects consumers using the strict option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "resolutions": {
     "node-forge": "^0.10.0",
+    "ssri": "^8.0.1",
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
💯 👍

PR #3862, opened 2021-04-01
[security] selectively bump is-svg from 3.0.0 to 4.3.1

naseberry:
CVE-2021-28092
high severity
Vulnerable versions: >= 2.1.0, < 4.2.2
Patched version: 4.2.2
The is-svg package 2.1.0 through 4.2.1 for Node.js uses a regular expression that is vulnerable to Regular Expression Denial of Service (ReDoS). If an attacker provides a malicious string, is-svg will get stuck processing the input for a very long time.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3605,6 +3605,11 @@ fast-levenshtein@^2.0.6:
   resolved "https://registry.yarnpkg.com/fast-levenshtein/-/fast-levenshtein-2.0.6.tgz#3d8a5c66883a16a30ca8643e851f19baa7797917"
   integrity sha1-PYpcZog6FqMMqGQ+hR8Zuqd5eRc=
 
+fast-xml-parser@^3.19.0:
+  version "3.19.0"
+  resolved "https://registry.yarnpkg.com/fast-xml-parser/-/fast-xml-parser-3.19.0.tgz#cb637ec3f3999f51406dd8ff0e6fc4d83e520d01"
+  integrity sha512-4pXwmBplsCPv8FOY1WRakF970TjNGnGnfbOnLqjlYvMiF1SR3yOHyxMR/YCXpPTOspNF5gwudqktIP4VsWkvBg==
+
 fastest-levenshtein@^1.0.12:
   version "1.0.12"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -4858,12 +4858,12 @@ is-string@^1.0.5:
   resolved "https://registry.yarnpkg.com/is-string/-/is-string-1.0.5.tgz#40493ed198ef3ff477b8c7f92f644ec82a5cd3a6"
   integrity sha512-buY6VNRjhQMiF1qWDouloZlQbRhDPCebwxSjxMjxgemYT46YMd2NR0/H+fBhEfWX4A/w9TBJ+ol+okqJKFE6vQ==
 
-is-svg@^3.0.0:
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/is-svg/-/is-svg-3.0.0.tgz#9321dbd29c212e5ca99c4fa9794c714bcafa2f75"
-  integrity sha512-gi4iHK53LR2ujhLVVj+37Ykh9GLqYHX6JOVXbLAucaG/Cqw9xwdFOjDM2qeifLs1sF1npXXFvDu0r5HNgCMrzQ==
+is-svg@^3.0.0, is-svg@^4.3.1:
```
jrmhaig, line 4861: Should version 3.0.0 be removed from this list?

naseberry: Unfortunately this is how yarn lock works when you selectively resolve a dependency.

See https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md for full spec

jsugarman: >If one of the dependencies has a new version available and the available version is within the specified version range in the package.json, then the newest dependency will be installed.

jsugarman: https://robertcooper.me/post/how-yarn-lock-files-work-and-upgrading-dependencies

jsugarman (approved):
👍

yarn.lock
jsugarman: LGTM

PR #3863, opened 2021-04-01
[CBO-1660] Migrate certification form to GOVUK design system

naseberry:
#### What
Migrate certification form to the GOVUK design system form builder

#### Ticket
[Migrate certification form to the GOVUK design system form builder](https://dsdmoj.atlassian.net/browse/CBO-1660)

#### Why
Part of a larger change to:

### DIFF
--- a/app/form_builders/gds_adp_form_builder.rb
+++ b/app/form_builders/gds_adp_form_builder.rb
@@ -1,0 +1,7 @@
+class GdsAdpFormBuilder < AdpFormBuilder
+  delegate :content_tag, :tag, :safe_join, :link_to, :capture, to: :@template
+
+  include ActionView::Helpers::FormTagHelper
+  include ActionView::Context
+  include GOVUKDesignSystemFormBuilder::Builder
+end
--- a/app/models/certification.rb
+++ b/app/models/certification.rb
@@ -18,20 +18,8 @@ class Certification < ApplicationRecord
   belongs_to :certification_type
 
   validates_with CertificationValidator
-  validate :at_least_one_boolean_selected
-
-  acts_as_gov_uk_date :certification_date, error_clash_behaviour: :override_with_gov_uk_date_field_error
 
   def perform_validation?
     true
-  end
-
-  private
-
-  def at_least_one_boolean_selected
-    return unless claim.is_a?(Claim::AdvocateClaim)
-
-    return if certification_type.present?
-    errors[:base] << 'You must select one option on this form'
   end
 end
--- a/app/validators/certification_validator.rb
+++ b/app/validators/certification_validator.rb
@@ -1,11 +1,15 @@
 class CertificationValidator < BaseValidator
   class << self
     def mandatory_fields
-      %i[certified_by certification_date]
+      %i[certification_type_id certified_by certification_date]
     end
   end
 
   private
+
+  def validate_certification_type_id
+    validate_presence(:certification_type_id, :blank) if @record&.claim&.agfs?
+  end
 
   def validate_certified_by
     validate_presence(:certified_by, :blank)
--- a/app/validators/certification_validator.rb
+++ b/app/validators/certification_validator.rb
@@ -8,7 +8,8 @@ class CertificationValidator < BaseValidator
   private
 
   def validate_certification_type_id
-    validate_presence(:certification_type_id, :blank) if @record&.claim&.agfs?
+    return unless @record&.claim&.agfs? && @record&.claim&.final?
+    validate_presence(:certification_type_id, :blank)
   end
 
   def validate_certified_by
--- a/app/views/external_users/certifications/new.html.haml
+++ b/app/views/external_users/certifications/new.html.haml
@@ -1,114 +1,43 @@
 = content_for :page_title, flush: true do
   = t(".page_title.#{present(@claim).type_identifier}")
 
-= render partial: 'errors/error_summary', locals: { ep: @certification, error_summary: t('.error_summary') }
-
 = render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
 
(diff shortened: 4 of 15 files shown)


app/views/external_users/certifications/new.html.haml
```diff
@@ -1,114 +1,43 @@
...
+      - if @claim.agfs?
+        = t('.agfs_claim_collection_html')
 
-            %li
-              this work has not been and will not be the subject of any other claim for remuneration from criminal legal aid.
+        = f.govuk_collection_radio_buttons :certification_type_id,
+          CertificationType.post_may_2015.agfs,
+          :id,
+          :name,
+          legend: { text: t('.certification_type_post_may'), size: 's' }
```
jsugarman: maybe horribly long but i think we should have the year for consistency
```suggestion
          legend: { text: t('.certification_type_post_may_2015'), size: 's' }
```

app/models/certification.rb
```diff
@@ -17,21 +17,16 @@ class Certification < ApplicationRecord
   belongs_to :claim, class_name: 'Claim::BaseClaim'
   belongs_to :certification_type
 
+  validates :certification_type_id, presence: true, if: :claim_agfs?
```
jsugarman: This should ideally follow the existing validator pattern that uses `CertificationValidator` 

also, though not from this PR,  but i think the `claim_agfs?` is out of date or wrong. I think all agfs claims may require this, but need to check that with the business. see related comment below

app/models/certification.rb
```diff
@@ -17,21 +17,16 @@ class Certification < ApplicationRecord
...
 
   private
 
-  def at_least_one_boolean_selected
-    return unless claim.is_a?(Claim::AdvocateClaim)
-
-    return if certification_type.present?
-    errors[:base] << 'You must select one option on this form'
+  def claim_agfs?
+    claim.is_a?(Claim::AdvocateClaim)
```
jsugarman: Not from this PR, but want to check this business logic as this implies only agfs "final" claims require certification type, which may be true but if so then the check could be `agfs? && final?` or similar i believe.

jsugarman: and method should be called `claim_agfs_final?` or similar

app/models/certification.rb
jsugarman: I am looking to add a commit to this effect ^^

app/form_builders/gds_adp_form_builder.rb
```diff
@@ -0,0 +1,7 @@
+class GdsAdpFormBuilder < AdpFormBuilder
```
jrmhaig, line 1: As ADP is an old name for CCCD can it be dropped from this name, and so become `GdsFormBuilder`?

I presume that the ultimate goal is that this form builder will completely replace `AdpFormBuilder`.

naseberry: This new form builder is to show the inheritance & composition of AdpFormBuilder and GOVUKDesignSystemFormBuilder.

This should be short lived and will be removed once we have migrate to and using the GOVUKDesignSystemFormBuilder as the default.

app/validators/certification_validator.rb
```diff
@@ -1,12 +1,16 @@
...
     def mandatory_fields
-      %i[certified_by certification_date]
+      %i[certification_type_id certified_by certification_date]
     end
   end
 
   private
 
+  def validate_certification_type_id
+    validate_presence(:certification_type_id, :blank) if @record&.claim&.agfs?
```
jsugarman: To match current logic  this needs to be 

```suggestion
    validate_presence(:certification_type_id, :blank) if @record&.claim&.agfs? && @record&.claim&.final?
```
or 
```suggestion
    validate_presence(:certification_type_id, :blank) if @record&.claim.is_a?(Claim::AdvocateClaim)
```

i prefer the less specific nature of the first myself, even if a bit messy

jsugarman: OR
```suggestion
  return unless if @record&.claim&.agfs? && @record&.claim&.final?
  validate_presence(:certification_type_id, :blank) 
```

spec/controllers/external_users/certifications_controller_spec.rb
```diff
@@ -156,10 +156,10 @@
     context 'invalid certification' do
       it 'redirects to new' do
         params = valid_certification_params(claim, certification_type)
-        params['certification']['certification_type_id'] = 9999
+        params['certification']['certification_type_id'] = nil
         post :create, params: params
         expect(response).to render_template(:new)
-        expect(assigns(:certification).errors.full_messages).to eq(['You must select one option on this form'])
+        expect(assigns(:certification).errors.messages[:certification_type_id]).to eq(['Choose a certification type'])
```
jsugarman: this should not be tested in a controller spec really - again, i know this is not you

jsugarman (changes requested):
See important comment on when to apply `certification_type_id` validation.

Also we should have some unit tests on the certification_validator. However it looks like there is no cerification_validator_spec. 🤦 and the unit tests are handled in both the controller spec (in one place) and the model spec (where it is more appropriate but ideally they would be in a new certification_validator_spec. and moved out of the model, with a model test to ensure it is using the CertificationValidator.

Do you want me to add something for this?

app/validators/certification_validator.rb
jsugarman: have applied commit for this

spec/controllers/external_users/certifications_controller_spec.rb
jsugarman: have applied commit for this

jsugarman (approved):
👍 looks good, have refactored the specs and ensured the same validation logic is applied now

PR #3864, opened 2021-04-01
[CBO-1754] Convert documents to PDF in a background job

jrmhaig:
#### What

When an evidence document is uploaded, use a background job to convert it to PDF instead of doing it inline.

#### Ticket

[Background task for creating converted preview documents](https://dsdmoj.atlassian.net/browse/CBO-1754)

#### Why

There are some problems with converting documents before sending a response back to the user:

* If the conversion fails then it cannot be retried and so the download needs to be repeated.
* To the user the download does not appear to complete until the conversion has finished and this can be several seconds. This is an unnecessary delay.

#### How

Create a new background job, `ConvertDocumentJob`.

--------

#### TODO (wip)

 - [x] Fix any tests that are broken due to newly created documents not having a converted preview
 - [x] Replace tests related to the conversion process, as appropriate
 - [x] Test that conversion job is being created and performed
 - [x] Cater for attempting to view a preview if a preview has not been created yet
   - [x] Do not display the 'preview' link if the preview does not exist

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -35,6 +35,8 @@ class DocumentsController < ApplicationController
   end
 
   def show
+    raise ActiveRecord::RecordNotFound, 'Preview not found' unless document.converted_preview_document.attached?
+
     redirect_to document.converted_preview_document.blob.service_url(disposition: :inline)
   end
 
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,11 +1,11 @@
 ---
 :logfile: ./log/sidekiq.log
 :queues:
+  - [convert_document, 3]
   - [mailers, 3]
   - [default, 2]
   - [claims, 2]
   - [stats_reports, 1]
-  - [convert_document, 1]
 development:
   :verbose: true
   :concurrency: 5
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -5,6 +5,7 @@
   - [default, 2]
   - [claims, 2]
   - [stats_reports, 1]
+  - [convert_document, 1]
 development:
   :verbose: true
   :concurrency: 5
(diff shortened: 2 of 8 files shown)


app/controllers/documents_controller.rb
```diff
@@ -35,6 +35,8 @@ def index
   end
 
   def show
+    raise ActiveRecord::RecordNotFound, 'Preview not found' unless document.converted_preview_document.attached?
```
jsugarman, line 38: are we wanting to error the system for this?

config/sidekiq.yml
```diff
@@ -5,6 +5,7 @@
   - [default, 2]
   - [claims, 2]
   - [stats_reports, 1]
+  - [convert_document, 1]
```
jsugarman: **not a blocker** might be worth moving these to the top too so they [are always processed first](https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues)

spec/requests/documents_spec.rb
```diff
@@ -83,11 +85,18 @@
...
 
+    it_behaves_like 'download evidence document' do
       let(:expected_disposition) { 'inline' }
     end
+
+    context 'when the preview does not exist' do
+      let(:document) { create :document, external_user: document_owner }
+      let(:document_owner) { external_user }
+
+      it { expect { show_document }.to raise_error(ActiveRecord::RecordNotFound) }
```
jsugarman, line 98: as for the other comment regarding the raising of the error. what happens if we do not raise the error here, will it not bubble up to sentry anyway? does the view hide the show link if there is no preview anyway?

jsugarman (approved):
looks ok.

one suggestion about job ordering and one concern about the error raising but 👍 .

might want to wait til monday to deploy but as you like.

spec/requests/documents_spec.rb
jrmhaig: The main reason to raise an error here is to generate a 404 response. And yes, the link does not appear if there is no preview so this is not likely to happen anyway.

PR #3869, opened 2021-04-12
Move sanity check into SQL query

jrmhaig:
#### What

Update the sanity check for Paperclip to Active Storage migration so that it is performed in the database.

#### Ticket

N/A

#### Why

Currently, each migrated Active Storage record is individually checked to make sure that the Paperclip version of the file name and checksum match the Active Storage versions. For a large number of files this is very time consuming and memory intensive.

#### How

Rewrite the `validate` method in `lib/tasks/rake_helpers/storage.rb` to perform the check in SQL instead of fetching all records and then checking them individually.


jsugarman (approved):
👍 Have tested on dev-lgfs

PR #3870, opened 2021-04-12
Add feeback during migration and prevent AS record deletion

jsugarman:
#### What/Why
feedback raw sql results to console to show progress
and prevent storage:rollback being run on production
as it results in unrecoverable files.


lib/tasks/storage.rake
```diff
@@ -8,6 +8,7 @@ namespace :storage do
 
   desc 'Rollback asset migration'
   task :rollback, [:model] => :environment do |_task, args|
+    production_protected
```
jrmhaig, line 11: 👍

jrmhaig (approved):
Looks good.

PR #3872, opened 2021-04-12
Purge attachments when records are deleted

jrmhaig:
Active Storage does not automatically purge blobs when the attachments
are removed. This PR will configure the `Message` and `Stats::StatsReport`
models to purge any linked attachments when the record is deleted.

### DIFF
--- a/spec/models/message_spec.rb
+++ b/spec/models/message_spec.rb
@@ -204,17 +204,17 @@ RSpec.describe Message, type: :model do
   describe '#destroy' do
     subject(:destroy_message) { message.destroy }
 
-    before { message }
+    let!(:message) { create(:message, trait) }
 
     context 'without an attachment' do
-      let(:message) { create(:message) }
+      let(:trait) { nil }
 
       it { expect { destroy_message }.not_to change(ActiveStorage::Attachment, :count) }
       it { expect { destroy_message }.not_to change(ActiveStorage::Blob, :count) }
     end
 
     context 'with an attachment' do
-      let(:message) { create(:message, :with_attachment) }
+      let(:trait) { :with_attachment }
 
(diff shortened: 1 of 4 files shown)


spec/models/message_spec.rb
```diff
@@ -201,6 +201,26 @@
     end
   end
 
+  describe '#destroy' do
+    subject(:destroy_message) { message.destroy }
+
+    before { message }
+
+    context 'without an attachment' do
+      let(:message) { create(:message) }
```
jsugarman: **Not a blocker** but can use bang to create the message immediately perhaps, not tried
```suggestion
      let!(:message) { create(:message) }
```

However, it might be clearer, since you do not access the `message` anyway, to use  a separate before hook rather than add the indirection.

```
before
 create(:message)
end
```
```
before
 create(:message, :with_attachment)
end
```

jsugarman (approved):
looks ok, just a suggestion on the tests

spec/models/message_spec.rb
jrmhaig: `message` is used in the subject on line 205 but it needs to be touched so that the `change` matchers work correctly - the `ActiveStorage::Attachment` and `ActiveStorage::Blob` records are not created until the message is created so the count would be incorrect at the start of the test.

Rubocop, for some reason, doesn't detect that `message` is being used and so using a `let!` results in a warning.

I think this would work:

```ruby
describe '#destroy' do
  context 'with an attachment' do
    let!(:message) { create(:message, :with_attachment) }

    it { expect { message.destroy }.to change(ActiveStorage::Attachment, :count).by(-1) }
    it { expect { message.destroy }.to change(ActiveStorage::Blob, :count).by(-1) }
  end
end
```

and it should satisfy Rubocop but it does without a `subject`.

jsugarman: your trait method is clear, so good to go

PR #3886, opened 2021-04-14
Treat zero `calculate_distance` as valid

jsugarman:
#### What
Treat zero `calculate_distance` as valid

#### Ticket

[CBO-1776](https://dsdmoj.atlassian.net/browse/CBO-1776)

#### Why
Currently with the google maps API broken
the request is returning nil which is converted
to zero and the expense considered invalid.

#### How
To allow for google maps API failures
without raising a validation error to the user
we must allow zeros to not be invalid.

This allows the user to enter a travel distance 
manually which case workers can then vet manually.

### DIFF
--- a/app/validators/expense_validator.rb
+++ b/app/validators/expense_validator.rb
@@ -66,7 +66,7 @@ class ExpenseValidator < BaseValidator
   end
 
   def validate_calculated_distance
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
:+1:

naseberry (approved):

PR #3887, opened 2021-04-14
Amend calculated_distance minimum validation

jsugarman:
#### What
Amend calculated_distance minimum validation

#### Ticket

N/A

#### Why
A simpler, and IMO clearer, way to loosen
the validation that causes Maps API failures to
prevent users proceeding.

#### How
Rather than ensure a non-zero value before attempting
to validate we can just set the minimum to be 0.0. A
failed API call, returning 0, would then be considered valid,
while a negative decimal would be invalid. In addition this
validates upper limit by default (of 200_000).

### DIFF
--- a/app/validators/expense_validator.rb
+++ b/app/validators/expense_validator.rb
@@ -66,8 +66,8 @@ class ExpenseValidator < BaseValidator
   end
 
   def validate_calculated_distance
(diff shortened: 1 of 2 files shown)


naseberry (approved):
👍

PR #3888, opened 2021-04-15
[CBO-1777] Update googlemaps-services for Ruby 2.7 support

jrmhaig:
#### What

Update the `googlemaps-services` gem to use this branch: https://github.com/jrmhaig/googlemaps-services/tree/update_http
#### Ticket

[Mileage calculation intermittent failures](https://dsdmoj.atlassian.net/browse/CBO-1777)

#### Why

The `googlemaps-services` gem has a dependency on the `http` gem pinned to a version that does not support Ruby 2.7. This results in intermittent failures. A PR has been submitted to `googlemaps-services` (https://github.com/amrfaissal/googlemaps-services/pull/13) but it has not yet been merged.

Once the PR has been merged and a new version released then we need to update `Gemfile` to use the released version again.


jsugarman:
I notice this gem is not very popular. from a quick view i wonder whether this one is worth a look as seesm more popular but even older
 
https://github.com/edwardsamuel/google-maps-services-ruby

jrmhaig:
@jsugarman I am currently looking at the [`geocoder`](https://github.com/alexreisner/geocoder) gem to see if we can use it instead.

**Update:** It looks as though our use case is out of the scope of `geocoder` so that is not the answer. See https://github.com/alexreisner/geocoder/issues/815

jsugarman (approved):
👍 as an interim measure this looks good

PR #3889, opened 2021-04-15
[CBO-1777] Remove googlemaps-services gem

jrmhaig:
#### What

Remove `googlemaps-services` gem.

#### Ticket

[Mileage calculation intermittent failures](https://dsdmoj.atlassian.net/browse/CBO-1777)

#### Why

The `googlemaps-services` gem appears to be abandoned so there may not be an update to support Ruby 2.7. All it does for us is make a request to the Google API so this could be done directly.

#### How

Replace the use of `Directions` from this gem with a `RestClient.get`.

#### Notes

I have move some files around but, unfortunately, the diff in Github is a bit confused making it look like some files were deleted, others created and other with significant modifications. The moved files are:

| Old filename | New filename
|---|---|
| `app/services/expenses/travel_distance_calculator.rb` | `app/services/distance_calculator_service.rb` |
| `app/services/maps/distance_calculator.rb` | `app/services/distance_calculator_service/google_api.rb` |
| `app/models/maps/directions_results.rb` | `app/services/distance_calculator_service/directions.rb` |
| `spec/services/expenses/travel_distance_calculator_spec.rb` | `spec/services/distance_calculator_service_spec.rb` |
| `spec/services/maps/distance_calculator_spec.rb` | `spec/services/distance_calculator_service/google_api_spec.rb` |
| `spec/models/maps/directions_results_spec.rb` | `spec/services/distance_calculator_service/directions_spec.rb` |

Additionally, the way that `GoogleApi` and `Direction`  (previously `DistanceCalculator` and `DirectionsResults`) are used is reversed. That is to say, where the distance was originally found with:

```ruby
distance = Maps::DistanceCalculator.call(origin, destination)
# Maps::DistanceCalculator queries the Google API and saves the
# results in a Maps::DirectionsResults instance
```

it is now found with:

```ruby
distance = DistanceCalculatorService::Directions.new(origin, destination).max_distance
# DistanceCalculatorService::Directions uses DistanceCalculatorService::GoogleAPI
# to call the API and then handles the results
```

--------

#### TODO (wip)

 - [x] Some error handling


app/services/distance_calculator_service.rb
```diff
@@ -0,0 +1,67 @@
+require 'dry/monads/all'
+
+class DistanceCalculatorService
+  include Dry::Monads
+
+  def initialize(claim, params)
+    @claim = claim
+    @params = params
+  end
+
```
jsugarman, line 12: can add convenience short cut

```
def self.call(claim, params)
  new(claim, params).call
end
```

app/services/distance_calculator_service/directions.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def results
+      @results ||= DistanceCalculatorService::DirectionsResult.new(routes)
+    end
+
+    def routes
+      @routes ||= begin
+        JSON.parse(directions.body)['routes']
+      rescue StandardError => e
+        log("Failed to calculating distance form #{@origin} to #{@destination}", error: e, level: :error)
```
jsugarman:
```suggestion
        log("Failed to calculate distance form #{@origin} to #{@destination}", error: e, level: :error)
```

jsugarman (approved):
Looks ok, assuming you have tested yourself on staging or similar. might be worth getting Alan to excercise too

Eventually much of this logic should be extracted to a gem, which i am working on

app/services/distance_calculator_service.rb
```diff
@@ -9,52 +9,38 @@ def initialize(claim, params)
...
-  def response(value=nil, error: nil)
-    OpenStruct.new(value: value, error: error)
-  end
 
-  def valid_inputs?
-    result = validate_inputs
-    self.validation_error = result if result.error
-    validation_error.nil?
+  def response
+    OpenStruct.new(value: nil, error: validate_inputs)
```
jsugarman: openstruct is not very effiecient, prefer struct if attributes are fixed

jrmhaig: I have updated this to use Struct instead of OpenStruct.

PR #3891, opened 2021-04-15
Bump `Max` for RSpec/NestedGroups 3 --> 4

jsugarman:
#### What
Bump `Max` for RSpec/NestedGroups 3 --> 4

#### Why
Alot of violations but 3 seems rather low
and test different claim types is often
clearer by adding a layer of context for
the claim type.


jrmhaig (approved):
:+1:

I wouldn't want to increase this much further as there are certainly cases where the nesting needs to be cleaned up.

PR #3893, opened 2021-04-16
Refactor distance caclulator and extend spec

jsugarman:
#### What
Refactor distance caclulator and extend spec

#### Ticket

[CBO-1777 mileage intermittent failures](https://dsdmoj.atlassian.net/browse/CBO-1777)

#### Why
replacement of the directions API gem
means we have placed a forked version in
so we should rerecord cassettes. The spec
does not cover the error handling path either
so this has been added.

### DIFF
--- a/app/services/maps/distance_calculator.rb
+++ b/app/services/maps/distance_calculator.rb
@@ -18,23 +18,26 @@ module Maps
     end
 
     def call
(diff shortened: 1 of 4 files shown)


jrmhaig:
This is going to conflict a lot with #3889, where these files are being moved around and refactored.

PR #3895, opened 2021-04-19
Update .rubocop_todo.yml

jrmhaig:
Specifically, permit RSpec/MultipleExceptions in two places in `spec/models/document_spec.rb`.

This is to allow the test:

```ruby
it 'schedules a ConvertDocumentJob' do
  expect { document_save }
    .to(have_enqueued_job(ConvertDocumentJob).with { |id| expect(id).to eq document.reload.to_param })
end
```

According to the [RSpec documentation](https://relishapp.com/rspec/rspec-rails/docs/matchers/have-enqueued-job-matcher#checking-passed-arguments-to-job,-using-a-block) this is the correct way to do this but Rubocop doesn't like it. This fix is to get the pipeline moving but I will see if I can find a workaround.


kmahern (approved):
👍

PR #3897, opened 2021-04-20
Prevent convert retry if document deleted

jrmhaig:
#### What

Silently terminate a document convert job if the document cannot be found.

#### Ticket

N/A

#### Why

If a document is uploaded and then deleted before the convert job has been performed this causes an error, as the document cannot be found, resulting in the job being retried. This is unnecessary.

#### How

Prevent the convert job from raising a 'record not found' error.

### DIFF
--- a/app/jobs/convert_document_job.rb
+++ b/app/jobs/convert_document_job.rb
@@ -2,7 +2,9 @@ class ConvertDocumentJob < ApplicationJob
   queue_as :convert_document
 
   def perform(document_id)
-    document = Document.find(document_id)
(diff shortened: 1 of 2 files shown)


jrmhaig:
~I've marked this as 'do not merge' for the moment because it looks as though there are cases where the convert job is being run before the document has been saved in the database. This causes the 'record not found' resulting in a retry, correctly.~

#3898 adds a 30 seconds delay to job so that this problem no longer exists, so this PR can now be merged.

kmahern (approved):
👍

PR #3898, opened 2021-04-20
Add delay to DocumentConvertJob call

jrmhaig:
#### What

Wait at least 30 seconds before attempting to convert a document. This is related to #3897 

#### Ticket

N/A

#### Why

Many of the backgrounded document convert jobs fail on the first attempt because of a race condition. The Sidekiq process fails to find the document in the database that has just been saved.

The conversion succeeds on retry but there is a possibility that the document has genuinely been deleted, in which case we want the job to silently fail (see #3897).

#### How

Delay the conversion by adding `set(wait: 30.seconds)` to the call to `ConvertDocumentJob`.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -95,7 +95,7 @@ class Document < ApplicationRecord
   end
 
   def convert_document
-    ConvertDocumentJob.perform_later(to_param)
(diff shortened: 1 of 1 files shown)


kmahern (approved):
👍

PR #3904, opened 2021-04-21
[CFP-92] Allow zip files with duplicate filenames

jrmhaig:
#### What

Append an index to evidence documents when added to a zip file for caseworkers to download.

#### Ticket

[Duplicate file names causes failure to download zip file](https://dsdmoj.atlassian.net/browse/CFP-92)

#### Why

The creation of a zip file fails if two files have the same name. Appending a file number avoids this.

#### How

If there are two files named `filename.doc` then they will be added to the zip file as `filename.doc` and `filename (1).doc`.

### DIFF
--- a/app/services/s3_zip_downloader.rb
+++ b/app/services/s3_zip_downloader.rb
@@ -14,11 +14,10 @@ class S3ZipDownloader
   private
 
   def build_zip_file(documents, bundle)
-    filename_counts = {}
     Dir.mktmpdir("#{@claim.case_number}-") do |tmp_dir|
       Zip::File.open(bundle, Zip::File::CREATE) do |zip_file|
-        documents.map(&:document).each do |document|
-          zip_file.add(get_filename(document.filename, filename_counts), local_file(document, tmp_dir))
+        documents.map(&:document).each_with_index do |document, i|
+          zip_file.add("#{i}_#{document.filename}", local_file(document, tmp_dir))
         end
       end
     end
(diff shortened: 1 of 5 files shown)


spec/factories/documents.rb
```diff
@@ -25,38 +25,35 @@
...
+    transient do
+      file_type { :pdf }
+      content_type { { pdf: 'application/pdf', docx: 'application/msword' }[file_type] }
+      sequence(:filename) { |n| "testfile#{n}.#{file_type}" }
+      file_source do
+        {
+          pdf: 'features/examples/longer_lorem.pdf',
+          docx: 'features/examples/shorter_lorem.docx'
+        }[file_type]
+      end
```
jsugarman: unless this is necessary fo this fix can you put this in a separate PR, as it seems like scope creep. The use of simple traits seems simpler in any event.

jrmhaig: I did this mainly so that I could set the filename, to ensure that two files had the same name. I'll see if I can restore the traits.

app/services/s3_zip_downloader.rb
```diff
@@ -14,10 +14,11 @@ def generate!
   private
 
   def build_zip_file(documents, bundle)
+    filename_counts = {}
     Dir.mktmpdir("#{@claim.case_number}-") do |tmp_dir|
       Zip::File.open(bundle, Zip::File::CREATE) do |zip_file|
         documents.map(&:document).each do |document|
-          zip_file.add(document.filename, local_file(document, tmp_dir))
+          zip_file.add(get_filename(document.filename, filename_counts), local_file(document, tmp_dir))
         end
```
jsugarman: Did the previous incarnation of this not work. i prefer its simplicity

```
documents.map(&:document).each_with_index do |document, i|
  zip_file.add("#{i}_#{document.filename}", local_file(document, tmp_dir))
end
```

jrmhaig: It did work but I thought that the `filename (1).doc` is closer to what people expect when downloading a second file with the same name. Adding the index to every file implies an ordering (although this probably isn't really a problem).

jrmhaig: I have changed it back now.

jsugarman (approved):
👍

PR #3905, opened 2021-04-22
[CFP-80] Remove paperclip

jrmhaig:
#### What

Remove the paperclip gem and any related code from the codebase.

#### Ticket

[Remove Paperclip and cleanup codebase](https://dsdmoj.atlassian.net/browse/CFP-80)

#### Why

We have fully migrated to using Active Storage for asset management so the Paperclip gem is no longer required.

#### How

See the checklist in `docs/active_storage_migration.md`.

--------

#### TODO (wip)

 - [ ] Remove dummy asset files on S3 used for testing in the dev, dev-lgfs and staging environments. This must be done before this PR is merged as it depends on the storage Rake tasks, which are being deleted.
 - [ ] Remove Paperclip fields from the database. These include the fields called `*_file_name`, `*_file_size` and `*_content_type` in the `stats_reports`, `messages` and `documents` tables.
 - [ ] Ticket to find and remove orphan document, message and stats_reports records.
 - [ ] Ticket to find and remove active storage records with no s3 object associated with them.


jsugarman (approved):
Looks ok

1.  has this be manually tested on dev-lgfs yet by yourself and others?

2. we need to remember to try remove these entirely as they are not used from what i can tell. Separate PR after or before this:
- app/interfaces/api/entities/document.rb
- app/interfaces/api/entities/message.rb

3. we should book in a date to remove the paperclip columns too - how long to leave it or immediately?

jrmhaig:
> 1. has this be manually tested on dev-lgfs yet by yourself and others?

I did put it on dev-lgfs for a bit and tested it myself. It hasn't been tested by others yet.

> 2. we need to remember to try remove these entirely as they are not used from what i can tell. Separate PR after or before this:
> 
> * app/interfaces/api/entities/document.rb
> * app/interfaces/api/entities/message.rb
> 

I think this is part of a bigger job to review what is in `app/interfaces/api/entities`. I have not yet worked out how they are used.

> 3. we should book in a date to remove the paperclip columns too - how long to leave it or immediately?

There isn't a great hurry to do this but as soon as Paperclip has been removed then these can be deleted.

PR #3908, opened 2021-04-26
[CFP-73] Migrate user forms to GOVUK design system

naseberry:
#### What
Migrate certification form to the GOVUK design system form builder

#### Ticket
[Migrate user forms to the GOVUK design system form builder](https://dsdmoj.atlassian.net/browse/CFP-73)

#### Why
Part of a larger change to:

### DIFF
--- a/app/views/external_users/admin/external_users/_form.html.haml
+++ b/app/views/external_users/admin/external_users/_form.html.haml
@@ -1,62 +1,50 @@
-= form_for [:external_users, :admin, @external_user] do |f|
+= form_with model: [:external_users, :admin, @external_user], builder: GdsAdpFormBuilder, local: true do |f|
 
   = f.fields_for :user do |user_fields|
-    .form-group
-      %a{ name: 'first_name' }
-      = user_fields.label :first_name, t('.first_name'), class: 'form-label-bold'
-      = user_fields.text_field :first_name, class: 'form-control'
-    .form-group
-      %a{ name: 'last_name' }
-      = user_fields.label :last_name, t('.last_name'), class: 'form-label-bold'
-      = user_fields.text_field :last_name, class: 'form-control'
+    = user_fields.govuk_error_summary
 
-    .form-group
-      %a{ name: 'email' }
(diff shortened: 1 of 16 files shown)


app/views/external_users/admin/external_users/_form.html.haml
```diff
@@ -1,62 +1,50 @@
...
+          - if current_user.persona.provider.chamber?
+            = f.govuk_collection_radio_buttons :vat_registered,
+              [['Yes','true'],['No','false']],
+              :last,
+              :first,
+              inline: true,
+              legend: { text: t('.vat_registered'), size: 's' }
+
+    = f.govuk_text_field :supplier_number,
+      hint: { text: t('.supplier_number_hint') },
```
jsugarman, line 47: do we need to worry about the safari bug here i.e. adding hints/`aria-describedby` impact on buttons?

spec/models/user_spec.rb
```diff
@@ -39,7 +39,7 @@
   it { is_expected.to validate_length_of(:email).is_at_most(80).with_message('Email must be 80 characters or less') }
   it { is_expected.to allow_value('email@addresse.foo').for(:email) }
   it { is_expected.not_to allow_value('foo').for(:email) }
-  it { is_expected.to validate_length_of(:password).is_at_least(8).with_message('Email must be at least 8 characters') }
+  it { is_expected.to validate_length_of(:password).is_at_least(8).with_message('Password must be at least 8 characters') }
```
jsugarman, line 42: oops, good catch

jsugarman (approved):
LGTM 👍  - one pointer regarding the recent safari bug and its impact

I guess we don't have much in the way of cucumber features to cover these but I may be looking to add some due to the sandbox bug

app/views/external_users/admin/external_users/_form.html.haml
naseberry: Not here @jsugarman as this radio collection has no hint.

PR #3909, opened 2021-04-26
Rake task to find unattached S3 files

jrmhaig:
#### What

Rake task to identify S3 files that are not attached.

#### Ticket

N/A

#### Why

This allows us to see if there are files in S3 that can be deleted. This is more likely to happen on the development environments.

#### How

```bash
rails storage:s3_status
```

This will create three files;

* `tmp/s3_files.csv`: List of files with one or more `ActiveStorage::Attachment` records. Columns are; `id` in `ActiveStorage::Blob`, filename, number of attachments, key, last modified time and file size.
* `tmp/s3_orphaned_files.csv`: List of files that are not linked to an `ActiveStorage::Attachment`. Columns are; key, last modified time and file size.
* `tmp/s3_summary.txt`: Information about number of files and sum of file sizes.

The task can be executed as `rails storage:s3_status[interactive]`. This will allow the task to be stopped after each 1000 files from S3 is checked.

The summary on staging (at the moment) looks like this:

```
Files attached to Active Storage attachments: 537
Size of attached files: 269 MB
Files not attached to Active Storage attachments: 18318
Size of non-attached files: 513 MB
```

### DIFF
--- a/lib/tasks/rake_helpers/storage.rb
+++ b/lib/tasks/rake_helpers/storage.rb
@@ -1,0 +1,56 @@
+module Storage
+  def self.s3_files(interactive)
+    Aws.config[:credentials] = Aws::Credentials.new(Settings.aws.s3.access, Settings.aws.s3.secret)
+    s3 = Aws::S3::Client.new
+    data = {
+      attached: {
+        rows: [],
+        total: 0,
+        size: 0
+      },
+      orphaned: {
+        rows: [],
+        total: 0,
+        size: 0
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


jsugarman (commented):
You not deleted these files in the [other PR, which is approved](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3905)?

jrmhaig:
> You not deleted these files in the [other PR, which is approved](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3905)?

Yes, that was probably a mistake but I'll deal with it in a rebase.

jsugarman (approved):

PR #3915, opened 2021-04-28
Fix Performance/MapCompact Rubocop offences

jrmhaig:
#### What

Replace `map { ... }.compact` with `filter_map { ... }`, as appropriate.

#### Ticket

N/A

#### Why

The newly-added Performance/MapCompact Rubocop cop resulted in some new items in the 'to do' list. See #3912

`filter_map` is a new method for enumerables that was added in Ruby 2.7. It acts like `map` but only returns entries with a truthy value (everything except false or nil). See https://ruby-doc.org/core-3.0.1/Enumerable.html#method-i-filter_map


app/models/claim/base_claim.rb
```diff
@@ -321,10 +321,8 @@ def representation_orders
 
     def earliest_representation_order
       return if defendants.empty?
-      defendants
-        .map(&:earliest_representation_order)
-        .compact
-        .sort_by(&:representation_order_date).first
+
+      defendants.filter_map(&:earliest_representation_order).min_by(&:representation_order_date)
```
jsugarman, line 325: and `min_by`, but i trust this works for dates!

https://ruby-doc.org/core-2.7.3/Enumerable.html#method-i-min_by

jsugarman: certainly seems to
```
arr = [Date.today - 1, Date.today - 2, Date.today]
arr.min_by{|x| x.to_date }
```

jsugarman (approved):
cool, TIL `filter_map`, `min_by`

PR #3918, opened 2021-04-28
Update python dependencies

jsugarman:
#### What
Update python dependencies

#### Why
Inline with available alpine v3.13
python packages

see https://pkgs.alpinelinux.org/packages?name=py3-pip*&branch=v3.13&repo=community&arch=x86_64
for package availability.

#### How
Use python3/py3-pip instead of python/py-pip as the later
are no longer available in alpine 3.13


jrmhaig (approved):
:+1:

PR #3928, opened 2021-04-29
Enforce expected order of misc fee types by default

jsugarman:
#### What
Enforce expected order of misc fee types by default

#### Why
Because the order of the fees appears to impact fee calculation
of uplifts.

#### Explanation
The misc fee types are ordered by their description on the page.
Local postgres 9.6.21 orders strings with parentheses slightly
differently:
```
 [[40, "Confiscation hearings (half day uplift)"],
  [38, "Confiscation hearings (half day)"],...
 ]
```

this does not appear to be the case on any of the environments
though as they are ordering as expected.
```
 [[38, "Confiscation hearings (half day)"],
  [40, "Confiscation hearings (half day uplift)"],...
 ]
```

The order of the fees on the page then has an impact on fee
calculation for uplifts, via the JS in some way, such that
if an uplift is above its non-uplift fee equivalent it will
not be recalculated.

Order should not impact fee calculation! Need to check the
JS and why the order is important...but avoiding yak shaving
atm.


jsugarman:
Closing as I believe the feature test is correct but should be failing due to some change that has impacted defendant uplift calculation. 

see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/3930

PR #3929, opened 2021-04-30
Bump api endpoint deprecation for another month

jrmhaig:
We need to remove these endpoints but need sign off from Product.


jsugarman (approved):
👍

PR #2144, opened 2017-11-21
Add posix-spawn gem

jsugarman:
Getting alot of errors during document upload:

`Errno::ENOMEM: Cannot allocate memory - file -b --mime '/tmp/*`

posix-spawn works with paperclips dependency cocaine:
see http://blog.sundaycoding.com/blog/2014/02/05/fighting-paperclip-errno-enomem-error/

PR #3931, opened 2021-04-30
Remove posix-spawn gem

jrmhaig:
#### What

Remove posix-spawn gem.

#### Ticket

[Remove unused Gems](https://dsdmoj.atlassian.net/browse/CFP-81)

#### Why

The posix-spawn gem was added to fix memory issues with the cocaine (now terrapin) gem, which is a dependency of Paperclip. See #2144. Paperclip is no longer used so posix-spawn is not required.

The ticket also mentions the `climate_control` gem. This was removed in #3826 

#### How

Remove `gem 'posix-spawn', '~> 0.3.15'` from `Gemfile`.


jsugarman:
open question - Is there a way we can monitor memory, cpu and unicorn workers on the servers to ensure there is no negative impact from removal of this gem.

jsugarman (approved):

PR #3934, opened 2021-05-04
Fix defendant uplifts on supplementary claims

jsugarman:
#### What
Ensure feature test behave as prod, which currently has fee calc
edge case bug, and fix that bug.

#### To duplicate bug
On staging
 - Create a scheme 11 supplementary claim
 - Add two defendants
 - Add Misc fee of "Confiscation hearing (half day)" with quantity of one - rate: £131
 - Add Misc fee of "Confiscation hearing (half day uplift)" with quantity of one - rate £52.40
 Last fee should have a rate of £26.20 (this works on local machine)

#### Ticket

[https://dsdmoj.atlassian.net/browse/CFP-105](https://dsdmoj.atlassian.net/browse/CFP-105)

#### Why
Invalid calculations for the Supplementary claim
"Confiscation hearing (half day uplift)" are causing
invalid amounts to be calculated and thereby potentially
claimed.

#### How
Suttle differences in ordering of strings with braces between local hosted postgres,
circleci alpine postgres images and AWS RDS DB instances are causing an ordering difference.
This would be fine except that a minor javascript oversight then causes a double counting of the first
fee type on the page - which happens to be the non-uplift fee quantity "Confiscation hearing (half day)".
This doubling then causes a doubling of the fee types defendant uplift "Confiscation hearing (half day uplift)"
if it is claimed/visible.

This Javascript bug has no impact if the uplift is the first fee type on the page.

Local postgres 9.6.21 on darwin puts the uplift above the non-uplift equivalent so no doubling occurs. The test
suite used alpine hosted postgres which also puts put the uplift above the non-uplift so no doubling occurs, however
when a postgres image hosted on debian is used the reverse ordering happens and the bug occurs, and the feature test
fails.

So this PR uses a debian version of postgres to match prod-like environments as closely as possible. This PR also
enforces the ordering of fee types so that postgres versions will not have an impact in any event. The result being
that non-uplift fee types should always appear above there uplift fee type.

Finally the javascript bug is fixed by removing a misplaced class, `.fx-fee-group`, on the outer container
for all the fee type checkboxes. This is causing the first fee type to be double-counted because of line
`231` of `app/webpack/javascripts/modules/external_users/claims/fee_calculator/FeeCalculator.UnitPrice.js`

### DIFF
--- a/app/models/fee/misc_fee_type.rb
+++ b/app/models/fee/misc_fee_type.rb
@@ -5,7 +5,8 @@ class Fee::MiscFeeType < Fee::BaseFeeType
   AGFS_SUPPLEMENTARY_TYPES = (AGFS_SUPPLEMENTARY_ONLY_TYPES + AGFS_SUPPLEMENTARY_SHARED_TYPES).freeze
   TRIAL_ONLY_TYPES = %w[MIUMU MIUMO].freeze
 
-  default_scope { order(description: :asc) }
+  # default_scope { order(description: :asc) }
+  default_scope { order(Arel.sql("regexp_replace(\"fee_types\".\"description\", '[()]', '', 'g')")) }
 
   scope :without_supplementary_only, -> { where.not(unique_code: AGFS_SUPPLEMENTARY_ONLY_TYPES) }
   scope :supplementary, -> { where(unique_code: AGFS_SUPPLEMENTARY_TYPES) }
--- a/app/models/fee/misc_fee_type.rb
+++ b/app/models/fee/misc_fee_type.rb
@@ -5,7 +5,6 @@ class Fee::MiscFeeType < Fee::BaseFeeType
   AGFS_SUPPLEMENTARY_TYPES = (AGFS_SUPPLEMENTARY_ONLY_TYPES + AGFS_SUPPLEMENTARY_SHARED_TYPES).freeze
   TRIAL_ONLY_TYPES = %w[MIUMU MIUMO].freeze
 
-  # default_scope { order(description: :asc) }
   default_scope { order(Arel.sql("regexp_replace(\"fee_types\".\"description\", '[()]', '', 'g')")) }
 
   scope :without_supplementary_only, -> { where.not(unique_code: AGFS_SUPPLEMENTARY_ONLY_TYPES) }
--- a/features/claims/advocate/scheme_eleven/advocate_supplementary_claim_submit.feature
+++ b/features/claims/advocate/scheme_eleven/advocate_supplementary_claim_submit.feature
@@ -29,10 +29,10 @@ Feature: Advocate tries to submit a supplementary claim for miscellaneous fees (
     And I should see the advocate categories 'Junior,Leading junior,QC'
     And the following miscellaneous fee checkboxes should exist:
       | section | fee_description |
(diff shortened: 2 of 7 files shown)


app/models/fee/misc_fee_type.rb
```diff
@@ -5,7 +5,7 @@ class Fee::MiscFeeType < Fee::BaseFeeType
   AGFS_SUPPLEMENTARY_TYPES = (AGFS_SUPPLEMENTARY_ONLY_TYPES + AGFS_SUPPLEMENTARY_SHARED_TYPES).freeze
   TRIAL_ONLY_TYPES = %w[MIUMU MIUMO].freeze
 
-  default_scope { order(description: :asc) }
+  default_scope { order(Arel.sql("regexp_replace(\"fee_types\".\"description\", '[()]', '', 'g')")) }
```
jrmhaig, line 8: Is the difference in ordering to do with collation? See [this Stack Overflow question.](https://stackoverflow.com/questions/56395711/postgresql-sort-a-character-field-with-special-characters-numbers-and-alphabe)

features/claims/advocate/scheme_eleven/advocate_supplementary_claim_submit.feature
```diff
@@ -44,16 +44,16 @@ Feature: Advocate tries to submit a supplementary claim for miscellaneous fees (
       | miscellaneous | Wasted preparation fee |
 
     When I select an advocate category of 'Junior'
-    And I choose the 'Confiscation hearings (half day uplift)' miscellaneous fee with quantity of '1'
     And I choose the 'Confiscation hearings (half day)' miscellaneous fee with quantity of '2'
+    And I choose the 'Confiscation hearings (half day uplift)' miscellaneous fee with quantity of '1'
```
jrmhaig, line 48: Does the ordering of these matter?
* If no, then do the steps need to be changed?
* If yes, then do both cases need to be tested?

app/models/fee/misc_fee_type.rb
jsugarman: you'd hope so, but I tried collation first and it did not work

jsugarman: tried again but none of the following do what i'd expect/want for the unit test
```
default_scope { order(Arel.sql("\"fee_types\".\"description\" COLLATE \"POSIX\""))}
default_scope { order(Arel.sql("\"fee_types\".\"description\" COLLATE \"C\""))}
```
or any of the regional/language collations either

jsugarman: it seems to be differences between darwin/alpine and debian/amazon-linux

features/claims/advocate/scheme_eleven/advocate_supplementary_claim_submit.feature
jsugarman: no they don't, as mentioned in the commit they are merely to reflect the expected ordering

jsugarman: yes they do need testing. this was the only test at the time that caught the problem and is a general blackbox feature test

jrmhaig (approved):

PR #3940, opened 2021-05-05
[CFP-108] Fix migrations after paperclip extraction

jrmhaig:
#### What

Modify database migrations that added Paperclip attachments to explicitly create the four paperclip fields.

#### Ticket

[Smoke test failure](https://dsdmoj.atlassian.net/browse/CFP-108)

#### Why

The paperclip gem provides a helper for database migration to create the required fields.

These migrations, therefore, fail when the paperclip gem is removed. The solution is to replace `attachment` with the four fields everywhere it appears.

#### How

Replace

```ruby
t.attachment document
```

with

```ruby
t.string documnet_file_name
t.string documnet_content_type
t.integer documnet_file_size
t.datetime documnet_updated_at
```

(or equivalent) wherever it appears in the database migrations.

I have added the newly created `db/schema.rb` as validation that these changes are correct. The only differences are;

* Slight column ordering change in two tables.
* Addition of a foreign key for add_foreign_key `injection_attempts` on the `claims` table. This is unrelated to this PR and I am unclear why this has appeared.

### DIFF
--- a/db/migrate/20150416144357_alter_documents_add_paperclip_attachement.rb
+++ b/db/migrate/20150416144357_alter_documents_add_paperclip_attachement.rb
@@ -1,8 +1,14 @@
 class AlterDocumentsAddPaperclipAttachement < ActiveRecord::Migration[4.2]
   def change
     change_table(:documents) do |t|
-      t.remove      :document
-      t.attachment  :document
+      t.remove :document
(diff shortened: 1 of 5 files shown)


jsugarman (approved):
👍

PR #3943, opened 2021-05-06
[CFP-94] Add backtrace to redrafting logging

jrmhaig:
#### What

Add the backtrace to the logging of a failed redrafting.

#### Ticket

[Fix bug preventing redrafting after a rejection](https://dsdmoj.atlassian.net/browse/CFP-94)

#### Why

Occasionally providers get an error when attempting to redraft a rejected claim and they are then required to re-create the claim from scratch. It is not clear what is causing the error other than a `ActiveStorage::FileNotFoundError` exception without context.

#### How

Add `error.backtrace` to the log that is reported in Kibana. Also, remove the 'rescue' in `ExternalUsers::ClaimsController#clone_rejected_to_new_draft' as this effectively resets the error's backtrace and makes it appear as though the error occurred in this method.

### DIFF
--- a/app/models/claims/cloner.rb
+++ b/app/models/claims/cloner.rb
@@ -108,8 +108,6 @@ module Claims::Cloner
     draft = duplicate
     draft.transition_clone_to_draft!(author_id: author_id)
     draft
-  rescue StandardError => e
-    raise("Claims::Cloner.clone_rejected_to_new_draft failed with error '#{e.message}'")
   end
 
   # `other_claim` can be a draft instance of any kind of claim scheme (agfs or lgfs).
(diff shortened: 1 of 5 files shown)


jsugarman (approved):
👍

app/models/claims/cloner.rb
```diff
@@ -108,8 +108,6 @@ def clone_rejected_to_new_draft(author_id:)
     draft = duplicate
     draft.transition_clone_to_draft!(author_id: author_id)
     draft
-  rescue StandardError => e
```
jsugarman, line 111: What do we lose by not throwing this error here? aside from specs failing

jrmhaig: Good question. This exception gets caught in `ClaimsController`. This rescue is just adding `Claims::Cloner.clone_rejected_to_new_draft failed with error` to the start of the message and, as I discovered, resetting the backtrace so that the error appears to have originated here rather than the actual error.

jsugarman (approved):

PR #3946, opened 2021-05-06
[CFP-97] Correct sign up terms and conditions link

jsugarman:
#### What
Point sign up terms and conditions link to vendor version

#### Ticket

relates to [CFP-97](https://dsdmoj.atlassian.net/browse/CFP-97)

#### Why
Previous work had replaced an inline rendering of the t and cs
with a link to them. The link was to the general t and cs rather
than a different "vendor sign up for sandbox" t and cs.

#### How
Move and rename the vendor terms and conditions partial so it
it is more explicitly named and alongside other boilerplate html.

Add basic route to the vendor t and cs to enable easy linking via
rails url helpers.

#### How it looks
Screenshot 2021-05-06 at 17 53 44


spec/requests/devise/registrations/user_signup_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    context 'when on api-sandbox' do
+      around do |example|
+        with_env('api-sandbox') do
+          example.run
+        end
+      end
+
+      it {
+        request
+        expect(response).to render_template('devise/registrations/new')
```
jsugarman, line 43: replace with url helper

spec/requests/devise/registrations/user_signup_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+
+RSpec.describe 'User sign up', type: :request do
+  include Capybara::RSpecMatchers
+
+  describe 'GET #new' do
+    subject(:request) { get new_user_registration_path }
+
+    context 'when NOT on api-sandbox' do
+      around do |example|
+        with_env('production') do
```
jrmhaig: This (copied from `spec/controllers/external_users/registrations_controller_spec.rb`) is causing problems with the dependabot update PR for `database_cleaner`. Version 2.0 of that gem introduced a safeguard to prevent database modifications from happening if the environment is detected to be production.

If it is at all possible (and simple) to do this another way it would be great, but if not this isn't a blocker as there are other instances of this issue elsewhere anyway.

spec/requests/devise/registrations/user_signup_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+          example.run
+        end
+      end
+
+      it {
+        request
```
jrmhaig, line 29: Can this `request` (and likewise below on lines 42 and 47) be moved into the `around` (or `before`) block?

jrmhaig (approved):
A couple of comments but I'm happy as it is if you decide not to make changes.

spec/requests/devise/registrations/user_signup_spec.rb
jsugarman: this only changes the `ENV['ENV']` rather than `ENV['RAILS_ENV']` environment variable (which is not a typical rails app env var) but I will have a look. it can use `staging` as a value in any event, with no impact to tests or their underlying intention.

spec/requests/devise/registrations/user_signup_spec.rb
jsugarman: i tried this and it was not working in the around block. using a before block would not work, i believe, as the request must be made after the env var is set. A before and after block might work but would meaning changing the ENV var helper possibly.

PR #3947, opened 2021-05-07
[CFP-97] Prevent T & Cs agreed validation for provider's user creation

jsugarman:
Fix bug requiring Terms and conditions agreement

#### What
 - Enforce terms and conditions being "agreed" only
when self-registering on api-sandbox.
 - Fix bug on (checkbox item) list of roles (affects all environments)
 - Feature test all happy path scenarios for user creation by providers

#### Ticket

[CFP-97](https://dsdmoj.atlassian.net/browse/CFP-97)

#### Why
This fixes 2 bugs introduced during the migration to
DfE formbuilder. 

1. validation of terms and conditions acceptance by admins
    creating users in their providers. acceptance is only required
    by self registering users on the api-sandbox.
2. Inaccurate list of roles on the user creation page.

Currently API sandbox users cannot create new users in a provider
as a result. It has been agreed that user creation should not require
agreeing terms_and_conditions unless they are signing
themselves up - self registration. Self registration
is only possible on the api-sandbox in any event.

When an admin for a provider creates a user in their
provider, or when a case worker admin creates an external
user or case worker on the sandbox (or anywhere else) they
should not need to agree terms and conditions.

#### How
Use virtual attribute to require T&Cs to be agreed,
default to false, and set this to true in the relevant
controller (registrations controller)


app/views/external_users/admin/external_users/_form.html.haml
```diff
@@ -25,24 +25,21 @@
...
+
+            - if role.eql?('advocate') && current_user.persona.provider.chamber?
+              = f.govuk_collection_radio_buttons :vat_registered,
+                [['Yes','true'],['No','false']],
+                :last,
+                :first,
+                inline: true,
+                legend: { text: t('.vat_registered'), size: 's' }
+
+      / TODO: The hint needs to be checked because an LGFS & AGFS "firm" might confuse this with an LGFS supplier number for the user. I believe the value must only be the AGFS supplier number for the firm (the default value is populated as such if it has one), or users if different. For an LGFS only firm, I believe this field is superfluos, as the users work under the supplier number of the firm alone.
```
jrmhaig: superfluos -> superfluous

Also (nit) can the long comment be broken over multiple lines, such as;

```suggestion
      / TODO: The hint needs to be checked because an LGFS & AGFS "firm" might confuse this
      /       with an LGFS supplier number for the user. I believe the value must only be the
      /       AGFS supplier number for the firm (the default value is populated as such if it
      /       has one), or users if different. For an LGFS only firm, I believe this field is
      /       superfluous, as the users work under the supplier number of the firm alone.
```

spec/models/user_spec.rb
```diff
@@ -1,32 +1,3 @@
-# == Schema Information
```
jrmhaig, line 1: 👍

app/models/user.rb
```diff
@@ -55,7 +56,7 @@ class User < ApplicationRecord
             acceptance: ['1', true],
             allow_nil: false,
             on: :create,
-            if: proc { Rails.host.api_sandbox? }
+            if: proc { terms_and_conditions_required? }
```
jrmhaig, line 59: :+1: This feels more reliable than checking the host.

jrmhaig (approved):
Just spelling correction and a (suggested) reformatting of a comment. Otherwise, looks good.

PR #3957, opened 2021-05-10
[CFP-30] Amend cronjobs to always use app-latest image

jsugarman:
#### What
Amend cronjobs to always use app-latest image

#### Why
Jobs pods spun up by the archvive_stale cronjob
are not being kept around.

I wonder if it is because when a new image is
applied during deploy this causes old job pods
to be removed.

We want see these logs to check archiving (or vacuuming)
was successful or not, and details thereof.

#### Ticket

[CFP-30] relates to vacuum db jobs for db upgrade

#### How
Always use the app latest image. This means the
config for the job will not be changed by the deploy
script. This means the existing completed job pods
will not be removed...hopefully - this may not be the
reason for their removal however.

It should be harmless to try this nonetheless as the
latest image should have the rake tasks that called
by the job.

[CFP-30]: https://dsdmoj.atlassian.net/browse/CFP-30


jrmhaig (approved):
Looks OK.

PR #3958, opened 2021-05-11
Some cleanup of `Claims::Calculations`

jrmhaig:
#### What

1. Remove the `calculate_expenses_total` method from `Claims::Calculations`
2. Make some methods private

#### Ticket

N/A

#### Why

1. All uses of Claims::Calculations#calculate_expenses_total were removed in #1893 but the method itself was not deleted.
2. Methods that are not used outside of the module do not need to be public. Explicitly making them private means that it is safer to make changes in the future.

### DIFF
--- a/app/models/claims/calculations.rb
+++ b/app/models/claims/calculations.rb
@@ -1,10 +1,4 @@
 module Claims::Calculations
-  def calculate_fee_total(fee)
-    return 0 unless fee
-    fee.calculate_amount
-    fee.amount || 0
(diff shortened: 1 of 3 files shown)


jsugarman (commented):
Looks OK, but hard to work out whats changed because you moved a load of other methods to be private in the same commit

jsugarman (approved):
I see now. I thought you had deleted `calculate_fee_total`.

seems ok

PR #3961, opened 2021-05-11
Add email confirmation on create

jsugarman:
#### What
Add email confirmation on create

#### Why
Since we do not confirm a users sign up email
by sending them an activation email (TBC), we
can, at least, have them confirm their address
in the form, as for passwords.


jsugarman:
Closing, until such time as this becomes an issue or a feature request

PR #3968, opened 2021-05-12
[CFP-114] User creation supplier number

naseberry:
#### What
Update user creation conditionally show supplier number

#### Ticket
[Supplier number requirement for User creation](https://dsdmoj.atlassian.net/browse/CFP-114)

#### Why
Supplier number should only be visible if advocate role checkbox is checked

### DIFF
--- a/app/views/external_users/admin/external_users/_form.html.haml
+++ b/app/views/external_users/admin/external_users/_form.html.haml
@@ -26,21 +26,28 @@
     - if current_user.persona.admin?
       = f.govuk_check_boxes_fieldset :roles, legend: { text: t('.roles'), size: 's' } do
         - @external_user.available_roles.each_with_index do |role, idx|
-          = f.govuk_check_box(:roles,
-            role,
-            label: { text: role.humanize },
-            link_errors: idx.eql?(0)) do
+          - if role.eql?('advocate')
+            = f.govuk_check_box(:roles,
+              role,
+              label: { text: role.humanize },
+              link_errors: idx.eql?(0)) do
 
-            - if role.eql?('advocate') && current_user.persona.provider.chamber?
-              = f.govuk_collection_radio_buttons :vat_registered,
-                [['Yes','true'],['No','false']],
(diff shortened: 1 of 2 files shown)


app/views/external_users/admin/external_users/_form.html.haml
```diff
@@ -26,21 +26,28 @@
...
-                [['Yes','true'],['No','false']],
-                :last,
-                :first,
-                inline: true,
-                legend: { text: t('.vat_registered'), size: 's' }
-
-    = f.govuk_text_field :supplier_number,
-      hint: { text: t('.supplier_number_hint') },
-      label: { text: t('.supplier_number') }
+          - if role.eql?('advocate')
```
jsugarman: i think this should go further down should it not, on line 35. and remove the else.

naseberry: I don't think so, this is how it was before, within the `govuk_check_box` block, thereby creating a 'conditionally revealing' checkbox.
This caused the page height to be resized as we encountered when clicking these checkboxes.
@jsugarman

jsugarman: is the supplier number only supposed to be visible under the advocates checkbox?

if so then it looks like this is unnecessarily repeating a block of code in the else. 

Am going to take a look locally

jsugarman: I think we need to deal with  page height resizing (or whatever is causing the minor movement issue when checking checkboxes) directly, rather than repeating code statements to work around it. You get that glitch anyway in an agfs firm when selecting the advocate check box on this branch.

jsugarman: @naseberry  So the `aria-expanded` of the hidden checkbox input is changed to true. but its the `govuk-checkboxes__conditional--hidden` being removed that causes the "glitch". Is that something we have control over.

jsugarman (approved):
👍

PR #3969, opened 2021-05-13
Fix indictment notification

naseberry:
#### What
Fix missing indictment notification.

#### Why
User were no longer notified of no indictment being attached on the supporting evidence step.

### DIFF
--- a/spec/javascripts/fixtures/supporting-evidence.html
+++ b/spec/javascripts/fixtures/supporting-evidence.html
@@ -21,7 +21,7 @@
     </fieldset>
 
     <div class="button-holder">
-        <input type="submit" name="commit_submit_claim" value="Continue" class="button left">
+        <button type="submit" name="commit_submit_claim" value="Continue" class="govuk-button" data-module="govuk-button">Save and continue</button>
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
👍  have sanity checked against a AGFS and  LGFS claim and seems to work as expected 

i do wonder what other impacts not having the name on the draft button might have caused 🤔

naseberry:
> i do wonder what other impacts not having the name on the draft button might have caused 🤔

My error, it happened when I changed the helper methods.
I thought that as well, and re-applied it, however, no JS reference on the draft button name.

spec/javascripts/fixtures/supporting-evidence.html
```diff
@@ -21,7 +21,7 @@ <h3>
     </fieldset>
 
     <div class="button-holder">
-        <input type="submit" name="commit_submit_claim" value="Continue" class="button left">
+        <button type="submit" name="commit_submit_claim" value="Continue" class="govuk-button" data-module="govuk-button">Save and continue</button>
```
jsugarman, line 24: lol, I have always found the usefulness of these unit tests questionable.

PR #3970, opened 2021-05-13
[CFP-124] Remove commit_save_draft param

jsugarman:
#### What
Remove commit_save_draft param

#### Ticket

[CFP-124](https://dsdmoj.atlassian.net/browse/CFP-124)

#### Why
Controller does not rely on this value so
it is irrelevant and confusing to keep in
there.

This came out of discovering a indirectly related
bug. The same commit(s) that had caused the bug
had removed the name: :commit_save_draft name
from the button.

I appears as this is not used by any controllers in
question in any event, but was at some point (TBC: a commit
link to when and where).

#### TODO
There is still a JS dependency in a jasmine test to remove
`spec/javascripts/supporting-evidence_spec.js`

### DIFF
--- a/spec/controllers/external_users/advocates/interim_claims_controller_spec.rb
+++ b/spec/controllers/external_users/advocates/interim_claims_controller_spec.rb
@@ -119,7 +119,7 @@ RSpec.describe ExternalUsers::Advocates::InterimClaimsController, type: :control
           }
 
           context 'and the form params were submitted as a draft' do
-            let(:form_action) { { } }
+            let(:form_action) { {} }
 
             it 'assigns @claim with the newly created record' do
               create_request
@@ -397,7 +397,7 @@ RSpec.describe ExternalUsers::Advocates::InterimClaimsController, type: :control
           }
 
           context 'and the form params were submitted as a draft' do
-            let(:form_action) { { } }
+            let(:form_action) { {} }
 
             it 'assigns @claim with the existent updated record' do
               update_request
--- a/spec/controllers/external_users/advocates/interim_claims_controller_spec.rb
+++ b/spec/controllers/external_users/advocates/interim_claims_controller_spec.rb
@@ -119,7 +119,7 @@ RSpec.describe ExternalUsers::Advocates::InterimClaimsController, type: :control
           }
 
           context 'and the form params were submitted as a draft' do
-            let(:form_action) { { commit_save_draft: 'Save as draft' } }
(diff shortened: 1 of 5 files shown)


spec/controllers/external_users/advocates/interim_claims_controller_spec.rb
```diff
@@ -119,7 +119,7 @@ def create_claim(*args)
           }
 
           context 'and the form params were submitted as a draft' do
-            let(:form_action) { { commit_save_draft: 'Save as draft' } }
+            let(:form_action) { { } }
```
jrmhaig:
```suggestion
            let(:form_action) { {} }
```

spec/controllers/external_users/advocates/interim_claims_controller_spec.rb
```diff
@@ -397,7 +397,7 @@ def create_claim(*args)
...
```
jrmhaig:
```suggestion
            let(:form_action) { {} }
```

jrmhaig (approved):
Good spot.

A couple of changes to fix Rubocop failures but otherwise approved.

jsugarman:
@naseberry the only relied on reference to `commit_save_draft` is now in a jasmine spec. The underlying js, in `NewClaim.js`, does not rely on this name nor does any ruby/backend code. 

I don't mind keeping the `commit_save_draft` name as I think it is a useful and clear selector, however, so am minded to keep it in. If we want to remove the name we would just have to amend the [jasmine unit test selector](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/spec/javascripts/supporting-evidence_spec.js#L15-L17) for that button. What do you think?

naseberry:
> @naseberry the only relied on reference to `commit_save_draft` is now in a jasmine spec. The underlying js, in `NewClaim.js`, does not rely on this name nor does any ruby/backend code.
> 
> I don't mind keeping the `commit_save_draft` name as I think it is a useful and clear selector, however, so am minded to keep it in. If we want to remove the name we would just have to amend the [jasmine unit test selector](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/spec/javascripts/supporting-evidence_spec.js#L15-L17) for that button. What do you think?


It is great that no ruby/backend code relies on this.

We will keep the front-end aspect as it is for now, I do believe there are questions being asked about modifying this (at which point we can review/rewrite).

@jsugarman

PR #3973, opened 2021-05-17
[CFP-126] Add litigator "no indictment" popup feature test

jsugarman:
#### What
Add litigator "no indictment" popup feature test

#### Ticket
[CFP-126](https://dsdmoj.atlassian.net/browse/CFP-126)

#### Why
This feature had been indvertently broken for while
due to suttle view/frontend changes that impacted it.

Jasmine unit tests had not picked this up as we need
an "integration" test to cover the functionality.

#### How
Modal popups are disabled by custom rack middleware
during test runs (rspec and cucumber) to prevent
the need to interact with them.

Since this feature needs to test/interact with a modal dialog
it re-enables modal dialogs/popups just for this particular
scenario. The step then verifies the text of the message and
accepts it to proceed. It will fail if no modal dialog
pops up or the modal does not have the expected text.

### DIFF
--- a/features/step_definitions/common_claim_steps.rb
+++ b/features/step_definitions/common_claim_steps.rb
@@ -79,6 +79,8 @@ When(/^I check the evidence boxes for\s+'([^']*)'$/) do |labels|
       @claim_form_page.evidence_checklist.check(label)
     end
   end
+  wait_for_ajax
+  sleep 1 # can't find a way around need for this when popups enabled.
 end
 
 When("I answer {string} to was prosecution evidence served on this case?") do |string|
--- a/features/step_definitions/supporting_evidence_steps.rb
+++ b/features/step_definitions/supporting_evidence_steps.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+When('I click "Continue" in the claim form and accept "no indictment" popup') do
+  accept_confirm(no_indictment_message) do
+    @claim_form_page.continue_button.click
+  end
+end
+
+When('I click "Continue" in the claim form and dismiss "no indictment" popup') do
+  dismiss_confirm(no_indictment_message) do
+    @claim_form_page.continue_button.click
+  end
+end
+
+def no_indictment_message
+  @no_indictment_message ||= "The evidence checklist suggests that no indictment has been attached.\n" +
+                             "This can lead to your claim being rejected.\n\n" +
+                             "Do you wish to proceed without attaching an indictment?"
+end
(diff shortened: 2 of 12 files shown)


features/support/overwrite_constants.rb
```diff
@@ -0,0 +1,63 @@
...
+module ConstantsHelper
+  def saved_constants
+    @saved_constants ||= {}
+  end
+
+  def undefined_constants
+    @undefined_constants ||= {}
+  end
+
+  def overwrite_constant(constant, value, object = Object)
```
jrmhaig, line 15: Is it possible to use [`stub_const` from RSpec?](https://relishapp.com/rspec/rspec-mocks/docs/mutating-constants)

`cucumber/rspec/doubles` is already required in `features/support/env.rb` and I think that would make it available in Cucumber tests (although I haven't tried it).

features/step_definitions/common_claim_steps.rb
```diff
@@ -79,6 +79,8 @@
       @claim_form_page.evidence_checklist.check(label)
     end
   end
+  wait_for_ajax
+  sleep 1 # can't find a way around need for this when popups enabled.
```
jrmhaig, line 83: :cry:

features/step_definitions/supporting_evidence_steps.rb
```diff
@@ -0,0 +1,19 @@
...
+  dismiss_confirm(no_indictment_message) do
+    @claim_form_page.continue_button.click
+  end
+end
+
+def no_indictment_message
+  @no_indictment_message ||= "The evidence checklist suggests that no indictment has been attached.\n" +
+                             "This can lead to your claim being rejected.\n\n" +
+                             "Do you wish to proceed without attaching an indictment?"
+end
```
jrmhaig, line 19: Can this be a constant? I'm not too familiar with Cucumber so perhaps this is how it has to be done.

features/support/overwrite_constants.rb
jrmhaig: I've had a look at this and it appears that `stub_const` doesn't work here.

jrmhaig (approved):
:+1:

features/step_definitions/supporting_evidence_steps.rb
jsugarman: does not have to be but a constant would probably have to be in the "World" (cucumber global??!) namespace, so would be polluting and offer no real advantage that i can think of.

PR #3974, opened 2021-05-17
[CFP-62] Migrate provider forms to GOVUKDesignSystemFormBuilder

naseberry:
#### What
Migrate provider form to the GOVUK design system form builder

#### Ticket
[Migrate provider forms to the GOVUK design system form builder](https://dsdmoj.atlassian.net/browse/CFP-62)

#### Why
Part of a larger change to:


app/views/provider_management/external_users/_form.html.haml
```diff
@@ -1,60 +1,34 @@
...
+
+    - if @provider.chamber?
+      = f.govuk_collection_radio_buttons :vat_registered,
+        [['Yes','true'],['No','false']],
+        :last,
+        :first,
+        legend: { text: t('.apply_vat'), size: 's' }
+
+      = f.govuk_text_field :supplier_number,
+        label: { text: t('.supplier_number') }
```
jsugarman, line 32: For discussion only: looks like this is like-for-like with current functionality so good in that respect, but are we doing the same thing differently on the provider side of things  - i.e. nesting these fields under the advocates checkbox.

features/provider_management.feature
```diff
@@ -11,16 +11,41 @@ Feature: Case worker can manage providers
...
-    Then I should see 'AGFS'
-    When I set the provider type to 'Chamber'
+    And I choose govuk radio 'Chamber' for 'Provider type'
+    Then I should not see 'AGFS'
     Then I should not see 'LGFS'
-    When I select the 'AGFS' fee scheme
+
+    When I click the Save details button
+    Then I should see 'Provider successfully created'
+    Then the page should be accessible
```
jsugarman: **nit** use `And` to "chain" additional `Then`'s

features/provider_management.feature
```diff
@@ -11,16 +11,41 @@ Feature: Case worker can manage providers
...
+
+    And I eject the VCR cassette
+
+  Scenario: A provider Manager can create a new firm
+    When I insert the VCR cassette 'features/provider_management'
+    Given I am a signed in case worker provider manager
     Then the page should be accessible
+    When I click the link 'Providers'
+    Then I should be on the provider index page
+    Then the page should be accessible
```
jsugarman: **nit** use `And` to "chain" additional `Then`'s

features/provider_management.feature
```diff
@@ -11,16 +11,41 @@ Feature: Case worker can manage providers
...
+
+    When I click the link 'Add a provider'
+    Then I should be on the new provider page
+    Then the page should be accessible skipping 'aria-allowed-attr'
```
jsugarman: **nit** use `And` to "chain" additional `Then`'s

features/provider_management.feature
```diff
@@ -11,16 +11,41 @@ Feature: Case worker can manage providers
...
+
+    When I set the provider name to 'Test Chambers'
+    And I choose govuk radio 'Firm' for 'Provider type'
+    Then I should see 'AGFS'
+    Then I should see 'LGFS'
```
jsugarman: **nit** use `And` to "chain" additional `Then`'s

features/users/new_external_user.feature
```diff
@@ -22,7 +22,7 @@ Feature: Create new external user
     Given I should not see 'VAT registered'
     And I should not see 'Supplier number'
     When I click govuk checkbox 'Advocate'
-    And the page should be accessible
+    And the page should be accessible skipping 'aria-allowed-attr'
```
jsugarman, line 25: **question:** not related to this PR directly but do we need this many accessibility checks. Will one per page not be sufficient? or is it checking only visible elements each time

app/webpack/javascripts/vendor/jquery.jq-element-revealer.js
```diff
@@ -1,207 +0,0 @@
-/*! Tiny Pub/Sub - v0.7.0 - 2013-01-29
- * https://github.com/cowboy/jquery-tiny-pubsub
```
jsugarman, line 2: good to know we are deleting a "cowboy" library 😂

jsugarman (commented):
Code changes look good from what i can tell and cucumber extensions are a good idea, and hopefully cover us.

Just one question in the comments, plus I would like to check this side by side with current behaviour and not sure I will have time before next week. But if Joe approves then do put it in while I am off.

app/views/shared/_supplier_numbers.html.haml
```diff
@@ -1,16 +1,13 @@
...
-        = link_to_add_association t('.add_supplier'), f, :lgfs_supplier_numbers, partial: 'shared/supplier_number_fields', render_options: { locals: { collection: @collection } }, class: 'govuk-button app-button--blue', data: { 'association-insertion-method': 'append', 'association-insertion-node': '.supplier-numbers-wrapper', module: 'govuk-button' }, role: 'button', draggable: 'false'
+      = f.fields_for :lgfs_supplier_numbers, @collection do |snf|
+        = render partial: 'shared/supplier_number_fields', locals: { f: snf, collection: @collection }
 
-      = govuk_detail t('.postcode_help.heading') do
-        = t('.postcode_help.text_html')
+    .govuk-form-group.form-actions
+      = link_to_add_association t('.add_supplier'), f, :lgfs_supplier_numbers, partial: 'shared/supplier_number_fields', render_options: { locals: { collection: @collection } }, class: 'govuk-button app-button--blue', data: { 'association-insertion-method': 'append', 'association-insertion-node': '.supplier-numbers-wrapper', module: 'govuk-button' }, role: 'button', draggable: 'false'
+
+    = govuk_detail t('.postcode_help.heading') do
```
jsugarman, line 12: **not a blocker** (at all) - this duplicate  current functionality but i notice that the help text is not near the field. Perhaps it should be?! design question perhaps.
 
Screenshot 2021-05-20 at 16 25 31

jsugarman (approved):
Have compared side by side and it looks like existing functionality on the provider side is broken for updates generally, so this is an obvious improvement regardless. I did notice, however, that changing a firm to a chamber does not remove the firm details or persist the fact that they are a chamber. I think this needs a bug raised to be looked at and a more systematic approach to identifying the broken functionality.

I also notice that supplier number errors are not linking (summary to detail). Is this something we want to handle as this stage as i am pretty sure we will hit this problem when it comes to linking nested errors on claims too (and they currently do work)

I do not believe this PR has done anything other than improve the situation though so 👍

features/users/new_external_user.feature
naseberry: @jsugarman Yes, as the feature changes the dom, we need to make sure that all changes are checked for accessibility.

app/views/shared/_supplier_numbers.html.haml
naseberry: @jsugarman we have done that on other views, the help text is then duplocated with each new cocoon section. One for design as it would be good to have a consistent approach.

PR #3975, opened 2021-05-18
[CFP-110] Revert to use custom partial for downtime notification banner

jsugarman:
#### What
Revert to use custom partial for downtime notification banner

#### Ticket

relates to [CFP-110](https://dsdmoj.atlassian.net/browse/CFP-110)

#### Why
`duplicate-id-aria` accessibility issue caused by having a sign-in flash notification
and the downtime banner displayed at the same time.

#### How
Resolved the accessibility issue of `duplicate-id-aria` (with sign-in flash and downtime banner both displayed) by reverting to using a custom partial, with a "custom" `id` and `aria-labelledby` for the title, namely `govuk-downtime-notification-banner-title`


naseberry (approved):
👍

PR #3985, opened 2021-05-24
Some cleanup of TimedTransitions

jrmhaig:
#### What

1. Remove the `is_dummy?` private method from `TimedTransitions::Transitioner`.
2. Remove the `current_state` attribute from `TimedTransitions:Specification`.
3. Replace `@@timed_transition_specifications` with a constant.

#### Ticket

N/A

#### Why

1. Rubocop was complaining [`Naming/PredicateName: Rename is_dummy? to dummy?`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/PredicateName) As it is a private method and is simply a wrapper for `@dummy` it can be removed completely.
2. `TimedTransitions::Specification#current_state` is not used anywhere.
3. [`Style/ClassVars`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ClassVars) recommends avoiding using class variables. Also, `@@timed_transition_specifications` is effectively a constant.

#### How

1. Replace `is_dummy?` with `@dummy` each time it is used.
2. Remove the `current_state` attribute and update `TimedTransitions::Transitioner` where specifications are set up.
3. Replace `@@timed_transition_specifications` by `TIMED_TRANSITION_SPECIFICATIONS`.

### DIFF
--- a/app/models/timed_transitions/transitioner.rb
+++ b/app/models/timed_transitions/transitioner.rb
@@ -2,14 +2,14 @@ module TimedTransitions
   class Transitioner
     attr_accessor :success
 
-    @@timed_transition_specifications = {
+    TIMED_TRANSITION_SPECIFICATIONS = {
       draft: Specification.new(Settings.timed_transition_stale_weeks, :destroy_claim),
       authorised: Specification.new(Settings.timed_transition_stale_weeks, :archive),
       part_authorised: Specification.new(Settings.timed_transition_stale_weeks, :archive),
       refused: Specification.new(Settings.timed_transition_stale_weeks, :archive),
       rejected: Specification.new(Settings.timed_transition_stale_weeks, :archive),
       archived_pending_delete: Specification.new(Settings.timed_transition_pending_weeks, :destroy_claim)
-    }
+    }.freeze
 
     def self.candidate_claims_ids
       Claim::BaseClaim.where(state: candidate_states)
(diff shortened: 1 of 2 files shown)


app/models/timed_transitions/transitioner.rb
```diff
@@ -41,24 +39,20 @@ def success?
 
     private
 
-    def is_dummy?
-      @dummy
-    end
-
     def log_level
-      is_dummy? ? :debug : :info
+      @dummy ? :debug : :info
```
jsugarman, line 43: **not a blocker** 

you could use something like to allow for both use of `dummy` and `dummy?` transparently.
```
attribute :dummy, :boolean, default: false
```

jsugarman (approved):
👍 looks fine, could go further and  extract the "factory" to the `Specification` class and retrieve with a `.for` class method
```
module TimedTransitions
  class Specification
     def self.for(transition)
        {
          draft: new(Settings.timed_transition_stale_weeks, :destroy_claim),
          authorised: new(Settings.timed_transition_stale_weeks, :archive),
          part_authorised: new(Settings.timed_transition_stale_weeks, :archive),
          refused: new(Settings.timed_transition_stale_weeks, :archive),
          rejected: new(Settings.timed_transition_stale_weeks, :archive),
          archived_pending_delete: new(Settings.timed_transition_pending_weeks, :destroy_claim)
        }[transition.to_sym]
     end

     ....
  end
end

```

app/models/timed_transitions/transitioner.rb
jrmhaig: This doesn't work as the class isn't an Active Record model. It should probably not be in `app/models`.

PR #3988, opened 2021-05-25
[CFP-120] Add environment variables to archive_stale cronjob

jrmhaig:
#### What

Add environment variables for accessing S3 and Redis to the configuration of the `archive_stale` cronjob.

#### Ticket

[Archiving stale claims errors](https://dsdmoj.atlassian.net/browse/CFP-120)

#### Why

When deleting archived claims the documents also need to be removed from S3. This requires access to the AWS S3 credentials. Redis access is also required as the job to delete the files it added to a background job.

#### How

Add the environment variable configuration to `kubernetes_deploy/cron_jobs/archive_stale.yaml`


jsugarman (approved):
👍

PR #4008, opened 2021-06-02
Remove @no-seed hook from features

naseberry:
#### What
Remove `@no-seed` hook from features
    
#### Why
CI Cucumber, if a @no-seed feature spec is run first,
db is not seeded, therefore specs requiring this data fails.
This temporarily removes it from features until we can review it.

### DIFF
--- a/features/authentication.feature
+++ b/features/authentication.feature
@@ -1,4 +1,4 @@
-@javascript @no-seed
+@javascript
 Feature: Caseworker can log in while active, but not once inactive
 
   Scenario: I log in as a case worker admin and I see the allocation page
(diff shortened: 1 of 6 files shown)


jrmhaig (approved):
Looks OK but I guess it is failing because it needs the node bump.

jsugarman:
This should not have done what you suggest @naseberry  - if any scenario without a @no-seeds hook is run it should seed for the first scenario.

The @no-seed is to allow fast running of features locally for those that do not require them.

I will look to revert. can you point me to runs that you think failed because of the this

PR #4010, opened 2021-06-03
[CFP-46] Replace grid system and layout class names

naseberry:
#### What
Replace GOV.UK Elements grid system and layout with GOV.UK Frontend

#### Ticket
[GOVUK layout - Grid system](https://dsdmoj.atlassian.net/browse/CFP-46)

#### Why
Part of a larger change to:


jsugarman (approved):
Have sanity checked side by side with prod for most common caseworker and AGFS flows and cannot see any styling problems and can see an improvement on the allocation grid layout when screen is narrow (old grid flows outside).

so 👍 from me

jrmhaig (approved):
Looks OK to me too. :+1:

PR #4015, opened 2021-06-08
[CFP-136] Change 'upto' to 'up to'

jrmhaig:
#### What

Change the text "Unused materials (upto 3 hours)" to "Unused materials (up to 3 hours)".

#### Ticket

Related to [Signpost unused material fees on misc fee page](https://dsdmoj.atlassian.net/browse/CFP-136)

#### Why

"upto" should be two words, as "up to".

#### How

All instances are changed, including the database seeds. These seeds need to be re-applied in all environments.

--------

#### TODO

 - [ ] Apply updated database seeds


jsugarman (commented):
Code changes look fine. 

Has the impact on fee calculation been tested - namely do advocate claims for this fee still return an amount, and the correct amount?

jsugarman (approved):
Happy for this to go in. 

May be safer to do out of hours and use the rake task you have already tried out and do on each environment as you push out so we do not miss one.

PR #4016, opened 2021-06-08
Replace activerecord `.count > 0` or similar with `.any?`

jsugarman:
#### What
Replace activerecord `.count > 0` or similar with `.any?`

#### Why
More efficient call that does

```
SELECT 1 as one
from TABLE
limit 1;
```

See [this interesting lightening talk]( https://youtu.be/XgA_2pJtJEM) for more

### DIFF
--- a/app/interfaces/api/v2/mi/agfs_scheme_ten_claims.rb
+++ b/app/interfaces/api/v2/mi/agfs_scheme_ten_claims.rb
@@ -27,12 +27,12 @@ module API
             desc 'Retrieve claims created with a scheme 10 rep order'
             get do
               if params[:format].eql?('csv')
-                records = claims.count.zero? ? [{ 'No scheme ten claims filed on' => params[:date] }] : claims
+                records = claims.any? ? claims : [{ 'No scheme ten claims filed on' => params[:date] }]
                 csv = build_csv_from(records)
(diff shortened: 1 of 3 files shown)


app/interfaces/api/v2/mi/agfs_scheme_ten_claims.rb
```diff
@@ -27,12 +27,12 @@ def claims
             desc 'Retrieve claims created with a scheme 10 rep order'
             get do
               if params[:format].eql?('csv')
-                records = claims.count.zero? ? [{ 'No scheme ten claims filed on' => params[:date] }] : claims
+                records = claims.any? ? claims : [{ 'No scheme ten claims filed on' => params[:date] }]
```
jrmhaig, line 30: Just FYI, you can use `.none?` as the exact opposite of `.any?`. Having said that, having the desired result first makes more sense.

jrmhaig (approved):
:tada:

app/interfaces/api/v2/mi/agfs_scheme_ten_claims.rb
jsugarman: will leave as `any?` for now - I prefer positive expressions myself

PR #4027, opened 2021-06-10
Bump govuk_design_system_formbuilder 2.5.1 to 2.6.0b1

jsugarman:
#### What
Bump govuk_design_system_formbuilder 2.5.1 to 2.6.0b1

#### Why
This bump handles two issues we raised and were fixed
in the gem

 - [X] do not render conditional divs and whitespace when radio or checkbox capture block is empty
 - [X] decrement dependency on rails 6.1+ to 6.0+

#### How
bump gem to and simplify two forms where govuk checkboxes and radio buttons
are used, by relying on the fix for empty capture blocks with conditionals.


naseberry (approved):
🎉   much better

jrmhaig (approved):
:+1:

PR #4029, opened 2021-06-11
[CFP-118] Correct multi-step controller spec

jsugarman:
#### What
Correct and refactor old controller spec

#### Ticket

#### Why
see individual commits but the specs are out of date, over complex
and fail on the rails 6.1 bump branch for reasons that are not clear

### DIFF
--- a/spec/controllers/external_users/advocates/claims_controller_spec.rb
+++ b/spec/controllers/external_users/advocates/claims_controller_spec.rb
@@ -151,47 +151,23 @@ RSpec.describe ExternalUsers::Advocates::ClaimsController, type: :controller do
           end
         end
 
-        # TODO: This does not represent the current form step logic
+        # NOTE: This is the most minimal data required to progress form steps
         #
         context 'multi-step form submit to LAA' do
           let(:case_number) { 'A20168888' }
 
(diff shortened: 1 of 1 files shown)


spec/controllers/external_users/advocates/claims_controller_spec.rb
```diff
@@ -150,60 +151,45 @@ def expense_params
...
-              expenses_attributes:
-              [
-                expense_params
-              ]
+              form_step: 'supporting_evidence'
             }
           end
 
-          let(:subject_claim) { Claim::AdvocateClaim.active.where(case_number: case_number).first }
+          let(:claim) { Claim::AdvocateClaim.active.where(case_number: case_number).first }
```
jrmhaig, line 174: (nit) Could this be a `subject`?

jrmhaig (approved):
:+1:

spec/controllers/external_users/advocates/claims_controller_spec.rb
jsugarman: in a controller spec i'd say the `subject` should be the `request` rather than the claim - e.g. at line 177

PR #4030, opened 2021-06-11
[CFP-118] Fix validation error preventing claim creation

jsugarman:
#### What
Fix validation error preventing claim creation on rails 6.1 bump

#### Ticket

#### Why
see individual commits but this spec breaks on rails 6.1 with the following
error

```
Failure/Error: claim.reload

ActiveRecord::RecordNotFound:
  Couldn't find Claim::LitigatorClaim without an ID
./spec/api/v2/cclf_claim_spec.rb:88:in `create_claim'
./spec/api/v2/cclf_claim_spec.rb:280:in `block (7 levels) in <top (required)>'
./spec/api/v2/cclf_claim_spec.rb:305:in `block (8 levels) in <top (required)>'
./spec/vcr_helper.rb:84:in `block (3 levels) in <top (required)>'
./spec/vcr_helper.rb:83:in `block (2 levels) in <top (required)>'
```

The fix is, however, to correct the validation error - which was
caused by not having a case type to which you can add an unused materials
fee.

### DIFF
--- a/spec/api/v2/cclf_claim_spec.rb
+++ b/spec/api/v2/cclf_claim_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe API::V2::CCLFClaim, feature: :injection do
     # TODO: this should not require build + save + reload
     # understand what the factory is doing to solve this
     claim = build(*args)
-    claim.save
+    claim.save!
     claim.reload
   end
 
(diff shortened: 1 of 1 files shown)


jrmhaig (approved):
Would be good to understand why it was not failing before, but looks OK.

jsugarman:
> Would be good to understand why it was not failing before, but looks OK.

yeah, I don't like not knowing but it I could not work it out and would have required some time-consuming delving into rails changes probably.

PR #4032, opened 2021-06-14
[CFP-118] Bump gov_uk_date_fields from 3.0.0 to ~> 3.1.0, < 4.0

jsugarman:
#### What
Bump gov_uk_date_fields dependency

#### Ticket

#### Why
So we can then bump the gem to have it function with
rails 6.1, which has deprecated ActiveModel::Errors.keys
in favour of ActiveModel::Errors.attribute_names.

This causes, for example, the following warning, which actually
comes from the gov_uk_date_fields gem.
```
DEPRECATION WARNING: ActiveModel::Errors#keys is deprecated and will be removed in Rails 6.2.

To achieve the same use:

  errors.attribute_names (called from block in _app_views_external_users_claims_transfer_fee__detail_fields_html_haml___3438986822973598706_6457400 at /home/circleci/repo/app/views/external_users/claims/transfer_fee/_detail_fields.html.haml:42)
DEPRECATION WARNING: ActiveModel::Errors#keys is deprecated and will be removed in Rails 6.2.
```


jrmhaig (approved):
:+1:

PR #4042, opened 2021-06-18
[CFP-136] [CFP-137] Signpost unused material fees

jrmhaig:
#### What

Add a notice on claims to indicate that unused materials fees may be claimed.

#### Ticket

* [Signpost unused material fees on misc fee page](https://dsdmoj.atlassian.net/browse/CFP-136)
* [Unused material fee flag on summary page](https://dsdmoj.atlassian.net/browse/CFP-137)

#### Why

Providers are being underpaid for work they are entitled to claim for.

#### How

A notice is added in the following places of a claim that is eligible for an unused materials fee:

* At the top of the claim summary
* In the claim summary, at the miscellaneous page section
* On the miscellaneous fees page

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -47,7 +47,7 @@ module ClaimsHelper
       page_hint: t('page_hint', scope: scope)
     }.tap do |headings|
       if display_unused_materials_notice_for(claim)
-        headings[:page_notice] = t('unused_materials_fee.notice', scope: scope)
+        headings[:page_notice] = t('unused_materials_fee.notice.short', scope: scope)
       end
       headings[:fees_calculator_html] = fees_calculator_html unless fees_calculator_html.nil?
     end
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -40,4 +40,21 @@ module ClaimsHelper
     end
     false
   end
+
+  def fee_shared_headings(claim, scope, fees_calculator_html = nil)
+    {
+      page_header: t('page_header', scope: scope),
+      page_hint: t('page_hint', scope: scope)
+    }.tap do |headings|
+      if display_unused_materials_notice_for(claim)
+        headings[:page_notice] = t('unused_materials_fee.notice', scope: scope)
+      end
+      headings[:fees_calculator_html] = fees_calculator_html unless fees_calculator_html.nil?
+    end
+  end
+
+  def display_unused_materials_notice_for(claim)
+    claim.eligible_misc_fee_types.map(&:unique_code).include?('MIUMU') &&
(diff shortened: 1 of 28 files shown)


app/helpers/claims_helper.rb
```diff
@@ -40,4 +40,21 @@ def messaging_permitted?(message)
...
+    }.tap do |headings|
+      if display_unused_materials_notice_for(claim)
+        headings[:page_notice] = t('unused_materials_fee.notice', scope: scope)
+      end
+      headings[:fees_calculator_html] = fees_calculator_html unless fees_calculator_html.nil?
+    end
+  end
+
+  def display_unused_materials_notice_for(claim)
+    claim.eligible_misc_fee_types.map(&:unique_code).include?('MIUMU') &&
```
jsugarman, line 69: 👍

app/helpers/claims_helper.rb
```diff
@@ -40,4 +40,21 @@ def messaging_permitted?(message)
...
+      page_hint: t('page_hint', scope: scope)
+    }.tap do |headings|
+      if display_unused_materials_notice_for(claim)
+        headings[:page_notice] = t('unused_materials_fee.notice', scope: scope)
+      end
+      headings[:fees_calculator_html] = fees_calculator_html unless fees_calculator_html.nil?
+    end
+  end
+
+  def display_unused_materials_notice_for(claim)
```
jsugarman: should this not be `display_unused_materials_notice?`

app/services/claims/fetch_eligible_misc_fee_types.rb
```diff
@@ -5,8 +5,8 @@ def initialize(claim)
     end
 
     def call
-      return nil unless claim
-      return nil if claim&.interim?
+      return unless claim
+      return [] if claim&.interim?
```
jsugarman, line 9: into scope creep territory with the commit that promoted these methods and amends this. Are you confident changing this behaviour is not going to break other areas of the app. 
Not sure feature tests and other integration specs cover it so have will need manual testing to check that it does not change behaviour in both the form steps and the various summary pages for all claim types (interims and transfers in particular) that may rely on the existing eligibility logic.

app/helpers/claims_helper.rb
jsugarman: have we missed copying the tests for it?

app/services/claims/fetch_eligible_misc_fee_types.rb
jrmhaig: The `display_unused_materials_notice?` helper depends on the `eligible_misc_fee_types` method existing for any claim type and it must return an array. The correct response for an interim claim is an empty array as it is not eligible for any miscellaneous fee types.

jsugarman (approved):
👍 good to go

PR #4044, opened 2021-06-21
Sort error summary order according to custom sequence in error_messages.en.yml

jsugarman:
#### What
Sort error summary errors according to custom sequence in error_messages.yml

#### Ticket

[came out of error handling review related to rails 6.1 bump](https://dsdmoj.atlassian.net/browse/CFP-118)

#### Why

The header errors method returns a sorted collection of
ErrorDetail objects. These objects internally sort by sequence
numbers defined via the `_seq` values in `error_messages.en.yml`.

No further ordering should be needed.


naseberry (approved):
👍 nice

PR #4047, opened 2021-06-21
Exclude id_sequence_resettable from coverage stats

jsugarman:
#### What
Exclude id_sequence_resettable from coverage stats

#### Why
Only used for local DB restore and not tested outside
of doing a restore itself.


jrmhaig (approved):
(nit) Could `lib/id_sequence_resettable.rb` be moved into `lib/tasks/rake_helpers/`?

Otherwise, OK (approved)

jsugarman:
@jrmhaig closing this as superseded by this by https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4051 which moves it to the rake_helpers as suggested

PR #4049, opened 2021-06-21
Only check existence of persisted documents

jrmhaig:
`ExternalUsers::ClaimsController#build_nested_resource` adds an empty instance of `Document` to `@claim.documents` if it is empty and this ensures that `@claim.document.any?` is true while `@claim.document.count` is zero (as this checks with an additional query to the database).

`@claim.document.any?(&:persisted?)` can be used to excluded the extra instance added by the controller.

## Before

Screenshot 2021-06-21 at 15 35 47

## After

Screenshot 2021-06-21 at 15 34 24


jsugarman (approved):
👍

PR #4051, opened 2021-06-21
Move id_sequence_resettable to lib/tasks/rake_helpers

jsugarman:
#### What
Move id_sequence_resettable to lib/tasks/rake_helpers

#### Why
Only used for local DB restore task and not tested outside
of doing a restore itself. This also removes it from
coverage stats.


jrmhaig (approved):
:+1:

PR #4054, opened 2021-06-22
Ensure that a test unique code has letters

jrmhaig:
`Faker::Lorem.characters` generates a random alphanumeric string which can, with a small probability, be entirely numeric. The Shoulda matcher used in the `#unique_code` tests in `spec/models/case_stage_spec.rb` fails if the test value does not contain letters as it cannot check case sensitivity. Adding the `min_alpha: 1` parameter to `Faker::Lorem.characters` will ensure that case sensitively can be checked.


jsugarman (approved):
👍 LGTM

PR #4055, opened 2021-06-22
Revert "Bump rails 6.0 to to 6.1"

jsugarman:
Reverts ministryofjustice/Claim-for-Crown-Court-Defence#3753

This was inadvertently a "rebase and merge" so am creating this revert now using github just in case we need to revert

Note: existing DB instances will need to have the migrations in this revert  rolled back on the servers (e.g. `rails db:rollback STEP=2`) BEFORE this reverted codebase can be deployed. Otherwise you cannot rollback and the app will be broken. Rolling back will also break the app until this revert is deployed.


jsugarman:
any problems arising will need forward fixing now

PR #4056, opened 2021-06-22
Add environment to rake task

jsugarman:
#### What
Add environment to rake task

#### Why
This was working previously but is currently broken.
Its unclear if this is a rails 6.1 issue or related
to rake version currenly bundled but the task was raising
the following error. The task should have an envn anyway.

```
bundle exec rake db:reload
=>
NoMethodError: undefined method `host' for Rails:Module
lib/tasks/rake_helpers/rake_utils.rb:30:in `production_protected'
lib/tasks/db.rake:63:in `block (2 levels) in <top (required)>'
```

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -59,7 +59,7 @@ namespace :db do
   task :reseed => [:clear, 'db:migrate', 'db:seed']
 
   desc 'CCCD task: clear the database, run migrations, seeds and reloads demo data'
-  task :reload do
+  task :reload => :environment do
     production_protected
 
     Rake::Task['db:clear'].invoke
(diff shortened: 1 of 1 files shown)


lib/tasks/db.rake
```diff
@@ -59,7 +59,7 @@ namespace :db do
   task :reseed => [:clear, 'db:migrate', 'db:seed']
 
   desc 'CCCD task: clear the database, run migrations, seeds and reloads demo data'
-  task :reload do
+  task :reload => :environment do
```
jrmhaig:
```suggestion
  task reload: :environment do
```

Not a blocker (considering that the rockets syntax is used everywhere else in the tasks).

jrmhaig (approved):
OK. Personal preference on syntax but not a major issue if you decide not to change it.

PR #4057, opened 2021-06-22
Fix disallowed sql in migrations run by smoke test

jsugarman:
#### What
Fix disallowed sql in migrations run by smoke test

#### Why

Was raising udring smoke test run, following rails 6.1 bump
```
Query method called with non-attribute argument(s): "offences.description COLLATE \"C\", offence_classes.class_letter COLLATE \"C\""
```


jrmhaig (approved):
:+1:

PR #4058, opened 2021-06-22
Error presenter refactor

jrmhaig:
#### What

Refactoring parts of the error presenter.

#### Ticket

N/A

#### Why

The changes to the Rails error handing in the upgrade to Rails 6.1 has highlighted some opportunities for cleaning up the error presenter.

#### How

Where `ErrorMessageTranslator` is used the method `#translation_found?` is used to determine whether it has been successful. If not, default errors are used instead. This check and setting of default values is unnecessary if it is handled inside the class. See commit https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4058/commits/f9fd14ab98bdf33acc71021764996b769a8c30b8. Also https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/5bd3df04023b1b79ee389c00d40eb59e063aea80, https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/549c187c8f92cc8a07c1533e7931a3f0cdf5def0 and https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/684d603beb6a8ca08a180c81496ec02c3ff40de5 prepare for this by making the unit tests more standards compliant.

#### Possible further work (not done here)

The `ErrorMessageTranslator` could possibly return an instance of `ErrorDetail`, so the `generate_messages` method in `ErrorPresenter` could look something like this:

```ruby
def generate_messages
  emt = ErrorMessageTranslator.new(@translations)

  @errors.each do |error|
    messages = emt.translate(error)
    next if @error_details[attribute] && @error_details[attribute][0].long_message.eql?(messages.long_message)
    messages.sequence = generate_sequence(error.attribute)
    @error_details[attribute] = messages
  end
end
```

It may also be possible to handle all of this inside `ErrorDetailCollection`.


jsugarman:
Closing in favour of https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4180

PR #4064, opened 2021-06-23
Use default url for form

jrmhaig:
#### What

Fix the new user form for providers at `/provider_management/providers/1/external_users/new`.

#### Ticket

[Unable to add new user](https://dsdmoj.atlassian.net/browse/CFP-172)

#### Why

Currently the new user form returns an internal server error (500) so it is not possible to create new users.

#### How

The `url` parameter for `form_with` in the view overrides the default submit path for the form and for the users form it is being set to the correct path for updating an existing user. When the user does not exist (in the case of a new user) the helper to create this path fails as the id is missing.

The `url` parameter is not necessary as the default path, for both new and existing users, is correctly generated based on the value of the `model` parameter.


jsugarman (approved):
Great, get this in as hotfix, but can you do a separate PR as part of this ticket to add a feature test. You can extend the `features/provider_management.feature` to add a 
```
Scenario: A provider manager can add users to provider
```

PR #4065, opened 2021-06-23
Scenario for adding new users to a provider

jrmhaig:
#### What

Add cucumber scenario for adding a new user for a provider.

#### Ticket

[Unable to add new user](https://dsdmoj.atlassian.net/browse/CFP-172)

#### Why

A user reported errors when adding new users for a provider and there is no feature test to check this.

#### How

A new scenario 'A provider Manager can add users to providers'.


jsugarman (approved):
👍 cheers

PR #4066, opened 2021-06-23
Simplify call to form_with in provider management

jsugarman:
#### What
Simplify call to form_with in provider management

#### Why
Pass namespaced model from which rails can
determine correct controller and action.

Also remove html novalidate option which
does not appear to have any effect.


jsugarman:
closing as is shared by external admin and caseworker provider management controllers so this breaks the external admin side of things

PR #4070, opened 2021-06-24
[CFP-174] Send email to newly created case worker

jrmhaig:
#### What

Send an email with the password reset link to a newly case worker.

#### Ticket

[https://dsdmoj.atlassian.net/browse/CFP-174](https://dsdmoj.atlassian.net/browse/CFP-174)

This ticket identifies issues on creation of new case workers as well as a password reset. This PR fixes the former issue but I have not been able to recreate the latter.

#### Why

The email is required to enable new users to set the password before the account can be used.

#### How

The email notification is being prepared but, due to a [previous commit,](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/41b129c50061e9fdcb1080205fa8d01cf73dcbef) the job to send it was not being enqueued.

### DIFF
--- a/spec/requests/case_workers/admin/case_workers_spec.rb
+++ b/spec/requests/case_workers/admin/case_workers_spec.rb
@@ -1,7 +1,10 @@
 RSpec.describe 'Caseworker admin', type: :request do
   let(:admin) { create(:case_worker, :admin) }
 
-  before { sign_in admin.user }
+  before do
+    ActiveJob::Base.queue_adapter = :test
+    sign_in admin.user
+  end
 
   describe 'POST /case_workers/admin/case_workers' do
     let(:create_case_workers_request) { post(case_workers_admin_case_workers_path, params: case_worker_params) }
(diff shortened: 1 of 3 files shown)


spec/requests/case_workers/admin/case_workers_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      it 'attempts to deliver an email' do
+        allow(DeviseMailer).to receive(:reset_password_instructions)
+        create_case_workers_request
+        expect(DeviseMailer).to have_received(:reset_password_instructions)
+      end
+
+      it 'enqueues a job to send the email' do
+        create_case_workers_request
+        expect(ActionMailer::DeliveryJob).to have_been_enqueued
+      end
```
jsugarman, line 47: 👍 does this break when `deliver_later` is removed?

jsugarman (approved):
👍

spec/requests/case_workers/admin/case_workers_spec.rb
jrmhaig: Yes

PR #4073, opened 2021-06-25
Remove state_machine in favour of state_machines

jrmhaig:
The [`state_machine`](https://github.com/pluginaweek/state_machine) gem is deprecated and was forked in 2015 as [`state_machines`.](https://github.com/state-machines/state_machines) See: https://github.com/pluginaweek/state_machine/issues/343

As it happens, we already include `state_machines` as a dependency on `state_machines-activerecord` and `state_machines-audit_trail` so all that needs to be done is to remove `state_machine` from the `Gemfile`.

As `state_machine` and `state_machines` provide the same functionality, and there may have been changes to the latter since 2015, there is a possibility that the ordering of library loading means that we have still been using the old version and so some testing is required to ensure that there are no breaking changes.


jsugarman:
Closed. See [CFP-195](https://dsdmoj.atlassian.net/browse/CFP-195) for details

PR #4074, opened 2021-06-25
Amend places where we should be using `#redeterminable?`

jsugarman:
#### What
Amend places where we should be using `#redeterminable?`

#### Ticket

[came out of/relates to CFP-109, redetermination messaging bugs](https://dsdmoj.atlassian.net/browse/CFP-109)

#### Why
The `#redeterminable?` method below applies the logic
used in these place plus the additional consideration of
not permitting redeterminations on Litigator interim claims,
which we PROBABLY should be applying everywhere.

```
def redeterminable?
  VALID_STATES_FOR_REDETERMINATION.include?(state) && !(lgfs? && interim?)
end
```

#### TODO (wip)

 - [x] pass test suite
 - [ ] test messaging functionality on all claim types is unchanged
       or write features tests to cover

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -35,9 +35,7 @@ module ClaimsHelper
 
   def messaging_permitted?(message)
     return true if message.claim_action.present?
-    if current_user_is_external_user?
-      return Claims::StateMachine::VALID_STATES_FOR_REDETERMINATION.exclude?(message.claim.state)
-    end
+    return !message.claim.redeterminable? if current_user_is_external_user?
     false
   end
 
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -87,7 +87,7 @@ class Message < ApplicationRecord
   end
 
   def process_claim_action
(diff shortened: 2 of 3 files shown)


app/helpers/claims_helper.rb
```diff
@@ -35,9 +35,7 @@ def show_message_controls?(claim)
 
   def messaging_permitted?(message)
     return true if message.claim_action.present?
-    if current_user_is_external_user?
-      return Claims::StateMachine::VALID_STATES_FOR_REDETERMINATION.exclude?(message.claim.state)
-    end
+    return !message.claim.redeterminable? if current_user_is_external_user?
```
jrmhaig, line 38: This could be:

```suggestion
    current_user_is_external_user? && !message.claim.redeterminable?
```

and the default `false` on line 39 is unnecessary (but Github will not let me include it in the suggestion). However, you may consider it clearer as it is. No problem either way.

jrmhaig:
closing as Joel on tuple as even though test pass there is a logic problem that needs investigating

jsugarman:
closing as even though test pass there is a logic problem that needs investigating

PR #4075, opened 2021-06-25
Remove simplecov-multi gem

jrmhaig:
The `simplecov-multi` gem is unused so can be removed. Also, it appears that it is now included in `simplecov`. See:

https://github.com/simplecov-ruby/simplecov#using-multiple-formatters

Also also, the HTML formatter is the default so it does not need to be set explicitly.


jsugarman (approved):
👍

PR #4080, opened 2021-06-29
[CFP-158] Fix govuk_design_system_formbuilder nested error linking

jsugarman:
#### What
Fix govuk_design_system_formbuilder nested error linking

#### Ticket

[CFP-156](https://dsdmoj.atlassian.net/browse/CFP-156)

#### Why
error linking between summary and field are broken nested fields when using the govuk formbuilder

#### How
- [X] Name nested errors following expectations from form builder
- [X] Use concern to override nested error attribute naming for validators on models using formbuilder on FE
- [X] Use DfE formbuilder 2.7.2

#### TODO
- [x] use new version of govuk formbuilder once [fix-multiple-captures](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/301) is merged and released
- [x] unit spec error naming via composition and inheritance
- [x] feature spec unhappy path for provider supplier number error linking?!

### DIFF
--- a/app/views/shared/_supplier_numbers.html.haml
+++ b/app/views/shared/_supplier_numbers.html.haml
@@ -1,13 +1,13 @@
 .govuk-grid-row
   .govuk-grid-column-full
     #supplier-numbering.supplier-numbers-wrapper.fx-numberedList-hook
-      - @collection = f.object.lgfs_supplier_numbers.present? ? f.object.lgfs_supplier_numbers : [f.object.lgfs_supplier_numbers.build]
+      - @lgfs_supplier_numbers = f.object.lgfs_supplier_numbers.present? ? f.object.lgfs_supplier_numbers : [f.object.lgfs_supplier_numbers.build]
 
-      = f.fields_for :lgfs_supplier_numbers, @collection do |snf|
-        = render partial: 'shared/supplier_number_fields', locals: { f: snf, collection: @collection }
+      = f.fields_for :lgfs_supplier_numbers, @lgfs_supplier_numbers do |snf|
+        = render partial: 'shared/supplier_number_fields', locals: { f: snf }
 
     .govuk-form-group.form-actions
-      = link_to_add_association t('.add_supplier'), f, :lgfs_supplier_numbers, partial: 'shared/supplier_number_fields', render_options: { locals: { collection: @collection } }, class: 'govuk-button app-button--blue', data: { 'association-insertion-method': 'append', 'association-insertion-node': '.supplier-numbers-wrapper', module: 'govuk-button' }, role: 'button', draggable: 'false'
+      = link_to_add_association t('.add_supplier'), f, :lgfs_supplier_numbers, partial: 'shared/supplier_number_fields', class: 'govuk-button app-button--blue', data: { 'association-insertion-method': 'append', 'association-insertion-node': '.supplier-numbers-wrapper', module: 'govuk-button' }, role: 'button', draggable: 'false'
 
     = govuk_detail t('.postcode_help.heading') do
       = t('.postcode_help.text_html')
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -218,6 +218,8 @@ en:
               incompatible_case_type: Fixed fee invalid on non-fixed fee case types
         supplier_number:
           attributes:
+            postcode:
+              invalid: Enter a valid postcode
             supplier_number:
               invalid: Enter a valid LGFS supplier number
         provider:
--- a/features/provider_management.feature
+++ b/features/provider_management.feature
@@ -62,7 +62,7 @@ Feature: Case worker can manage providers
     And I fill in 'Supplier number' with '11111'
 
     When I click the button 'Save details'
-    Then the following error details should exist:
+    Then the following govuk error details should exist:
       | field_type | field_locator | error_text | linked_id |
       | field | Provider name| Enter a provider name | provider-name-field-error |
       | field | Supplier number | Enter a valid LGFS supplier number | provider-lgfs-supplier-numbers-attributes-0-supplier-number-field-error |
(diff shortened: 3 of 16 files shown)


app/views/shared/_supplier_numbers.html.haml
```diff
@@ -1,13 +1,13 @@
 .govuk-grid-row
   .govuk-grid-column-full
     #supplier-numbering.supplier-numbers-wrapper.fx-numberedList-hook
-      - @collection = f.object.lgfs_supplier_numbers.present? ? f.object.lgfs_supplier_numbers : [f.object.lgfs_supplier_numbers.build]
+      - @lgfs_supplier_numbers = f.object.lgfs_supplier_numbers.present? ? f.object.lgfs_supplier_numbers : [f.object.lgfs_supplier_numbers.build]
```
naseberry, line 4: 👍

config/locales/en.yml
```diff
@@ -218,6 +218,8 @@ en:
               incompatible_case_type: Fixed fee invalid on non-fixed fee case types
         supplier_number:
           attributes:
+            postcode:
+              invalid: Enter a valid postcode
```
naseberry, line 222: Well spotted.

features/provider_management.feature
```diff
@@ -44,11 +42,46 @@ Feature: Case worker can manage providers
...
+    And I fill in 'Supplier number' with '11111'
+
+    When I click the button 'Save details'
+    Then the following govuk error details should exist:
+      | field_type | field_locator | error_text | linked_id |
+      | field | Provider name| Enter a provider name | provider-name-field-error |
+      | field | Supplier number | Enter a valid LGFS supplier number | provider-lgfs-supplier-numbers-attributes-0-supplier-number-field-error |
+      | fieldset | Is the provider VAT registered? | Choose VAT registration state | provider-vat-registered-field-error |
+
+    # long version
```
naseberry: Are these comments **to be or not to be**?

features/step_definitions/govuk_component_steps.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
naseberry, line 1: Would it be ideal to extend this new step definition  to include govuk textbox and other govuk form inputs?

naseberry:
I've manually tested the provider views on the dev-lgfs environment, error linking now works as intended with  nested forms.

I would suggest we get a second review from a BE developer.

👍 Great work.

PS - I noticed the error summary order was incorrect, will this be in a separate PR?

features/provider_management.feature
jsugarman: yes, this is the question of preference i wanted to raise in the commits comment with you guys
> I prefer the more succinct cucumber table as a reusable
pattern for future unhappy path scenarios. However,
this could be more confusing during failures so would
like other devs views and preferences input.

What do you think?

features/step_definitions/govuk_component_steps.rb
jsugarman: I think so, I added [this commit](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4080/commits/701de5ef718c1f931293e17df210e950a2bbb79e) to move other steps I found to this file. Are there others you are thinking-of/have-found?

jsugarman:
Cheers
> PS - I noticed the error summary order was incorrect, will this be in a separate PR?

Error order is not controlled in the provider form and falls back to default rails/activemodel behaviour I believe. 

As part of the spike I used to test custom error handling I believe it would work if providers were validated using our "custom error handling", but they currently do not. I have not tried to implement the custom error handling on this form as yet as yak shaving will undoubtedly ensue. 

I think we could apply custom error handling to this form as part of the ticket to handle formbuilder working with custom error handling as 1. a prelude to applying formbuilder to claim forms 2. to standardise our error handling further across different objects (we currently use either the rails-way or our-way)

spec/validators/claim/base_claim_sub_model_validator_spec.rb
```diff
@@ -4,6 +4,10 @@
   let(:claim) { FactoryBot.create :claim }
   let(:defendant) { claim.defendants.first }
 
+  it_behaves_like 'a custom CCCD associated error handler' do
+    subject(:validator) { described_class.new }
```
jrmhaig: Does this need to be a `subject`? I don't think it is being used as the subject of any tests. Likewise for the other places where the new shared examples are used.

Not a major issue. Maybe it is clearer like this.

features/provider_management.feature
jrmhaig: The tabular form is more understandable. What happens for a failure? Does it just indicate a failure without specifying which line? I don't think that is a problem, especially if there are just a few lines.

jsugarman: its not too bad, it will fail on the first row and expectation but from my testing it is pretty easy to identify the failure

spec/validators/claim/base_claim_sub_model_validator_spec.rb
jsugarman: have added [this commit](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4080/commits/db6be30299f56577ee458f158b4c7b0a315b56e4)  which makes it more consistent and places the subject in its "proper" context  - i.e. the class under test.

jrmhaig (approved):
:+1: Looks OK.

features/provider_management.feature
naseberry: ^^ I agree with Joe, the tabular form is understandable for me

features/step_definitions/govuk_component_steps.rb
naseberry: I was thinking more of https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/features/step_definitions/common_steps.rb#L204

Adding specificity like the other govuk steps by finding the relevant class name, but then again, I guess the reference above seem loose and robust.
What do you think?

PR #4089, opened 2021-06-30
Enable new cops in Rubocop

jrmhaig:
With each update to the `rubocop` and `rubocop-*` gems we are explicitly enabling new cops to silence warnings. By setting new cops to be enabled we can reduce the size of `.rubocop.yml` and any new offences can be seen in the Dependabot PR without checking out locally.


jsugarman (approved):
👍  happy to try it out

PR #4093, opened 2021-07-01
[CFP-170] Add the submitted claim report to the reports page

jrmhaig:
#### What

Add the submitted claims report to the reports page

#### Ticket

[Last 3 months submitted claims](https://dsdmoj.atlassian.net/browse/CFP-170)

#### Why

Put the report somewhere where it can be accessed.

#### How

I have tried to tidy up the existing stats reports generating services to pull out some common code into a single place. This is done in multiple commits so that the steps can be seen more clearly and, if necessary, they can be broken into separate PRs.

##### Step 1 - 6bea52c398dc494884396e8457cb972ea21ec6c2

A new report is added to show a count of all submitted claims per week for the past twelve weeks. The rake task to create the submitted claims report is updated to use the same report generation instead of its own helper.

This new report has been created following the pattern of the provisional assessment report and rejections refusals report, which both work in the same way. Specifically, the `Stats::ProvisionalAssessmentReportGenerator` and `Stats::RejectionsRefusalsReportGenerator` classes are services that are identical except for the class that is used for generating the data. With this in mind I have pulled out the common functionality of these into `Stats::ReportGenerator::Base` and built the new service `Stats::ReportGenerator::SubmittedClaims` on top of this.

##### Step 2 - 9ca10f41570d717bdee27c13d379ccde460b1d2c

`Stats::ProvisionalAssessmentReportGenerator` is mostly identical to the `Stats::ReportGenerator::Base` created in step 1 and so, after moving it into the same namespace, it can be made to inherit from it. This means that it can be seen to work in the same way as the submitted claims report.

##### Step 3 - 6925ed085b3eec635ed4dfa1f2345111160c5780

The same as step 2 but for the rejections refusal report.

##### Step 4 - Further work, not done here.

The management information report (`app/services/stats/management_information_generator.rb`) should work the same way as the other three reports but it has been written differently. If it is rewritten to follow the same pattern then it may be possible to eliminate the need for four separate report generator classes, and instead pass the relevant report class to a single report generator. For example;

```ruby
Stats::ReportGenerator.call(reporter: Report::SubmittedClaims)
# instead of
Stats::ReportGenerator::SubmittedClaims.call
```

(see `generate_new_report` in `app/services_stats_stats_report_generator.rb`)


spec/controllers/case_workers/admin/management_information_controller_spec.rb
```diff
@@ -10,7 +10,7 @@
       before { get :index }
 
       it 'assigns available_report_types' do
-        expect(assigns(:available_report_types).keys).to match_array(%w[management_information provisional_assessment rejections_refusals])
+        expect(assigns(:available_report_types).keys).to match_array(%w[management_information provisional_assessment rejections_refusals submitted_claims])
```
jsugarman: **nit** 
```
let(:expected_report_types) { %w[management_information provisional_assessment rejections_refusals submitted_claims] }
...
expect(assigns(:available_report_types).keys).to match_array(expected_report_types)
```

spec/services/stats/csv_exporter_spec.rb
```diff
@@ -2,42 +2,84 @@
...
         row_4_column_1,row_4_column_2,row_4_column_3
         CSVOUTPUT
         expect(service.call).to eq(expected_output)
       end
+
+      context 'when no headers are provided' do
+        let(:options) { {} }
+
+        it 'returns a CSV string with only the provided data' do
+          expected_output = <<-CSVOUTPUT.strip_heredoc
```
jsugarman: **nit**
easier to read if you `let` the expected_output

spec/services/stats/report_generator/submitted_claims_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe Stats::ReportGenerator::SubmittedClaims, type: :service do
+  describe '.call' do
+    let(:result) { described_class.call }
```
jsugarman:
```suggestion
    subject(:result) { described_class.call }
```

spec/services/stats/report_generator/submitted_claims_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'counts the number of submitted claims for last week' do
+      expect(CSV.parse(result.content).last).to match_array(['21/06/2021', '2'])
+    end
+
+    it 'has results for the past 12 weeks' do
+      t = Time.zone.parse('21 June 2021')
+      last_twelve_mondays = Array.new(12) { |i| (t - i.weeks).strftime('%d/%m/%Y') }
+      expect(CSV.parse(result.content).map(&:first)).to match_array(['Week starting', *last_twelve_mondays])
+    end
+  end
```
jsugarman:
```
let(:rows) { CSV.parse(result.content) }

...
expect(rows.first)
...
expect(rows.last)
...
expect(rows.map(&:first))
```

spec/services/stats/csv_exporter_spec.rb
```diff
@@ -2,42 +2,84 @@
...
+          %w[row_3_column_1 row_3_column_2 row_3_column_3],
+          %w[row_4_column_1 row_4_column_2 row_4_column_3]
+        ]
+      }
 
-      it 'returns a CSV string with only the provided data' do
+      subject(:service) { described_class.new(data, options) }
+
+      it 'returns a CSV string with the provided data including the headers' do
         expected_output = <<-CSVOUTPUT.strip_heredoc
```
jsugarman: nit
easier to read if you let the expected_output

spec/services/stats/csv_exporter_spec.rb
```diff
@@ -2,42 +2,84 @@
...
+        row_4_column_1,row_4_column_2,row_4_column_3
+        CSVOUTPUT
+        expect(service.call).to eq(expected_output)
+      end
+
+      context 'when no headers are provided' do
+        let(:options) { {} }
+
+        it 'returns a CSV string with only the provided data' do
+          expected_output = <<-CSVOUTPUT.strip_heredoc
```
jsugarman: nit
easier to read if you let the expected_output

spec/services/stats/csv_exporter_spec.rb
```diff
@@ -2,42 +2,84 @@
...
+          { column_1: 'row_2_column_1', column_2: 'row_2_column_2', column_3: 'row_2_column_3' },
+          { column_1: 'row_3_column_1', column_2: 'row_3_column_2', column_3: 'row_3_column_3' },
+          { column_1: 'row_4_column_1', column_2: 'row_4_column_2', column_3: 'row_4_column_3' }
+        ]
+      }
+
+      subject(:service) { described_class.new(data, options) }
+
+      it 'returns a CSV string with the provided data including the headers' do
+        expected_output = <<-CSVOUTPUT.strip_heredoc
```
jsugarman: nit
easier to read if you let the expected_output

jsugarman (changes requested):
Refactoring looks good

1. some spec refactoring suggestions.
2. we should have a shared example to encompass the `Base` class i'd say:
```
it_behaves_like 'a report generator'
```
3. where is the claim being put? how does the user retrieve it? The caseworker reports are produced on demand, except for management information, rather than via a cronjob.
lets discuss this!

kubernetes_deploy/cron_jobs/submitted_claims_report.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: batch/v1beta1
```
jsugarman: as discussed we do  not need to schedule this report

kubernetes_deploy/scripts/cronjob.sh
```diff
@@ -28,7 +28,7 @@ function _cronjob() {
   fi
 
   case "$1" in
-    archive_stale | vacuum_db)
+    archive_stale | vacuum_db | submitted_claims_report)
```
jsugarman: remove

kubernetes_deploy/scripts/cronjob.sh
```diff
@@ -3,7 +3,7 @@ function _cronjob() {
   usage="cronjob -- apply job in the specified environment
   Usage: cronjob job environment [branch]
   Where:
-    job [archive_stale|clean_ecr|vacuum_db]
+    job [archive_stale|clean_ecr|vacuum_db|submitted_claims_report]
```
jsugarman: remove

spec/services/stats/report_generator_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'rails_helper'
+
+RSpec.shared_examples 'package stats in CSV' do
```
jsugarman: I think it would be clearer to name it something like this

```suggestion
RSpec.shared_examples 'a stats report CSV exporter' do
```

spec/services/stats/report_generator_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  describe '.call' do
+    subject(:result) { described_class.call(report, **options) }
+
+    let(:options) { {} }
+
+    context 'with a provisional assessment report' do
+      let(:report) { 'provisional_assessment' }
+      let(:reporter) { Reports::ProvisionalAssessments }
+
+      include_examples 'package stats in CSV'
```
jsugarman:
```suggestion
      it_behaves_like 'a stats report CSV exporter'
```

spec/services/stats/report_generator_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    context 'with a rejections refusals report' do
+      let(:report) { 'rejections_refusals' }
+      let(:reporter) { Reports::RejectionsRefusals }
+
+      include_examples 'package stats in CSV'
```
jsugarman:
```suggestion
      it_behaves_like 'a stats report CSV exporter'
```

spec/services/stats/report_generator_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    context 'with a submitted claims report' do
+      let(:report) { 'submitted_claims' }
+      let(:reporter) { Reports::SubmittedClaims }
+
+      include_examples 'package stats in CSV'
```
jsugarman:
```suggestion
      it_behaves_like 'a stats report CSV exporter'
```

spec/services/stats/csv_exporter_spec.rb
```diff
@@ -1,43 +1,62 @@
 require 'rails_helper'
 
+RSpec.shared_examples 'csv exporter output' do
+  it { is_expected.to eq expected_output_with_header }
```
jsugarman: quite hard to follow the indirection
clearer to have a 

```
context 'when headers are provided' do
  let(:options) { { headers: %w[column_1 column_2 column_3] } }
  ...
end
```

lib/tasks/claims.rake
```diff
@@ -143,9 +143,7 @@ namespace :claims do
   end
 
   desc 'Create a report of submitted claims'
-  task :submitted_claims_report, [:filename] => :environment do |_task, args|
-    require_relative 'rake_helpers/submitted_claims'
-
-    RakeHelpers::SubmittedClaims.write args[:filename]
+  task submitted_claims_report: :environment do |_task, args|
+    StatsReportGenerationJob.perform_later('submitted_claims')
   end
```
jsugarman: I don't think we need this anymore, if we are not using the cronjob?!

jsugarman (changes requested):
Just some shared example renaming and i don't think we need that rake task anymore?!

lib/tasks/claims.rake
jrmhaig: Probably not.

jsugarman (approved):
Assuming we want case workers to be able to acces this and aside from a  rather confusing naming issue with a `Stats::ReportGenerator` and a `Stats::StatsReportGenerator`   I think this is good to go.

PR #4098, opened 2021-07-02
[CFP-158] Move capybara config to own file and extend

jsugarman:
#### What
Move capybara config to own file and extend

#### Ticket
relates to [CFP-158](https://dsdmoj.atlassian.net/browse/CFP-158)

#### Why
There is a lot of capybara config for cucumber so this separates
it out to its own file as a first step. 

It would be useful to extend capybara with govuk component 
helpers as we migrate to using the form builder and apply its styling. To 
assist with this we can mixin extensions using composition via the config.

The matcher file that is mixed in here is just using using some tweaked
extensions taken from VCD.

### DIFF
--- a/spec/support/capybara_extensions/matchers.rb
+++ b/spec/support/capybara_extensions/matchers.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Extensions to capybara can be written here
+# and they will be mixed into existing
+# helpers/matchers by the `features/support/capybara.rb`
+#
+# Idea comes from:
+# https://github.com/DavyJonesLocker/capybara-extensions
+#
+
+require_relative 'govuk_component_matchers'
+
+module CapybaraExtensions
+  module Matchers
+    include GOVUKComponent::Matchers
+  end
+end
--- a/features/support/capybara.rb
+++ b/features/support/capybara.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+# Load custom capybara extensions
+#
+# Idea comes from:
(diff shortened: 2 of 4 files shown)


spec/support/capybara_extensions/matchers.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+# Idea comes from:
+# https://github.com/DavyJonesLocker/capybara-extensions
+#
+
+require_relative 'govuk_design_system_form_builder_matchers'
+
+module CapybaraExtensions
+  module Matchers
+    include GOVUKDesignSystemFormBuilder::Matchers
```
jsugarman: change to govukcomponent matchers

spec/support/capybara_extensions/matchers.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Extensions to capybara can be written here
+# and they will be mixed into existing
+# helpers/matchers by the `spec/support/capybara.rb`
```
jsugarman:
```suggestion
# helpers/matchers by the `features/support/capybara.rb`
```

naseberry (approved):
LTGM.
I like the separation.

naseberry:
As you've separated the capybara config to `features/support/capybara.rb`, do you need to require/import this into `features/support/env.rb`?
@jsugarman

jsugarman:
> As you've separated the capybara config to `features/support/capybara.rb`, do you need to require/import this into `features/support/env.rb`?
> @jsugarman

it already is. otherwise the feature tests would fail but have checked anyway. Not quite sure how though!

PR #4108, opened 2021-07-07
[CFP-214] Set unknown_asset_fallback to false

jrmhaig:
#### What

```ruby
Rails.application.config.assets.unknown_asset_fallback = false
```

is the default for Rails 5.1+. CCCD is not using sprockets so this does not have any effect.

#### Ticket

[Rails 6.1 post upgrade: handle config for Rails 5.1+](https://dsdmoj.atlassian.net/browse/CFP-214)

### DIFF
--- a/config/initializers/new_framework_defaults_5_1.rb
+++ b/config/initializers/new_framework_defaults_5_1.rb
@@ -1,0 +1,6 @@
+# From https://edgeguides.rubyonrails.org/configuring.html#configuring-assets
+# config.assets.unknown_asset_fallback allows you to modify the behavior of
+# the asset pipeline when an asset is not in the pipeline, if you use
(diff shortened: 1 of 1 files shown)


jsugarman (approved):
LGTM, i assume all these will either be "promoted" to application config or handled via `load_defaults` eventually?!

jrmhaig:
> LGTM, i assume all these will either be "promoted" to application config or handled via `load_defaults` eventually?!

Yes. Once this and #4107 have been merged and we are happy that there are no problems then `config/application.rb` can be updated with `config.load_defaults 5.1` and `config/initializers/new_framework_defaults_5_1.rb` can be deleted.

jsugarman:
sanity tested on hosted env with no obvious problems

PR #4109, opened 2021-07-07
Remove json importer code

jsugarman:
#### What
Remove `json_import` and `json_import_web_edited` and all references to it

#### Ticket
[original ticket for removing JSON importer/uploader](https://dsdmoj.atlassian.net/browse/CBO-608)

#### Why
System no longer has json imports and there are
no claims on production that have this as a `claim#source`
value any longer. We can therefore safely delete the code.

This is a follow up to https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/2879


jrmhaig (approved):
🎉

PR #4111, opened 2021-07-08
[CFP-215] Enable AES-256-GCM cipher

jrmhaig:
#### What

Rails 5.2 adopts AES-256-GCM as the default cipher for cookies and message.
Currently the old cipher is being used but this will change when the new
defaults are adopted.

This PR explicitly selects the new cipher.

See https://www.bigbinary.com/blog/rails-5-2-uses-aes-256-gcm-authenticated-encryption-as-default-cipher-for-encrypting-messages

#### Ticket

#### TODO

- [x] Deploy to an environment and test. For example, changing the encryption on cookies may affect currently logged in users.

### DIFF
--- a/config/initializers/new_framework_defaults_5_2.rb
+++ b/config/initializers/new_framework_defaults_5_2.rb
@@ -14,3 +14,13 @@ Rails.application.config.action_view.form_with_generates_ids = true
 # #cache_key method that is accompanied by a changing version in the
 # #cache_version method.
 Rails.application.config.active_record.cache_versioning = true
+
+# config.action_dispatch.use_authenticated_cookie_encryption controls whether
+# signed and encrypted cookies use the AES-256-GCM cipher or the older
+# AES-256-CBC cipher. It defaults to true.
+Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
+
+# config.active_support.use_authenticated_message_encryption specifies whether
+# to use AES-256-GCM authenticated encryption as the default cipher for
+# encrypting messages instead of AES-256-CBC.
+Rails.application.config.active_support.use_authenticated_message_encryption = true
--- a/config/initializers/cookie_rotation.rb
+++ b/config/initializers/cookie_rotation.rb
@@ -15,7 +15,7 @@ return unless Rails.application.secrets.old_secret_key_base.present?
 Rails.application.configure do
   # Not technically necessary, as this is the current default,
   # but it is to line up with the config below.
-  config.action_dispatch.use_authenticated_cookie_encryption = false
+  config.action_dispatch.use_authenticated_cookie_encryption = true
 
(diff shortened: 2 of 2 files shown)


config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,9 @@
+# config.action_dispatch.use_authenticated_cookie_encryption controls whether
+# signed and encrypted cookies use the AES-256-GCM cipher or the older
+# AES-256-CBC cipher. It defaults to true.
+Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
+
+# config.active_support.use_authenticated_message_encryption specifies whether
+# to use AES-256-GCM authenticated encryption as the default cipher for
+# encrypting messages instead of AES-256-CBC.
+Rails.application.config.active_support.use_authenticated_message_encryption = true
```
jsugarman: **nit** linebreak

jrmhaig:
Tested on dev-lgfs.

IMO, I don't expect this to have any effect.

jsugarman (approved):
As when we rotated the secret I expect this will log users out. I will test but we then need to decide whether to do a late night deploy and accept the inconvenience to users or [rotate the keys as described here](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-changing-to-use-sha256) and similar to what we have already done

jsugarman:
Is key generation encryption impacted by this and do we need to [amend this line](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/config/initializers/cookie_rotation.rb#L26) and [this line](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/config/initializers/cookie_rotation.rb#L18) to reflect this change to  `aes-256-gcm` as well

jsugarman:
> Is key generation encryption impacted by this and do we need to [amend this line](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/config/initializers/cookie_rotation.rb#L26) and [this line](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/config/initializers/cookie_rotation.rb#L18) to reflect this change to `aes-256-gcm` as well

Myself and peter  have test the impact of this on the session cookie and it does not log the user out when deployed! so good to go

jsugarman:
added a commit to:

>Amend cookie rotation initializer inline with use_authenticated_cookie_encryption

>The cookie rotation initializer is only used when we want to
rotate an old secret_key_base. It should, however, use the new
defaults in line with the apps config.

>see [key generator cookie rotation](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-changing-to-use-sha256)
for details.

PR #4112, opened 2021-07-08
[CFP-215] Set method for generating non-sensitive digests

jrmhaig:
#### What

config.active_support.hash_digest_class allows configuring the digest class to use to generate non-sensitive digests, such as the ETag header.

The default method changes with Rails 5.2 to OpenSSL::Digest::SHA1. This PR prepares for the Rails 5.2 defaults being adopted.

#### Ticket

#### TODO

- [x] Deploy to an environment and test.


jrmhaig:
Tested on dev-lgfs

IMO, this will have no noticeable effect. It is possible that a cached page view may be invalidated unnecessary. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag for information about ETags.

jsugarman (commented):
The[ latest version is actually SHA256](hash_digest_class) so wondering whether we should bump to that

jsugarman (approved):

PR #4113, opened 2021-07-08
[CFP-215] Set default_protect_from_forgery to true

jrmhaig:
#### What

Prior to 5.2 the `ApplicationController` required

```
protect_from_forgery prepend: true, with: :exception
```

This is now added to `ActionController::Base`, but only if default_protect_from_forgery is set to true. This becomes the default setting when the 5.2 defaults are loaded. This commit prepares for this by setting it explicitly and updating the controllers accordingly.

#### Ticket

#### TODO

- [x] Deploy to an environment and test. Specifically,
  - check that form submission with an invalid CSRF token fails
  - check that the API, and any other urls that do not need require CSRF, still works

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,7 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
-  protect_from_forgery prepend: true, with: :exception unless ENV['DISABLE_CSRF'] == '1'
+  skip_forgery_protection if ENV['DISABLE_CSRF'] == '1'
 
   include CookieConcern
   before_action :set_default_cookie_usage
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,6 +1,6 @@
 class ErrorsController < ApplicationController
   skip_load_and_authorize_resource only: %i[not_endpoint not_found internal_server_error dummy_exception]
-  protect_from_forgery prepend: true, except: %i[not_endpoint not_found internal_server_error]
+  skip_forgery_protection only: %i[not_endpoint not_found internal_server_error]
 
   def not_endpoint
     logger.info("Data POSTed to root with API key: #{not_endpoint_params[:api_key]}") if params.present?
--- a/app/controllers/geckoboard_api/application_controller.rb
+++ b/app/controllers/geckoboard_api/application_controller.rb
@@ -1,5 +1,4 @@
 class GeckoboardApi::ApplicationController < ActionController::Base
-  protect_from_forgery prepend: true, with: :exception
   before_action :authenticate_token!
 
   private
--- a/config/initializers/new_framework_defaults_5_2.rb
+++ b/config/initializers/new_framework_defaults_5_2.rb
@@ -1,3 +1,7 @@
 # config.active_support.hash_digest_class allows configuring the digest class
 # to use to generate non-sensitive digests, such as the ETag header.
 Rails.application.config.active_support.hash_digest_class = OpenSSL::Digest::SHA1
+
+# config.action_controller.default_protect_from_forgery determines whether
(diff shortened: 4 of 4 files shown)


jrmhaig:
Confirmed that and incorrect CSRF prevents a form submission and the API still works.

It is possible that the `skip_forgery_protection` lines in `app/controllers/errors_controller.rb` and `app/controllers/application_controller.rb` are not necessary.

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,7 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
-  protect_from_forgery prepend: true, with: :exception unless ENV['DISABLE_CSRF'] == '1'
+  skip_forgery_protection if ENV['DISABLE_CSRF'] == '1'
```
jsugarman, line 4: check if default behaviour is `with: :null_session` rather than  `with: :exception` and the impact of this

app/controllers/errors_controller.rb
```diff
@@ -1,6 +1,6 @@
 class ErrorsController < ApplicationController
   skip_load_and_authorize_resource only: %i[not_endpoint not_found internal_server_error dummy_exception]
-  protect_from_forgery prepend: true, except: %i[not_endpoint not_found internal_server_error]
+  skip_forgery_protection only: %i[not_endpoint not_found internal_server_error]
```
jsugarman, line 3: check if default behaviour is `with: :null_session` rather than  `with: :exception` and the impact of this

app/controllers/application_controller.rb
jsugarman: appears to be defaulting to raising `ActionController::InvalidAuthenticityToken` which because we have a `rescue_from ActionController::InvalidAuthenticityToken` in application_controller it redirects back to referrer or unauthenticated root path

app/controllers/errors_controller.rb
jsugarman: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4113/files#r669152854

app/controllers/application_controller.rb
jsugarman: as covered here https://github.com/rails/rails/pull/29742/files#diff-41177c56a83ce1447bfba614c4b7f29f3deef57f2312a9e0591763c489e76099R73-R79

app/controllers/geckoboard_api/application_controller.rb
```diff
@@ -1,5 +1,4 @@
 class GeckoboardApi::ApplicationController < ActionController::Base
-  protect_from_forgery prepend: true, with: :exception
```
jsugarman, line 2: Out of scope but this endpoint and other geckoboard controllers should be `skip_forgery_protection` as they are either used by the geckoboard API (which cannot provide a CSRF authentication token) or are get requests which should/need not use CSRF protection anyway `rails routes | grep geckoboard` - see https://stackoverflow.com/questions/9446946/csrf-protection-with-http-get-requests-in-rails

jsugarman: see https://stackoverflow.com/questions/9446946/csrf-protection-with-http-get-requests-in-rails

jsugarman (approved):
Have sanity checked myself both by:

1. trying CSRF attacks - amending/removing the authentication token, causing an InvalidAuthenticityToken to be raised and rescued with a redirect to referrer
2. checking rails ordering of before actions - these do what is needed by verifying the authentication token before doing anything application specific, as `prepend: true` was used to achieved previously.

So i do not expect this PR to have any impact.

I believe their could be a bug related to [previous and current use of protect_from_forgery in geckoboard related endpoints, as mentioned here](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4113/files#r669157052). I will raise a separate PR for this.

PR #4115, opened 2021-07-08
[CFP-215] Set cache_versioning to true

jrmhaig:
#### What

See https://www.bigbinary.com/blog/rails-adds-support-for-recyclable-cache-keys

Set `Rails.application.config.active_record.cache_versioning` to true. This will be the default when the defaults for Rails 5.2 are loaded so this PR prepares for this change.

#### Ticket

#### TODO

- [x] Deploy to an environment and test.


jrmhaig:
I don't think we are actually using these cache keys. See https://www.bigbinary.com/blog/rails-adds-support-for-recyclable-cache-keys and https://www.bigbinary.com/blog/activerecord-relation-cache-key

jsugarman (approved):
👍  I agree, it appears we do not use active_record caching explicitly anyway. There is a an implicit [ActiveRecord SQL Cache](https://www.honeybadger.io/blog/rails-activerecord-caching/) but this is not impacted but this change from what i can tell.

PR #4117, opened 2021-07-12
[CFP-214] Remove unneeded `local: true` from form_with calls

jsugarman:
#### What
Remove unneeded `local: true` from form_with calls

#### Ticket

relates to [CFP-214](https://dsdmoj.atlassian.net/browse/CFP-214)

#### Why
We have adopted [rails 6.1 default for form_with remote calls in this PR](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4107)

```
form_with_generates_remote_forms = false
```

As such the `local: true` is superfluos as it is the default.
Going forward any forms that require ajax type logic will need
to explicitly add `remote: true` to their `form_with` blocks.

see https://guides.rubyonrails.org/configuring.html for details

--------

#### TODO (wip)

 - [x] sanity test


kmahern (approved):
lgtm (and presumably something we also need to do on VCD) 👍

jsugarman:
> lgtm (and presumably something we also need to do on VCD) 👍

thanks and yes, quite possibly if you have `form_with(..., ..., local: true)` anywhere and `form_with_generates_remote_forms = false` set

PR #4125, opened 2021-07-14
[CFP-215] add missing geckoboard request specs

jsugarman:
#### What
Add missing request specs

#### Ticket

[CFP-215](https://dsdmoj.atlassian.net/browse/CFP-215)

#### Why
They were missing plus am wanting to test they work
with CSRF forgery protection enbled or disabled as the
controller should probably implement `skip_forgery_protection`.

I believe CSRF protection is not needed for these requests because:
a. they are only GET requests, and GET's are irrelevant to CSRF attacks
b. the json formatted responses are intended for consumption by geckoboard API and requests will, therefore, never include a CSRF authentication token


naseberry (approved):
Looks good, however, I thought we were going to remove these routes?

jsugarman:
> Looks good, however, I thought we were going to remove these routes?

Yes, indeed. we probably should but is beyond the scope of this work as PM may find a sue for them if they are aware of them
and there are separate tickets in backlog related to Geckoboard.

I was merely trying to add specs to cover the existing codebase and ensure it works with or without forgery protection, as it should. t does bump the coverage by 0.2%.

PR #4130, opened 2021-07-19
[CFP-216] Adopt rails 6.0 default for `collection_cache_versioning` (true)

jsugarman:
#### Warning
This setting has no effect because the govuk_notify_rails gems
(by the looks of it) is misbehaving and causing the rails framework
to load too early - see this [issue comment](https://github.com/rails/rails/issues/39855#issuecomment-659670294) and this
[useful debugging issue comment](https://github.com/rails/rails/issues/37030#issuecomment-524511912) for more information and details.

#### What
Set rails 6.0 default for `collection_cache_versioning` (true)

#### Ticket

#### Why
Improved caching for SQL queries

Enables the same cache key to be reused when the object being
cached of type ActiveRecord::Relation changes by moving the
volatile information (max updated at and count) of the relation's
cache key into the cache version to support recycling cache key.


naseberry (approved):
Thanks for the explanation, it is good to go.

PR #4135, opened 2021-07-19
[CFP-216] Set rails 6.0 default `action_mailer.delivery_job` ("ActionMailer::MailDeliveryJob")

jsugarman:
#### What
Set rails 6.0 default `action_mailer.delivery_job` ("ActionMailer::MailDeliveryJob")

Changes value from baseline:
```
ActionMailer::DeliveryJob --> "ActionMailer::MailDeliveryJob"
```

#### Ticket

#### Why
Inline with rails 6.0 defaults

Not clear if this has any effect as we use govuk_notify_rails
for mail sending anyway.

 - [x] check setting "sticks"


jsugarman:
have tested mail sending on staging successfully with new config.

naseberry (approved):
I was able to conduct a 'password reset' on the staging environment.
Although a single test, it seems that GOVUK mailer has not been affected.

PR #4141, opened 2021-07-20
[CFP-217] Adopt rails 6.1 default of cookies_same_site_protection (:lax)

jsugarman:
#### What
Adopt rails 6.1 default of cookies_same_site_protection (:lax)

#### Ticket

#### Why

Improved CSRF protection

 > Configures the default value of the SameSite attribute
   when setting cookies. When set to nil, the SameSite
   attribute is not added.

see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
for more.

#### TODO (wip)

 - [X] check config sticks (SameSite header exists on any page)
 - [x] Have FE consider/review impact


naseberry (approved):
👍  

Tested on api-sandbox, a value of 'Lax' is now being set for SameSite attribute.
There are no effects to setting this.

PR #4157, opened 2021-07-27
[CFP-213] Adopt rails 5.0 default of true for `belongs_to_required_by_default`

jsugarman:
#### What
Adopt rails 5.0 default of true for `belongs_to_required_by_default`

#### Ticket

[came out of CFP-213](https://dsdmoj.atlassian.net/browse/CFP-213)

#### Why

This will require quite a bit of remedial work to
implement but would be nice to adopt this new
default.

By removing the config we accept the default,
since rails 5.0 of `true`.

[Relevant rails config documentation](https://guides.rubyonrails.org/configuring.html#for-5-0-baseline-defaults-from-below-and)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,14 +42,11 @@ module AdvocateDefencePayments
     #
     config.autoloader = :classic
     config.action_mailer.deliver_later_queue_name = :mailers
-    config.active_record.belongs_to_required_by_default = false
     config.active_storage.queues.analysis = :active_storage_analysis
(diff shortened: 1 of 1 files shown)


jsugarman:
Remedial action needed - something that could be worked on in spare time or for a change.

jsugarman:
closing for now as too much remedial work required.

 - will rails ever remove this option and default to true forever?!
 - what advantages are there. reverse looks `provider.supplier_numbers.first` --> `supplier_number.provider`??

PR #4164, opened 2021-07-28
[CFP-162] Govuk Form builder migration: custom error handling

jsugarman:
#### What
Govuk Form builder migration: custom error handling
 - migrate provider forms to custom error handler
 - inject custom error presenter into govuk-formbuilder
 - create functional and accessible error linking
   between error summaries and field level errors

#### Ticket

depends on [spike CFP-197](https://dsdmoj.atlassian.net/browse/CFP-197)
[implementation CFP-162](https://dsdmoj.atlassian.net/browse/CFP-162)

#### Why
To unblock migration to the govuk-formbuilder we must be able to have customized error messages for the
summary

As part of work to use custom error handling with the `govuk-formbuilder` I am migrating
the existing provider forms to use custom error handling.

This is because the provider form:
 - is a relatively simple form with nested attributes
 - does not use custom error handling
 - does use the govuk-formbuilder
 - does not currently handle nested errors in an accessible way.
 -  make error handling consistent with claim forms
 - provide a proof of concept for migrating claim forms to `govuk-formbuilder` while retaining our custom error handling.

##### Enable govuk-formbuilder to use custom error presenter

The [govuk-formbuilder](https://github.com/DFE-Digital/govuk-formbuilder) master branch now enables presenter injection
  
**Summary errors** are generated by injecting our custom error presenter into the govuk_error_summary helper method.
The formbuilder will call `formatted_error_messages` to generate attribute error message pairs. The formbuilder will 
use the attribute error message pairs to generate the summary errors with a link to the relevant attribute. 

see the full guide (once updated) for details and an interactive example.

**Field errors** cannot, as yet, be customised using the formbuilder beyond
the standard rails error lookup mechanism. This generally works in any event
but will need reviewing when it comes to migrating the claims forms.

Field error presenter injection may be possible in future iterations but for now 
I have created an alternative that relies on two facts:

1. rails validation errors FOR FIELD ERRORS will be stored in conventional
translation locations location - https://guides.rubyonrails.org/i18n.html#error-message-scopes

2. govuk-formbuilder retrieves error messages set on the attribute and
displays them on the field level for any fields, nested or otherwise.

The presenter or translator can then be used for summary and api errors only. field
errors must be stored in the conventional location.

The translator converts each error message on an object, underscoring and
downcasing it, before it the looks up that error "key" on the attribute in
translations.

So an activemodel error of:
```
activerecord:
  errors:
    models:
      provider:
        attributes:
          role:
            blank: 'Choose at least one role'
```
becomes a translator lookup in custom `provider.yml` translations of:
```
role:
  choose_at_least_one_role:
    long: blah
    short: _cannot_be_looked_up
    api: blah
```

**The result is:**
1. Field errors therefore come from rails conventional location
2. Summary errors come from injecting error presenter into `govuk_error_summary` helper
3. API errors are specific to CCCD and will continue to function as normal

**When we come to migrate the claim forms we:**
1.  will need to add activerecord.errors.models.[model_name].attributes.[attribute_name] errors to replace custom "short" error messages and update custom error message keys as outlined above (no small task)
2. will need to instantiate and inject the error presenter into govuk_error_summary for each page form (assuming we can get the gem amended)


app/presenters/error_presenter.rb
```diff
@@ -1,18 +1,26 @@
 class ErrorPresenter
   attr_reader :error_details
 
-  def initialize(claim, message_file = nil)
-    @claim = claim
-    @errors = claim.errors
-    message_file ||= "#{Rails.root}/config/locales/error_messages.#{I18n.locale}.yml"
+  def initialize(object, message_file = nil)
+    @object = object
```
jrmhaig: I don't think that `@object` is actually used anywhere so this line can be removed.

jrmhaig (approved):
I've had a look over this again and checked on dev-lgfs. Everything seems OK. :+1:

PR #4165, opened 2021-08-03
Standardize messages files names

jrmhaig:
#### What

Rename the error messages file for claims to follow the same pattern as the new file for providers. Then allow the error presenter to automatically find the correct file based on the model name.

**Note:** I didn't realise that `ErrorMessageTranslator` is also used in `app/interfaces/api/error_response.rb`, and the messages file isn't just for the claims model. Therefore renaming `config/locales/error_messages.en.yml` to `config/locales/en/error_messages/claim.yml` is not clear unless it is then split up.


jsugarman:
Closing as partially implemented in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4180 and the magic of deriving file name from class of object may confuse and may not work for namespaced  claim objects in any event

Such as:
```
Claim::AdvocateClaim --> claim_advocate_claim
```

PR #4174, opened 2021-08-04
[CFP-103] Remove Paperclip fields from messages

jrmhaig:
#### What

Remove fields used by the Paperclip gem from the `messages` table.

#### Ticket

[Remove Paperclip fields from the database](https://dsdmoj.atlassian.net/browse/CFP-103)

#### Why

We moved from Paperclip to Active Storage for document storage several weeks ago and there have been no issues that would require us to roll back the change. The Paperclip specific fields can now be cleared out of the database.

#### How

A database migration that will be applied automatically when deployed to each environment; dev-lgfs, dev, staging, api-sandbox and production.


jsugarman (approved):
👍

PR #4176, opened 2021-08-05
[CFP-103] Remove report field from stats_reports table

jrmhaig:
#### What

Remove the `report` field from the `stats_reports` table.

#### Ticket

Related to: [Remove Paperclip fields from the database](https://dsdmoj.atlassian.net/browse/CFP-103)

#### Why

The `report` field in the `Stats::StatsReport` model used to contain the data of the report but that was moved into AWS a long time ago. This field is therefore unused now, although errors from failed report generation may get written there.

#### How

Remove the field from the database and stop attempting to write errors to it.

### DIFF
--- a/db/migrate/20210805124942_remove_report_from_stats_reports.rb
+++ b/db/migrate/20210805124942_remove_report_from_stats_reports.rb
@@ -1,0 +1,5 @@
+class RemoveReportFromStatsReports < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :stats_reports, :report, :string, after: :report_name
+  end
+end
--- a/db/migrate/20210805124942_remove_report_from_stats_reports.rb
+++ b/db/migrate/20210805124942_remove_report_from_stats_reports.rb
@@ -1,5 +1,5 @@
 class RemoveReportFromStatsReports < ActiveRecord::Migration[6.1]
   def change
-    remove_column :stats_reports, :report, :string, after: :report_name
+    remove_column :stats_reports, :report, :string
   end
 end
--- a/app/models/stats/stats_report.rb
+++ b/app/models/stats/stats_report.rb
@@ -1,15 +1,3 @@
-# == Schema Information
-#
-# Table name: stats_reports
-#
-#  id           :integer          not null, primary key
-#  report_name  :string
-#  report       :string
-#  status       :string
-#  started_at   :datetime
-#  completed_at :datetime
-#
-
 module Stats
   class StatsReport < ApplicationRecord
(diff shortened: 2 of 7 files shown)


jsugarman (commented):
Seems ok but what errors do we have in that field on production at the moment, and where does the notification go?

db/migrate/20210805124942_remove_report_from_stats_reports.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveReportFromStatsReports < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :stats_reports, :report, :string, after: :report_name
```
jrmhaig: I was hoping that `after: :report_name` would make a `db:rollback` put the field back in the same place in the database table but apparently it doesn't. :-(

jrmhaig:
> Seems ok but what errors do we have in that field on production at the moment, and where does the notification go?

The errors are when the stats report generation fails and currently there are none:

```ruby
irb(main):001:0> Stats::StatsReport.pluck(:report)
=> [nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil]
```

The errors will be reported in Sentry, as they result from exceptions, and I believe that the `notify_error` method will send a message to Slack, although I do not know much about that.

**Update:** We could write the errors to the file in AWS to replicate what was originally intended, if that is of interest. I guess that this was an oversight when the reports were moved.

db/migrate/20210805124942_remove_report_from_stats_reports.rb
jsugarman: never heard of an `after:` option. do you have a link to its documentation?

jsugarman (approved):
👍  Happy for this to go in

However
1. it would be good to understand where the notification now goes 
2. if that `after:` is not need can we remove it.
3. fixe for the cop violation/advisory

db/migrate/20210805124942_remove_report_from_stats_reports.rb
jrmhaig: Hmmm, it appears that `after` is a MySQL extension and it works with the `mysql` adapter but not `pg`.

It is really for a `create_column` to make sure that the new column is created in the correct place relative to other columns, so on the `remove_column` it is an instruction for the rollback.

PR #4177, opened 2021-08-05
[CFP-240] refactor custom error handler specs

jsugarman:
#### What
Refactor unit tests for custom error handling code

#### Ticket

[relates to CFP-240](https://dsdmoj.atlassian.net/browse/CFP-240)

#### Why
The error message translator and error presenter specs can be improved:
 - for clarity following [betterspecs](https://www.betterspecs.org/) practices.
 - to cover missing cases
 - to facilitate refactoring of the translater and presenters in particular

### DIFF
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -22,22 +22,25 @@ RSpec.describe Provider, type: :model do
   let(:chamber) { create(:provider, :chamber) }
   let(:agfs_lgfs) { create(:provider, :agfs_lgfs) }
 
-  it { should have_many(:external_users) }
-  it { should have_many(:claims) }
+  it { is_expected.to have_many(:external_users) }
+  it { is_expected.to have_many(:claims) }
 
-  it { should validate_presence_of(:provider_type) }
-  it { should validate_presence_of(:name) }
-  it { should validate_uniqueness_of(:name).ignoring_case_sensitivity.with_message(:not_unique) }
+  it { is_expected.to validate_presence_of(:provider_type) }
+  it { is_expected.to validate_presence_of(:name) }
+  it { is_expected.to validate_uniqueness_of(:name).ignoring_case_sensitivity.with_message(:not_unique) }
 
-  it { should delegate_method(:advocates).to(:external_users) }
-  it { should delegate_method(:admins).to(:external_users) }
+  it { is_expected.to delegate_method(:advocates).to(:external_users) }
+  it { is_expected.to delegate_method(:admins).to(:external_users) }
 
   context '#destroy' do
     before { create(:external_user, :advocate, provider: chamber) }
+
     it 'destroys external users' do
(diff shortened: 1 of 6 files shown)


spec/models/provider_spec.rb
```diff
@@ -214,70 +216,86 @@
...
-  context 'AGFS supplier number validation for a chamber' do
-    context 'for a blank supplier number' do
+  context 'when validating AGFS supplier number for a chamber' do
+    context 'with a blank supplier number' do
+      before { chamber.firm_agfs_supplier_number = '' }
+
       it 'returns no errors' do
-        chamber.firm_agfs_supplier_number = ''
         expect(chamber).to be_valid
       end
```
jrmhaig, line 225: (nit)

```ruby
it { is_expected.to be_valid }
```

or

```ruby
it 'returns no errors' do
  expect { chamber.save }.not_to raise_error
end
```

I'd favour the first as the current version isn't actually testing "returns no errors".

**Update:** I've noticed that `chamber` isn't the subject here (although it is elsewhere) so this will not work without other changes.

jsugarman: There is a lot more i could have done but need to stop yak shaving.

spec/models/supplier_number_spec.rb
```diff
@@ -44,37 +33,49 @@
...
-  context 'validates supplier number format' do
+  context 'when validating supplier number' do
     let(:format_error) { ['Enter a valid LGFS supplier number'] }
 
-    it 'fails for incorrect format' do
-      allow(subject).to receive(:supplier_number).and_return('ABC123')
-      expect(subject).not_to be_valid
-      expect(subject.errors[:supplier_number]).to eq(format_error)
+    before do
+      allow(supplier).to receive(:supplier_number).and_return(supplier_number)
```
jrmhaig: Shouldn't really stub the thing under test. I think this is the equivalent (without stubbing):

```suggestion
      supplier.supplier_number = supplier_number
```

spec/presenters/error_message_translator_spec.rb
```diff
@@ -67,171 +86,160 @@
     }
   end
 
-  let(:emt) { ErrorMessageTranslator.new(translations, key, error) }
+  it { is_expected.to respond_to(:long_message, :short_message, :api_message) }
```
jrmhaig: I don't think this is necessary as all these three methods are tested elsewhere.

jsugarman: actually they are not explicitly tested anywhere so this is as much for documentary purposes IMO

jsugarman: removed

spec/models/supplier_number_spec.rb
jsugarman: changed

jrmhaig (approved):

PR #4178, opened 2021-08-05
[CFP-240] Refactor and modify custom error translation

jsugarman:
#### What
Refactor error message translation and enable it
to work with rails validation errors.

#### Ticket

[CFP-240](https://dsdmoj.atlassian.net/browse/CFP-240)

#### Why

 - To enable custom error handling to work with both
   existing custom validation errors and those that
   are generated from standard rails validations. The
   later are typically of the form.

    ```
    foos_attributes_4_bars_attributes_3_baz
    ```
    or
    ```
    foo.bar.baz
    ```

 - To break up this complex class and apply SOLID principles

This class is currently doing way too much and not easy to understand
or maintain or test. General SOLID principle violations abound.

In attempting to reduce the smells I have written tests to cover
the various possible imputs and expected outputs before
attempting to extract logic to other classes.

I have extracted:
 - translator logic to `ErrorMessage::Translator` class
 - fallback message logic to `ErrorMessage::Fallback` class
 - the notion of a "key" to a `ErrorMessage:Key` class
 - generic helpers for sharing to `ErrorMessage::Helper` module

Further worker needed includes:
 - Refactor ErrorPresenter to use the helper mixin
 - Refactor ErrorPresenter to remove duplicate logic related to submodels
 - Refactor ErrorPresenter to move file loading out if sensible


spec/presenters/error_message/translator_spec.rb
```diff
@@ -0,0 +1,374 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/BlockLength
```
jrmhaig: `shared_examples` is listed as an exception for `Metrics/BlockLength` in `.rubocop.yml` so it makes sense to add `shared_context` too.

spec/presenters/error_message/key_spec.rb
```diff
@@ -0,0 +1,313 @@
...
+  end
+
+  describe '#all_model_indices' do
+    subject { error_key.all_model_indices }
+
+    context 'when single nested' do
+      context 'with rails nested error' do
+        let(:key) { 'foos_attributes_4_bar' }
+
+        it { is_expected.to eq({ 'foo' => '4' }) }
```
jrmhaig, line 245: Can this be

```suggestion
        it { is_expected.to eq({ foo: 4 }) }
```

(Will the index ever be anything other than an integer?)

Likewise in all other cases.

app/presenters/error_message/translator.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    def get_messages(translations, key, error)
+      if key.submodel?
+        translation_subset, submodel_key = extract_submodel_attribute(key)
+        get_messages(translation_subset, submodel_key, error)
+      elsif translation_exists?(translations, key, error)
+        @long_message = translations[key][error]['long']
+        @short_message = translations[key][error]['short']
+        @api_message = translations[key][error]['api']
```
jsugarman, line 51: Use a factory to remove conditional:

@long_message, @short_mesasge, @api_message = GetMessage.new(@key, @error).messages
@long.messages ... = message_klass.new()

app/presenters/error_presenter.rb
```diff
@@ -21,38 +21,25 @@ def initialize(claim, message_file = nil)
...
-      )
-    else
-      OpenStruct.new(
-        long: generate_standard_long_message(error),
-        short: generate_standard_short_message(error),
-        api: generate_standard_api_message(error)
-      )
-    end
+  def translate(error)
+    ErrorMessage::Translator.new(@translations, error.attribute, error.message)
```
jsugarman, line 36: @translator = ErrorMessage::Translator.new(@translations)

@translator.translate(error)

ErrorMessage::Translator.new(error.attribute, error.message)

app/presenters/error_message/translator.rb
```diff
@@ -0,0 +1,84 @@
...
+  class Translator
+    include ErrorMessage::Helper
+
+    attr_reader :long_message,
+                :short_message,
+                :api_message,
+                :translations,
+                :translator
+
+    def initialize(translations, key, error)
```
jsugarman, line 22: - we could accept translations
- we could have translate method on instance that returns struct/hash/ErrorMessage::Message

```
instance = ErrorMessage::Translator.new(translations)
message = instance.translate(error)
message.long
message.short
message.api
```

app/presenters/error_message/translator.rb
jsugarman: agree to leave this conditional removal for now

app/presenters/error_presenter.rb
```diff
@@ -21,38 +21,25 @@ def initialize(claim, message_file = nil)
...
   end
 
   def last_parent_attribute(_translations, key)
-    attribute = ErrorMessageTranslator.association_key(key)
+    attribute = ErrorMessage::Translator.association_key(key)
     while attribute =~ SUBMODEL_REGEX
```
jsugarman, line 43: use helper instead

app/presenters/error_message/key.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def parse
+      attribute = association_key(self)
+      all_model_indices = {}
+
+      while attribute.match(numbered_model_regex)
+        model = Regexp.last_match(1)
+        model.slice!('_attributes')
+        model = model.singularize
+        all_model_indices[model] = Regexp.last_match(3)
```
jsugarman, line 49: consider
```suggestion
        all_model_indices[model.to_sym] = Regexp.last_match(3).to_int
```

spec/presenters/error_message/translator_spec.rb
jsugarman: its also the `let` block that it is violating this cop, so I have added that exclusion too since we exclude most/all rspec block methods anyway.

.rubocop.yml
```diff
@@ -49,6 +49,7 @@ Metrics/BlockLength:
     - resource
     - context
     - describe
+    - let
```
jrmhaig, line 52: I wouldn't include `let` here. I would have an inline exception for this (or include it in `.rubocop_todo.yml`), with a view to trying to fix it in the future.

NB, `spec/presenters/error_message_translator_spec.rb` was already in `rubocop_todo.yml` but then the file was renamed.

jsugarman: we are ignoring pretty much everything else for rspec though, including `it` and `specify`, which are arguably more important to restrict the size of -  i mention this in commit, so am not sure why we would not just add this too?!

app/presenters/error_message/translator.rb
jsugarman: May have time to get this in over the weekend

jrmhaig (approved):
I have some outstanding comments but they are not blockers, considering that this is part of a much bigger clean-up.

app/presenters/error_presenter.rb
jsugarman: see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4180

app/presenters/error_message/translator.rb
jsugarman: see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4180

PR #4179, opened 2021-08-06
Update schema to reflect existing foreign key

jsugarman:
Update schema to reflect existing foreign key

#### What
Update schema to reflect existing foreign key

#### Ticket

n/a

#### Why
When I run migrations (i think this happened in the past too) it changes the
`schema.rb` to add this line:
```
add_foreign_key "injection_attempts", "claims"
```

From what I can tell this FK exists on all instance dbs anyway…
```
=> \d claims
...
Referenced by:
    TABLE "injection_attempts" CONSTRAINT "fk_rails_fa68ab3e28" FOREIGN KEY (claim_id) REFERENCES claims(id)
```
So perhaps we should commit this change to the schema

There are also some inconsequential column ordering changes

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -220,8 +220,8 @@ ActiveRecord::Schema.define(version: 2021_08_04_123914) do
     t.datetime "updated_at"
     t.date "date_to"
     t.uuid "uuid", default: -> { "uuid_generate_v4()" }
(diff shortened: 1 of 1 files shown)


jrmhaig (approved):
I have been seeing this too but thought it was just due to my local database.

PR #4180, opened 2021-08-07
[CFP-240] Refactor ErrorMessage and namespace objects

jsugarman:
#### What
Modify `ErrorMessage::Translator` and other `ErrorMessage` namespaced objects

#### Ticket

[CFP-240](https://dsdmoj.atlassian.net/browse/CFP-240)

#### Why
To reduce code smell and facilitate injection of custom error handling presenter into
[govuk-formbuilder](https://github.com/DFE-Digital/govuk-formbuilder)

#### How
Following on from previous work this separates instantiation and retrieval of error messages 
to facilitate instance reuse. The primary aim being as below

```ruby
translator = ErrorMessage::Translator.new(translations)
message = translator.message(key, error)
message.long --> "A long message"
message.short --> "A short message"
message.api --> "An api message"
```

This then enables further refactoring, here and in future.


app/interfaces/api/error_response.rb
```diff
@@ -45,45 +45,37 @@ def initialize(object)
...
-    end
+    def build_error_response
+      raise 'unable to build error response as no errors were found' if @model.errors.empty?
 
-    def submodel_prefix
-      submodel_instance_num = ''
-      m = @model
+      fetch_translated_error_messages
+      @body = error_messages
+      @status = 400
```
jrmhaig, line 53: (nit)

```suggestion
      @status = :bad_request
```

Also see line 40, above, which could have `? :unauthorized : :bad_request`

(possible future work but definitely not to do here) Might there be any value in having different status codes depending on the error?

app/presenters/error_message/key.rb
```diff
@@ -52,5 +65,13 @@ def parse
...
       { model: model, attribute: attribute, all_model_indices: all_model_indices }
     end
+
+    def numbered_model_regex
+      /^(\S+?)(_(\d+)_)(\S+)$/
+    end
+
+    def unnumbered_model_regex
+      /^(\S+?)\.(\S+)$/
+    end
```
jrmhaig, line 75: Consider having these as constants, which can then be frozen for a slight performance gain.

app/presenters/error_message/key.rb
```diff
@@ -32,14 +34,25 @@ def all_model_indices
...
     end
 
+    # Support for keys in nested attribute format
+    #
+    # Examples:
+    # fixed_fee.date_attended_1_date --> fixed_fee_0_date_attended_0_date
+    # defendant.representation_order.maat_reference --> defendant_0_representation_order_0_maat_reference
+    #
+    def association_key
+      return self if to_s.index('.').blank?
```
jrmhaig: (nit) Line space after guard clause. (Doesn't Rubocop pick this up?)

Also, I haven't checked this but this may work:

```suggestion
      return self unless self.include?('.')
```

app/presenters/error_message/message.rb
```diff
@@ -0,0 +1,61 @@
...
+  class Message
+    def initialize(long, short, api, key)
+      @long = long
+      @short = short
+      @api = api
+      @key = key
+    end
+
+    def long
+      substitute_submodel_numbers_and_names(@long, @key)
```
jrmhaig:
```suggestion
      substitute_submodel_numbers_and_names(@long)
```

See below.

app/presenters/error_message/message.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def short
+      substitute_submodel_numbers_and_names(@short, @key)
```
jrmhaig:
```suggestion
      substitute_submodel_numbers_and_names(@short)
```

See below.

app/presenters/error_message/message.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def api
+      substitute_submodel_numbers_and_names(@api, @key)
```
jrmhaig:
```suggestion
      substitute_submodel_numbers_and_names(@api)
```

See below.

app/presenters/error_message/message.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    private
+
+    def substitute_submodel_numbers_and_names(message, key = @key)
+      key.all_model_indices.each do |model_name, number|
+        int = number.to_i
+        int += 1 if key.zero_based?
```
jrmhaig: See suggestions on lines 15, 19 and 23.

```suggestion
    def substitute_submodel_numbers_and_names(message)
      @key.all_model_indices.each do |model_name, number|
        int = number.to_i
        int += 1 if @key.zero_based?
```

app/interfaces/api/error_response.rb
jsugarman: > (possible future work but definitely not to do here) Might there be any value in having different status codes depending on the error?

possibly but anything major would need checking with vendors.

the rails status code helpers are only for use with the likes of `render status: :bad_request` in a controller i believe
https://makandracards.com/makandra/49265-list-of-status-symbols-for-rendering-in-rails

It does not work here in any event, from tests which all fail with this.

app/presenters/error_message/key.rb
jsugarman: can freeze them in place. not sure of implications.
`/^(\S+?)\.(\S+)$/.freeze`

app/presenters/error_message/key.rb
jsugarman: did this, which works. remove redundant `self`
```
return self unless include?('.')
...
```

app/presenters/error_message/message.rb
jsugarman: 👍  yeah this was a leftover from a previous refactoring which had the old and new styles working simultaneously

app/presenters/error_message/message.rb
jsugarman: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4180#discussion_r686817255

app/presenters/error_message/message.rb
jsugarman: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4180#discussion_r686817255

app/interfaces/api/error_response.rb
jrmhaig: I had forgotten that this is Grape, not a Rails controller.

app/presenters/error_message/key.rb
jrmhaig: Sorry, I'm having second thoughts about the freezing - not sure if it has much effect on a regex. Is there a reason not to have these as constants, though? I'm thinking about the the number of times the regex instance needs to be created.

jsugarman: i don't think it makes a lot of difference to be honest which is why i moved it out of constants, for readability. network round trips are always going to be the limiting factor over such minor optimisations.

jsugarman: Do you want me to remove the freeze? I am not really inclined to revert back to using constants

jrmhaig: Yes please. I don't think that the freeze is doing anything here.

jrmhaig (approved):

PR #4181, opened 2021-08-10
Fix lgfs summary fields

naseberry:
#### What
Fix claim summary table layout.

#### Ticket
n/a

#### Why
Summary page, LGFS Fixed fee table was broken due to conditional logic.

Table columns (date, quantity and rate) should not be shown for LGFS misc fees but do for AGFS misc fees and AGFS/LGFS fixed fees.

| Before      | After |
| ----------- | ----------- |
| summary_table_error      | Screenshot 2021-08-10 at 11 21 45       |


jsugarman (approved):
🎖️

PR #4189, opened 2021-08-11
[CFP-243] Fix document upload feedback

jrmhaig:
#### What

Following the removal of Paperclip fields (#4175) the user feedback on the evidence upload form was broken as it depended on this information.

#### Ticket

[Document upload description ) #252888](https://dsdmoj.atlassian.net/browse/CFP-243)

#### Why

The `document_file_name` field, which was one of the Paperclip fields, was used to display the filename to the user.

#### How

Instead of returning the entire document model in response to a successful upload only return the id and filename, now taken from the Active Storage attachment.

#### Before

Screenshot 2021-08-11 at 12 41 10

#### After

Screenshot 2021-08-11 at 12 41 19

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -23,7 +23,7 @@ class DocumentsController < ApplicationController
     @document = Document.new(document_params.merge(creator_id: current_user.id))
 
     if @document.save_and_verify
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
💥  LGTM

naseberry (approved):

PR #4196, opened 2021-08-12
[CFP-240] Move ErrorPresenter and dependencies to ErrorMessage namespace

jsugarman:
#### What
Move ErrorPresenter and dependencies to ErrorMessage namespace

#### Ticket

[extension of CFP-240](https://dsdmoj.atlassian.net/browse/CFP-240)

#### Why
Group Custom error handling classes into their own
module/namespace for organisational clarity


jrmhaig (approved):
(nit) I'd be inclined to keep the presenter itself in `app/presenters` while putting all the extra bits in a subdirectory but I'm not going to object to it being in `app/presenters/error_message`.

On the whole, :+1: * :100: for namespacing and making things a bit more understandable.

PR #4206, opened 2021-08-23
Regenerate rubocop todo

jsugarman:
Following various refactoring, violations have
changed, and in many cases been fixed.

Generated using:

```
rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 1000
```


jrmhaig (approved):
:+1:

PR #4209, opened 2021-08-24
[CFP-150] Send report to Slack after stale claim archive job

jrmhaig:
#### What

Send a message to Slack after the completion of the archive stale claim job, which runs overnight.

#### Ticket

[Alerting for errors in archive stale claims job](https://dsdmoj.atlassian.net/browse/CFP-150)

#### Why

We currently have no visibility on how the stale claim archive job runs.

#### How

As a first version, the Slack notification contains:

* Title; "Stale claim archiver completer" with optional " with failures"
* Message and fallback; "*n* transactions processed (*n* failed)"

#### TODO (possibly as this work or later)

 - [ ] Also show numbers for how many were archived and how many were deleted (also counts for state before archive/delete?)
 - [ ] Break down failures into categories and display error messages
 - [ ] Add attachments to the alert with details of each failure (too spammy?)


app/models/timed_transitions/batch_transitioner.rb
```diff
@@ -76,5 +93,9 @@ def log(level, message, options)
         message
       end
     end
+
+    def slack_notifier
+      @slack_notifier ||= SlackNotifier.new('laa-cccd-alerts', formatter: SlackNotifier::Formatter::Transitioner.new)
+    end
```
jsugarman: **not a blocker:** for decoupling we could invert dependencies by injecting this as a `notifier` option in the `initialize` options

spec/models/timed_transitions/batch_transitioner_spec.rb
```diff
@@ -1,87 +1,140 @@
...
+  let(:transitioner22) { instance_double('Transitioner 22', success?: true) }
+  let(:transitioner878) { instance_double('Transitioner 878', success?: true) }
 
-    context 'dummy' do
-      let(:batch_transitioner) { BatchTransitioner.new(dummy: true) }
+  before do
+    allow(TimedTransitions::Transitioner).to receive(:candidate_claims_ids).and_return(claim_ids)
+    allow(Claim::BaseClaim).to receive(:find).with(22).and_return(claim22)
+    allow(Claim::BaseClaim).to receive(:find).with(878).and_return(claim878)
+    allow(TimedTransitions::Transitioner).to receive(:new).with(claim22, dummy).and_return transitioner22
```
jsugarman, line 14: **not you and not a blocker** but having these id specific claims is not pleasant

spec/models/timed_transitions/batch_transitioner_spec.rb
```diff
@@ -1,87 +1,140 @@
...
-      end
+  it 'runs the transitioner for the first claim' do
+    batch_transitioner_run
+    expect(transitioner22).to have_received(:run)
+  end
+
+  it 'creates a transitioner for the second claim' do
+    batch_transitioner_run
+    expect(TimedTransitions::Transitioner).to have_received(:new).with(claim878, dummy)
+  end
```
jsugarman, line 34: 🎖️ much better break up of tests

spec/models/timed_transitions/batch_transitioner_spec.rb
```diff
@@ -1,87 +1,140 @@
...
+            true
+          )
+          allow(SlackNotifier).to receive(:new).and_return(slack_notifier)
+          allow(slack_notifier).to receive(:build_payload)
+          allow(slack_notifier).to receive(:send_message)
+        end
+
+        it 'creates Slack attachment with tally of succeeded and failed transitions' do
+          batch_transitioner_run
+          expect(slack_notifier).to have_received(:build_payload).with(processed: 3, failed: 2)
```
jsugarman: ✅

jsugarman (approved):
👍  LGTM. definite improvement for the tests

Couple of non-blocking potential refactorings while you are working in the area and we should dry run this with failures on an environment. I think to duplicate an error you could update the evidence checklist items for a claim to include `2`, on the given environment. This is something i know has caused failures in the past. See `DocType` model where type 2 was removed a long time ago.

PR #4222, opened 2021-08-25
Revert "Bump webpacker from 5.4.0 to 5.4.2"

jsugarman:
Reverts ministryofjustice/Claim-for-Crown-Court-Defence#4212


jsugarman:
The change this reverts is currently in master and on production and is not causing an issue that we are aware of. Closing!

PR #4227, opened 2021-08-26
Test breakage

jsugarman:
DO NOT MERGE

### DIFF
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -25,4 +25,10 @@ environment.plugins.append('ConcatPlugin', new ConcatPlugin({
   ]
 }))
 
+environment.config.merge({
+  output: {
+    filename: 'js/[name]-[hash].js'
+  }
+})
+
 module.exports = environment
--- a/package.json
+++ b/package.json
@@ -35,11 +35,13 @@
     "webpack-concat-plugin": "^3.0.0"
   },
   "devDependencies": {
+    "@webpack-cli/serve": "^1.5.2",
     "standard": "^16.0.3",
     "stylelint": "^13.13.1",
     "stylelint-config-gds": "^0.1.0",
     "stylelint-order": "^4.1.0",
-    "webpack-dev-server": "^4.0.0"
+    "webpack-dev-server": "^4.0.0",
+    "webpack-cli": "^4.8.0"
   },
   "resolutions": {
     "css-what": "^5.0.1",
@@ -52,9 +54,18 @@
     "webpack": "^4.44.2"
   },
   "standard": {
-    "env": [ "jasmine", "jquery" ],
-    "globals": [ "GOVUK", "moj" ],
-    "ignore": [ "app/webpack/javascripts/vendor/", "spec/javascripts/helpers/" ]
+    "env": [
+      "jasmine",
+      "jquery"
+    ],
+    "globals": [
+      "GOVUK",
+      "moj"
+    ],
+    "ignore": [
+      "app/webpack/javascripts/vendor/",
+      "spec/javascripts/helpers/"
+    ]
   },
   "@comments": {
     "@dependencies": {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "stylelint": "^13.13.1",
     "stylelint-config-gds": "^0.1.0",
     "stylelint-order": "^4.1.0",
-    "webpack-dev-server": "^3.11.2"
+    "webpack-dev-server": "^4.0.0"
   },
   "resolutions": {
     "css-what": "^5.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1080,6 +1080,11 @@
   resolved "https://registry.yarnpkg.com/@csstools/convert-colors/-/convert-colors-1.4.0.tgz#ad495dc41b12e75d588c6db8b9834f08fa131eb7"
   integrity sha512-5a6wqoJV/xEdbRNKVo6I4hO3VjyDq//8q2f9I6PBAvMesJHFauXDorcNCsr9RzvsZnaWi5NYCcfyqP1QeFHFbw==
 
+"@discoveryjs/json-ext@^0.5.0":
+  version "0.5.3"
+  resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.3.tgz#90420f9f9c6d3987f176a19a7d8e764271a2f55d"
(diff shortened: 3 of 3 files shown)


.circleci/config.yml
```diff
@@ -251,6 +251,31 @@ commands:
           success_message: ":tada: deploy of <$CIRCLE_BUILD_URL|$CIRCLE_BRANCH> to << parameters.environment >> successful!"
           failure_message: ":red_circle: deploy of <$CIRCLE_BUILD_URL|$CIRCLE_BRANCH> to << parameters.environment >> failed!"
 
+  ui-smoke-test:
```
Paul2021V: I'm unclear on how this change will catch the issue you had with Webpack in an earlier environment based on the information we went through on the postmortem.

I may be misremembering but ISTR that when Peter walked us through the problem, he said it only manifested in prod and all the other environments worked OK which would imply the _devDependencies_ block was being used all the way through to staging or some other factor was at play which had the same effect. 

If my recollection is correct, the 200 test is great for checking the app is relatively healthy when the configuration works but it won't catch the issue any sooner if it happens again because it didn't manifest in any of the non-prod environments.

I'm happy to discuss this further if you'd like. If my recollection is off, feel free to ignore my ramblings!

jsugarman: sorry just saw this. @Paul2021V  it did manifest in all the hosted environments. All of the environments are "production" from the point of view of the framework, Rails, and dependencies. It is just the host name that differs. Configuration wise they are all pretty similar too baring use of secrets and accounts on external services

jsugarman: You can see examples of the ui smoke test failing here
https://app.circleci.com/pipelines/github/ministryofjustice/Claim-for-Crown-Court-Defence?branch=js%2Fbroken

Paul2021V: If it manifested in all environments, why was it only caught when it went to prod? Am I missing something? 

Were there no tests/PLV type actions carried out in the non-prod environments before it deployed to prod? 

At the postmortem walkthrough it was stated that the service had passed all tests in the lower environments.

jsugarman: I am not sure what postmortem you are referring to. Was I in attendance? 

> If it manifested in all environments, why was it only caught when it went to prod?

It went to prod because: 
1. no automated tests caught it
2. ping did not catch it
2. it was not **manually** sanity tested on any environment

> Am I missing something?

??? 
We did not, prior to this, automate tests of actual hosted environments/namespaces/ingresses, we test the codebase and the image. build and deploy can fail too in certain circumstances. The front end and production impacting (not test or development) impact of the error, relating to assets but which i am still not clear on, meant all passed. 

We also have a liveness/readiness ping test of all hosted environments (lower or otherwise) which did not fail but would also have avoided the breakage had it done so. It do not catch it because ping does not require assets.

> PLV??? production live verification??

See point 2. above. Ideally there would have been even for supposedly trivial dependency bumps, as this was. This test automates a basic sanity test of an actual hosted environment, for the first time, which is not ideal and will be improved upon in future iterations.

Are you guys automating tests of actual hosted environments in CCR/CCLF? Do you have any automated [smoke] tests of production? Do you do ping tests? What is your interest in this?

Paul2021V: I apologise if I've overstepped my bounds in asking these questions, they were intended to help clarify my confusion regarding how it got to prod when I was of the understanding that the problem hadn't manifested in non-prod and if that was the case, how a 200 check would catch the issue. 

With that in mind, I'll answer your last question first I think. 

My interest is that I'm a new dev to MoJ and I'm trying to get a little bit of wider knowledge about how things hang together and work here, I might be asking "dumb" questions but it's my way of gaining the knowledge that I need to do my job in a new contract. It's purely a professional interest following up on the PM/Walkthrough that Peter gave the teams and it's to help me understand how the teams work here, what our state of play is regarding our testing and deployment processes and then in the future hopefully being able to contribute more in helping to get all teams in a better place with those practices.

In previous contracts I've worked, a service failing to start would not pass testing practices because even the developers would do basic manual checks of a service and application even if the automated test coverage was very high and everything passed. After the developer did the checks to ensure a deployment looks sane in each environment up to the agreed point with a tester; we might push it through the lower environments based on the automation and dev checks alone, providing we could show supporting evidence. 

At the agreed environment, the service would be passed over to the tester who would have pre-discussed the changes with the team, even going as far as having a code walk through to explain the change in some of the more complex requirements. A service wouldn't go out to prod without a tester looking at the change in at least one environment. After that, it would go through a change control approval process where several people outside of the development and test teams would review the ticket, evidence provided etc. and approve the deployment. The least favourite part of the process for most devs. This whole thing still wasn't foolproof and issues still got through to prod that shouldn't have got there but it was a very rare case when it happened. We live in an imperfect world after all.

PLV checks are indeed production/post live verification checks. Normally these would take 5-10 min and be a final visual confirmation that all is well after a deployment. The checks would include things like checking AWS healthy host counts, ECS for task churn etc. plus a manual login to the service and a run through of the part of the service where the change was made if possible. Screenshots would be taken and attached to the deployment ticket as part of the change control protocols showing proof of life of the service and that the checks had been completed. In lower environments we wouldn't necessarily do all of those checks, sometimes a suite of Postman tests executed as part of the pipeline would be enough.

As for the postmortem in this case, I don't recall you being there. It wasn't a full PM at the time and wasn't meant to be at that stage because it was the day after the incident. At that stage, it was more a walk through for other devs to understand what the problem was and why it occurred, going down to the code level change. I think it's the walkthrough where my misunderstanding came in regarding it not manifesting in non-prod environments.

jessicabrentnall: Hi both.

I'm sure the questioning is all meant in the spirit of learning from each other across teams and not as a challenge to the work the teams are doing. We have very different estates, but can learn from each others code and processes. I wasn't sure about the process either to be honest, but have been a bit stretched for time so I've learnt from this exchange.

I'll book in a call with you both and Ken (new DM) when everyone is back to see if there are ways we can work together more effectively going forwards.

jsugarman: Thanks Jess, yes I was going to suggest an appointment/call would have been more appropriate. 

We do have significantly different estates and approaches between Stack's, Platforms/PaaS' and CI, at least. Always happy to share. I may be available tomorrow now (Jury service put me on call this pm again 😮‍💨 )

jsugarman:
closing PR but keeping branch as it would be good to have a smoke test that catches this error without deploying to an actual environment.

PR #4229, opened 2021-08-27
gitignore public/assets

jsugarman:
#### What
gitignore public/assets

#### Ticket

n/a

#### Why

`bundle exec rails assets:precompile` can generate various
files directly under public/assets that we do no want or need
in the repo. assets are precompiled as part of docker build
process in nay event.


jrmhaig (approved):
:+1:

PR #4230, opened 2021-08-27
Add slack alerts channel webhook secret

jsugarman:
#### What
Add slack alerts channel webhook secret

#### Ticket
[relates to CFP-210](https://dsdmoj.atlassian.net/browse/CFP-210)

#### Why
This is purely to record the webhook somewhere and to enable
cloud platforms to retrieve and use it to direct prometheus
alertmanager routes to.

In addition this webhook could be used in app's slack notifier
to post messages to the alert channel.


jrmhaig (approved):
:+1:

PR #4232, opened 2021-08-27
[CFP-210] Change slack channel for prometheus alerts

jsugarman:
#### What
Change slack channel for prometheus alerts

#### Ticket

[relates to CFP-210](https://dsdmoj.atlassian.net/browse/CFP-210)

#### Why

The `severity` entry is used by cloud platforms
infrastructure in some way that is obscured from us
to route alerts to the "receiver" (the webhook I guess)

#### How
We just need to change the "severity" level (can also call it
a "slack channel receiver").

See this PR that creates a "severity" called laa-cccd-alerts
that points to a webhook that sends output to the
 \#laa-cccd-alerts channel

https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/1347

### TODO

- [x] apply to dev
- [x] apply to dev-lgfs
- [x] apply to staging
- [x] apply to api-sandbox
- [x] apply to production
- [ ] merge PR 

**Note**: merging this PR won't actually do anything since we do not apply `prometheusrules` via the pipeline. we can do it manually on the console

```
# describe rules in namespace
kubectl describe prometheusrules prometheus-custom-rules-cccd -n <namespace>

# apply rule
kubectl apply -f kubernetes_deploy/<env>/prometheus-custom-rules.yaml -n <namespace>
e.g. k apply -f kubernetes_deploy/staging/prometheus-custom-rules.yaml -n cccd-staging
```

### DIFF
--- a/kubernetes_deploy/api-sandbox/prometheus-custom-rules.yaml
+++ b/kubernetes_deploy/api-sandbox/prometheus-custom-rules.yaml
@@ -14,7 +14,7 @@ spec:
       expr: 100 * kube_resourcequota{job="kube-state-metrics",type="used",namespace="cccd-api-sandbox"} / ignoring(instance, job, type) (kube_resourcequota{job="kube-state-metrics",type="hard",namespace="cccd-api-sandbox"} > 0) > 90
       for: 1m
       labels:
-        severity: laa-court-get-paid
+        severity: laa-cccd-alerts
       annotations:
         message: cccd-api-sandbox is using {{ printf "%0.0f" $value}}% of its {{ $labels.resource }} quota.
         runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubequotaexceeded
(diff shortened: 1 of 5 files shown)


kubernetes_deploy/api-sandbox/prometheus-custom-rules.yaml
```diff
@@ -14,30 +14,30 @@ spec:
       expr: 100 * kube_resourcequota{job="kube-state-metrics",type="used",namespace="cccd-api-sandbox"} / ignoring(instance, job, type) (kube_resourcequota{job="kube-state-metrics",type="hard",namespace="cccd-api-sandbox"} > 0) > 90
       for: 1m
       labels:
-        severity: laa-court-get-paid
+        severity: laa-cccd-alerts
```
jrmhaig, line 17: The previous version was `laa-court-get-paid` but this channel does not exist. Does `court` have some special purpose?

jrmhaig (approved):
I have left a comment about something I though odd with the old and new setting of `severity`. I'm approving on the basis that this is probably correct.

kubernetes_deploy/api-sandbox/prometheus-custom-rules.yaml
jsugarman: the name is not related to the channel. it relates to a slack webhook and its "receiver" that CP create and give a name. I have just used the convention of naming the webhook after the channel, which CP have then used for the custom severity level. 

Its all a bit disconnected but we do no have insight or access to their creation of the "receiver".

PR #4241, opened 2021-09-02
[CFP-255] Add smoke test of UI to CI/CD pipeline

jsugarman:
#### What
Add circle executor, command and job for ui smoke test

#### Ticket

[CFP-255](https://dsdmoj.atlassian.net/browse/CFP-255)

#### Why
Basic smoke test of the UI to mitigate against the risk of
deploying changes that break due to problems that do no 
show up in development and tests.

Follows a P1 incident caused by this [FE dependency update](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4227) where test build and deploy were all green but app
enters a TOO_MANY_REDIRECTS state due to all pages have asset management
issues.

Future iterations would ideally test a running instance prior to
any deployment or break during deployment.


jsugarman:
@jrmhaig I spoke to PM about this and he considered it worthwhile having until a better solution can be found.

jrmhaig (approved):
Looks good :+1:

PR #4242, opened 2021-09-02
[CFP-150] Only report to Slack for failure

jrmhaig:
#### What

Only report results of overnight claims archiving job when there are failures.

#### Ticket

[Alerting for errors in archive stale claims job](https://dsdmoj.atlassian.net/browse/CFP-150)

#### Why

Avoid spamming the Slack alerts channel whilst keeping watch on all the environments.

#### How

Only send the notification when failures have been found.


jsugarman (approved):
👍

PR #4243, opened 2021-09-03
Correct namespace name in promethues alert

jsugarman:
#### What
Correct namespace name in promethues alert

#### Why
This alert shows up as
```
Alert: cccd-dev An HTTP 5xx error has occurred
```
but should be
```
Alert: cccd-dev-lgfs An HTTP 5xx error has occurred
```


jrmhaig (approved):
🚀

PR #4244, opened 2021-09-07
[CFP-162] Refactoring custom error handling

jsugarman:
#### What
Refactor specs and add conversion method

#### Ticket

[CFP-162](https://dsdmoj.atlassian.net/browse/CFP-162)

#### Why
Specs examples not using describe blocks and
had too many tests per example

Conversion method belongs on the ErrorMessage::Detail class.

### DIFF
--- a/app/presenters/error_message/detail_collection.rb
+++ b/app/presenters/error_message/detail_collection.rb
@@ -32,7 +32,7 @@ module ErrorMessage
     # when a presenter instance is injected in to govuk_error_summary.
     #
     def formatted_error_messages
-      summary_errors.map { |detail| [detail.attribute, detail.long_message] }
+      summary_errors.map(&:to_summary_error)
     end
 
     def summary_errors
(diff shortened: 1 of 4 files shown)


app/presenters/error_message/detail_collection.rb
```diff
@@ -32,7 +32,7 @@ def short_messages_for(fieldname)
     # when a presenter instance is injected in to govuk_error_summary.
     #
```
jrmhaig, line 34: nit, and only because I went to check what the official documentation said (which I know you wrote anyway) (and yes, I should really have looked for the previous PR)

```suggestion
    # See https://govuk-form-builder.netlify.app/introduction/error-handling/#custom-summary-error-presenter-injection
    #
```

jrmhaig (approved):
:tada:

app/presenters/error_message/detail_collection.rb
jsugarman: Thanks, yes I had meant to add that once it was released

PR #4252, opened 2021-09-08
Cleaner cleaners

jrmhaig:
#### What

Group all the claim cleaners together.

#### Ticket

N/A

#### Why

The claim cleaners are used to remove invalid elements from various types of claim when they are updated via the API. At the moment they are written specifically for each claim type and in the application file structure they exist along side the claim models. There is some common functionality and they could be standardised but it is not made clear by the current organisation.

The cleaners each do one or more of the following;

1. Remove the values set to fields that are invalid for the claim type
2. Delete fees that are not valid for the claim type
3. Delete disbursements for interim claims (when the interim fee type is `IWARR`)

If the cleaners can be standardised then it will be simpler to update according to business logic. Furthermore, this could help us to modify the API to return sensible errors to clients for invalid input rather than silently deleting.

#### How

* Move all the cleaners into a new directory `app/models/claim/concerns/cleaner` and rename the classes in the `Claim::Cleaner::*` namespace
* Pull out some of the common functionality into a new class `Claim::Cleaner` for all the cleaners to be based on


jsugarman:
Looks good, but these are not `concerns`. They are more like services 

https://blog.appsignal.com/2020/09/16/rails-concers-to-concern-or-not-to-concern.html

jsugarman:
> For simplicity, tasks 2 and 3 listed in the 'Why' section are grouped as deleting expenses (is this the correct and logical terminology?).

Disbursements, expenses and fees are logically different

jrmhaig:
> > For simplicity, tasks 2 and 3 listed in the 'Why' section are grouped as deleting expenses (is this the correct and logical terminology?).
> 
> Disbursements, expenses and fees are logically different

I have split out 'expenses' into 'fees' and 'disbursements'.

jrmhaig:
> Looks good, but these are not `concerns`. They are more like services

Moved into `app/services` now.

app/services/claim_cleaner_service/clear_cracked_details.rb
```diff
@@ -1,25 +1,11 @@
...
-    def call
-      clear_inapplicable_fields
-    end
-
+class ClaimCleanerService
+  module ClearCrackedDetails
     private
 
-    # TODO: Hardship claim - can be shared with advocate final claims
     def clear_inapplicable_fields
```
jsugarman: this method belongs on the claim cleaner, rather than the mixin

jsugarman (changes requested):
having a namespace and inheritable class being one and the same is rather odd.
```
class MyClass
  class MySubClass < MyClass
  end
end
```

I think from a naming and code organisation point of view it would be better to have something like below.

```
# app/services/cleaners/base_claim_cleaner.rb
module Cleaners
  class BaseClaimCleaner 
  end
end
```

Then for each claim type
```
# app/services/cleaners/advocate_claim_cleaner.rb
module Cleaners
  class AdvocateClaimCleaner  < BaseClaimCleaner
  end
end
```

💯  on the mixin but could be better named
```
# app/services/cleaners/[concerns]/cracked_detail_cleanable.rb
module Cleaners
  module CrackedDetailCleanable
    ...
  end
end
```


Note also that the cleaners are used not just for "cleaning" API submissions but handling the potential flows through the multi step form. e.g. If a user creates a final claim for a fixed fee then goes back and changes it to a grad fee then the fixed fee is removed silently by the cleaner.

Testing wise, since this is a significant refactor, we should probably think about moving the integration tests (e.g. advocate_claim_spec#cleaner) to more unit testing like specs too.

jsugarman:
I still think we should call them `app/services/cleaners/...claim_cleaner.rb` - e.g. `advocate_claim_cleaner.rb` - from a naming point of view. To not confuse with, eg,`advocate_claim`. Similar patterns exist for `mailers`, `controllers`, `validators`, `presenters` and `serializers` (to name the most obvious ones in the repo) - the outer namespace `cleaners` is more just to group them. Could even have `app/cleaners` but that would require loading config changes too i think.

jrmhaig:
> I still think we should call them `app/services/cleaners/...claim_cleaner.rb` - e.g. `advocate_claim_cleaner.rb` ...

They are all named like that but I did push changes at one point when I hadn't renamed everything properly, so you may have been looking at that version.

spec/support/shared_examples/models/shared_examples_for_base_claim.rb
```diff
@@ -44,4 +44,18 @@
     it { is_expected.to accept_nested_attributes_for(:assessment) }
     it { is_expected.to accept_nested_attributes_for(:redeterminations) }
   end
+
+  describe 'cleaner' do
```
jsugarman:
```suggestion
  describe '#cleaner' do
```

app/models/claim/base_claim.rb
```diff
@@ -609,7 +609,9 @@ def reset_transfer_court_details
       self.transfer_case_number = nil
     end
 
-    def cleaner; end
+    def cleaner
+      Cleaners::BaseClaimCleaner.new(self)
```
jsugarman: 🤔  not sure we need this on the `base_claim`. how many claim types actually use cleaners?

spec/services/cleaners/advocate_claim_cleaner_spec.rb
```diff
@@ -1,6 +1,42 @@
...
+  it { expect { call_cleaner }.to change(claim, :trial_cracked_at_third).to nil }
+end
+
+RSpec.shared_examples 'does not clear cracked details' do
+  it { expect { call_cleaner }.not_to change(claim, :trial_fixed_notice_at) }
+  it { expect { call_cleaner }.not_to change(claim, :trial_fixed_at) }
+  it { expect { call_cleaner }.not_to change(claim, :trial_cracked_at) }
+  it { expect { call_cleaner }.not_to change(claim, :trial_cracked_at_third) }
+end
+
```
jsugarman: 👍  to the shared examples

spec/services/cleaners/advocate_claim_cleaner_spec.rb
```diff
@@ -12,102 +48,72 @@
   describe '#call' do
     subject(:call_cleaner) { cleaner.call }
 
-    context 'with an "Appeal against sentence" case type including a fixed fee' do
-      let(:claim) do
-        create(:draft_claim, case_type: create(:case_type, :appeal_against_sentence), fixed_fees: fixed_fees)
-      end
-      let(:fixed_fees) { [build(:fixed_fee, :fxase_fee, :with_date_attended, rate: 9.99)] }
+    let(:claim) do
+      build(
```
jsugarman: does this work if you use `create` here too. Just thinking about hooks and their impact on all this

spec/services/cleaners/advocate_claim_cleaner_spec.rb
```diff
@@ -12,102 +48,72 @@
...
+        :draft_claim,
```
jsugarman: for clarity I would use
:advocate_final_claim, :draft

The older `<state>_claim` factories should be considered deprecated as they pre-date the existence of any  other claim type (i.e. when there was only one claim type) and are therefore not very descriptive

spec/services/cleaners/advocate_claim_cleaner_spec.rb
```diff
@@ -1,6 +1,42 @@
...
+end
+
+RSpec.shared_examples 'clear cracked details' do
+  it { expect { call_cleaner }.to change(claim, :trial_fixed_notice_at).to nil }
+  it { expect { call_cleaner }.to change(claim, :trial_fixed_at).to nil }
+  it { expect { call_cleaner }.to change(claim, :trial_cracked_at).to nil }
+  it { expect { call_cleaner }.to change(claim, :trial_cracked_at_third).to nil }
+end
+
+RSpec.shared_examples 'does not clear cracked details' do
```
jsugarman: nit - as with other negative tests I am wary of false positives and like think about if there is positive way this could be achieved

spec/services/cleaners/advocate_claim_cleaner_spec.rb
jrmhaig: The problem is that create would run the cleaner before the test, which is why the old versions of the tests set up the claim with one case type and then updated it.

Maybe something like this would work?

```ruby
let(:claim) do
  create(:advocate_final_claim, :draft, case_type: case_type).tap do |claim|
    claim.fixed_fees = [build(:fixed_fee, :fxase_fee, :with_date_attended, rate: 9.99)]
    # etc
  end
end
```

app/models/claim/base_claim.rb
jrmhaig: The main reason for this is so that the `it_behaves_like 'a base claim'` works for all the claim types tests without having to conditionally skip the relevant part.

As a null object, perhaps it would be better to have a new class `Cleaners::NullClaimCleaner` in addition to (and based on) the `BaseClaimCleaner`, to make it clearer what the purpose of this is.

Incidentally, this means that the save navigation operator is not required on line 129, but I haven't changed that (yet).

spec/services/cleaners/advocate_claim_cleaner_spec.rb
jrmhaig: I'll have a look at this. Would a `expect { ... }.not_to change(...).from(old_value)` satisfy?

app/services/cleaners/cracked_detail_cleanable.rb
```diff
@@ -0,0 +1,12 @@
+module Cleaners
+  module CrackedDetailCleanable
+    private
+
+    def clear_cracked_details
+      claim.trial_fixed_notice_at = nil if trial_fixed_notice_at
+      claim.trial_fixed_at = nil if trial_fixed_at
+      claim.trial_cracked_at = nil if trial_cracked_at
+      claim.trial_cracked_at_third = nil if trial_cracked_at_third
```
jsugarman: **nit** if missing is delegated to claim in the `BaseClaimCleaner` do we need `claim.whatever`?

jsugarman: oh you might due to assignment actually but perhaps `self.whatever = `

app/services/cleaners/litigator_claim_cleaner.rb
```diff
@@ -0,0 +1,17 @@
+module Cleaners
+  class LitigatorClaimCleaner < BaseClaimCleaner
+    private
+
+    def destroy_invalid_fees
+      return if case_type.blank?
+
+      if case_type.is_fixed_fee?
+        graduated_fee&.destroy
+        claim.graduated_fee = nil
```
jsugarman: **nit** as with related comment
```suggestion
        self.graduated_fee = nil
```

app/services/cleaners/litigator_claim_cleaner.rb
```diff
@@ -0,0 +1,17 @@
...
+      else
+        fixed_fee&.destroy
+        claim.fixed_fee = nil
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4252/files#r720222724)

app/services/cleaners/litigator_hardship_claim_cleaner.rb
```diff
@@ -0,0 +1,22 @@
+module Cleaners
+  class LitigatorHardshipClaimCleaner < BaseClaimCleaner
+    private
+
+    def clear_inapplicable_fields
+      clear_ppe if case_type.present? && ppe_not_required?
+    end
+
+    def ppe_not_required?
+      claim.case_stage.unique_code.eql?('NOPTPHNOPPE') && hardship_fee.present?
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4252/files#r720222724)

app/services/cleaners/litigator_hardship_claim_cleaner.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def hardship_fee
+      @hardship_fee ||= claim.fees.find_by(type: 'Fee::HardshipFee')
```
jsugarman: would this work?
```suggestion
      @hardship_fee ||= fees.find_by(type: 'Fee::HardshipFee')
```

spec/services/cleaners/advocate_claim_cleaner_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  it { expect { call_cleaner }.not_to(change { claim.misc_fees.size }) }
+end
+
+RSpec.shared_examples 'clear basic fees' do
+  it { expect { call_cleaner }.not_to(change { claim.basic_fees.size }) }
+  it { expect { call_cleaner }.to change { claim.basic_fees.flat_map(&:dates_attended).size }.to 0 }
+end
+
+RSpec.shared_examples 'does not clear basic fees' do
+  it { expect { call_cleaner }.not_to(change { claim.basic_fees.size }) }
```
jsugarman: would `size` change since we are only clearing the basic fees, not destroying them?

Withstanding that:
```suggestion
  it { expect { call_cleaner }.not_to change { claim.basic_fees.size } }
```

app/services/cleaners/advocate_supplementary_claim_cleaner.rb
```diff
@@ -1,20 +1,8 @@
...
-      destroy_ineligible_misc_fees
-    end
-
+module Cleaners
+  class AdvocateSupplementaryClaimCleaner < BaseClaimCleaner
     private
 
-    def destroy_ineligible_misc_fees
+    def destroy_invalid_fees
       misc_fees.delete(ineligible_misc_fees)
```
jsugarman, line 10: 🤔  not you and not for now but the `.delete` has  the potential to introduce a bug as hooks won't be called. I don't think there are any impacts to worry about for here as the only hook i can find sets the amount on the fee (which is being deleted)

spec/services/cleaners/interim_claim_cleaner_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    context 'without a warrant fee' do
+      let(:fee_type) { :disbursement }
+
+      it { expect { call_cleaner }.not_to change { claim.disbursements.size }.from(1) }
+    end
+
+    context 'with a warrant fee' do
+      let(:fee_type) { :warrant }
+
+      it { expect { call_cleaner }.to change { claim.disbursements.size }.from(1).to(0) }
```
jsugarman, line 23: yes, this looks correct logic to me. interim claims for warrants cannot claim disbursements, but interim claims for  Trial start, Effective PCMH or "Disbursement only" can claim them. 

If you choose one of the last three and add a disbursement then navigate back and change to a warrant then the disbursements need to be cleaned/deleted.

app/models/claim/base_claim.rb
jsugarman: **nit** I do like null objects and it would be clearer

spec/services/cleaners/advocate_claim_cleaner_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  describe '#call' do
+    subject(:call_cleaner) { cleaner.call }
+
+    let(:claim) do
+      create(
+        :advocate_final_claim, :draft,
+        case_type: build(:case_type, :fixed_fee),
+        fixed_fees: [create(:fixed_fee, :fxase_fee, :with_date_attended, rate: 9.99)]
+      )
+    end
```
jsugarman, line 45: line break between multiline lets - i would have though rubocop would have picked that up?

spec/services/cleaners/advocate_hardship_claim_cleaner_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+require 'services/cleaners/cleaner_shared_examples'
+
+RSpec.describe Cleaners::AdvocateHardshipClaimCleaner do
+  subject(:cleaner) { described_class.new(claim) }
+
+  describe '#call' do
+    subject(:call_cleaner) { cleaner.call }
+
+    let(:claim) { create(:advocate_hardship_claim) }
```
jsugarman, line 10: **nit** line break

spec/support/shared_examples/models/shared_examples_for_base_claim.rb
```diff
@@ -44,4 +44,18 @@
     it { is_expected.to accept_nested_attributes_for(:assessment) }
     it { is_expected.to accept_nested_attributes_for(:redeterminations) }
   end
+
+  describe '#cleaner' do
+    subject(:claim) { create(:draft_claim) }
```
jsugarman: this creates an advocate final claim so is not a shared example for any claim type. The subject should be set by the calling spec.

following this refactors logic a base claim is just expected to respond to #cleaner, and claim types expected to call a specific cleaner.

so,  remove subject and, optionally, add
```
  it { is_expected.to respond_to(:cleaner) }
```

spec/services/cleaners/litigator_hardship_claim_cleaner_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+RSpec.describe Cleaners::LitigatorHardshipClaimCleaner do
+  subject(:cleaner) { described_class.new(claim) }
+
+  describe '#call' do
+    subject(:call_cleaner) { cleaner.call }
+
+    let(:claim) do
+      create(
+        :litigator_hardship_claim,
+        case_stage: create(:case_stage, :pre_ptph_with_evidence)
```
jsugarman: **nit** can probably use build here as with interim fee examples?!

app/models/claim/base_claim.rb
jsugarman: are you going to apply the nullobject pattern here?

jsugarman (changes requested):
A few nits but also couple of issues that should be resolved. we could pair on this and/or discuss the review.

spec/services/cleaners/advocate_claim_cleaner_spec.rb
jrmhaig: 1. Yes, this test is probably superfluous with the following test, which checks the `dates_attending`. I'll remove it.
2. The brackets are required on this (and others) as otherwise the block after `change` is ambiguous. https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/AmbiguousBlockAssociation

spec/services/cleaners/advocate_claim_cleaner_spec.rb
jrmhaig: Added.

spec/services/cleaners/advocate_hardship_claim_cleaner_spec.rb
jrmhaig: Added.

spec/services/cleaners/advocate_claim_cleaner_spec.rb
jsugarman: > The brackets are required on this (and others) as otherwise the block after change is ambiguous.

from that link @jrmhaig you supplied it specifically excludes the `change` method
> IgnoredMethods: [change]

```
# good
expect { do_something }.to change { object.attribute }
```

jsugarman: > Yes, this test is probably superfluous with the following test, which checks the dates_attending. I'll remove it.

It should be tested but the test would need to check it does not change the attribute values instead of the size of the association collection.

jsugarman (approved):
👍  if you squash up a bit before merge that would be good but not a blocker

PR #4259, opened 2021-09-13
[Outdated - do not merge] [CFP-244] Automatically remove graduated fees that are incorrectly added to transfer claims

jrmhaig:
#### What

Automatically delete any graduate fees from transfer claims.

#### Ticket

[Incorrect fee on transfer claim (#252755)](https://dsdmoj.atlassian.net/browse/CFP-244)

#### Why

Transfer claims cannot have graduated fees but some vendor clients are adding them automatically. Ideally the API would return an error to the client if an invalid fee is added to a claim but the current behaviour is to silently clean up the input and changing this could cause problems for vendors.

#### How

Create a new cleaner class for transfer claims that will delete any attached graduated fees that are found.

#### TODO

- [ ] Confirm business rule; any graduated fee on a transfer claim should be removed
- [ ] Deploy to api-sandbox and get provider to test


jsugarman (approved):
👍

jrmhaig:
Replaced by #4323

PR #4265, opened 2021-09-13
Update feedback form MVC.

imtiazAhmd:
#### What
UI Changes for User Satisfaction Form based on the Design Reference.

#### Ticket
[User satisfaction form changes](https://dsdmoj.atlassian.net/browse/CFP-194)

#### Why
As part of our review into data quality for user satisfaction, we have become aware of a survey monkey API that could be used to submit data from CCCD-side forms to Survey Monkey. This would allow us to control the flow in the app and adhere to GDS design patterns ensuring a high level of accessibility and design consistency. 

|Before| After   |
|---|---|
| screencapture-staging-claim-crown-court-defence-service-justice-gov-uk-feedback-new-2021-09-14-15_43_02  |  screencapture-localhost-3000-feedback-new-2021-09-14-15_42_39 |

--------

#### TODO (wip)

 - [x] API, API Specs
 - [x] Remove Rating Collection Method
 - [x] Remove Zendesk from Feedback Form
 - [X] Front End changes


app/controllers/feedback_controller.rb
```diff
@@ -12,21 +13,15 @@ def new
   def create
     @feedback = Feedback.new(merged_feedback_params)
     if @feedback.save
-      submit_feedback_google_event if @feedback.is?('feedback')
       redirect_to after_create_url, notice: 'Feedback submitted'
     else
+      flash[:error] = @feedback.error
```
naseberry: Feedback and Bug report both share the same controller.

The bug report view uses `govuk_error_summary` to display errors, therefore this flash would also show on the bug report view, you may need to add a conditional check for feedback type.

Screenshot 2021-09-30 at 14 57 46

app/controllers/feedback_controller.rb
```diff
@@ -64,14 +59,17 @@ def after_create_url
...
       :case_number,
       :referrer,
-      :email
+      :email,
+      reason: []
     )
   end
+
+  def setup_page
+    @feedback_form = FeedbackForm.new if type == 'feedback'
```
naseberry, line 68: The `eql?` operator is the same as `==`, but which should be used for our standards?

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,72 @@
...
+    )
+  end
+
+  def ratings
+    Section.new(
+      60_742_964, :radio,
+      [
+        Answer.new('5', 'Very satisfied', 505_488_046),
+        Answer.new('4', 'Satisfied', 505_488_047),
+        Answer.new('3', 'Neither satisfied nor dissatisified', 505_488_048),
```
naseberry: Grammar: `dissatisified` should be `dissatisfied`

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,72 @@
...
+        Answer.new('1', 'Very dissatisfied', 505_488_050)
+      ]
+    )
+  end
+
+  def reasons
+    Section.new(
+      60_745_386, :checkboxes,
+      [
+        Answer.new('3', 'Submit an LGFS claim', 505_511_336),
```
naseberry: Grammar: `an` should be `a`

app/views/feedback/feedback.html.haml
```diff
@@ -18,17 +18,31 @@
         %p
           = t('.callout')
           = render partial: 'feedback/claim_edit_alert' if referrer_is_claim?(@feedback.referrer)
+        
+      %p.govuk-body{:class => "govuk-!-margin-bottom-6"}
```
naseberry: Standard: No rocket hashes in views
`:class => "govuk-!-margin-bottom-6"` should be `class: 'govuk-!-margin-bottom-6'`

features/support/stub_survey_monkey_requests.rb
```diff
@@ -0,0 +1,5 @@
+require 'webmock/cucumber'
+
+Before('@stub_survey_monkey_success') do
+  @called_zendesk = stub_request(:post, %r{\Ahttps://api.eu.surveymonkey.com/v3/.*} ).and_return(status: 201, body: { id: 123 }.to_json )
```
naseberry: 💯

lib/survey_monkey/configuration.rb
```diff
@@ -0,0 +1,25 @@
+module SurveyMonkey
```
naseberry, line 1: 💯

app/controllers/feedback_controller.rb
jsugarman: I prefer the `eql?` method, but not a big deal IMO

features/support/stub_survey_monkey_requests.rb
jsugarman: is the response required/used, plus the name would be wrong if it is needed
```suggestion
  stub_request(:post, %r{\Ahttps://api.eu.surveymonkey.com/v3/.*} )
    .and_return(status: 201, body: { id: 123 }.to_json )
```

We should  also have a negative/error test and stub

features/feedback.feature
```diff
@@ -1,30 +1,31 @@
...
 
-  Scenario: An advocate is unable to sign in and selects to submit feedback and is prompted for their email and completes it
-    Given I have not signed in
-    When I click the link 'feedback'
-    Then I should see 'What is your email address?'
-    When I fill in the 'feedback' form with email of 'test@example.com'
-    Then I expect ZendeskSender to receive a description with an email
-    And I see confirmation that my 'feedback' was received
+    Then I see confirmation that my 'feedback' was received
     And I should be on the sign in page
```
jsugarman, line 18: I think we need a feature, deleted here, to cover the flow when a user is not signed in but submits feedback and is prompted for an email address. No?

app/forms/feedback_form.rb
jrmhaig: Is that correct? I would read it as "an el-gee-ef-es claim".

app/controllers/feedback_controller.rb
imtiazAhmd: @jrmhaig

features/support/stub_survey_monkey_requests.rb
jrmhaig: I have added a negative test.

features/step_definitions/feedback_steps.rb
```diff
@@ -47,12 +20,16 @@
   end
 end
 
+Then('I see a warning that my feedback was not submitted successfully') do
+  expect(page).to have_content "Unable to submit feedback [1050]"
```
jsugarman: can this be loosened to use a regex for the `id` (i assume that is some sort of id).

```suggestion
  expect(page).to have_content(/Unable to submit feedback \[\d+\]/)
```

app/forms/feedback_form.rb
imtiazAhmd: https://www.a-or-an.com/a_an/l
🤔

features/step_definitions/feedback_steps.rb
jsugarman: **nit**: i see it is a specific error code that provides us with info on the error, but:
a.  it might be worth loosening anyway
b. perhaps we should not bother showing this to the user anyway?

app/models/feedback.rb
```diff
@@ -43,8 +36,8 @@ def is?(type)
 
   def save
     return false unless valid?
-    ZendeskSender.send!(self) unless is_feedback_with_empty_comment?
-    true
+    return save_feedback if feedback?
+    save_bug_report
```
jsugarman, line 40: As discussed, is this the time to split the bug report out entirely to its own controller and model so that we can isolate the two entirely?

features/feedback.feature
```diff
@@ -1,30 +1,48 @@
...
+    And I choose govuk radio 'Very satisfied' for 'How satisfied have you been in your experience of Claim for crown court defence today?'
+    And I fill in 'Tell us about your experience of using the service today' with 'This is great!'
+    And I click govuk checkbox 'Other (please specify)'
+    And I fill in 'Enter your comment' with 'Something Else'
+    And I click the button 'Send'
+
+    Then I see a warning that my feedback was not submitted successfully
+    And I should be on the feedback page
+
+# TODO: Introduce a new test case for bug report
```
jsugarman: needs reimplementing

jsugarman (commented):
Looks good. a few comments for wider discussion, particularly about whether we should split the bug report out entirely into a separate controller and model.

app/forms/feedback_form.rb
jsugarman: "Submit an LGFS claim" would appear to be grammatically correct.

features/step_definitions/feedback_steps.rb
jrmhaig: The code can be one of several values and we are showing it to the user so that, if we get an error report, we can more easily find the cause. See https://developer.surveymonkey.com/api/v3/#error-codes

features/feedback.feature
naseberry: An email address is no longer required, feedback is anonymous regardless if a user is signed in or not.

app/models/feedback.rb
naseberry: We also discussed separating the feedback and bug report into different mvc.
I don't think it would be a quick task, nor was it part of the task (unforeseen) and therefore out of scope.
I would welcome a new task to action this separation.

Thoughts?

features/feedback.feature
imtiazAhmd: @jsugarman This is a new test case. I couldn't implement it now as the zendesk doesn't work locally. We need to implement this test as part of the work where we split the bug report out to its own controller and model.

jrmhaig: I have created a ticket - CFP-281

app/models/feedback.rb
jrmhaig: I have created a ticket - CFP-282

features/feedback.feature
jsugarman: if so then it looks like we would not need these in the feedback.html.haml then?

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4265/files#diff-3788e17baf72ad638ba083b1035c354a10c2a746eef7aebc93feb43552990b36R12-R13

Also we may not need this as from what i can tell bug reports do not pass `user_id` (only email) and feedback does not need them from what you say. dead code?
```
# feedback_controller.rb
  def email_from_user_id
    User.active.find(params[:user_id])&.email
  rescue ActiveRecord::RecordNotFound
    nil
  end
```

features/feedback.feature
jsugarman: @imtiazAhmd we can just stub zendesk anyway

here is a [PR against this branch](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4302) to add the feature in. works locally

app/controllers/feedback_controller.rb
jsugarman: is this resolved?

jrmhaig: Yes: 404f039a71445f247d2a45e86396f21f2c2c5305

features/feedback.feature
jsugarman: have [PR this changed here](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4304) but needs sanity testing and could go in after the fact!

jsugarman (approved):
Looks Good code-wise.

One additional PR to clean up dead code. see unresolved comment but otherwise 👍

naseberry (approved):
I've sanity tested on staging:

- Empty/Blank submission
- Individual option selection per question submission (12)
- Multiple options selection submission
- Verified error response [1014] (removed all API permissions)
- Verification of response details
- Verification of API activity

app/controllers/feedback_controller.rb
naseberry: @jsugarman @jrmhaig this issue has been re-introduced, see staging env

PR #4271, opened 2021-09-17
[CFP-194] Submit feedback to Survey Monkey

jrmhaig:
#### What

Backend changes for submitting feedback form to Survey Monkey.

#### Ticket

[User satisfaction form changes](https://dsdmoj.atlassian.net/browse/CFP-194)

#### Why

The backend component for #4265

#### How

Submission to Survey Monkey is handled by the `SurveyMonkeySender` class using the `SurveyMonkey` library.

```ruby
# Create a new Survey Monkey response
monkey = SurveyMonkey::Response.new
# Add form data
monkey.add_page(
  :feedback,
  tasks: '2',
  ratings: '1',
  comments: "Comments from the 'Tell us about your experience' feedback text box",
  reasons: ['1', '3', { other: 'Other reasons for option 3' }]
)
# Submit the form
monkey.submit
```

`SurveyMonkey` is configured in `config/initializer/survey_monkey.rb`, where details of acceptable options together with their ids can be set.

#### TODO

- [x] What is the correct value for `ip_address` in the data sent to Survey Monkey?
- [x] Add logging to Survey Monkey library
- [x] Display error message and Survey Monkey error code after failure


kubernetes_deploy/dev-lgfs/app-config.yaml
```diff
@@ -17,4 +17,5 @@ data:
   SETTINGS__AWS__POLL_MESSAGE_COUNT: '1'
   SETTINGS__AWS__POLL_MESSAGE_WAIT_TIME: '10'
   SETTINGS__AWS__S3__REGION: 'eu-west-2'
+  SURVEY_MONKEY_COLLECTOR_ID: 330142588
```
jsugarman: This is the same collector id for all non-prod environments. we do no collect zendesk except for prod and staging I think? fewer attack surfaces amongst other things

lib/survey_monkey/answer/checkboxes.rb
```diff
@@ -0,0 +1,27 @@
+module SurveyMonkey
+  module Answer
+    class Checkboxes
+      def initialize(**args)
```
jsugarman: probably better named `**kwargs` or `**options` but regardless  double-splat parameters, `**args`, makes providing the hash optional  - is that what you want?

>...if you need to have a hash passed, then it’s better to use a positional parameter, but if your code assumes that in some cases when the method is called, a hash will not be passed, then you can use a parameter with double splat operator.

https://medium.com/@sologoubalex/parameter-with-double-splat-operator-in-ruby-d944d234de34

config/initializers/survey_monkey.rb
```diff
@@ -0,0 +1,41 @@
+SurveyMonkey.configure do |config|
+  config.root_url = 'https://api.eu.surveymonkey.com/v3/'
+  config.bearer = ENV['SURVEY_MONKEY_AUTHORIZATION_BEARER']
+  config.collector_id = ENV['SURVEY_MONKEY_COLLECTOR_ID']
+  config.logger = Rails.logger
+  config.verbose_logging = true
+
+  config.register_page(
+    :feedback, 25473840,
```
jsugarman: perhaps this could be wrapped up into an object that exists in the app dir instead of the lib dir.

```
class SurveryMonkeyFeedbackPage
  def name
    :feedback
  end
  
  def id
    25473840
  end
  
  def body
     tasks.merge(ratings).....
  end

  def tasks
    {
      tasks: {
        id: task_id,
        format: :radio,
        answers: {
          ...
        }
      }
    }
  end

  def task_id
    60742936
  end
  ...
end
```

```
page = SurveryMonkeyFeedbackPage.new
config.register_page(page.name, page.id, page.body)
```

lib/survey_monkey/response.rb
```diff
@@ -0,0 +1,39 @@
...
+      @pages << SurveyMonkey.page_by_name(page).answers(**responses)
+    end
+
+    private
+
+    def parse_response(response)
+      body = JSON.parse(response.body)
+      if response.success?
+        SurveyMonkey.log(:info, "Response submitted. #{body['analyze_url']}")
+        { id: JSON.parse(response.body)['id']&.to_i, success: true }
```
jsugarman: you already have body parsed. Are you expecting nil (or non-existent key) to return nil
```suggestion
        { id: body['id']&.to_i, success: true }
```
OR
```suggestion
       { id: body.fetch('id', nil)&.to_i, success: true }
```

lib/survey_monkey/response.rb
```diff
@@ -0,0 +1,39 @@
...
+      parse_response(response)
+    end
+
+    def add_page(page, **responses)
+      @pages << SurveyMonkey.page_by_name(page).answers(**responses)
+    end
+
+    private
+
+    def parse_response(response)
```
jsugarman:
```suggestion
    def parse(response)
```

lib/survey_monkey/response.rb
```diff
@@ -0,0 +1,39 @@
...
+      survey_response = { pages: @pages.map(&:to_h) }
+      SurveyMonkey.log(:info, "Sending response. #{survey_response.inspect}")
+
+      response = SurveyMonkey.connection.post(
+        "collectors/#{SurveyMonkey.collector_id}/responses",
+        survey_response.to_json,
+        content_type: 'application/json'
+      )
+
+      parse_response(response)
```
jsugarman:
```suggestion
      parse(response)
```

spec/lib/survey_monkey/configuration_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  describe '#register_page' do
+    subject(:register_page) { SurveyMonkey.configure { |config| config.register_page(new_page, new_id) } }
+
+    context 'with the re-registering of the same page name' do
+      let(:new_page) { :test_survey }
+      let(:new_id) { 456 }
+
+      it 'replaces an existing page' do
+        expect { register_page }.to change { SurveyMonkey.page_by_name(new_page).id }.from(123).to 456
```
jsugarman:
```suggestion
        expect { register_page }.to change { SurveyMonkey.page_by_name(new_page).id }.from(123).to(456)
```

spec/lib/survey_monkey_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  describe '.page_by_name' do
+    subject(:page) { described_class.page_by_name(page_name) }
+
+    context 'with a known page' do
+      let(:page_name) { :test_survey }
+
+      it { expect(page.id).to eq 123 }
```
jsugarman:
```suggestion
      it { expect(page.id).to eq(123) }
```

app/models/feedback.rb
```diff
@@ -54,8 +56,14 @@ def is?(type)
 
   def save
     return false unless valid?
-    ZendeskSender.send!(self) unless is_feedback_with_empty_comment?
-    true
+    if feedback?
+      response = SurveyMonkeySender.send_response(self)
```
jsugarman: If we are following the ZendeskSender naming convention and also for brevity I thin `send!` is a better name
```suggestion
      response = SurveyMonkeySender.send!(self)
```

app/services/survey_monkey_sender.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.send_response(feedback)
+    new(feedback).send_response
+  end
+
+  def initialize(feedback)
+    @monkey = SurveyMonkey::Response.new
+    @monkey.add_page(:feedback, **payload(feedback))
+  end
+
+  def send_response
```
jsugarman: as mentioned above
```suggestion
  def send!
```

config/initializers/survey_monkey.rb
```diff
@@ -0,0 +1,41 @@
+SurveyMonkey.configure do |config|
+  config.root_url = 'https://api.eu.surveymonkey.com/v3/'
+  config.bearer = ENV['SURVEY_MONKEY_AUTHORIZATION_BEARER']
```
jsugarman: nit - is this a a bearer token, just thinking!
```suggestion
  config.bearer = ENV['SURVEY_MONKEY_BEARER_TOKEN']
```

app/services/survey_monkey_sender.rb
```diff
@@ -0,0 +1,31 @@
+class SurveyMonkeySender
+  def self.send_response(feedback)
+    new(feedback).send_response
+  end
+
+  def initialize(feedback)
+    @monkey = SurveyMonkey::Response.new
```
jsugarman: Naming wise, ignoring the amusement factor, i think @response or @survey, @survey_response is clearer
 
```suggestion
    @survey = SurveyMonkey::Response.new
```

```suggestion
    @response = SurveyMonkey::Response.new
```

lib/survey_monkey/question/checkboxes.rb
```diff
@@ -0,0 +1,32 @@
+module SurveyMonkey
+  module Question
+    class Checkboxes < Base
+      def initialize(id, answers:, **_args)
```
jsugarman: as with other incarnations of this I think it is clearer to either pass all `options` and then use or set var from `options[:answers]` or  call with `answers: options[:answers]`, and remove the unused `**_args` from the definition

jsugarman (changes requested):
A lot of stuff to look at so have probably missed things but it is good work and aside from the individual comments I wonder whether:

1. we are missing a `FeedbackPage` object generally
2. whether we could delegating methods more to avoid use of things like 

```
SurveyMonkey.page_by_name
SurveyMonkey.log
```

kubernetes_deploy/dev-lgfs/app-config.yaml
jrmhaig: :+1: I had put this in all environments but I'll set it just to staging.

lib/survey_monkey/response.rb
```diff
@@ -0,0 +1,39 @@
+module SurveyMonkey
+  class Response
+    delegate :connection, :collector_id, :page_by_name, :log, to: :SurveyMonkey
```
jsugarman: you can spec these with, for example (although i have not tried that with a Module/Class

```
it { is_expected.to delegate_method(:connection).to(:SurveyMonkey) }
```

app/services/survey_monkey_sender.rb
```diff
@@ -0,0 +1,31 @@
+class SurveyMonkeySender
+  def self.call(feedback)
+    new(feedback).call
+  end
+
+  def initialize(feedback)
+    @survey_response = SurveyMonkey::Response.new
```
jsugarman: **nit** if not injecting the dependency (could do that too) then it is neater to isolate the dependency

```
def initialize(feedback)
  response.add_page(:feedback, **payload(feedback))
end

private

def response
  @response ||= SurveyMonkey::Response.new
end
```

app/services/survey_monkey_sender.rb
```diff
@@ -0,0 +1,31 @@
...
+    @survey_response.add_page(:feedback, **payload(feedback))
+  end
+
+  def call
+    @survey_response.submit
+  end
+
+  private
+
+  def payload(feedback)
```
jsugarman: **nit** if you initialize a `@feedback` var you could remove the passing of it to the method. As a private method it is only ever going to use what the object is initialised with

app/controllers/feedback_controller.rb
```diff
@@ -67,4 +68,8 @@ def feedback_params
       reason: []
     )
   end
+
+  def setup_page
+    @feedback_page = FeedbackForm.new if type == 'feedback'
```
jsugarman: **nit** I think it is clearer to call the var  `@feedback_form`

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,67 @@
...
+        Answer.new('3', 'Submit an LGFS claim', 505_511_336),
+        Answer.new('2', 'Submit an AGFS claim', 505_511_337),
+        Answer.new('1', 'Other (please specify)', 505_511_338, true)
+      ]
+    }
+  end
+
+  private
+
+  def survey_monkey_tamplate_for(section)
```
jsugarman: off two minds whether `template_for` is a better name, but either way! it is all  coupled to survey monkey anyway.

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,67 @@
...
+    {
+      tasks: survey_monkey_tamplate_for(tasks),
+      ratings: survey_monkey_tamplate_for(ratings),
+      reasons: survey_monkey_tamplate_for(reasons),
+      comments: { id: 60_742_937, format: :text }
+    }
+  end
+
+  def tasks
+    {
```
jsugarman: probably over optimisation but we could have
```
Section = Struct.new(:id, :format, :answers)

...

Section.new(60_742_936, 
                      :radio, 
                      [Answer.new('3', 'Yes', 505_487_572),
                       Answer.new('2', 'No', 505_487_573),
                       Answer.new('1', 'Partially', 505_487_574)]
```


```
# in view
@feedback_form.tasks.answers do |answer|
  ...

```

not sure if you would nee to `to_h` this for when you need it as a hash?! does it work for the Answer structs?

config/initializers/survey_monkey.rb
```diff
@@ -5,37 +5,6 @@
...
-      format: :checkboxes,
-      answers: {
-        '3' => 505511336, # Submit an LGFS claim
-        '2' => 505511337, # Submit an AGFS claim
-        '1' => { id: 505511338, other: true } # Other reasons
-      }
-    }
-  )
+  page = FeedbackForm.new
+  config.register_page(page.name, page.id, **page.template)
```
jsugarman, line 9: does the double-splat deconstruct the answers array of structs into the expected answers hash

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,67 @@
+class FeedbackForm
```
jsugarman, line 1: We will need some  specs for this class

jsugarman (changes requested):
Looks good

- need feedback_form spec
- question on struct to hash deconstruction - is it doing it transparent?

lib/survey_monkey/response.rb
jrmhaig: This doesn't work as `delegate_method` expects the argument of `to` to be an instance method.

I will, however, make these delegates private as they do not need to be public.

app/controllers/feedback_controller.rb
jrmhaig: :+1: I had originally called the class `FeedbackPage` but then renamed it.

app/forms/feedback_form.rb
jrmhaig: :+1: I don't think it is needed as a hash.

config/initializers/survey_monkey.rb
jrmhaig: `page.template` isn't a struct at this point, so that should be OK.

app/forms/feedback_form.rb
jrmhaig: I've added some basic tests, although this is effectively a configuration file at the moment. I'll have a look at decoupling the survey monkey template formatter from this too.

lib/survey_monkey/response.rb
```diff
@@ -1,6 +1,6 @@
 module SurveyMonkey
   class Response
-    delegate :connection, :collector_id, :page_by_name, :log, to: :SurveyMonkey
+    delegate :connection, :collector_id, :page_by_name, :log, to: :SurveyMonkey, private: true
```
jsugarman, line 3: nice, did not know about `private: true`

spec/forms/feedback_form_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    subject { form.id }
+
+    it { is_expected.to eq 25_473_840 }
+  end
+
+  describe '#template' do
+    subject(:template) { form.template }
+
+    it { is_expected.to be_a Hash }
+    it { expect(template.values.map(&:class).uniq).to match_array([Hash]) }
```
jsugarman, line 23: **nit** might this work and be more readable
```suggestions
    it { expect(template.values).to all(be_instance_of(Hash)) } 
```

app/views/feedback/feedback.html.haml
```diff
@@ -23,11 +23,11 @@
         = t('.report_fault_html', bug_report_link: new_feedback_path(type: 'bug_report'))
 
       = f.govuk_collection_radio_buttons :task,
-        @feedback_page.tasks[:answers], :key, :label,
+        @feedback_form.tasks.answers, :key, :label,
```
jsugarman, line 26: 💯

app/forms/feedback_form.rb
```diff
@@ -11,57 +11,55 @@ def id
...
 
-  def survey_monkey_tamplate_for(section)
+  def template_for(section)
     {
-      id: section[:id],
-      format: section[:format],
-      answers: section[:answers].map do |answer|
-        [answer.key, answer.other ? { id: answer.id, other: true } : answer.id]
-      end.to_h
+      id: section.id,
```
jsugarman, line 67: 🏆

app/forms/form_answer.rb
```diff
@@ -0,0 +1,16 @@
+class FormAnswer
```
jsugarman: **nit** if we use a separate class, rather than struct this a. should have a spec b. probably better named feedback_form_answer, or namespaced as it is quite specific to the feedback form currently. 

I personally think a Struct would be more suitable and implicitly handle the naming/namespacing and testing grey areas.

```
class FeedbackForm
  Answer = Struct.new(:key, :label, :id, :other) do
    def initialize(:key, :label, :id, other: false)
      super
    end
    
    def id
      return { id: id, other: true } if other
      id
    end
  end
 ...
end
```

app/forms/feedback_form.rb
```diff
@@ -0,0 +1,65 @@
...
+    )
+  end
+
+  private
+
+  def template_for(section)
+    {
+      id: section.id,
+      format: section.format,
+      answers: section.answers.map { |answer| [answer.key, answer.id] }.to_h
```
jsugarman, line 69: 👍 missed this -  this is where the `to_h` is called, that converts all the structs to hashes

jsugarman (approved):
Great, approved, albeit I have not sanity tested this on any environment.

Couple of nits you might want to consider otherwise 👍  - I think the Answer Struct would make things neater.

app/forms/form_answer.rb
jrmhaig: I'll move it back into a struct. I had a thought about going further with this but it was getting overly complicated.

PR #4273, opened 2021-09-21
[CFP-198] add AGFS and LGFS management information reports

jsugarman:
#### What
Provide 2 additional MI reports, only AGFS and only LGFS

#### Ticket

[CFP-198](https://dsdmoj.atlassian.net/browse/CFP-198)

#### Why
The number of returned by the existing MI report is
prohibitively large, making the case worker use of it
,including fitlering and extracting to other reports
difficult and cumbersome.

The report is used by 2 distinct teams, one for AGFS
claims and one for LGFS so providing two separate
reports, split by scheme, provides more applicable datasets
of a smaller size and thereby thereby easing case worker
manipulation of the data.

#### How
Pass in scheme as an optional scope on the base relational
query.

### DIFF
--- a/app/services/stats/management_information_generator.rb
+++ b/app/services/stats/management_information_generator.rb
@@ -10,6 +10,7 @@ module Stats
 
     def initialize(options = {})
       @format = options.fetch(:format, DEFAULT_FORMAT)
+      @claim_scope = options.fetch(:claim_scope, :all)
     end
 
     def call
@@ -25,8 +26,8 @@ module Stats
       Settings.claim_csv_headers.map { |header| header.to_s.humanize }
     end
 
-    def active_non_draft_claims
-      Claim::BaseClaim.active.non_draft
+    def claims
+      Claim::BaseClaim.active.non_draft.send(@claim_scope)
     end
 
     # TODO: separate data retrieval from exporting the data itself
@@ -35,7 +36,7 @@ module Stats
       log_info('Report generation started...')
       content = CSV.generate do |csv|
         csv << headers
-        active_non_draft_claims.find_each do |claim|
+        claims.find_each do |claim|
           ManagementInformationPresenter.new(claim, 'view').present! do |claim_journeys|
             claim_journeys.each { |journey| csv << journey } if claim_journeys.any?
           end
--- a/scheduled_tasks/lgfs_management_information_generation_task.rb
+++ b/scheduled_tasks/lgfs_management_information_generation_task.rb
@@ -1,0 +1,14 @@
+require 'chronic'
+
+# https://github.com/ssoroka/scheduler_daemon for help
+class LgfsManagementInformationGenerationTask < Scheduler::SchedulerTask
+  every '1d', first_at: Chronic.parse('next 2:30 am')
+
+  def run
+    LogStuff.info { 'LGFS Management Information Generation started' }
+    Stats::StatsReportGenerator.call('lgfs_management_information')
+    LogStuff.info { 'LGFS Management Information Generation finished' }
+  rescue StandardError => e
+    LogStuff.error { 'LGFS Management Information Generation error: ' + e.message }
+  end
+end
--- a/spec/services/stats/stats_report_generator_spec.rb
+++ b/spec/services/stats/stats_report_generator_spec.rb
@@ -2,9 +2,7 @@ require 'rails_helper'
 
 RSpec.describe Stats::StatsReportGenerator, type: :service do
   describe '.call' do
-    subject(:call_report_generator) { described_class.call(report_type) }
-
-    let(:report_type) { 'management_information' }
+    subject(:call) { described_class.call(report_type) }
 
     context 'with an invalid report type' do
       let(:report_type) { 'some-report-type' }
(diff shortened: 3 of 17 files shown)


spec/services/stats/stats_report_generator_spec.rb
```diff
@@ -2,87 +2,141 @@
...
+
+      context 'with management information report' do
+        let(:report_type) { 'management_information' }
+
+        before { allow(Stats::ManagementInformationGenerator).to receive(:call).and_return(mocked_result) }
+
+        it 'calls management information generator' do
+          call
+          expect(Stats::ManagementInformationGenerator).to have_received(:call)
+        end
```
jrmhaig: Considering that there are now tests for when `Stats::ManagementInformationGenerator` is called with a `claim_scope` may be it is worth explicitly checking that there is no argument in this case.

Either;

```suggestion
        it 'calls management information generator' do
          call
          expect(Stats::ManagementInformationGenerator).to have_received(:call).with(no_args)
        end
```

or

```suggestion
        it 'does not include a claim scope' do
          call
          expect(Stats::ManagementInformationGenerator).to have_received(:call).with(hash_excluding(:claim_scope))
        end
```

scheduled_tasks/lgfs_management_information_generation_task.rb
```diff
@@ -0,0 +1,14 @@
...
+# https://github.com/ssoroka/scheduler_daemon for help
+class LgfsManagementInformationGenerationTask < Scheduler::SchedulerTask
+  every '1d', first_at: Chronic.parse('next 3:30 am')
+
+  def run
+    LogStuff.info { 'AGFS Management Information Generation started' }
+    Stats::StatsReportGenerator.call('lgfs_management_information')
+    LogStuff.info { 'AGFS Management Information Generation finished' }
+  rescue StandardError => e
+    LogStuff.error { 'AGFS Management Information Generation error: ' + e.message }
```
jrmhaig: AGFS -> LGFS

```suggestion
    LogStuff.info { 'LGFS Management Information Generation started' }
    Stats::StatsReportGenerator.call('lgfs_management_information')
    LogStuff.info { 'LGFS Management Information Generation finished' }
  rescue StandardError => e
    LogStuff.error { 'LGFS Management Information Generation error: ' + e.message }
```

scheduled_tasks/lgfs_management_information_generation_task.rb
```diff
@@ -0,0 +1,14 @@
+require 'chronic'
+
+# https://github.com/ssoroka/scheduler_daemon for help
+class LgfsManagementInformationGenerationTask < Scheduler::SchedulerTask
+  every '1d', first_at: Chronic.parse('next 3:30 am')
```
jrmhaig: (nit) Maybe run at a different time from the AGFS task. I see this is already 30 minutes after the complete MI report task.

```suggestion
  every '1d', first_at: Chronic.parse('next 4 am')
```

app/services/stats/management_information_generator.rb
```diff
@@ -10,6 +10,7 @@ def self.call(options = {})
 
     def initialize(options = {})
       @format = options.fetch(:format, DEFAULT_FORMAT)
+      @claim_scope = options.fetch(:claim_scope, :all)
```
jrmhaig, line 13: (nit, and possibly not really part of this PR) Unless there are lots of arguments I would prefer;

```ruby
def initialize(format: 'csv', claim_scope: :all)
  @format = format # And the DEFAULT_FORMAT constant can be removed
  @claim_scope = claim_scope
end
```

(Also `:csv` instead of `'csv'` for the default format but that may have a knock-on effect)

spec/services/stats/stats_report_generator_spec.rb
jsugarman: 👍

scheduled_tasks/lgfs_management_information_generation_task.rb
jsugarman: 👍

scheduled_tasks/lgfs_management_information_generation_task.rb
jsugarman: 👍  I have moved them to 2am and 2:30am (note this is server time which is UTC, so add an hour) to be before the main one which can take 30 to 45+mins.

app/services/stats/management_information_generator.rb
jsugarman: Unless we change `StatsReportGenerator.initialize` to match this would require its call to be amended as below, which just shifts the yuckiness. options hashes are a pretty common ruby idiom, albeit increasingly out of fashion for kwargs.

```
# in Stats::StatsReportGenerator
ManagementInformationGenerator.call(format: options[:format], claim_scope: options[:claims_scope])
```

And amending the generalised `Stats::StatsReportGenerator` as below does not feel right to me at the moment as claim_scope is only applicable to MI reports.
```
module Stats
  class StatsReportGenerator
  ...
    def initialize(report_type, format: 'csv', claims_scope: :all)
      @report_type = report_type
      @options = options
    end
    ...
   end
end
```

The next tranch of work may involve further refactoring which could smooth this out.

jrmhaig: OK. No problem.

jrmhaig (approved):
I've checked on dev-lgfs and confirmed that the old MI report has all the same elements as the combined new lgfs and agfs reports.

PR #4287, opened 2021-09-23
[CFP-234] Migrate caseworker allocation table to Design System 1/3

naseberry:
#### What
Migrate allocation table markup to the GOVUK Design System Table markup.

#### Ticket
[Migrate allocation table markup to the GOVUK Design System Table](https://dsdmoj.atlassian.net/browse/CFP-234)

#### Why
Part of a larger change to:


jrmhaig:
Comparing between dev and dev-lgfs (with this branch) I notice that the warning for 'Claim not injected' has changed. Specifically, the text is indented further into the row and there is no longer a red bar on the left. Is this intended?

**Old**
Screenshot 2021-09-28 at 10 11 30

**New**
Screenshot 2021-09-28 at 10 11 38

jsugarman (approved):
Have sanity tested fundamental function of the allocation page and all appears to function as expected

Style wise, comparing dev to dev-lgfs, I notice the table loading message has changed and the up/down column order. Both of which seem to be an improvement IMO. 

One thing to consider - the order icons are not very clear (grey) from a possible accessibility point of view?! 


Changed table loading styling - seems to be an improvement if intended?!
Screenshot 2021-09-28 at 16 45 17 (2)

imtiazAhmd (approved):
Sanity checked on `dev-lgfs`

PR #4288, opened 2021-09-24
Fix dump and restore in line with paperclip attribute removal

jsugarman:
#### What
Fixes the dump task to work against current DB schema

#### Ticket

came out of MI report split ticket [CFP-198](https://dsdmoj.atlassian.net/browse/CFP-198)

#### Why
Was broken and need to refresh dev-lgfs


jrmhaig (approved):
Looks good.

PR #4290, opened 2021-09-24
[CFP-208] Fix non-rendering integers in `case_details/summary` view

jsugarman:
#### What
Fix non-rendering integers in `case_details/summary` view

#### Ticket

[CFP-208](https://dsdmoj.atlassian.net/browse/CFP-208)

#### Why
The following fields, at least, will not render, because they are not String objects

- [X] estimated_trial_length
- [X] actual_trial_length
- [X] retrial_estimated_length
- [X] retrial_actual_length

*In explanation*
The passing of a curly brace block with this component
helper does not work for non-string objects.

This is due to the block being used by a capture statement.
The implementation of [capture](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/actionview/lib/action_view/helpers/capture_helper.rb#L46) means it will only output the result if it is a `String` object.

This [commit](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/036af87b09b72f574d3173bc2fae8b75c3cb6def) introduced
the `govuk_summary_list_row` component and replaced use of, for example,
```ruby
.govuk-summary-list__row
  %dt.govuk-summary-list__key
    = t('estimated_trial_length', scope: trial_detail_fields)
  %dd.govuk-summary-list__value
    = claim&.estimated_trial_length
```
with
```ruby
= govuk_summary_list_row( t('actual_trial_length', scope: trial_detail_fields) ) { claim&.actual_trial_length }
```
...but since claim&.actual_trial_length is an integer it is not "captured" by the helper.

hamls '=' method would return a string so the code below would work
```ruby
= govuk_summary_list_row( t('actual_trial_length', scope: trial_detail_fields) ) do
  = claim&.actual_trial_length
```

if a single line with curly braces block is to be used then the object must be explicitly cast to a string, as below
```ruby
= govuk_summary_list_row( t('estimated_trial_length', scope: trial_detail_fields) ) { claim&.estimated_trial_length&.to_s }
```

In the end we need to either have view or feature specs to define the expected
view result so such refactoring do not silently break the views expected behaviour.

Could also handle this explicitly by casting in the helpers themselves.

#### TODO (wip)
 - [X] fix most obvious breakages
 - [x] look for other possible non-rendering objects in summary page
 - [x] add case_details/_summary view spec or extend feature tests to cover expected behaviour

Further Todos:
 - view-spec/feature-test all summary partials and conditional logic


lib/govuk_component/shared_helpers.rb
```diff
@@ -10,5 +10,10 @@ def prepend_classes(classes_to_prepend, options = {})
       options[:class] = classes.join(' ')
       options
     end
+
+    def capture_output
+      output = proc { yield&.to_s }
```
naseberry, line 15: 💯

naseberry (approved):
Great coverage of the summary view 🥇

PR #4294, opened 2021-09-28
Define claim state factorybot traits globally

jsugarman:
These are applicable to any claim and their
traits can therefore be shared globally.


spec/factories/claim/litigator_hardship_claims.rb
```diff
@@ -6,30 +6,22 @@
...
-        frozen_time do
-          claim.submit!
-          claim.allocate!
-          claim.assessment.update(fees: 24.2, expenses: 8.5)
-          claim.authorise!
-          claim.redetermine!
-        end
-      end
-    end
+    # trait :redetermination do
```
jsugarman: remove

jsugarman: not used

.rubocop.yml
```diff
@@ -64,6 +64,9 @@ Metrics/BlockLength:
     - trait
     - define
 
+Rails/FilePath:
+  EnforcedStyle: arguments
```
jrmhaig: :+1:

spec/factories/claim/shared/common_claim_traits.rb
```diff
@@ -0,0 +1,75 @@
+# These methods are going to the global namespace so do not abuse it and
```
jrmhaig: This comment (I think) related to the `claim_state_common_traits` method, which you have now removed.

spec/factories/claim/advocate_claims.rb
```diff
@@ -91,24 +83,18 @@
       after(:create) { |claim| allocate_claim(claim); claim.reload }
     end
 
-    trait :allocated do
-      after(:create) { |claim| allocate_claim(claim); claim.reload }
-    end
-
+    # DEPRECATED see shared traits
```
jrmhaig, line 86: :+1: for the comment. I was unaware that these factories were deprecated. Thanks.

jrmhaig (commented):
Just one comment that I think is no longer required. Please remove or clarify. Otherwise, looks good.

spec/factories/claim/shared/common_claim_traits.rb
jsugarman: Although I have named this file as a generic holder of traits common to claims it is, in fact, currently only storing common claim state transition traits/helpers. The comment applies generally to all "global" traits as they can cause clashes, which is why i put the comment in here as well. 

Happy to remove the comment if you think it adds no value?

spec/factories/claim/advocate_claims.rb
jsugarman: 👌  They can, and are, still used but are a duplication of logic that is/should-be generic to all claims. I only left them here to avoid the yak-shaving involved in remove them.

We should remove them when time permits.

spec/factories/claim/shared/common_claim_traits.rb
jrmhaig: If I understood correctly, the comment was copied over from `spec/factories/claim/base_claims.rb` where it was referring to the `claim_state_common_traits` methods (which contained these trait definitions) as well as a few other methods. It is applicable to the methods, which are still in `base_claims.rb`, but not these traits.

jsugarman: anything defined directly inside `FactoryBot.define do` will go into the global namespace and therefore could have suffer collisions i believe. 

I will remove the comment as it is obviously confusing.

jrmhaig (approved):

PR #4300, opened 2021-09-30
Apply rubocop `Rails/FilePath` `argument` style

jsugarman:
#### What
Apply rubocop `Rails/FilePath` `argument` style

#### Why

To adhere to one style throughout the app, which mainly
uses the argument style, baring these examples.

Also rely on Pathname object to be usable everywhere a
string path is usable. This should be generally true
and tests should seriously fail if not but will need sanity
testing.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,6 +64,9 @@ Metrics/BlockLength:
     - trait
     - define
 
+Rails/FilePath:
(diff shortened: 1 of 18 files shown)


jrmhaig (approved):
:+1: I prefer the 'arguments' style. Not sure why the alternative is the default.

PR #4302, opened 2021-10-04
Implement basic bug report feature

jsugarman:
Since they currently use the same controller this
could be added here, and split out later to its own
`bug_report.feature` file.


naseberry (approved):
👍  Glad to have the bug report feature

jrmhaig (approved):

imtiazAhmd (approved):
Checked Locally, Works Flawlessly 👍

PR #4303, opened 2021-10-05
[CFP-274] Resolve Dependabot alerts surrounding jQuery security vulnerabilities

imtiazAhmd:
#### What
Focussing on the successful upgrade of jQuery to the latest stable version and ensuring the existing functionality/spec continues to work.

**Given** we bump jQuery to version 3.5.0 or later
**When** it is merged into the master branch
**Then** the jQuery Dependabot alert should be resolved
#### Ticket
[Resolve Dependabot alerts surrounding jQuery security vulnerabilities](https://dsdmoj.atlassian.net/browse/CFP-274)

#### Why
The current fixed version is old, some methods being used by the app is deprecated and the version has some noted security vulnerabilities

### How
We are upgrading jQuery to 3.6.0 and introducing jQuery Migrate. This new library will restore the APIs that were removed and additionally shows warnings in the browser console when removed and/or deprecated APIs are used. So We can bump the version to the latest and fix the issues simultaneously, as old code will continue to work until the next major release. 

#### TODO

 - [x] Fix failing Jasmin Tests


naseberry (approved):
💯 

Solid code changes to update the dependency and fix failing specs.
I've also sanity test on staging, covering areas such as:

- File upload (Dropbox)
- Datatables
- Autocomplete
- Show&Hide
- Offences search (fetch)
- Distance calculator

jsugarman (approved):
Can't say I am familiar but I have quickly sanity tested on staging and so has peter so 👍

config/webpack/environment.js
```diff
@@ -8,6 +8,7 @@ environment.plugins.append('ConcatPlugin', new ConcatPlugin({
   fileName: 'application.bundle.js',
   filesToConcat: [
     'jquery',
+    process.env.RAILS_ENV === 'development' ? 'jquery-migrate' : 'jquery-migrate/dist/jquery-migrate.min.js',
```
jrmhaig, line 11: Is this line required or is it there for debugging?

imtiazAhmd: Required, This line is to make sure on Production, we only load minify version of the jQuery Migrate. Which has no Warning Logs (Which deprecated APIs are still in use) Enabled. We would only like to see those warnings in the Dev environment.

jrmhaig (approved):
Like @jsugarman I'm not too familiar with this but it is good to see version numbers going up. 😄

PR #4304, opened 2021-10-05
Bug report changes on top of feedback changes

jsugarman:
Add success and failure for bug report submission

Inline with feedback form changes this changes the
bug report to handle success and failure in a similar
manner.

**Also**, removes code which is no longer needed
or was already "dead". And breaks out
feature tests for the bug report and extends to include
the logged in and logged out workflows.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -64,7 +64,10 @@ class Feedback
     ZendeskSender.send!(self)
     @response_message = 'Fault reported'
   rescue ZendeskAPI::Error::ClientError => e
-    @response_message = "Unable to submit fault report [#{e}]"
+    @response_message = 'Unable to submit fault report'
+    LogStuff.error(class: self.class, action: 'save', error_class: e.class, error: e.to_s) do
+      'Bug report submisson failed!'
+    end
     false
   end
 
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -9,7 +9,7 @@ class Feedback
 
   attr_accessor :email, :referrer, :user_agent, :type
   attr_accessor :event, :outcome, :case_number
-  attr_accessor :task, :rating, :comment, :reason, :other_reason, :response_failed_message
+  attr_accessor :task, :rating, :comment, :reason, :other_reason, :response_message
 
   validates :type, inclusion: { in: FEEDBACK_TYPES.keys.map(&:to_s) }
   validates :event, :outcome, presence: true, if: :bug_report?
@@ -35,7 +35,7 @@ class Feedback
   end
 
   def save
-    return false unless valid?
+    return unless valid?
     return save_feedback if feedback?
     save_bug_report
   end
@@ -52,20 +52,23 @@ class Feedback
 
   def save_feedback
     response = SurveyMonkeySender.call(self)
-    @response_failed_message = "Unable to submit feedback [#{response[:error_code]}]" unless response[:success]
+    @response_message = if response[:success]
+                          'Feedback submitted'
+                        else
+                          "Unable to submit feedback [#{response[:error_code]}]"
+                        end
     response[:success]
   end
 
   def save_bug_report
-    ZendeskSender.send!(self) unless is_feedback_with_empty_comment?
-    true
+    ZendeskSender.send!(self)
+    @response_message = 'Fault reported'
+  rescue ZendeskAPI::Error::ClientError => e
+    @response_message = "Unable to submit fault report [#{e}]"
+    false
   end
 
   def feedback_type_attributes
     FEEDBACK_TYPES[type.to_sym]
-  end
-
-  def is_feedback_with_empty_comment?
-    feedback? && comment.to_s.empty?
   end
 end
--- a/spec/models/feedback_spec.rb
+++ b/spec/models/feedback_spec.rb
@@ -107,13 +107,19 @@ RSpec.describe Feedback, type: :model do
       context 'when zendesk submission fails' do
         before do
           allow(ZendeskAPI::Ticket).to receive(:create!).and_raise ZendeskAPI::Error::ClientError, 'oops, something went wrong'
+          allow(LogStuff).to receive(:error)
         end
 
         it { expect(bug_report.save).to be_falsey }
 
         it 'stores failure message on object' do
           bug_report.save
-          expect(bug_report.response_message).to eq('Unable to submit fault report [oops, something went wrong]')
+          expect(bug_report.response_message).to eq('Unable to submit fault report')
+        end
+
+        it 'logs error details' do
+          bug_report.save
+          expect(LogStuff).to have_received(:error).with(class: described_class, action: 'save', error_class: ZendeskAPI::Error::ClientError, error: 'oops, something went wrong')
         end
       end
     end
--- a/spec/models/feedback_spec.rb
+++ b/spec/models/feedback_spec.rb
@@ -3,7 +3,6 @@ require 'rails_helper'
 RSpec.describe Feedback, type: :model do
   let(:params) do
     {
-      email: 'example@example.com',
       user_agent: 'Firefox',
       referrer: '/index'
     }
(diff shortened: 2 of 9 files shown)


imtiazAhmd (approved):
looks harmless to me. needs sanity testing.

app/models/feedback.rb
```diff
@@ -52,13 +53,21 @@ def description
...
-    @response_failed_message = "Unable to submit feedback [#{response[:error_code]}]" unless response[:success]
+    if response[:success]
+      @response_succeeded_message = 'Feedback submitted'
+    else
+      @response_failed_message = "Unable to submit feedback [#{response[:error_code]}]"
+    end
     response[:success]
   end
 
+  # TODO: is_feedback_with_empty_comment? is invalid here ??? should always be false for bug reports!
```
jsugarman: lets do this in this branch!

app/models/feedback.rb
```diff
@@ -52,20 +53,23 @@ def description
 
   def save_feedback
     response = SurveyMonkeySender.call(self)
-    @response_failed_message = "Unable to submit feedback [#{response[:error_code]}]" unless response[:success]
+    if response[:success]
+      @response_succeeded_message = 'Feedback submitted'
+    else
+      @response_failed_message = "Unable to submit feedback [#{response[:error_code]}]"
+    end
```
jrmhaig: Only one of `@response_succeeded_message` and `@response_failed_message` is used so maybe there only needs to be one `@response_message`. The return value from `save` already indicates whether it is a success or failure.

You could have this (although it may be clearer still to have it over multiple lines):

```ruby
    @response_message = response[:success] ? 'Feedback submitted' : "Unable to submit feedback [#{response[:error_code]}]"
```

spec/models/feedback_spec.rb
```diff
@@ -40,16 +40,12 @@
     end
 
     describe '#is?' do
-      context 'feedback' do
-        it 'is true for feedback' do
-          expect(feedback.is?(:feedback)).to eq(true)
-        end
+      context 'with feedback type' do
+        it { expect(feedback.is?(:feedback)).to be true }
```
jrmhaig, line 43: (nit) `be_truthy`

spec/models/feedback_spec.rb
```diff
@@ -40,16 +40,12 @@
...
       end
 
-      context 'bug report' do
-        it 'is false for feedback' do
-          expect(feedback.is?(:bug_report)).to eq(false)
-        end
+      context 'with bug report type' do
+        it { expect(feedback.is?(:bug_report)).to be false }
```
jrmhaig, line 47: (nit) `be_falsey`

app/models/feedback.rb
```diff
@@ -52,20 +53,23 @@ def description
...
     response[:success]
   end
 
   def save_bug_report
-    ZendeskSender.send!(self) unless is_feedback_with_empty_comment?
-    true
+    ZendeskSender.send!(self)
+    @response_succeeded_message = 'Fault reported'
+  rescue ZendeskAPI::Error::ClientError => e
+    @response_failed_message = "Unable to submit fault report [#{e}]"
```
jrmhaig: :+1: for giving the user feedback when there is a submission failure.

We probably do not want to be putting the `ZendeskAPI::Error::ClientError` directly in here as we do not have control over what information is included. I got this;

Screenshot 2021-10-06 at 15 18 40

which gives unnecessary information to the user. It would be helpful to have an error code, like Survey Monkey, but I don't see that anywhere.

spec/models/feedback_spec.rb
```diff
@@ -59,7 +55,7 @@
       params.merge(type: 'bug_report', case_number: 'XXX', event: 'lorem', outcome: 'ipsum')
```
jrmhaig: The email (included in the params on line 6) could be moved to here as it is not required now for the feedback part.

spec/models/feedback_spec.rb
```diff
@@ -68,86 +64,109 @@
...
+          expect(bug_report.response_succeeded_message).to eq('Fault reported')
         end
       end
 
-      context 'when invalid' do
-        it 'returns false' do
-          subject.event = nil
-          subject.outcome = nil
-          expect(subject.save).to eq(false)
+      context 'when bug report has not outcome' do
```
jrmhaig:
```suggestion
      context 'when bug report has no outcome' do
```

spec/models/feedback_spec.rb
```diff
@@ -59,7 +55,7 @@
       params.merge(type: 'bug_report', case_number: 'XXX', event: 'lorem', outcome: 'ipsum')
     end
 
-    subject { Feedback.new(bug_report_params) }
+    subject(:bug_report) { Feedback.new(bug_report_params) }
```
jrmhaig: (While you are here)

```suggestion
    subject(:bug_report) { described_class.new(bug_report_params) }
```

Also, move above the preceding `let`.

jrmhaig (commented):
Looks good. A few comments, most of which you may decide to take on board but some are regarding further clean-up and not specifically to do with what you have changed. The only one I'd really want changing is the message given to users following a Zendesk submission failure.

app/models/feedback.rb
jsugarman: > Only one of @response_succeeded_message and @response_failed_message is used so maybe there only needs to be one @response_message.

👍  good shout!

Use of the  ternary was considered and decided against as it would 
>...be clearer still to have it over multiple lines...

spec/models/feedback_spec.rb
jsugarman: I originally used `be_truthy` but, in this context, raised a rubocop-rspec violation ([pipline error](https://app.circleci.com/pipelines/github/ministryofjustice/Claim-for-Crown-Court-Defence/6432/workflows/2f6c3cff-848c-4f95-9e63-bb97af0d9ac1/jobs/49890))and its suggested fix does not look right when i tried it so chose the lesser of two evils - it should always be a boolean in any event due to `==` comparison.

```
spec/models/feedback_spec.rb:158:11: C: [Correctable] RSpec/PredicateMatcher: Prefer using be_is matcher over is?.
          expect(bug_report.is?(:feedback)).to be_falsey
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/models/feedback_spec.rb:164:11: C: [Correctable] RSpec/PredicateMatcher: Prefer using be_is matcher over is?.
          expect(bug_report.is?(:bug_report)).to be_truthy
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

spec/models/feedback_spec.rb
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4304#discussion_r723456200)

app/models/feedback.rb
jsugarman: yeah, we discussed this in tuple session. happy to remove but perhaps log the detail instead (as sentry will not capture any error detail dues to us handling in-app)?

spec/models/feedback_spec.rb
jrmhaig: OK. This should go when this is split into two parts anyway.

app/models/feedback.rb
jsugarman: done now.

_the ternary also violates rubocop line length_

app/models/feedback.rb
jsugarman: have moved error detail to log now @jrmhaig

jsugarman: we should probably exclude the error code from survey monkey for feedback too BTW @jrmhaig as it is meaningless to a user and could be logged instead to track survey error incident.

jrmhaig (approved):
:+1:

PR #4314, opened 2021-10-06
[CFP-286] Use black text for messages in print stylesheet

jrmhaig:
#### What

Set the text colour to black in the print stylesheet for right-flushed messages.

#### Ticket

[Missing messages when converting to PDF #257595](https://dsdmoj.atlassian.net/browse/CFP-286)

#### Why

At some point the print stylesheet was altered so that the messages from caseworkers, which on the web are displayed as white-on-blue, started printing as white-on-white.

NB, on Mac it appears to print light-gray-on-white instead of white-on-white. I presume this issue was seen on a Windows machine.

#### How

Update the `normalize-button` mixin to set the paragraph text colour for the print stylesheet at the same point where the background is set to white.

Currently deployed on dev-lgfs

#### Before

Screenshot 2021-10-06 at 13 04 18

#### After

Screenshot 2021-10-06 at 13 39 07

### DIFF
--- a/app/webpack/stylesheets/_messages.scss
+++ b/app/webpack/stylesheets/_messages.scss
@@ -32,6 +32,10 @@
       border: 1px solid $black;
       background: none;
 
+      p {
+        color: $text-colour;
+      }
+
       &:after {
         border: none;
       }


naseberry (approved):
:

PR #4322, opened 2021-10-11
Load scheme date helpers

jrmhaig:
#### What

Load the `SchemeDateHelpers' module into Factory Bot.

#### Ticket

N/A

#### Why

The rake task to reload the sample data (`rails db:reload`) fails when using the defendants factory because the `scheme_date_for` is missing.


jsugarman (approved):
Can't think of a better place, with centralising it somewhere outside of the spec folder and requiring it. At least it shows the explicit requirement

PR #4329, opened 2021-10-13
Regenerate Rubocop todo list to update exceptions

jrmhaig:
#### What

Regenerate the `.rubocop_todo.yml` file

#### Ticket

N/A

#### Why

The `.rubocop_todo.yml` file lists Rubocop offences that are accepted, with the intention of fixing later. Several of these have been fixed and so can be removed from the list.

#### How

```bash
$ bundle exec rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 1000
...
1243 files inspected, 10782 offenses detected, 3011 offenses auto-correctable
```


jsugarman (approved):
👍

PR #4330, opened 2021-10-14
[CFP-284] Use correct caseworker name in MI report

jrmhaig:
#### What

In the MI report, use the name of the subject of the final transition of an incomplete journey as the caseworker's name.

#### Ticket

[MI discrepancy #256749](https://dsdmoj.atlassian.net/browse/CFP-284)

#### Why

The MI report includes lines for all 'journeys' of a claim. The journeys are, for example;

* Provider submits a claim, it is allocated to a case worker and the case worker completes the assessment as rejected, authorised, etc. One journey.
* The provider requests a redetermination or written reasons, the claim is allocated to a case worker and the case worker responds as appropriate. Another journey.

There can be multiple journeys for a single claim and the case worker listed in the MI report for each should be the case worker who is allocated the case. However, it was found that if the final journey has been allocated but not completed then the final case worker was appearing for *all* journeys.

#### How

When the journey is allocated but not completed then the name of the subject of the most recent transition for the claim is used.


app/presenters/concerns/management_information_reportable.rb
```diff
@@ -84,8 +84,9 @@ def rejection_reason
     end
 
     def case_worker
-      return claim.case_workers.first.name if claim.allocated?
-      COMPLETED_STATES.include?(@journey&.last&.to) ? @journey&.last&.author&.name : 'n/a'
+      return @journey.last.subject&.name if @journey.last.to == 'allocated'
```
jsugarman, line 87: this will work better for the new version of the MI report too: 👍

jsugarman (approved):
Looks good from a code and logic perspective. You will just need to check with the business and confirm (before and after the release) that this is what they need.

PR #4331, opened 2021-10-15
Update sidekiq gem

jrmhaig:
#### What

Update sidekiq gem.

#### Ticket

N/A

#### Why

See https://github.com/advisories/GHSA-grh7-935j-hg6w

There is a moderate severity vulnerability for sidekiq that is fixed in 6.2.1. I was expecting this to be picked up by dependabot on Wednesday but it wasn't.

Changelog; https://github.com/mperham/sidekiq/blob/main/Changes.md


jsugarman (approved):
👍

PR #4332, opened 2021-10-18
Remove dependabot sidekiq bump ignore

jsugarman:
#### What
Remove dependabot sidekiq bump ignore

#### Why
Reinstate dependabot bumps for sidekiq

This was transfered over during update from version 1
to version 2 of dependabot and was probably originally
due to problems with the sidekiq dependency at the time.


jrmhaig (approved):
:+1:

Can more be removed from this? Eg, we are not using New Relic.

PR #4337, opened 2021-10-18
Remove remaining dependabot ignores

jsugarman:
#### What
Remove remaining dependabot ignores

#### Why
Should be ok to remove these as we are either not using them
(newrelic_rpm, simplecov-console) or they may be ok to bump now.

We will need to be careful with govuk_frontend_toolkit as it is unmaintained
except for majro bugs.

Pretty sure we do not need sdoc and it may have been ignored for that reason.

parrallel_tests should be ok to bump but would need pulling locally and running
before merging as is not used in CI.


jrmhaig (approved):
:+1:

PR #4340, opened 2021-10-18
[CFP-299] MI report version 2

jsugarman:
#### What
Create version 2 of the MI report using aggregatable SQL query

#### Tickets

[CFP-247](https://dsdmoj.atlassian.net/browse/CFP-247) MI spike
[CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299) MI refactor
relates to [CFP-300](https://dsdmoj.atlassian.net/browse/CFP-299) MI weekly statistics

#### Why
1. to enable aggregations of the MI report
2. to improve performance of the MI report

#### TODO:

- [x] checks diffs of old and new versions for discrepancies - 0 diffs

#### Questions for caseworkers during catchup:

- [x]  can we use truncate the date for 6 month cut off. current MI report (v1) uses `Time.zone.now` to filter transitions which is sensitive to both the time the report started and time the claim is processed (it is a long time to process the claims). We should have it focused on any transitions for the day. This would be with a cut-off period of midnight 6 months earlier, where those at midnight are included, those at 11:59:59pm of previous day are not, to the second, taking daylight saving/BST into account. Caseworker (JW) and PM (JA) agrees.

- [x]  can we round the claim total to 2 decimal places. PM (JA) agrees

~~- [x]  can we get rid of the "af1/lf1 processed_by" field as this information is already available in the report as the caseworker who processed the previous redetermination (i.e. one row up). There could be issues in using the previous record due to the 6 month cut off though. i.e. previous record may be excluded due to date.~~

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -352,6 +352,7 @@ Metrics/ModuleLength:
     - 'app/helpers/application_helper.rb'
     - 'app/models/claims/calculations.rb'
     - 'app/models/claims/state_machine.rb'
+    - 'app/services/stats/management_information/concerns/journey_queryable.rb'
     - 'spec/models/fee/base_fee_spec.rb'
     - 'spec/support/validation_helpers.rb'
 
--- a/app/models/claim/base_claim.rb
+++ b/app/models/claim/base_claim.rb
@@ -247,6 +247,10 @@ module Claim
       will_save_change_to_case_transferred_from_another_court?
     end
 
+    def self.claim_types
+      agfs_claim_types | lgfs_claim_types
+    end
+
     def self.agfs_claim_types
       [Claim::AdvocateClaim,
        Claim::AdvocateInterimClaim,
--- a/app/presenters/concerns/management_information_reportable.rb
+++ b/app/presenters/concerns/management_information_reportable.rb
@@ -38,7 +38,7 @@ module ManagementInformationReportable
     end
 
     def claim_total
-      total_including_vat.to_s
+      format('%.2f', total_including_vat)
     end
 
     def submission_type
@@ -76,7 +76,7 @@ module ManagementInformationReportable
     def state_reason_code
       reason_code = @journey.last.reason_code
       reason_code = reason_code.flatten.join(', ') if reason_code.is_a?(Array)
-      reason_code
+      reason_code.presence
     end
 
     def rejection_reason
@@ -106,11 +106,12 @@ module ManagementInformationReportable
     end
 
     def af1_lf1_processed_by
-      previous_redetermination_steps.present? ? previous_redetermination_steps.last.author_name : ''
+      previous_redetermination_steps.present? ? previous_redetermination_steps.last.author_name : nil
     end
 
     def misc_fees
-      claim.misc_fees.map { |f| f.fee_type.description.tr(',', '') }.join(' ')
+      misc_fees = claim.misc_fees.map { |f| f.fee_type.description.tr(',', '') }.join(' ')
+      misc_fees.presence
     end
 
     def redetermination_steps
--- a/app/presenters/concerns/management_information_reportable.rb
+++ b/app/presenters/concerns/management_information_reportable.rb
@@ -38,7 +38,7 @@ module ManagementInformationReportable
     end
 
     def claim_total
-      format('%.2f', total_including_vat)
+      format('%.2f', total_including_vat.round(4))
     end
 
     def submission_type
--- a/app/presenters/concerns/management_information_reportable.rb
+++ b/app/presenters/concerns/management_information_reportable.rb
@@ -22,7 +22,7 @@ module ManagementInformationReportable
     end
 
     def case_type_name
-      case_type&.name || ''
+      case_type&.name
     end
 
     def bill_type
--- a/app/services/stats/management_information/concerns/claim_type_filterable.rb
+++ b/app/services/stats/management_information/concerns/claim_type_filterable.rb
@@ -9,9 +9,9 @@ module Stats
         delegate :claim_types, :agfs_claim_types, :lgfs_claim_types, to: :'::Claim::BaseClaim'
 
         def claim_type_filter
-          return in_statement_for(claim_types) if scheme.blank?
           return in_statement_for(agfs_claim_types) if scheme.eql?('AGFS')
-          in_statement_for(lgfs_claim_types)
+          return in_statement_for(lgfs_claim_types) if scheme.eql?('LGFS')
+          in_statement_for(claim_types)
         end
 
         def in_statement_for(arr)
--- a/app/services/stats/management_information/concerns/claim_type_filterable.rb
+++ b/app/services/stats/management_information/concerns/claim_type_filterable.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Stats
+  module ManagementInformation
+    module ClaimTypeQueryable
+      extend ActiveSupport::Concern
+
+      included do
+        delegate :claim_types, :agfs_claim_types, :lgfs_claim_types, to: :'::Claim::BaseClaim'
+
+        def claim_type_filter
+          return in_statement_for(claim_types) if scheme.blank?
+          return in_statement_for(agfs_claim_types) if scheme.eql?('AGFS')
+          in_statement_for(lgfs_claim_types)
+        end
+
+        def in_statement_for(arr)
+          arr.map(&:to_s).join('\', \'').prepend('(\'').concat('\')')
+        end
+      end
+    end
+  end
+end
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -95,7 +95,10 @@ module Stats
               c.*,
               c2.scheme as scheme,
               p.name as organisation,
-              ct.name as case_type_name,
+              coalesce(ct.name, (select ct2.name
+                                   from case_types ct2
+                                  where ct2.id = cs.case_type_id)
+              ) as case_type_name,
               c2.scheme || ' ' || c2.sub_type as bill_type,
               round(c.total + c.vat_amount, 2)::varchar as claim_total,
               main_defendant.name as main_defendant,
@@ -111,6 +114,8 @@ module Stats
               ON p.id = creator.provider_id
             LEFT OUTER JOIN case_types AS ct
               ON ct.id = c.case_type_id
+            LEFT OUTER JOIN case_stages cs
+              ON cs.id = c.case_stage_id
             LEFT JOIN LATERAL journeys(c.id) j
               ON TRUE
             LEFT JOIN LATERAL (
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -139,7 +139,7 @@ module Stats
               from representation_orders r, defendants d2
               where d2.claim_id = c.id
                 and r.defendant_id = d2.id
-              order by r.representation_order_date asc
+              order by r.representation_order_date, r.created_at asc
               fetch first row only
             ) earliest_representation_order ON TRUE
             LEFT JOIN LATERAL (
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -97,7 +97,7 @@ module Stats
               p.name as organisation,
               ct.name as case_type_name,
               c2.scheme || ' ' || c2.sub_type as bill_type,
-              (c.total + c.vat_amount)::varchar as claim_total,
+              round(c.total + c.vat_amount, 2)::varchar as claim_total,
               main_defendant.name as main_defendant,
               earliest_representation_order.maat_reference as maat_reference,
               earliest_representation_order.representation_order_date as rep_order_issued_date,
@@ -153,7 +153,7 @@ module Stats
               fetch first row only
             ) previous_redetermined_decision ON TRUE
             LEFT JOIN LATERAL (
-              select string_agg(ft.description, ' ') as descriptions
+              select string_agg(ft.description, ' ' order by f.id) as descriptions
               from fees f, fee_types ft
               where f.claim_id = c.id
                 and f.fee_type_id = ft.id
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -51,14 +51,7 @@ module Stats
                     on t.subject_id = subjects.id
                   where t.claim_id = in_claim_id
                   and t.to not in (select * from unnest(filtered_states))
-                  and t.created_at >= (now() - '6 months'::interval) at time zone 'utc' at time zone 'Europe/London'
-                  --
-                  -- postgres created_at is utc, now is utc
-                  -- to gain MI report v1 equivalent we need to compare created_at to now() Europe/London
-                  -- to do this more simply we need to change MI report v1 to truncate dates to midnight (see old version 1)
-                  -- and truncate these dates in this statement, as below
-                  -- "and DATE_TRUNC('day', t.created_at at time zone 'utc' at time zone 'Europe/London') >= DATE_TRUNC('day', (current_date - '6 months'::interval) at time zone 'utc' at time zone 'Europe/London')"
-                  --
+                  and DATE_TRUNC('day', t.created_at at time zone 'utc' at time zone 'Europe/London') >= DATE_TRUNC('day', (current_date - '6 months'::interval) at time zone 'utc' at time zone 'Europe/London')
                 )
                 select t.claim_id,
                        jsonb_agg(to_jsonb(t) order by t.created_at asc) as transitions
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -101,6 +101,7 @@ module Stats
               ) as case_type_name,
               c2.scheme || ' ' || c2.sub_type as bill_type,
               round(c.total + c.vat_amount, 2)::varchar as claim_total,
+              c.last_submitted_at at time zone 'utc' at time zone 'Europe/London' as last_submitted_at,
               main_defendant.name as main_defendant,
               earliest_representation_order.maat_reference as maat_reference,
               earliest_representation_order.representation_order_date as rep_order_issued_date,
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -51,7 +51,14 @@ module Stats
                     on t.subject_id = subjects.id
                   where t.claim_id = in_claim_id
                   and t.to not in (select * from unnest(filtered_states))
-                  and DATE_TRUNC('day', t.created_at) > DATE_TRUNC('day', current_date - '6 months'::interval)
+                  and t.created_at at time zone 'utc' at time zone 'Europe/London' >= (now() - '6 months'::interval) at time zone 'utc' at time zone 'Europe/London'
+                  --
+                  -- postgres created_at is utc, now is utc
+                  -- to gain MI report v1 equivalent we need to compare created_at Europe/London to now() Europe/London
+                  -- to do this more simply we need to change MI report v1 to truncate dates to midnight (see old version 1)
+                  -- and truncate these dates in this statement, as below
+                  -- "and DATE_TRUNC('day', t.created_at at time zone 'utc' at time zone 'Europe/London') >= DATE_TRUNC('day', (current_date - '6 months'::interval) at time zone 'utc' at time zone 'Europe/London')"
+                  --
                 )
                 select t.claim_id,
                        jsonb_agg(to_jsonb(t) order by t.created_at asc) as transitions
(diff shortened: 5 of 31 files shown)


app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,122 @@
...
+    class Presenter
+      # OPTIMIZE: share this array more centrally?!
+      # see also Claims::StateMachine::CASEWORKER_DASHBOARD_UNALLOCATED_STATES
+      SUBMITTED_STATES = %w[submitted redetermination awaiting_written_reasons].freeze
+
+      # OPTIMIZE: share this array with the SQL query itself?!
+      # see also Claims::StateMachine::CASEWORKER_DASHBOARD_COMPLETED_STATES
+      COMPLETED_STATES = %w[rejected refused authorised part_authorised].freeze
+
+      # OPTIMIZE: could instantiate a "record/claim_journey" struct from record hash for clearer calls?!
```
jsugarman: remove, using method missing instead

app/services/stats/stats_report_generator.rb
```diff
@@ -29,24 +48,12 @@ def call
     attr_reader :report_type, :options
 
     def validate_report_type
-      raise InvalidReportType unless StatsReport::TYPES.include?(report_type.to_s)
+      raise InvalidReportType unless StatsReport.names.include?(report_type.to_s)
```
jrmhaig: The `StatsReport.names` method was part of a commit that you didn't include here so this needs to remain as `StatsReport::TYPES`.

```suggestion
      raise InvalidReportType unless StatsReport::TYPES.include?(report_type.to_s)
```

Also, the `v2` versions of the reports need to be added to `StatsReport::TYPES`.

.rubocop_todo.yml
```diff
@@ -352,9 +352,11 @@ Metrics/ModuleLength:
     - 'app/helpers/application_helper.rb'
     - 'app/models/claims/calculations.rb'
     - 'app/models/claims/state_machine.rb'
+    - 'app/services/stats/management_information/concerns/journey_queryable.rb'
     - 'spec/models/fee/base_fee_spec.rb'
     - 'spec/support/validation_helpers.rb'
 
+
```
jsugarman: remove

app/models/claim/base_claim.rb
```diff
@@ -247,6 +247,10 @@ def case_transferred_from_another_court_changed?
       will_save_change_to_case_transferred_from_another_court?
     end
 
+    def self.claim_types
```
jsugarman, line 250: needs spec

app/services/stats/stats_report_generator.rb
jsugarman: done now

spec/factories/case_stages.rb
```diff
@@ -6,6 +6,15 @@
...
 
+    trait :agfs_pre_ptph do
+      description { 'Pre PTPH' }
+      unique_code { 'PREPTPH' }
+      position { 10 }
+      case_type_id { create(:case_type, :discontinuance).id }
+      roles { %w(agfs) }
+    end
+
+    # do not use obsolete
```
jsugarman:
```suggestion
    # obsolete - do not use
```

spec/services/stats/management_information/daily_report_query_spec.rb
```diff
@@ -0,0 +1,528 @@
...
+      end
+    end
+
+    describe ':bill_type' do
+      subject { response.pluck(:bill_type) }
+
+      context 'with advocate final claim' do
+        before { create(:advocate_final_claim, :submitted) }
+
+        it { is_expected.to match_array(['AGFS Final']) }
```
jsugarman: use `contain_exactly`

spec/services/stats/management_information/daily_report_query_spec.rb
```diff
@@ -0,0 +1,528 @@
...
+      end
+
+      context 'with advocate supplementary claim' do
+        before { create(:advocate_supplementary_claim, :submitted) }
+
+        it { is_expected.to match_array(['AGFS Supplementary']) }
```
jsugarman: use `contain_exactly`

spec/services/stats/management_information/daily_report_query_spec.rb
```diff
@@ -0,0 +1,528 @@
...
+      context 'with litigator final claim' do
+        before { create(:litigator_final_claim, :submitted) }
+
+        it { is_expected.to match_array(['LGFS Final']) }
+      end
+
+      context 'with litigator transfer claim' do
+        before { create(:litigator_transfer_claim, :submitted) }
+
+        it { is_expected.to match_array(['LGFS Transfer']) }
```
jsugarman: use `contain_exactly`

spec/services/stats/management_information/presenter_spec.rb
```diff
@@ -0,0 +1,302 @@
+# frozen_string_literal: true
+
+RSpec.describe Stats::ManagementInformation::Presenter do
+  subject(:presenter) { described_class.new(record) }
+
+  # OPTIMIZE: to use factory build instead of create and hash the claim object instead
+  # of querying the database. This would spead the tests up at the cost of being a less
```
jsugarman:
```suggestion
  # of querying the database. This would speed the tests up at the cost of being a less
```

app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+      def submission_type
+        journey.first[:to] == 'submitted' ? 'new' : journey.first[:to]
+      end
+
+      def claim_total
+        format('%.2f', record[:claim_total])
+      end
+
+      def transitioned_at
+        submissions.present? ? submissions.first[:created_at].strftime('%d/%m/%Y') : 'n/a'
```
jsugarman: try the optimization below using `find` instead of `first`

app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+        record[:last_submitted_at]&.strftime('%d/%m/%Y')
+      end
+
+      def originally_submitted_at
+        record[:originally_submitted_at]&.strftime('%d/%m/%Y')
+      end
+
+      # why `.last`, there should be only one allocation per journey?! see optimize below
+      def allocated_at
+        allocations.present? ? allocations.last[:created_at].strftime('%d/%m/%Y') : 'n/a'
```
jsugarman: try the optimization below using `find` instead of `first`

app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      # why `.first`, there should be only one completion per journey?! see optimize below
+      def completed_at
+        completions.present? ? completions.first[:created_at].strftime('%d/%m/%Y %H:%M') : 'n/a'
```
jsugarman: try the optimization below using `find` instead of `first`

app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      private
+
+      # OPTIMIZE: there should be only one submission per journey as each journey
+      # slice starts with a submission state (or possibly not if the submission was
+      # over 6 months ago). So could use `.find` to return first found and change name and
+      # callers to rely on it.
+      def submissions
+        @submissions ||= journey.select { |transition| SUBMITTED_STATES.include?(transition[:to]) }
```
jsugarman: try the optimization below using `find` instead of `select`

app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      # OPTIMIZE: there should be only one allocation per journey as deallocated
+      # allocations (and deallocations themselves) are removed. So could use `.find`
+      # to return first found and change name and callers to rely on it.
+      def allocations
+        @allocations ||= journey.select { |transition| transition[:to] == 'allocated' }
```
jsugarman: try the optimization below using `find` instead of `select`

app/services/stats/management_information/presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      # OPTIMIZE: there should be only one completion per journey as a journey
+      # is defined as a slice ending in a completion (or remainder, when it is the last
+      # journey). So could use `.find` to return first found and change name and
+      # callers to rely on it.
+      def completions
+        @completions ||= journey.select { |transition| COMPLETED_STATES.include?(transition[:to]) }
```
jsugarman: try the optimization below using `find` instead of `select`

app/services/stats/stats_report_generator.rb
```diff
@@ -2,6 +2,25 @@ module Stats
   class StatsReportGenerator
     class InvalidReportType < StandardError; end
 
+    # rubocop:disable Metrics/MethodLength
+    def self.for(report_type)
+      Hash.new({ class: ReportGenerator, args: [report_type] }).merge(
```
jsugarman: default
```suggestion
      Hash.new({ class: ReportGenerator, default_args: [report_type] }).merge(
```

app/services/stats/stats_report_generator.rb
```diff
@@ -33,20 +52,8 @@ def validate_report_type
...
-      case report_type.to_sym
-      when :agfs_management_information
-        options[:claim_scope] = :agfs
-      when :lgfs_management_information
-        options[:claim_scope] = :lgfs
-      end
-
-      ManagementInformationGenerator.call(options)
+      generator = self.class.for(report_type)
+      generator[:class].call(*generator[:args])
```
jsugarman:
```suggestion
      generator[:class].call(*generator[:default_args])
```

app/services/stats/management_information/concerns/claim_type_filterable.rb
```diff
@@ -0,0 +1,23 @@
...
+    module ClaimTypeQueryable
+      extend ActiveSupport::Concern
+
+      included do
+        delegate :claim_types, :agfs_claim_types, :lgfs_claim_types, to: :'::Claim::BaseClaim'
+
+        def claim_type_filter
+          return in_statement_for(claim_types) if scheme.blank?
+          return in_statement_for(agfs_claim_types) if scheme.eql?('AGFS')
+          in_statement_for(lgfs_claim_types)
```
jrmhaig: (nit) It might make more sense for `claim_types` to be the 'catch all' option, so;

```ruby
          return in_statement_for(lgfs_claim_types) if scheme.eql?('LGFS')
          return in_statement_for(agfs_claim_types) if scheme.eql?('AGFS')
          in_statement_for(claim_types)
```

app/services/stats/management_information/concerns/claim_type_filterable.rb
```diff
@@ -0,0 +1,23 @@
...
+        end
+
+        def in_statement_for(arr)
+          arr.map(&:to_s).join('\', \'').prepend('(\'').concat('\')')
```
jrmhaig, line 18: Maybe I'm being over paranoid (I know that these values are not user inputs) but this is being inserted into an SQL query. Can this be;

```ruby
arr.map { |type| ActiveRecord::Base.connection.quote(type.to_s) }.join(', ').prepend('(').concat(')')
```

app/presenters/concerns/management_information_reportable.rb
```diff
@@ -106,11 +106,12 @@ def rep_order_issued_date
...
 
     def af1_lf1_processed_by
-      previous_redetermination_steps.present? ? previous_redetermination_steps.last.author_name : ''
+      previous_redetermination_steps.present? ? previous_redetermination_steps.last.author_name : nil
     end
 
     def misc_fees
-      claim.misc_fees.map { |f| f.fee_type.description.tr(',', '') }.join(' ')
+      misc_fees = claim.misc_fees.map { |f| f.fee_type.description.tr(',', '') }.join(' ')
+      misc_fees.presence
```
jrmhaig, line 114: I think this would also work, avoiding the creation of a temporary variable;

```ruby
claim.misc_fees.map { |f| f.fee_type.description.tr(',', '') }.join(' ').presence
```

spec/presenters/management_information_presenter_spec.rb
```diff
@@ -102,8 +102,8 @@ def debug_transitions_for(claim)
 
       it 'total (inc VAT)' do
         presenter.present! do |claim_journeys|
-          expect(claim_journeys.first).to include(claim.total_including_vat.to_s)
-          expect(claim_journeys.second).to include(claim.total_including_vat.to_s)
+          expect(claim_journeys.first).to include(format('%.2f', claim.total_including_vat))
```
jrmhaig, line 105: (nit) Multiple exceptions in a single example (see [RSpec/MultipleExpectations](https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/MultipleExpectations)).

(This is pre-existing and occurs various places elsewhere in the file, as well as multiple other Rubocop offences)

app/services/stats/management_information/concerns/journey_queryable.rb
```diff
@@ -39,7 +39,7 @@ def create_journeys_func
                   select t.claim_id,
                          t.from,
                          t.to,
-                         t.created_at,
+                         t.created_at at time zone 'utc' at time zone 'Europe/London' as created_at,
```
jrmhaig, line 42: :+1: Opting to display 'local' time instead of UTC.

spec/factories/case_stages.rb
```diff
@@ -14,6 +23,7 @@
       roles { %w(agfs) }
     end
 
+    # obsolete - do not use
```
jrmhaig: I cannot find this trait used in any tests. Can it be removed?

spec/factories/case_stages.rb
```diff
@@ -78,6 +88,7 @@
...
```
jrmhaig: I cannot find this trait used in any tests. Can it be removed?

spec/factories/case_stages.rb
```diff
@@ -86,6 +97,7 @@
       roles { %w(lgfs) }
     end
 
+    # obsolete - do not use
```
jrmhaig: I have found a single use of this trait in `spec/services/cleaners/litigator_hardship_claim_cleaner_spec.rb` (line 30). Perhaps that instance can be changed to `agfs_pre_ptph` and this this trait removed.

spec/services/stats/management_information/daily_report_query_spec.rb
```diff
@@ -114,21 +114,25 @@
       subject { response.pluck(:case_type_name) }
 
       context 'with a claim without a case type' do
-        let!(:claim) { create(:litigator_transfer_claim, :submitted) }
+        before { create(:litigator_transfer_claim, :submitted) }
```
jrmhaig, line 117: :+1:

spec/services/stats/management_information/daily_report_query_spec.rb
```diff
@@ -392,35 +392,48 @@
...
+        create(:advocate_final_claim, :allocated).tap do |claim|
+          claim.deallocate!
+          claim.allocate!
+          claim.deallocate!
+          claim.allocate!
+        end
+
+        expect(journey_tos).to eq([%w[submitted allocated]])
+      end
+      # rubocop:enable RSpec/ExampleLength
```
jrmhaig, line 436: The long example length can be avoided with;

```ruby
    context 'with a claim allocated and deallocated twice' do
      before do
        create(:advocate_final_claim, :allocated).tap do |claim|
          claim.deallocate!
          claim.allocate!
          claim.deallocate!
        end
      end

      it { expect(journey_tos).to eq([%w[submitted]]) }
    end

    context 'with a claim allocated and deallocated twice and then allocated' do
      before do
        create(:advocate_final_claim, :allocated).tap do |claim|
          claim.deallocate!
          claim.allocate!
          claim.deallocate!
          claim.allocate!
        end

        it { expect(journey_tos).to eq([%w[submitted allocated]]) }
      end
    end
```

jrmhaig (commented):
A few comments from looking over the commits individually. I'll have another look over the whole diff.

I have only briefly looked over the SQL.

app/services/stats/management_information/concerns/claim_type_filterable.rb
jsugarman: 👍 i did this on the sql_query branch actually separate but will apply here too

app/services/stats/management_information/concerns/claim_type_filterable.rb
jsugarman: will have a look at what can be done, but i note this was taken from another place where it is performaed as part of an API call for searching....where sql injection is more likely to be a problem - search for `in_statement_for`

app/presenters/concerns/management_information_reportable.rb
jsugarman: it would, but i prefer to break up the line for length and clarity

spec/presenters/management_information_presenter_spec.rb
jsugarman: we will also be getting rid of this file in the near future 🤞

spec/factories/case_stages.rb
jsugarman: 👍  yes, i will remove the ones that are not used. There are some obsoletes that are used which I would be inclined to leave alone. Incidentally the db, on prod too, has case stages with unique codes marked as obsolete - which roughly reflects these traits but which need cleaning up too. - search for `obsolete|OBSOLETE` generally in the code base

spec/factories/case_stages.rb
jsugarman: see previous related comment

spec/factories/case_stages.rb
jsugarman: thats the one i found, will  have a look to see if it can be changed

spec/services/stats/management_information/daily_report_query_spec.rb
jsugarman: i was trying to avoid using contexts too much as it looks neater in rspec output, but will have another look

jsugarman: also, given complexity of the system i'd be tempted to bump the limit. These kind of rules are guides rather than must haves IMO

jsugarman:
> A few comments from looking over the commits individually. I'll have another look over the whole diff.
> 
> I have only briefly looked over the SQL.

to do a diff you can use dev-lgfs BUT the main report for v1 or v2 is cannot be regenerated via the interface due to memory consumption problems, the scheduled job for v2 also fails for similar reasons i imagine - because it is holding more in memory at one time than v1 I guess. I will looking at bumping the memory temporarily but for now you can use the separate AGFS and LGFS version.

app/services/stats/management_information/concerns/claim_type_filterable.rb
jsugarman: These result in the same output, so would inclined to keep the current  briefer version

```
arr.map { |type| ActiveRecord::Base.connection.quote(type.to_s) }.join(', ').prepend('(').concat(')')
=> "('Claim::AdvocateClaim', 'Claim::AdvocateInterimClaim', 'Claim::AdvocateSupplementaryClaim', 'Claim::AdvocateHardshipClaim')"
arr.map(&:to_s).join('\', \'').prepend('(\'').concat('\')')
=> "('Claim::AdvocateClaim', 'Claim::AdvocateInterimClaim', 'Claim::AdvocateSupplementaryClaim', 'Claim::AdvocateHardshipClaim')"
```

This line is what #quote ends up doing i believe https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L219 which is effectively what the `.join('\', \'')` does anyway.

spec/services/stats/management_information/daily_report_query_spec.rb
jsugarman: rspec output from current is cleaner so would prefer to keep
```
    :journey
      excludes state transitions to archived_pending_review
      excludes state transitions to archived_pending_delete
      excludes deallocations and deallocated allocations
      excludes deallocations and all but last allocation
      excludes state transitions to draft
```

spec/factories/case_stages.rb
jsugarman: removed 1 that no longer exists and another that has no claims with it on prod. The remaining two would need dealing with in a separate PR as functionality needs to change to be able to remove the tests and the traits they rely on. One of which has 5 claims still on prod in `archived_pending_review` state.

app/services/stats/management_information/concerns/claim_type_filterable.rb
jsugarman: or, even briefer
```ruby
 "(\'#{arr.map(&:to_s).join('\', \'')}\')"
```

jrmhaig: OK. We need to be careful if this is ever repurposed for another search with a different `arr`.

You could avoid an arbitrary list being passed to `in_statement_for` by making it private and updating `claim_type_filter` to take an optional argument like; 

```ruby
def claim_type_filter(types: scheme)
  return in_statement_for(agfs_claim_types) if types.eql?('AGFS')
  return in_statement_for(lgfs_claim_types) if types.eql?('LGFS')
  in_statement_for(claim_types)
end
```

If you decide not to go with this I am happy to resolve the conversation and approve.

jrmhaig (approved):

PR #4377, opened 2021-11-01
[CFP-300] MI statistics reports

jsugarman:
#### What
New reports for "Management information statistics". These are counts that match the required outputs (see attachment to ticket)

#### Ticket

[CFP-300](https://dsdmoj.atlassian.net/browse/CFP-300)
relates to:
  [CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299)
  [CFP-247](https://dsdmoj.atlassian.net/browse/CFP-247)

#### Why
Automates the manual aggregations that caseworkers currently perform on a weekly basis. They want a months
worth to cover the differences in the days that they sum the figures for inputs compared to outputs:

1. inputs are counts based filtering by "submitted at"
2. outputs are counts based on filtering by "completed at"

#### How
Use SQL to return a **single** set of data for the MI report. This single query can then be used in Common table expressions (CTEs) to filter and aggregate the set to meet various user defined requirements for statistics.

--------

#### TODO (wip)

 - [x] design details needed off the back of user research and business analysis
 - [x] change to generating a months worth from start date
 - [x] Generate AGFS monthly statistics based on originally submitted date
 - [x] Generate LGFS monthly statistics based on originally submitted date
 - [x] Generate AGFS monthly statistics based on completed at date
 - [x] Generate LGFS monthly statistics based on completed at date
 - [x] spec all count queries
 - [x] enable user to supply date for month
 - [ ] "Rollup" Saturday and Sunday into Monday's counts
 - [X] optimization: do not create journeys function unless necessary - see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4459
 - [x] optimization: use group by
 - [ ] ~~optimization: share AGFS/LGFS query logic?!~~


lib/tasks/claims.rake
```diff
@@ -66,6 +66,17 @@ namespace :claims do
...
+  task :demo_data_without_seed, :num_claims_per_state, :num_external_users do |_task, args|
+    params = {num_claims_per_state: 1, num_external_users: 1}.merge(args)
+    Rake::Task['claims:sample_users'].invoke
+    Rake::Task["claims:demo_data:advocates"].invoke(params[:num_claims_per_state], params[:num_external_users])
+    # Rake::Task["claims:demo_data:advocate_interims"].invoke(params[:num_claims_per_state], params[:num_external_users]) # TODO
+    Rake::Task['claims:demo_data:litigators'].invoke(params[:num_claims_per_state], params[:num_external_users])
+    Rake::Task['claims:demo_data:interims'].invoke(params[:num_claims_per_state], params[:num_external_users])
+    Rake::Task['claims:demo_data:transfers'].invoke(params[:num_claims_per_state], params[:num_external_users])
+  end
+
```
jsugarman: remove this is now the demo_data tasks behaviour, after a rebase

.rubocop_todo.yml
```diff
@@ -355,6 +355,7 @@ Metrics/ModuleLength:
     - 'spec/models/fee/base_fee_spec.rb'
     - 'spec/support/validation_helpers.rb'
 
+
```
jrmhaig: (nit) This probably slipped in accidentally.

app/controllers/case_workers/admin/management_information_controller.rb
```diff
@@ -30,13 +31,32 @@ def generate
...
 
+  def report_params
+    params.permit(
+      :report_type,
+      :day
+    )
+  end
+
+  def validate_and_set_date
+    @day ||= Date.iso8601("#{report_params['day(1i)']}-#{report_params['day(2i)']}-#{report_params['day(3i)']}")
```
jrmhaig: Is there a reason not to use

```ruby
@day ||= Date.new(report_params['day(1i)'].to_i, report_params['day(2i)'].to_i, report_params['day(3i)'].to_i)
# Or something like
# @day ||= Date.new(report_params.slice('day(1i)', 'day(2i)', 'day(3i)').map(&:to_i))
```

Note that the year needs to be a 4-digit year, although I think that is already the case with the `Date.iso8601` version.

app/models/stats/stats_report.rb
```diff
@@ -1,14 +1,30 @@
...
-               agfs_management_information
-               lgfs_management_information
-               management_information_v2
-               agfs_management_information_v2
-               lgfs_management_information_v2
-               provisional_assessment
-               rejections_refusals
-               submitted_claims].freeze
+    Report = Struct.new(:name, :date_required, keyword_init: true) do
+      def initialize(name:, date_required: false)
```
jrmhaig, line 4: For possible future extensions a list of required fields could be passed here, and then the `date_required?` method updated accordingly.

app/models/stats/stats_report.rb
```diff
@@ -1,14 +1,30 @@
...
+        super
+      end
+
+      def date_required?
+        date_required
+      end
+
+      def to_s
+        name
+      end
```
jrmhaig: The `name` could be given as a symbol and (arguably) it may be beneficial to do so. In which case `to_s` would not return a string. To make sure this works correctly,

```suggestion
      def to_s
        name.to_s
      end
```

app/models/stats/stats_report.rb
```diff
@@ -1,14 +1,30 @@
...
+               Report.new(name: 'agfs_management_information'),
+               Report.new(name: 'lgfs_management_information'),
+               Report.new(name: 'management_information_v2'),
+               Report.new(name: 'agfs_management_information_v2'),
+               Report.new(name: 'lgfs_management_information_v2'),
+               Report.new(name: 'agfs_management_information_weekly_statistics', date_required: true),
+               Report.new(name: 'lgfs_management_information_weekly_statistics', date_required: true),
+               Report.new(name: 'provisional_assessment'),
+               Report.new(name: 'rejections_refusals'),
+               Report.new(name: 'submitted_claims')].freeze
```
jrmhaig: (nit) As noted with relation to the `Report#to_s` method, maybe the names could be given a symbols instead of strings. This may have knock-on effects if there are places where a string and not a symbol is required but I don't think I see any.

app/services/stats/management_information/daily_report_count_query.rb
```diff
@@ -0,0 +1,54 @@
...
+      def filter_by(date_column_filter)
+        @query_set.each_with_object([]) do |(name, query), results|
+          result = { name: name.to_s.humanize, filter: date_column_filter.to_s.humanize }
+          @date_range.each do |day|
+            result[day.iso8601] = query.call(day: day.iso8601,
+                                             date_column_filter: date_column_filter).first['count']
+          end
+          results.append(result)
+        end
+      end
```
jrmhaig: This `each_with_object` looks like a `map`, so a possible simplification (not at all tested) could be;

```ruby
      def filter_by(date_column_filter)
        @query_set.map do |(name, query)|
          {
            name: name.to_s.humanize,
            filter: date_column_filter.to_s.humanize,
          }.tap do |result|
            @date_range.each do |day|
              result[day.iso8601] = query.call(day: day.iso8601,
                                                 date_column_filter: date_column_filter).first['count']
            end
          end
          # Or, instead of the `tap`:
          # }.merge(query_over_date_range(query, date_column_filter))
        end
      end
```

Another alternative has a map for the `@query_set` too;

```ruby
      def filter_by(date_column_filter)
        @query_set.map do |(name, query)|
          @date_range.map(&:iso8601).map do |day|
            [
              day,
              query.call(day: day, date_column_filter: date_column_filter).first['count']
            ]
          end.to_h.merge(name: name.to_s.humanize, filter: date_column_filter.to_s.humanize)
        end
      end
```

spec/jobs/stats_report_generation_job_spec.rb
```diff
@@ -4,13 +4,15 @@
   subject(:job) { described_class.new }
 
   describe '#perform' do
-    subject(:perform) { job.perform(report_type) }
+    subject(:perform) { job.perform('any_old_report_name', my_option: 1) }
 
-    let(:report_type) { 'provisional_assessment' }
+    before do
+      allow(Stats::StatsReportGenerator).to receive(:call).with(instance_of(String), instance_of(Hash))
```
jrmhaig: (nit) I don't think you need the `with` part on the `allow`, unless there are different spies depending on the arguments.

spec/requests/case_workers/admin/management_information_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+    let(:report_type) { 'management_information' }
+
+    it 'returns http redirect status' do
+      subject
+      expect(response).to have_http_status(:redirect)
+    end
+
+    it 'redirects to users root url' do
+      is_expected.to redirect_to(case_workers_root_url)
+    end
```
jrmhaig, line 15: The second test (redirects to users root url) includes testing that the response is a redirect so the first test is not required. Likewise below.

spec/requests/case_workers/admin/management_information_spec.rb
```diff
@@ -0,0 +1,240 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples 'case worker not authorized' do
```
jrmhaig, line 3: These two shared examples are always used together. Maybe there could be a single `shared_examples` with two contexts.

app/services/stats/management_information/daily_report_count_generator.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Stats
+  module ManagementInformation
+    class DailyReportCountGenerator
```
jrmhaig, line 7: It may be possible to share code between this and app/services/stats/management_information/daily_report_generator.rb (and others?) either by a mixin or a base class (or both).

app/services/stats/management_information_generator.rb
```diff
@@ -2,15 +2,15 @@
...
+    def self.call(**kwargs)
+      new(kwargs).call
     end
 
-    def initialize(options = {})
-      @format = options.fetch(:format, DEFAULT_FORMAT)
-      @scheme = options.fetch(:scheme, :all)
+    def initialize(**kwargs)
+      @format = kwargs.fetch(:format, :csv)
+      @scheme = kwargs.fetch(:scheme, :all)
```
jrmhaig, line 13: If there are only two arguments, each with a default value, perhaps it makes more sense to have

```suggestion
    def initialize(format: :csv, scheme: :all)
      @format = format
      @scheme = scheme
```

app/services/stats/stats_report_generator.rb
```diff
@@ -4,30 +4,43 @@ class InvalidReportType < StandardError; end
...
+      new(kwargs).call
     end
 
-    def initialize(report_type, options = {})
-      @report_type = report_type
-      @options = options
+    def initialize(**kwargs)
+      @report_type = kwargs[:report_type]
+      @generator = self.class.for(report_type)
+      @options = kwargs
```
jrmhaig, line 43: Does `@options' need to include the report type? If not,

```suggestion
    def initialize(report_type:, **kwargs)
      @report_type = report_type
      @generator = self.class.for(report_type)
      @options = kwargs
```

(This will also make `report_type` a required argument)

app/services/stats/stats_report_generator.rb
```diff
@@ -45,15 +58,14 @@ def call
...
 
     def validate_report_type
-      raise InvalidReportType unless StatsReport::TYPES.include?(report_type.to_s)
+      raise InvalidReportType unless StatsReport.names.include?(report_type.to_s)
     end
 
     def generate_new_report
-      generator = self.class.for(report_type)
-      generator[:class].call(*generator[:default_args])
+      generator[:class].call(options.merge(*generator[:default_args]))
```
jrmhaig: The default args defined in `self.for` do not need to be inside an array. This this line becomes;

```suggestion
      generator[:class].call(options.merge(generator[:default_args]))
```

The empty `default_args` for two of the reports can be `{}`. I think this caused problems before the `merge` was added.

app/services/stats/management_information/daily_report_count_generator.rb
```diff
@@ -14,7 +14,7 @@ def self.call(**kwargs)
       def initialize(**kwargs)
         @scheme = kwargs[:scheme]&.to_s&.upcase
         @day = kwargs[:day]
-        raise ArgumentError, 'scheme must be "agfs" or "lgfs"' if @scheme.present? && %w[AGFS LGFS].exclude?(@scheme)
+        raise ArgumentError, 'scheme must be "agfs" or "lgfs"' if %w[AGFS LGFS].exclude?(@scheme)
```
jrmhaig: I think this change means that combined agfs & lgfs reports cannot be generated. Is this intentional?

If this is correct then the safe navigators two lines above are not required as `nil.to_s` will successfully return an empty string, and it no longer matters in the query whether `@scheme` is `nil` or `''`.

app/services/stats/management_information/daily_report_count_query.rb
```diff
@@ -1,40 +1,27 @@
...
 
-      def initialize(options = {})
-        @scheme = options[:scheme]&.to_s&.upcase
-        @date_range = options[:date_range]
-        raise ArgumentError, 'scheme must be "agfs" or "lgfs"' if @scheme.present? && %w[AGFS LGFS].exclude?(@scheme)
-        raise ArgumentError, 'date range must be supplied' if @date_range.blank?
+      def initialize(**kwargs)
+        @scheme = kwargs[:scheme]&.to_s&.upcase
+        @date_range = kwargs[:date_range]
+        raise ArgumentError, 'scheme must be "agfs" or "lgfs"' if %w[AGFS LGFS].exclude?(@scheme)
```
jrmhaig: This is being checked in `DailyReportCountGenerator`. Does it need to be done again here? (Is it the intention that it will be called from somewhere else?)

The `.to_s.upcase` of `@scheme` has also already been done in `DailyReportCountGenerator`.

app/services/stats/management_information/daily_report_count_generator.rb
```diff
@@ -39,22 +39,27 @@ def generate_csv
...
 
-      def week_range
-        @day.beginning_of_week(:saturday)..@day.end_of_week(:saturday)
+      def row_for(rec)
+        date_range.map(&:iso8601).map { |date| rec[date] }.prepend(rec[:name])
+      end
+
+      # TODO: make interval an arg with default of 1.month
+      def date_range
+        @day..(@day + 1.month)
```
jrmhaig: `week_range` was Saturday to Friday (exactly 7 days). This is now from `@day` to the same day in the next month, which is one calendar month plus one day. I think you might want `@day + 1.month - 1.day`.

app/services/stats/management_information/concerns/journey_queryable.rb
```diff
@@ -100,6 +100,7 @@ def journeys_query
               round(c.total + c.vat_amount, 4) as claim_total,
               c.last_submitted_at at time zone 'utc' at time zone 'Europe/London' as last_submitted_at,
               c.original_submission_date at time zone 'utc' at time zone 'Europe/London' as originally_submitted_at,
```
jrmhaig, line 102: To avoid confusion, can `originally_submitted_at` be called something based on the original column name. E.g. `original_submission_date_local` or `original_submission_date_london`.

app/services/stats/management_information/daily_report_count_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+
+        keys.map do |key|
+          Date.parse(key).strftime("%d/%m/%Y\n%A")
+        rescue Date::Error
+          key
+        end
+      end
+
+      def date_range
+        @start_at..(@start_at + @duration)
```
jrmhaig, line 63: I made a comment on this on a previous commit but this line has been changed so that might have disappeared.

If you have 'today' up to 'today plus 1 month (eg)' you will get a calendar month plus one day. To have a time period of the correct duration the last day needs to be removed. However, check whether the query is using inclusive or exclusive ranges.

```suggestion
        @start_at..(@start_at + @duration - 1.day)
```

spec/services/stats/management_information/daily_report_count_generator_spec.rb
```diff
@@ -52,6 +52,10 @@
       let(:kwargs) { { day: start_date, scheme: 'agfs' } }
       let(:start_date) { 1.month.ago.to_date }
 
+      let(:expected_headers) do
+        (start_date..(start_date + 1.month)).to_a.map { |d| d.strftime("%d/%m/%Y\n%A") }.prepend('Name')
```
jrmhaig: I think this might want to be `start_date + 1.month - 1.day`.

app/services/stats/management_information/queries/agfs/af1_high_value_query.rb
```diff
@@ -0,0 +1,37 @@
...
+            WITH journeys AS (
+              #{journeys_query}
+            )
+            SELECT count(*)
+            FROM journeys j
+            WHERE j.scheme = 'AGFS'
+            AND j.journey -> 0 ->> 'to' = 'submitted'
+            AND date_trunc('day', j.#{@date_column_filter}) = '#{@day}'
+            AND NOT j.disk_evidence
+            AND j.claim_total::float >= 20000.00
```
jrmhaig: (nit) The comment at the top says `> 20,000` but this is checking `>= 20000.00`. My guess is that the comment should be updated as there are other cased that are `< 20000.00`.

app/services/stats/management_information/queries/agfs/af2_high_value_query.rb
```diff
@@ -0,0 +1,36 @@
...
+            WITH journeys AS (
+              #{journeys_query}
+            )
+            SELECT count(*)
+            FROM journeys j
+            WHERE j.scheme = 'AGFS'
+            AND j.journey -> 0 ->> 'to' = 'redetermination'
+            AND date_trunc('day', j.#{@date_column_filter}) = '#{@day}'
+            AND NOT j.disk_evidence
+            AND j.claim_total::float >= 20000.00
```
jrmhaig: See comment in `app/services/stats/management_information/queries/agfs/af1_high_value_query.rb`

app/services/stats/management_information/queries/base_count_query.rb
```diff
@@ -0,0 +1,42 @@
...
+        ActiveRecord::Base.connection.execute(create_journeys_func)
+      end
+
+      def sql_quote(column_name)
+        ActiveRecord::Base.connection.quote_column_name(column_name)
+      end
+
+      def query
+        raise RuntimeError
+      end
```
jrmhaig, line 39: I'm not sure this is required. If this method is missing in a child then a `NoMethodError` exception will be raise, which gives more information than `RuntimeError`.

app/services/stats/management_information/queries/lgfs/lf1_high_value_query.rb
```diff
@@ -0,0 +1,37 @@
...
+            WITH journeys AS (
+              #{journeys_query}
+            )
+            SELECT count(*)
+            FROM journeys j
+            WHERE j.scheme = 'LGFS'
+            AND date_trunc('day', j.#{@date_column_filter}) = '#{@day}'
+            AND j.journey -> 0 ->> 'to' = 'submitted'
+            AND NOT j.disk_evidence
+            AND j.claim_total::float >= 20000.00
```
jrmhaig: See comment in `app/services/stats/management_information/queries/agfs/af1_high_value_query.rb`

app/services/stats/management_information/queries/lgfs/lf2_high_value_query.rb
```diff
@@ -0,0 +1,37 @@
...
+            WITH journeys AS (
+              #{journeys_query}
+            )
+            SELECT count(*)
+            FROM journeys j
+            WHERE j.scheme = 'LGFS'
+            AND j.journey -> 0 ->> 'to' = 'redetermination'
+            AND date_trunc('day', j.#{@date_column_filter}) = '#{@day}'
+            AND NOT j.disk_evidence
+            AND j.claim_total::float >= 20000.00
```
jrmhaig: See comment in `app/services/stats/management_information/queries/agfs/af1_high_value_query.rb`

spec/factories/case_types.rb
```diff
@@ -156,6 +156,10 @@
       lgfs_roles
     end
 
+    trait :hearing_subsequent_to_sentence do
```
jrmhaig, line 159: :+1: clearer trait name.

Maybe replace all instances of `:hsts` and then remove that trait?

app/views/case_workers/admin/management_information/_stats_report_with_date_form.html.haml
```diff
@@ -0,0 +1,22 @@
+- i18n_scope = 'case_workers.admin.management_information.index'
+
+- if report_rec&.started_at
+  = t('report_generated_at_html',
+      scope: i18n_scope,
+      datetime: report_rec.started_at.iso8601,
+      time: report_rec.started_at.strftime(Settings.report_date_format))
+
+  %p
```
naseberry, line 4: empty `<p>` tag

app/views/case_workers/admin/management_information/_stats_report_with_date_form.html.haml
```diff
@@ -0,0 +1,22 @@
...
+  = govuk_link_button(t('download', scope: i18n_scope),
+                      case_workers_admin_management_information_download_url(report_type: report_object.name, format: :csv))
+- else
+  = t('unavailable_report', scope: i18n_scope)
+  %p
```
naseberry, line 13: empty `<p>` tag

app/views/case_workers/admin/management_information/_stats_report.html.haml
```diff
@@ -0,0 +1,17 @@
...
+
+- if report_rec&.started_at
+  = t('report_generated_at_html',
+      scope: i18n_scope,
+      datetime: report_rec.started_at.iso8601,
+      time: report_rec.started_at.strftime(Settings.report_date_format))
+- else
+  = t('unavailable_report', scope: i18n_scope)
+
+%p
```
naseberry: empty `<p>` tag

app/services/stats/management_information/daily_report_count_generator.rb
jrmhaig: This is no longer relevant.

app/services/stats/management_information/daily_report_count_query.rb
jrmhaig: This is no longer relevant.

app/controllers/case_workers/admin/management_information_controller.rb
jsugarman: There are options and here and Date.new is one (also looked Date.civil). I ended up with this as this is the format that that needs sending to the SQL query. In the end as long as it Validates the date and raises an error when invalid then it should be fine.


the above method  requires leading zeros, so not ideal. `Date.new` as above would allow negative numbers, which is not ideal either.

Will have a play and test

app/models/stats/stats_report.rb
jsugarman: example?

app/services/stats/management_information/queries/base_count_query.rb
jsugarman: i believe if the method is missing in the child/subclass then the superclasses method will be called, no?

spec/factories/case_types.rb
jsugarman: yak-shaving but could do

app/views/case_workers/admin/management_information/_stats_report_with_date_form.html.haml
jsugarman: should probably be a `<br≥`? There were spacing issues without this i believe

app/views/case_workers/admin/management_information/_stats_report.html.haml
jsugarman: should probably be a <br≥? There were spacing issues without this i believe

app/services/stats/management_information/concerns/journey_queryable.rb
jsugarman: These names were already applied as part of the previous refactoring work. They were named after what their csv header column is to avoid confusion in that context. I can't say I am keen on naming them after the timezone either.

app/services/stats/management_information/daily_report_count_generator.rb
jsugarman: I was aware of this and that is correct. However, the default duration is open for discussion generally and from a user perspective may be clearer to give them 1.month inclusive - they would only use a given weeks in any event. I think to improve speed we should we should see if they agree to two weeks as default anyway.

app/services/stats/management_information/daily_report_count_query.rb
jsugarman: Will take another look - the second is good but i find the play of concise versus clear to fall in each_with_object's favour. also looking to change this whole query to use a group by "date" if I have time.

```
        @query_set.each_with_object([]) do |(name, query), results|
          result = { name: name.to_s.humanize, filter: date_column_filter.to_s.humanize }
          @date_range.map(&:iso8601).each do |day|
              result[day] = query.call(day: day, date_column_filter: date_column_filter).first['count']
          end
          results.append(result)
        end
```

app/services/stats/management_information/queries/agfs/af1_high_value_query.rb
jsugarman: The business document from which that is a copy/paste shows they want one filter for > 20k and one for < 20k, so = 20k never gets looked for. in reality this is of negligible importance but I amended to include 20k in the  above lot for completeness.

we can get rid of these comments entirely unless you think they add some value?

.rubocop_todo.yml
jsugarman: done

app/models/stats/stats_report.rb
jsugarman: @jrmhaig ^^

app/models/stats/stats_report.rb
jsugarman: done via delegation now

app/models/stats/stats_report.rb
jsugarman: is a bit of a yak-shave with costs in other ways (as used by controller, which always returns strings) but have done this here now on commit 73257b52a96bdc1b35663b0d64cc1f83889b0cca

app/views/case_workers/admin/management_information/_stats_report_with_date_form.html.haml
jsugarman: have added a fix here 8016efdcf

app/views/case_workers/admin/management_information/_stats_report.html.haml
jsugarman: fix here 8016efdcf

spec/services/stats/management_information/daily_report_count_generator_spec.rb
jsugarman: as https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4377#discussion_r757494540

spec/requests/case_workers/admin/management_information_spec.rb
jsugarman: I prefer more "modular". shared examples can get too big and are difficult to debug as well

spec/requests/case_workers/admin/management_information_spec.rb
jsugarman: i am aware but you can redirect without a redirect status i believe. This being the test differences

app/models/stats/stats_report.rb
jrmhaig: It was just a thought that a particular report may have options for (e.g.) date range (weekly, monthly) or some other filters. On second thoughts, this probably isn't worth changing now.

app/services/stats/management_information/daily_report_count_query.rb
jsugarman: have amended here f483625fe

app/services/stats/management_information_generator.rb
jsugarman: This was just a name change of argument on a class (for consistencies sake) that will hopefully be trashed soon in any event. lets not sweat the small stuff!

app/controllers/case_workers/admin/management_information_controller.rb
jrmhaig: > I ended up with this as this is the format that that needs sending to the SQL query.

Unless I am missing something,

```
Date.iso8601("#{report_params['day(1i)']}-#{report_params['day(2i)']}-#{report_params['day(3i)']}")
Date.new(report_params['day(1i)].to_i, report_params['day(2i)'].to_i, report_params['day(3i)'].to_i)
```

are identical. Both can be made into an ISO8601 with `@day.iso8601`.

> In the end as long as it Validates the date and raises an error when invalid then it should be fine.

That's true. I was just wondering why it was being formatted so that it could then be parsed to create the Date instance.

(Not a blocker)

app/views/case_workers/admin/management_information/_stats_report_with_date_form.html.haml
naseberry: Remember we can use the spacing overrides: https://design-system.service.gov.uk/styles/spacing/

app/services/stats/stats_report_generator.rb
jsugarman: I personally find it neater/more-concise to just pass all the kwargs, in any order, to the generators and they can use the report_type if they need to or not (for logging for example, although i do not believe we are at the moment).

spec/requests/case_workers/admin/management_information_spec.rb
jrmhaig: > ... you can redirect without a redirect status i believe.

I was not aware of that.

app/services/stats/management_information/queries/base_count_query.rb
jrmhaig: This is the top level so there is no superclass. This will raise a `RuntimeError` exception, which I think is less helpful than the `NoMethodError` exception that would be raised if this method doesn't exist.

spec/factories/case_types.rb
jrmhaig: Not a blocker.

app/services/stats/management_information/daily_report_count_generator.rb
jsugarman: There are some timezone edge cases to consider too, as below, but i have posed the question to the PMs

```
> Date.new(2021,10,30) + 1.month
=> Tue, 30 Nov 2021
> Date.new(2021,10,31) + 1.month
=> Tue, 30 Nov 2021
> Date.new(2021,10,30) + 1.month - 1.day
=> Mon, 29 Nov 2021
> Date.new(2021,10,31) + 1.month - 1.day
=> Mon, 29 Nov 2021
```

app/services/stats/management_information/daily_report_count_query.rb
jrmhaig: OK. Good to stick with `each_with_object`.

app/services/stats/stats_report_generator.rb
jrmhaig: :+1: for passing on the report type to generators in the options.

app/controllers/case_workers/admin/management_information_controller.rb
jsugarman: I have changed to use Date.new to allow leading or non-leading zeros, as that might be  a pain for users

the slice works as below, which feels to messy

```
Date.new(*report_params.to_h.slice('start_at(1i)', 'start_at(2i)', 'start_at(3i)').values.map(&:to_i))
```

jsugarman: see fc677ca19

app/services/stats/management_information/queries/base_count_query.rb
jsugarman: I could change to NoMethodError but having the method in the superclass is for documentary purposes as typical for template method pattern. So if you define a subclass and do not define a query method and call it you get RunTimeEror. You should never be used this abstract class, which i could define as such too.

app/services/stats/management_information/daily_report_count_generator.rb
jrmhaig: This isn't a timezone issue. It is more an issue about what '1 month on' means for 30th or 31st of a 31 day month when the next monthly only has 30 days.

* One month on from 30 June, meaning '30 days from the start of the June', would be 30 July.
* One month on from 30 June, meaning 'final day of June', would be 31 July.

Methinks that the calendar month is not a convenient time duration. Can we convince them to use 4-week periods instead? Or stick strictly to months beginning on the 1st of the month?

A bit more fun:

```
> Date.new(2022, 1, 28) + 1.month
=> Mon, 28 Feb 2022
> Date.new(2022, 1, 29) + 1.month
=> Mon, 28 Feb 2022
> Date.new(2022, 1, 30) + 1.month
=> Mon, 28 Feb 2022
> Date.new(2022, 1, 31) + 1.month
=> Mon, 28 Feb 2022
```

app/services/stats/management_information/concerns/journey_queryable.rb
jrmhaig: Yes, I missed it in the previous refactoring work.

I didn't realise that they were named after the csv header columns. This isn't a blocker but I can foresee that the slight language change between `original_submission_date` and `originally_submitted_at`, for what is essentially the same data in different time zones, could cause confusion.

app/services/stats/management_information/queries/agfs/af1_high_value_query.rb
jrmhaig: I think it is useful to have those comments. I was going to suggest having something in `docs/` in case the Google spreadsheet disappears but it makes more sense to be here.

app/services/stats/stats_report_generator.rb
jsugarman: done

app/services/stats/management_information/daily_report_count_generator.rb
jsugarman: will do exact calendar month for now. but may change depending on feedback anyway

app/services/stats/management_information/queries/agfs/af1_high_value_query.rb
jsugarman: will leave comment in but update the > or < to reflect code logic

app/services/stats/management_information/daily_report_count_generator.rb
jsugarman: done

app/services/stats/management_information/daily_report_count_generator.rb
jsugarman: done now. Have made the default duration `1.month - 1.day`, see e9ec5e6e6. This leaves the class open to being used for any duration, but defaulting to that.

spec/services/stats/management_information/daily_report_count_generator_spec.rb
jsugarman: see above

app/services/stats/management_information/queries/lgfs/lf2_high_value_query.rb
jsugarman: fixed in 82c600f2a

app/services/stats/management_information/queries/lgfs/lf1_high_value_query.rb
jsugarman: fixed in 82c600f2a

app/services/stats/management_information/queries/agfs/af2_high_value_query.rb
jsugarman: fixed in 82c600f2a

app/services/stats/management_information/queries/agfs/af1_high_value_query.rb
jsugarman: fixed in 82c600f2a

app/services/stats/management_information/daily_report_count_generator.rb
jsugarman: will have a quick look at but with only two things sharing the logic it may not be worth it (rule of three+), plus may be changing the count generator further.

jrmhaig (approved):

PR #4378, opened 2021-11-02
Factory pattern for stats report generators

jrmhaig:
#### What

Apply factory pattern to create the stats report generator based on the name of the report.

#### Ticket

[CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299) MI refactor

#### Why

`Stats::StatsReportGenerator` selects the correct report generator using a hash containing the generator classes together with specific arguments that are required for each. This hash is already quite long and needs to allow for two different styles of argument list. As more reports are added this will become longer and, possibly, more complex.

#### How

Extract the creation of report generators into factories, under the `Stats::StatsReportFactory` namespace, which can be selected based on the report name.

For clarity, `GENERATOR_FACTORIES` provides a mapping from the report name to the factory and the correct factory is found with;

```ruby
GENERATOR_FACTORIES[report_name.to_sym]
```

but, provided that the factories have consistent naming, this could be removed and replaced with

```ruby
Stats::StatsReportFactory.const_get(report_type.camelize)
```

and the mapping hash removed.

### DIFF
--- a/app/services/stats/report_generator.rb
+++ b/app/services/stats/report_generator.rb
@@ -5,24 +5,14 @@ module Stats
     end
 
     def call
-      data = report_klass.call
-      output = Stats::CsvExporter.call(data, headers: report_klass::COLUMNS)
+      data = @report_klass.call
+      output = Stats::CsvExporter.call(data, headers: @report_klass::COLUMNS)
       Stats::Result.new(output, @format)
     end
 
-    def initialize(report, format: 'csv')
-      @report = report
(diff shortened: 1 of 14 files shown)


jsugarman:
Thanks for this. it is interesting but I think it adds more layers of abstraction to an already fairly gnarly abstracted system for report generation generally. A single method holding all the "config" in one place is easier to read IMO. 

I note we are looking to remove 3 of the reports as and when the users are satisfied with the v2 of the MI report in any event so the factory method will shrink and we can revisit if deemed important.

PR #4403, opened 2021-11-03
Enable annotations in views in development env

jrmhaig:
#### What

Enable `annotate_rendered_view_with_filenames` in the development environment.

#### Ticket

N/A

#### Why

The `annotate_rendered_view_with_filenames` option was added to Action View in version 6.1. When this is enabled annotations are added to the HTML that is generated to indicate which views and partials have been rendered. The latest version of the `haml` gem is compatible with this option and so we can use it. The "inspect" view in the browser now shows something like this:

Screenshot 2021-11-03 at 15 04 11

#### How

Set

```ruby
config.action_view.annotate_rendered_view_with_filenames = true
```

in `config/environments/development.rb` only.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -82,7 +82,7 @@ Rails.application.configure do
   # config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
-  # config.action_view.annotate_rendered_view_with_filenames = true
+  config.action_view.annotate_rendered_view_with_filenames = true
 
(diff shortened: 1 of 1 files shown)


imtiazAhmd (approved):

jsugarman (approved):
we can try it out. 

Might get quite verbose in the html when inspecting, given the number of partials. Similar info can be can be gleaned from meta-request gem in conjunction with the rails panel chrome extension (which i tend to use to identify view file names)

PR #4404, opened 2021-11-04
Separate demo_data generation from db:reload task

jsugarman:
#### What
Separate demo_data generation from db:reload task

#### Ticket

depended on by [CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299)

#### Why

demo_data errors when run contiguosly with, for example:
```
ActiveRecord::AssociationTypeMismatch: Claim::BaseClaim(#1442500) expected, got #<Claim::AdvocateClaim id: 1,
```

Plus a reload should just set up db and seeds for app, not
demonstration data - for clarity.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -37,7 +37,8 @@ bin/setup
 - Load data
 
 ```
-rake db:drop db:create db:reload
+rails db:drop db:create db:reload
+rails claims::demo_data
 ```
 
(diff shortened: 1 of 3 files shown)


jrmhaig (approved):
Looks good. Opens up the possibility of having different sets of demo data.

PR #4405, opened 2021-11-05
[CFP-330] Anonymise claim_state_transitions

jsugarman:
#### What
Anonymise claim_state_transitions reason_text

#### Ticket

[CFP-330](https://dsdmoj.atlassian.net/browse/CFP-330)
came out of [CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299)

#### Why
approx 2.5% of claim_state_transitions
will have a reason_text attribute. And
it is possible this could contain
sensitive information so it should be
anonymised.

Note that this table contains a large number
of records so it will slow the dump process
down.

Also:
 - enable local db anonymised dump for testing purposes.
 - default file extension resolution to pdf for invalid content_type
 - overwrite reason code array because it causes a write error
 - * quote all columns and table names to avoid reserved word conflicts (this needs testing against a prod load)

### DIFF
--- a/lib/tasks/dump.rake
+++ b/lib/tasks/dump.rake
@@ -169,7 +169,7 @@ namespace :db do
           # typically less than 5% will have reasons and fewer with additional text
           ClaimStateTransition.where.not(reason_text: nil).find_each(batch_size: batch_size) do |claim_state_transition|
             claim_state_transition.reason_code = ['other_refuse'] if claim_state_transition.reason_code.present?
-            claim_state_transition.reason_text = Faker::Lorem.sentence(word_count: 10, supplemental: false) if claim_state_transition.reason_text.present?
+            claim_state_transition.reason_text = Faker::Lorem.sentence(word_count: 10) if claim_state_transition.reason_text.present?
             writer.call(claim_state_transition)
           end
 
--- a/lib/tasks/dump.rake
+++ b/lib/tasks/dump.rake
@@ -107,6 +107,9 @@ namespace :db do
 
       s3_bucket = S3Bucket.new(host)
       dump_files = s3_bucket.list('tmp').select { |item| item.key.match?('dump') }
+
+      abort("#{dump_files.size} dump file(s) found!".yellow) if dump_files.empty?
+
       dump_files.sort_by(&:last_modified).reverse[start..].map do |object|
         print "Deleting #{object.key}..."
         object.delete
--- a/lib/tasks/dump.rake
+++ b/lib/tasks/dump.rake
@@ -38,7 +38,7 @@ namespace :db do
     end
 
     desc 'Create anonymised database dump, compress (gzip) and upload to s3 - run on host via job (see run_job)'
-    task :anonymised => :environment do
+    task :anonymised, [:local] => :environment do |_task, args|
       cmd = 'pg_dump --version'
       puts '---------------------------'
       print "Using: #{%x(#{cmd})}".yellow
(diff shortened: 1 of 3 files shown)


lib/tasks/dump.rake
```diff
@@ -38,7 +38,7 @@ namespace :db do
     end
 
     desc 'Create anonymised database dump, compress (gzip) and upload to s3 - run on host via job (see run_job)'
-    task :anonymised => :environment do
+    task :anonymised, [:local] => :environment do |_task, args|
```
jrmhaig: This is (usually?) called from `kubernetes_deploy/scripts/job.sh` (by running `rake db:dump:run_job`) where an argument isn't provided.

If it is expected to be run directly is it worth adding something to `docs/development.md`?

lib/tasks/dump.rake
```diff
@@ -47,7 +47,13 @@ namespace :db do
       filename = File.join('tmp', "#{Time.now.strftime('%Y%m%d%H%M%S')}_dump.psql")
 
       shell_working "exporting unanonymised database data to #{filename}..." do
-        cmd = "pg_dump $DATABASE_URL --no-owner --no-privileges --no-password #{sensitive_table_exclusions} #{unneeded_table_exclusions} -f #{filename}"
+        if args.local.eql?('local')
+          with_config do |host, db, user|
+            cmd = "pg_dump --no-owner --no-privileges --no-password #{sensitive_table_exclusions} #{unneeded_table_exclusions} #{db} -f #{filename}"
+          end
```
jrmhaig: `host` and `user` appears to be unused. Are they missing from the command or can they be removed (on lines 271 and 273)?

jrmhaig (commented):
Looks OK. I've added a couple of comments.

lib/tasks/dump.rake
jsugarman: yeah, i mean to tidy this up. it was from previous attempts.

lib/tasks/dump.rake
```diff
@@ -38,16 +38,22 @@ namespace :db do
     end
 
     desc 'Create anonymised database dump, compress (gzip) and upload to s3 - run on host via job (see run_job)'
-    task :anonymised => :environment do
+    task :anonymised => :environment do |_task, args|
```
jsugarman: removed unused arguments

lib/tasks/dump.rake
```diff
@@ -236,6 +270,12 @@ namespace :db do
 
     private
 
+    def with_config
```
jsugarman: remove and tidy up above too

lib/tasks/dump.rake
jsugarman: it would have defaulted to no args but, in any event, I have amended to infer localhost environment and not require the argument

lib/tasks/dump.rake
jsugarman: amended

lib/tasks/rake_helpers/dump_file_writer.rb
```diff
@@ -54,7 +54,7 @@ def add_values(attribute_names)
   end
 
   def prepare_sql
-    table.compile_insert(data).to_sql.gsub('"', '') + ';'
+    table.compile_insert(data).to_sql + ';'
```
jrmhaig, line 57: A word of warning. I tried to find out what was going on here, and how the string is quoted, and the entire [documentation for the `compile_insert` method](https://www.rubydoc.info/docs/rails/Arel/Crud) says:

> FIXME hopefully we can remove this

My guess is that this is an internal method and not intended to be used.

It might be possible to do something more future-proof like;

```ruby
model.create(data).to_sql
```

(but this is beyond the scope of this PR, and also that will make changes to the running database, which is not desired)

jrmhaig (approved):

lib/tasks/rake_helpers/dump_file_writer.rb
jsugarman: 👍  feel free to change, test locally and PR. ultimate test is to run against prod though so is time consuming to work on

PR #4406, opened 2021-11-05
[CFP-173] Exclude unused materials fees for transfer claim with guilty plea

jrmhaig:
#### What

Do not offer the unused materials fees (up to 3 hours and over 3 hours) for transfer claims that have a guilty plea.

#### Ticket

[Transfer guilty plea claims Unused material fee flag ](https://dsdmoj.atlassian.net/browse/CFP-173)

#### Why

A transfer claim with the following details:

* Transfer Detail: This section contains transfer details.
* Litigator type: New
* Was the case elected? No
* When did you start acting? Up to and including PCMH transfer
* How did the case conclude? Guilty plea

should not be eligible for an unused material fees due to the guilty plea. However, CCCD currently offers these (up to 3 hours and over 3 hours) in the miscellaneous fees section.

#### How

Add a filter to the list of eligible misc fees to remove unused materials fee types if;

* Claim is a transfer fee
* The case conclusion is 'Guilty plea'

--------

#### TODO (wip)

 - [x] Do I need to check if 'litigator type' is 'new'?
   - Case conclusion option is only available for 'new' litigator type
 - [x] Do I need to check if 'case elected' is 'no'?
   - Case conclusion option is only available when case elected is 'no'
 - [x] Do I need to check if 'start acting' is 'Up to and including PCMH transfer'?
   - This is checked in 7d380fcf956c504b092654c406ef134cfbf10405
   - Unused materials fees are not valid on any transfer claim with a 'guilty plea' so this check is reverted in 4e9712c4875a53de522239da22cb2fa79e87f0cb


jrmhaig:
Regarding the question about whether the 'started acting' needs to be checked, it turns out that a 'Guilty' case conclusion is not valid for some values of 'started acting'. Shouldn't we be filtering options? (Not as part of this ticket)

app/services/claims/fetch_eligible_misc_fee_types.rb
```diff
@@ -66,15 +62,17 @@ def lgfs_trial_fee_misc_fee_types
       Fee::MiscFeeType.lgfs.where(unique_code: LGFS_TRIAL_CLAIM_ELIGIBILITY)
     end
 
-    def trial_fee_filter(relation)
-      return relation if case_type&.is_trial_fee? || case_type.nil?
-      relation.without_trial_fee_only
+    def apply_trial_fee_filter?
+      !case_type&.is_trial_fee? && case_type
```
jsugarman: if you test case_type first could we get away with removing the safe navigator?
`case_type && !case_type.is_trial_fee?`

app/services/claims/fetch_eligible_misc_fee_types.rb
```diff
@@ -41,7 +41,11 @@ def agfs_scheme_scope
     end
 
     def eligible_lgfs_misc_fee_types
-      filter_trial_only_types(lgfs_fee_types_by_claim_type, apply_clar_rep_order_filter? || apply_trial_fee_filter?)
+      filter_trial_only_types(lgfs_fee_types_by_claim_type, apply_filter_for_lgfs?)
+    end
+
+    def apply_filter_for_lgfs?
+      apply_clar_rep_order_filter? || apply_trial_fee_filter? || apply_transfer_guilty_plea_filter?
```
jsugarman, line 48: could use Enumerable#any?, but just a preference
`[apply_clar_rep_order_filter?, apply_trial_fee_filter?, apply_transfer_guilty_plea_filter?].any?

app/services/claims/fetch_eligible_misc_fee_types.rb
```diff
@@ -71,6 +75,11 @@ def apply_clar_rep_order_filter?
         (claim.earliest_representation_order_date < Settings.clar_release_date.to_date.beginning_of_day)
     end
 
+    def apply_transfer_guilty_plea_filter?
+      claim.transfer? && claim.transfer_detail.case_conclusion_id &&
+        Claim::TransferBrain.case_conclusion_by_id(claim.transfer_detail.case_conclusion_id) == 'Guilty plea'
```
jsugarman: relying on a description is no ideal, but relying on `case_conclusion == 50` is not great either 🤔

spec/services/claims/fetch_eligible_misc_fee_types_spec.rb
```diff
@@ -198,6 +198,20 @@
...
+          before do
+            claim.transfer_detail.update(
+              case_conclusion_id: Claim::TransferBrain.case_conclusion_id('Guilty plea')
+            )
+          end
+
+          it 'returns all LGFS misc fee types except defendant uplifts' do
+            is_expected.to match_array %w[MICJA MICJP MIEVI MISPF]
+          end
+        end
```
jsugarman: Are you confident the tests in here cover the other scenarios exhaustively/sufficiently so we can be confident that no breakages will ensue?

app/models/claim/transfer_brain.rb
```diff
@@ -69,6 +69,10 @@ def self.case_conclusion_ids
       CASE_CONCLUSIONS.keys
     end
 
+    def self.case_conclusion(detail)
```
jsugarman, line 72: 👍

jsugarman (commented):
Looks ok, A couple of comments but mainly the question of confidence given the levels of abstraction involved. How much sanity testing has been done on this. It might be good to get angela to test it on an environment to check fro possible  regressions.

spec/services/claims/fetch_eligible_misc_fee_types_spec.rb
jrmhaig: This comment is not on the latest version of the code but the question is still valid. I need to check which combinations of transfer stage and case conclusion are possible and what the behaviour should be in each case.

app/services/claims/fetch_eligible_misc_fee_types.rb
jrmhaig: I prefer the former.

There is a (very small) performance benefit of `test_one || test_two || test_three` as it will stop evaluating when the first `true` is returned. Using `Enumerable#any` will evaluate the entire array.

app/services/claims/fetch_eligible_misc_fee_types.rb
jrmhaig: Using named constants is an option. I.e.

```ruby
claim.transfer_detail.case_conclusion_id == Claim::TransferBrain::CASE_CONCLUSION_GUILTY_PLEA
```

I'd prefer this to just a number, and there are other places where it could be used.

jsugarman (commented):
Code wise looks fine. Just need to be confident we are exhaustively/sufficiently covering all scenarios. we can pair on sanity testing and coverage as discussed

jsugarman:
can you cherry-pick that last commit regarding the cleaners  out to a separate PR as it is not related to this PR.

jrmhaig:
> can you cherry-pick that last commit regarding the cleaners out to a separate PR as it is not related to this PR.

Agreed. I was hoping it was a small change but it seems to be more complicated.

jsugarman (approved):
👍

PR #4408, opened 2021-11-08
[CFP-325] Add Canary report (Do not merge - replaced by #4472)

jrmhaig:
#### What

Add a Canary token to the MI reports page.

Some notes are [on Confluence here](https://dsdmoj.atlassian.net/wiki/spaces/CFP/pages/3770319057/Thinkst+Canary+setup).

**Note;** All the commits related to the API library, in `lib/thinkst_canary` and `spec/lib/thinkst_canary` have been cherry-picked into #4469 
**Note 2;** This has been rebased on `jh/thinkst_canary_lib` and now this PR has been updated to merge into that branch

#### Ticket

[Add canary reports for MI reports](https://dsdmoj.atlassian.net/browse/CFP-325)

#### Why

To detect if someone with access to the MI reports page is scraping the reports.

#### How

* Add a hidden link to a new report 'reports_access_details' that would appear (if it were not hidden) the same as other report download links.
* Create a Rake task to create a new MI report from a supplied file.

The Canary file can be created with:

```ruby
rails canary:create_reports_access_details
```

#### TODO

- [ ] Get a suitable dummy file


app/controllers/case_workers/admin/management_information_controller.rb
```diff
@@ -11,6 +11,7 @@ def index
     @available_report_types = Stats::StatsReport::TYPES.index_with do |report_type|
       Stats::StatsReport.most_recent_by_type(report_type)
     end
+    @reports_access = Stats::StatsReport.most_recent_by_type('reports_access_details')
```
jsugarman: be good to isolate all this logic into one. I am thinking

```
before_action :set_report_access_details, only: :index 
...
private
...
def set_report_access_details
    @reports_access = Stats::StatsReport.most_recent_by_type('reports_access_details')
end
```

app/controllers/case_workers/admin/management_information_controller.rb
```diff
@@ -34,6 +35,8 @@ def generate
 
   def validate_report_type
     return if Stats::StatsReport::TYPES.include?(params[:report_type])
+    return if params[:report_type] == 'reports_access_details'
```
jsugarman: i'd say preferable to put this type into `Stats::StatsReport::TYPES` and keep the controller thin.

app/views/case_workers/admin/management_information/index.html.haml
```diff
@@ -24,6 +24,12 @@
 
           = govuk_link_button_secondary(t('.update_report'), case_workers_admin_management_information_generate_url(report_type: report_type))
 
+    .form-section.hidden
```
jsugarman, line 27: You are actually touching the same files that my branch touched. I have changed quite a lot. I broke the page into partials (one for reports with date fields and one for others). This will cause some rebasing headaches for who ever goes in second, but also my branch has changed to use structs for report types, which could be extended to simplify this by doing something like

```
if report_object.hidden?
  .form-section.hidden
else
  .form-section

     ....

```

This would require adding it to the TYPES constant and treating it largely as any report in the controller.

Another alternative, more isloated approach, if added to TYPES is

```
@available_report_types.except{blah}.each do ||
   ...
@available_report_types.select{blah}.each do ||
  .form-section.hidden
    %h2.govuk-heading-l{ id: 'heading-reports_access_details' }
      = t(".report_types.reports_access_details_html")
    %p
      = govuk_link_button(t('.download'), case_workers_admin_management_information_download_url(report_type: 'reports_access_details'))
```

These could be extracted to partials to further isolate. 

should probably pair on that as it will impact my PR when rebasing

app/controllers/case_workers/admin/management_information_controller.rb
jsugarman: are we even using @reports_access anywhere - does not look like it

docs/thinkst_canary.md
```diff
@@ -0,0 +1,77 @@
+# Thinkst Canary API
+
+## Canarytoken Factory
+
+To facilitate the automation of Canary token creation without requiring the
+use of the main auth token a factory auth token can be used, which is a limit
```
jsugarman:
```suggestion
use of the main auth token a factory auth token can be used, which is a limited
```

docs/thinkst_canary.md
```diff
@@ -0,0 +1,77 @@
+# Thinkst Canary API
```
jsugarman, line 1: have you linked to this page from the development.md, or anywhere else?

docs/thinkst_canary.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+To download the file and save it;
+
+```ruby
+File.open('test_file_canary.docx', 'wb') do |file|
+  file.puts token.download
+end
+```
+
```
jsugarman, line 76: be good to mention the rake task shortcuts here too with a short "How to [re]create document canary" para. with code snippet.

lib/tasks/canary.rake
```diff
@@ -0,0 +1,69 @@
+namespace :canary do
+  desc 'Create a new Canary factory auth token'
+  task :create_factory_auth, [:memo, :flock_id] => :environment do |_task, args|
+    abort 'Memo required' if args[:memo].nil?
+
+    require_relative 'rake_helpers/canary_configuration'
```
jsugarman, line 6: Just wondering if it is worth moving the file here to an initializer!? 🤔  It would be yet another thing to load, slowing startup down further (for tests and rake tasks alike), but would mean we don't need this line (and others below), nor anywhere else we might need to call on the Canary classes generally.

Can always move it later i guess. what do you think?

lib/thinkst_canary/token/null_token.rb
```diff
@@ -0,0 +1,11 @@
+module ThinkstCanary
+  module Token
+    class NullToken < Base
+      def initialize(**options)
```
jsugarman: nit. seems to be the modern preference for such things and more consistent with other classes created in this PR
```suggestion
      def initialize(**kwargs)
```

spec/lib/thinkst_canary/factory_generator_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    let(:factory_options) { { flock_id: 'test_flock', memo: 'test_memo' } }
+    let(:factory_response) { { 'factory_auth' => 'factory_auth_token', 'result' => 'success' } }
+
+    before do
+      allow(ThinkstCanary.configuration).to receive(:query).and_return(factory_response)
+    end
+
+    it { is_expected.to be_a ThinkstCanary::Factory }
+    it { expect(create_factory.factory_auth).to eq(factory_response['factory_auth']) }
```
jsugarman: I personally prefer repeating the hard coded/stubbed response explicitly for clarity if it is static. Otherwise it just feels like you are testing the stub and detracts from the documentary aspect of the test, IMO

```suggestion
    it { expect(create_factory.factory_auth).to eq('factory_auth_token') }
    it { expect(create_factory.flock_id).to eq('test_flock') }
    it { expect(create_factory.memo).to eq('test_memo') }
```

spec/lib/thinkst_canary/factory_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    before do
+      allow(ThinkstCanary.configuration).to receive(:query)
+
+      delete
+    end
+
+    it do
+      expect(ThinkstCanary.configuration).to have_received(:query).with(
+        :delete, '/api/v1/canarytoken/delete_factory',
+        params: { factory_auth: factory_options[:factory_auth] }
```
jsugarman: see other comment but not a blocker

spec/lib/thinkst_canary/factory_generator_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    it { expect(create_factory.flock_id).to eq(factory_options[:flock_id]) }
+    it { expect(create_factory.memo).to eq(factory_options[:memo]) }
+
+    it 'makes a POST request for a factory' do
+      create_factory
+
+      expect(ThinkstCanary.configuration).to have_received(:query).with(
+        :post, '/api/v1/canarytoken/create_factory',
+        params: { flock_id: factory_options[:flock_id], memo: factory_options[:memo] }
```
jsugarman: see other comment but not a blocker

jsugarman (commented):
Looks good, couple of things to look at and consider

1. alternative handling of the report type in controller and view, possible pairing to do
2. possible "promotion" of canary config to be in a rails initializer
3. some smaller nits and comments on the tests

app/controllers/case_workers/admin/management_information_controller.rb
jrmhaig: > 

Apparently not. It was originally used for the 'last created date' but I removed that. It doesn't even appear to have made it into a commit.

lib/tasks/canary.rake
jrmhaig: This did occur to me but the configuration isn't required for the app so I didn't think it was suitable.

If `ThinkstCanary` becomes its own gem (a possibility if other teams want to use it) then it would make sense for it to have the `require: false` options in `Gemfile`, and then be required explicitly. An initializer in this case would cause problems.

We can talk about this.

app/views/case_workers/admin/management_information/index.html.haml
jrmhaig: It will be easier for me to resolve the conflicts so I'll wait for your PR to go in first.

app/controllers/case_workers/admin/management_information_controller.rb
jrmhaig: I had done this initially to avoid the (remote) possibility of this report type being used with the regenerate url. As it turns out, this didn't work as I expected so it might as well be in the `TYPE` array.

jrmhaig: I have added `report_access_details` to `Stats::StatsReport::TYPES` but this hasn't removed the requirement for the additional check as it should only work for the download url and not the generate url. It may be possible to do this better once #4377 has been merged into master and this PR gets rebased.

jrmhaig: A side effect of adding to `TYPES` is that it now appears visibly on the download page. It would be possible to fix this but I think it may benefit from the `Report` struct that is introduced in #4377 so I will wait for that before changing anything more.

PR #4425, opened 2021-11-12
Use sql_query gem for SQL templates

jsugarman:
#### What
use the sql_query gem to organize sql and allow for embedded ruby

#### Ticket

[CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299)

#### Why
Not sure it is the way to go but is an interesting option
for handling raw SQL queries generally.

#### How
Install sql_query gem, configure to use app/sql_templates default
folder for lookups and then call `SqlQuery.new(name|path_to_sql_file).execute`
wherever that SQL needs to be executed.


app/services/stats/management_information/daily_report_query.rb
```diff
@@ -28,7 +28,8 @@ def call
       private
 
       def journeys
-        ActiveRecord::Base.connection.execute(journeys_query)
+        # ActiveRecord::Base.connection.execute(journeys_query)
```
jsugarman: remove

mpw5:
This was a speculative PR from several years ago by a developer who is no longer on the team. Closing as we won't be implementing it.

PR #4427, opened 2021-11-16
Add error message to MI report Slack failure notifications

jrmhaig:
#### What

Add the error message to the Slack notification when an MI report fails. Also, send the notification to #laa-cccd-alerts instead of #cccd_development

#### Ticket

N/A

#### Why

The alert currently looks like this;

Screenshot 2021-11-16 at 10 13 05

To actually find the error, though, it is necessary to log on to the pod and look through the logs.

#### How

The error is already being passed to the Slack notifier class, `Subscribers::Slack`, but it is not being used. This has been changed to add the message to the payload.

The alert now looks like this;

Screenshot 2021-11-16 at 11 40 37

FYI, this was generated by

```ruby
begin
  2/0
rescue => e
  ActiveSupport::Notifications.instrument('call_failed.stats_report', id: 999, name: 'just-a-test', error: e)
end
```

### DIFF
--- a/app/models/subscribers/slack.rb
+++ b/app/models/subscribers/slack.rb
@@ -3,7 +3,7 @@ module Subscribers
     def process
       report_id = event.payload[:id]
       report_name = event.payload[:name]
-      slack_notifier = SlackNotifier.new('cccd_development', formatter: SlackNotifier::Formatter::Generic.new)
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍

PR #4441, opened 2021-11-17
Order main defendant by created_at

jsugarman:
#### What
Order main defendant by created_at

#### Ticket

[change to CFP-299](https://dsdmoj.atlassian.net/browse/CFP-299)

#### Why
Diffs with prod data versions for reports indicate
a non-deterministic return for main defendant. This
may not fix the diff completely but creates consistency for the
version 2 of the report.

Evidence from diffs shows it may resolve some of diffs
plus rails model `first` calls (`defendants.first`) supposedly
orders by `id` by default, which will match `defendant.order(:created_at)`).


jrmhaig (approved):
:+1:

PR #4452, opened 2021-11-24
Set the surname of the test advocate

jrmhaig:
#### What

Fix one of the flickering tests.

#### Ticket

N/A

#### Why

One of the tests (spec/controllers/external_users/claims_controller_spec.rb:475) flickers because it is searching for the advocates surname and expecting nothing to be returned but the surname is randomly chosen and 'Smith' is used as a defendant name.

#### How

The solution is to fix the name of the advocate to ensure that a result is not found.

### DIFF
--- a/spec/controllers/external_users/claims_controller_spec.rb
+++ b/spec/controllers/external_users/claims_controller_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe ExternalUsers::ClaimsController, type: :controller do
 
   before { sign_in advocate.user }
 
-  context 'list views' do
+  describe 'list views' do
     let!(:advocate_admin) { create(:external_user, :admin, provider: advocate.provider) }
     let!(:other_advocate) { create(:external_user, :advocate, provider: advocate.provider) }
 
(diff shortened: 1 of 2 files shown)


spec/factories/external_users.rb
```diff
@@ -17,11 +17,12 @@
   factory :external_user do
     transient do
       build_user { true }
+      last_name { Faker::Name.last_name }
     end
 
     after(:build) do |eu, evaluator|
       if evaluator.build_user
-        eu.user ||= build(:user, first_name: Faker::Name.first_name, last_name: Faker::Name.last_name, email: Faker::Internet.email, password: 'password', password_confirmation: 'password')
+        eu.user ||= build(:user, first_name: Faker::Name.first_name, last_name: evaluator.last_name, email: Faker::Internet.email, password: 'password', password_confirmation: 'password')
```
jsugarman: will this not need to fallback if the evaluator has not supplied a last name
```suggestion
        eu.user ||= build(:user, first_name: Faker::Name.first_name, last_name: evaluator.last_name || Faker::Name.last_name, email: Faker::Internet.email, password: 'password', password_confirmation: 'password')
```

jrmhaig: If this is not set in the `create` or `build` then it gets set on line 20, above.

spec/controllers/external_users/claims_controller_spec.rb
```diff
@@ -448,28 +469,26 @@ def advocate
...
-        @allocated_claim = create(:allocated_claim, external_user: advocate)
-        create(:defendant, claim: @allocated_claim, first_name: 'Pete', last_name: 'Adams')
+        create(:defendant, claim: archived_claim, first_name: 'John', last_name: 'Smith')
+        create(:defendant, claim: draft_claim, first_name: 'John', last_name: 'Smith')
+        create(:defendant, claim: allocated_claim, first_name: 'Pete', last_name: 'Adams')
       end
 
-      context 'in all claims' do
-        context 'by defendant name' do
+      describe  'in all claims' do
```
jsugarman: `describe` should be for a method  `#method_name|.class_method_name`

spec/controllers/external_users/claims_controller_spec.rb
```diff
@@ -482,21 +501,21 @@ def advocate
...
-              expect(assigns(:claims).sort_by(&:state)).to eq([@allocated_claim, @draft_claim])
+              expect(assigns(:claims).sort_by(&:state)).to eq([allocated_claim, draft_claim])
             end
           end
         end
       end
 
-      context 'in archive' do
-        context 'by defendant name' do
+      describe 'in archive' do
```
jsugarman: describe should be for a method #method_name|.class_method_name

spec/factories/external_users.rb
jsugarman: right! i'd be tempted to go further then and make the first_name, and even email transient with default, as for last name. would be more consistent and less confusing.

jsugarman (changes requested):

spec/factories/external_users.rb
jrmhaig: I could do this. I was going for the minimum possible change but it makes sense to be able to set the other fields too.

jrmhaig: Actually, I had second thought and made it simpler with bb6347a86908a8e1c86d868bca1065203f7228a5

jsugarman (approved):
👍

PR #4457, opened 2021-11-25
[CFP-300] Improve MI report link purpose

naseberry:
#### What
Add context to each `Download` and `Update` link on MI report page

#### Ticket

n/a

#### Why

- To improve WCAG Success Criterion 2.4.4: Link Purpose (In Context)
- To improve event tracking


jsugarman (commented):
@naseberry just to say i have touched and change this file a lot, breaking into partials, so would be ideal if you could PR this against my `mi-weekly-statistics-report-v2` branch - which will require a rewrite as a result

app/views/case_workers/admin/management_information/_stats_report.html.haml
```diff
@@ -1,4 +1,5 @@
 - i18n_scope = 'case_workers.admin.management_information.index'
+- action_context = t("report_types.#{report_object}_html", scope: i18n_scope)
```
jsugarman, line 2: i was trying to avoid any dynamic translation lookup as it is not testable by i18n-tasks gem. will need to look at this to see if we can break it down to avoid that

naseberry:
Superseded by https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4473

PR #4458, opened 2021-11-26
[CFP-343] Fix for previously upload table

imtiazAhmd:
#### What
Upload a supporting evidence document to a claim that already has a document attached breaks the table layout
#### Ticket
[CFP-343](https://dsdmoj.atlassian.net/browse/CFP-343)

#### Steps to reproduce

- Log in as an external user
- Open any draft claim
- While on the claim summary page, click change *Supporting evidence*
- If no files have have been previously uploaded, upload a supported file
- Click Save and continue
- While on the claim summary page, click change *Supporting evidence*
- Upload a supported file
- Visually inspect the file tables in the UI for error (see attached example)

#### How
Both Previously uploaded table and Document waiting to be saved table are using the same className that was using as a selector for targeting an action to update the tables. 

selector now changed to unique id to mitigate this issue.

### DIFF
--- a/app/webpack/javascripts/modules/external_users/claims/Dropzone.js
+++ b/app/webpack/javascripts/modules/external_users/claims/Dropzone.js
@@ -51,7 +51,9 @@ moj.Modules.Dropzone = {
   },
 
   toggleFileStatus: function () {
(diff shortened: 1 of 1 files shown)


naseberry (approved):
👍 

As @jrmhaig identified the issue, I would try to get his approval as well.

jrmhaig (approved):

PR #4459, opened 2021-11-26
[CFP-300] MI statistics - function prep optimization

jsugarman:
**What**
Optimize MI query preparation

**Why**
Journeys postgres DB function is dropped and replaced
on every call. This is an unnecessary overhead most of the
time as only changes to the function signature or body
would require a drop and replace respectively.

This slows tests and there are race conditions running
on production(-like) load balanced servers where this can
result in errors such as below. I believe this is due to
trying to replace a function in use. It is unclear what the
DROP command could do in such instances.

```
Error: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "pg_proc_proname_args_nsp_index"
DETAIL:  Key (proname, proargtypes, pronamespace)=(journeys, 23, 16403) already
```

This change means it will only create the function if
1. There is no function of that name (irrespective of signature)
2. or if the app setting flag `Settings.replace_journeys_func?` is set to true.
   see below.

**The `Settings.replace_journeys_func?`**

Add a settings flag to forcibly cause recreation of the `journeys` function
every time a report that uses it is run.

We default replacing the function to false as it already exists
and will not require replacing unless changes are made to its
signature or body.

**Usage**
When changes are made to the 'journeys` function signature or body then this flag
must be enabled to recreate the function, the app deployed and a report run
against that environment. This will update the function on the environment which the
report is run.

1. A simple alternative is to run the following on each environment
```
Stats::ManagementInformation::DailyReportQuery.new.send(:recreate_journeys_func)
```

2. A version controlled alternative is to use a rails migration
 Example (untested):
 ```
 class ReplaceJourneysFunction < ActiveRecord::Migration[6.1]
  include Stats::ManagementInformation::JourneyQueryable
  def up
    send(:recreate_journeys_func)
  end

  def down
   raise ActiveRecord::IrreversibleMigration, 'function change cannot be undone!'
  end
end
```

### DIFF
--- a/app/services/stats/management_information/queries/base_count_query.rb
+++ b/app/services/stats/management_information/queries/base_count_query.rb
@@ -25,11 +25,6 @@ module Stats
 
       private
 
-      def prepare
-        ActiveRecord::Base.connection.execute(drop_journeys_func)
-        ActiveRecord::Base.connection.execute(create_journeys_func)
-      end
-
       def sql_quote(column_name)
         ActiveRecord::Base.connection.quote_column_name(column_name)
       end
--- a/app/services/stats/management_information/concerns/journey_queryable.rb
+++ b/app/services/stats/management_information/concerns/journey_queryable.rb
@@ -9,84 +9,8 @@ module Stats
 
       included do
         def prepare
-          ActiveRecord::Base.connection.execute(drop_journeys_func)
-          ActiveRecord::Base.connection.execute(create_journeys_func)
-        end
-
-        def drop_journeys_func
-          <<~SQL
-            DROP FUNCTION IF EXISTS journeys(integer);
-          SQL
-        end
-
-        def create_journeys_func
-          <<~SQL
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: app/services/stats/management_information/queries/base_count_query.rb [hunk 1]
```diff
@@ -12,10 +12,11 @@ def self.call(**kwargs)
         new(kwargs).call
       end
 
-      def initialize(date_range:, date_column_filter:)
+      def initialize(date_range:, date_column_filter:, replace_journeys_func: false)
```
jrmhaig: lets remove the instance level logic and just use the app flag as a trigger

jrmhaig (approved):

app/services/stats/management_information/queries/base_count_query.rb
jsugarman: done

PR #4461, opened 2021-11-30
[CFP-263] Jasmine migration to node

imtiazAhmd:
#### What
The Jasmine Ruby gems are deprecated. There will be no further releases after the end of the Jasmine 3.x series. 
We need to migrate to the `jasmine-browser-runner` npm package, which is the direct replacement for the jasmine gem.

#### Ticket

[CFP-263](https://dsdmoj.atlassian.net/browse/CFP-263)

--------

#### TODO (wip)

 - [x] Retire current Jasmine gem.
 - [x] Migrate to jasmine-browser-runner.
 - [x] Run jasmine test in CI
 - [x] Clean up gem related jasmine config.
 - [x] Jasmine tests should run as expected after migration

--------

> **NOTE**: 
> If MacOS gatekeeper cannot verify the `chromedriver` use this command to exclude from the check
> `xattr -d com.apple.quarantine /usr/local/bin/chromedriver`
>  Installed chromedriver might be older than the browser installed on the system. 
>  Updating chromedriver will resolve the issue.
> `brew update` and `brew upgrade chromedriver`


config/webpacker.yml
```diff
@@ -10,7 +10,7 @@ default: &default
 
   # Additional paths webpack should lookup modules
   # ['app/assets', 'engine/foo/app/assets']
-  additional_paths: ['node_modules/govuk-frontend/govuk', 'node_modules/govuk_frontend_toolkit']
+  additional_paths: ['node_modules/govuk-frontend/govuk', 'node_modules/govuk_frontend_toolkit', 'spec/javascripts']
```
naseberry, line 13: Does `spec/javascripts` need to be added or is this redundant, especially as you are already loading all js specs via entry file `app/webpack/packs/specs.js`

package.json
```diff
@@ -41,6 +41,8 @@
     "webpack-concat-plugin": "^3.0.0"
   },
   "devDependencies": {
+    "jasmine-browser-runner": "^0.9.0",
```
naseberry, line 46: To make the this change more pragmatic and readable, I would suggest we set jasmine as a test script in our package.json (also inline with Jasmine's documentation)
`"scripts": { "test:jasmine": "jasmine-browser-runner runSpecs" }`

naseberry (commented):
Great work, we may also be able to remove jasmine references in our guard file

config/webpacker.yml
imtiazAhmd: refer to the documentation step 5: [Jasmine](https://jasmine.github.io/setup/browser.html#use-with-rails)

package.json
imtiazAhmd: Two scripts added, `test:jasmine` which will run the test in a browser and `test:jasmine-headless` for running in CLI

imtiazAhmd:
> 

Removed `jasmine` related config from the `guard`

naseberry (approved):
LGTM 👍

jrmhaig (approved):

jsugarman (approved):
Got it working following the PR guidance

PR #4469, opened 2021-12-01
[CFP-325] Thinkst Canary API lib

jrmhaig:
#### What

Library to communicate with the Canarytoken API.

This is made up of commits cherry-picked from #4408 

#### Ticket

[Add canary reports for MI reports](https://dsdmoj.atlassian.net/browse/CFP-325)

#### Why

Provide a way to communicate with the Canarytoken API that can be used in various places for creating and managing of canaries.

#### How

See docs/thinkst_canary.md

With an existing factory auth string, a Canarytoken factory can be created with;

```ruby
factory = ThinkstCanary::Factory.new(
  factory_auth: <factory_auth_token>,
  flock_id: 'flock:123...'
)
```

Alternatively, if a factory auth string does not exist then one can be generated with;

```ruby
factory_generator = ThinksCanary::FactoryGenerator.new
factory = factory_generator.create_factory(
  flock_id: 'flock:123...',
  memo: 'My shiny new token factory'
)
# => An instance of ThinkstCanary::Factory
```

Using the factory created above, new Canarytokens can be created with;

```ruby
token = factory.create_token(
  kind: <kind of token>,
  memo: 'My shiny new Canary token',
  <other options>
)
token_file_contents = token.download
```


docs/thinkst_canary.md
```diff
@@ -0,0 +1,77 @@
+# Thinkst Canary API
+
+## Canarytoken Factory
+
+To facilitate the automation of Canary token creation without requiring the
+use of the main auth token a factory auth token can be used, which is a limit
```
jsugarman:
```suggestion
use of the main auth token a factory auth token can be used, which is a limited
```

docs/thinkst_canary.md
```diff
@@ -0,0 +1,77 @@
...
+An existing factory auth token can be used with;
+
+```ruby
+factory = ThinkstCanary::Factory.new(
+  factory_auth: <factory_auth_token>,
+  flock_id: 'flock:123...'
+)
+```
+
+**TODO:** For a `factory` as defined above, query the API to fetch the `flock_id` and `memo`.
```
jsugarman: Should this be done: example snippet for this would be good, or link to relevant documentation

jsugarman (commented):
no blockers. just a question on a **todo** in the documentation and then happy to approve

docs/thinkst_canary.md
jrmhaig: There isn't currently a way to fetch details of a factory auth token from the API. I raised this via their support email and they say they are looking into it.

I put this TODO here because I think it would be useful to do but the API doesn't support it yet. I have added a note to this effect: ad36b5bec286c3cbf4727f9790d40f88322dbb97

jsugarman (approved):
👍

PR #4482, opened 2021-12-10
[CFP-300] Ensure count queries have a scheme associated

jsugarman:
#### What
Ensure count queries have a scheme associated

#### Why
Without a `scheme` method with appropriate value the `journeys_query` will
query all journeys across AGFS and LGFS. This roughly doubles the number
of records requiring counting and can therefore significantly
increase query time.

#### How
I have amended the `claim_type_filterable` mixin to add a class
method `acts_as_scheme` which can be used over-ride the default
`scheme` method for any class including it. The overridden
`scheme` method will always return the value passed to
`acts_as_scheme` (but as up-cased string).

### DIFF
--- a/spec/services/stats/management_information/claim_type_filterable_spec.rb
+++ b/spec/services/stats/management_information/claim_type_filterable_spec.rb
@@ -1,0 +1,204 @@
+# frozen_string_literal: true
+
+RSpec.describe Stats::ManagementInformation::ClaimTypeFilterable do
+  let(:agfs_in_statement) do
+    <<~STATEMENT.squish.squeeze(' ')
+      ('Claim::AdvocateClaim',
+      'Claim::AdvocateInterimClaim',
+      'Claim::AdvocateSupplementaryClaim',
+      'Claim::AdvocateHardshipClaim')
+    STATEMENT
+  end
+
+  let(:lgfs_in_statement) do
+    <<~STATEMENT.squish.squeeze(' ')
+      ('Claim::LitigatorClaim',
+      'Claim::InterimClaim',
+      'Claim::TransferClaim',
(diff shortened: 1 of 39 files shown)


spec/services/stats/management_information/claim_type_filterable_spec.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
```
jsugarman, line 1: move to `spec/.../concerns./..` to match path in app?!

spec/services/stats/management_information/daily_report_query_spec.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# require_relative 'shared_examples_for_claim_type_filterable'
```
jsugarman: delete

spec/services/stats/management_information/claim_type_filterable_spec.rb
jsugarman: rspec expects it to be in location base on module name not file path

jrmhaig:
Looks good. Can I clarify, for my understanding, are these equivalent?

```ruby
class NewQuery
  include ClaimTypeFilterable
  acts_as_scheme :agfs
end
```

and

```ruby
class NewQuery
  include ClaimTypeFilterable

  def initialize
    self.scheme = :agfs
  end
end
```

jsugarman:
Yes, pretty much, except act_as_scheme :agfs, will prevent `self.scheme = :agfs` from doing anything as `scheme` will always return AGFS thereafter

jrmhaig (approved):

PR #4484, opened 2021-12-10
Test summary and field error translation retrieval

jsugarman:
This gets it working

Screenshot 2021-12-10 at 17 45 03


config/locales/en/models/claim.yml
```diff
@@ -0,0 +1,26 @@
...
+# activerecord.errors.messages
+# errors.attributes.[attribute_name]
+# errors.messages
+---
+
+en:
+  activerecord:
+    errors:
+      models:
+        claim/base_claim:
```
jsugarman, line 6: this is what we were missing @naseberry  - the i18n-debug gem help to identify it

jsugarman:
closed in favour of #4504

PR #4491, opened 2021-12-15
[CFP-326] Consolidate canary shared examples

jsugarman:
#### What
Consolidate shared examples

#### Ticket

[Add canaries to S3 storage](https://dsdmoj.atlassian.net/jira/software/c/projects/CFP/boards/624?modal=detail&selectedIssue=CFP-326)

#### Why
Makes more sense to have just the one shared examples
set for this behaviour at the moment.


jrmhaig (approved):
OK, these can be consolidated.

If we ever extend this to work with non-file based Canarytokens then this may be split again.

PR #4496, opened 2021-12-16
[CFP-326] Consolidate shared examples for canarytoken

jrmhaig:
#### What
Consolidate shared examples

This is a copy of #4491, which should have been merged but wasn't due to a rebase.

#### Ticket

[Add canaries to S3 storage](https://dsdmoj.atlassian.net/jira/software/c/projects/CFP/boards/624?modal=detail&selectedIssue=CFP-326)

#### Why
Makes more sense to have just the one shared examples
set for this behaviour at the moment.


jsugarman (approved):
👍

PR #4498, opened 2021-12-17
[CFP-295] Update to Ruby 3.0

jrmhaig:
#### What

Update to Ruby 3.0

#### Ticket

#### Why

Ruby 3.1 will be released soon and we are still on 2.7.

#### How

There are a number of changes that are not backward compatible. This PR, which may or may not be merged in its current form, is intended to see what needs to be fixed.

* Commit 65b5fa5a95076f2b165e1ec4ee58db4010d9e3a3
  * `Dir[]` is now sorted by default.
  * Regexp and Range literals are now frozen objects.
  * New `Hash#except` method can replace `.reject { |k, _v| k == 'something' }`
* Commit f2e5289b7cba81945357f812cb14a18ef2827930
  * `bigdecimal` is a core library but there appears to be a bug in the version in Ruby 3.0 that causes problems with floating point numbers


naseberry (approved):
code changes LGTM, and sanity checked on `staging`

PR #4501, opened 2021-12-20
[CFP-359] Summary list

naseberry:
#### What

Migrate the description list markup to the GOVUK Design System Summary list markup.

#### Ticket

[GOVUK Migration - Summary list](https://dsdmoj.atlassian.net/browse/CFP-359)

#### Why

Part of a larger change to:

- improve code maintenance
- componentize Design system components
- remove deprecated code


config/locales/sidebars.en.yml
```diff
@@ -3,6 +3,7 @@ en:
   sidebar_common:
     - &sidebar_fixed_fees Fixed fees
     - &sidebar_graduated_fees Graduated fees
+    - &sidebar_hardship_fees Hardship fees
```
jsugarman, line 6: 🤔 hardship fees are not a kind of fee, from a business perspective, they are merely the same kinds of fees claimed on a hardship claim

jsugarman: although line 24 does call them that, in which case that line should change to use the anchor too

imtiazAhmd (approved):

PR #4502, opened 2021-12-20
[CFP-365] Replace cloud platform tool and authentication mechanism

jsugarman:
#### What
`setup-kube-auth` is deprecated and unavailable for use
on the new "live" EKS cluster. We therefore need to upgrade
to the newer tool that is supported for use on the new cluster.

#### Ticket

[CFP-365](https://dsdmoj.atlassian.net/browse/CFP-365)

#### Why
`setup-kube-auth` is deprecated and unavailable for use
on the new "live" EKS cluster. We therefore need to upgrade
to the newer tool that is supported for use on the new cluster.

#### How
see [CP guide](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/deploying-an-app/using-circleci-for-continuous-deployment.html#continuous-deployment-of-an-application-using-circleci-and-helm)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ executors:
   cloud-platform-executor:
     resource_class: small
     docker:
-    - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
+    - image: ministryofjustice/cloud-platform-tools
       environment:
         GITHUB_TEAM_NAME_SLUG: laa-get-paid
         REPO_NAME: cccd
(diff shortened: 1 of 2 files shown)


jrmhaig:
~Generally OK. Just one question from the commit message:~

> ~Remove no-longer-functional cloud-platform context var~

~Can you point out where this context var was? I cannot see one being removed.~

Ignore that. I see it now.

jrmhaig (approved):
👍

imtiazAhmd (approved):

PR #4503, opened 2021-12-21
No Canarytoken trigger for local copy of page

jrmhaig:
#### What

Do not trigger the Canarytoken if the protocol is `file:`, indicating that a user has saved a pages to the hard drive and then viewed it from there.

#### Ticket

N/A

#### Why

When users save a page from CCCD to their local drives and then view them a Canarytoken alert is triggered. This is because `document.domain` does not match any domains on the acceptable list.

#### How

Skip alerting the Canarytoken if the protocol is `file:`.

I could have explicitly matched on `http:` but by excluding `file:` it leaves open the possibility for a hacker to create a link with a more esoteric protocol.


jsugarman (approved):
👍 that'll do it. if we need to exclude any more we can abstract out to a constant or other helper

PR #4504, opened 2021-12-21
[CFP-54] amend validation and errors inline with govuk-formbuilder requirements

jsugarman:
#### What
a. add case number/urn validation errors
b. validate case type object presence (not Foreign key/FK) but with error on FK
b. validate case stage id (FK) absence

#### Why
**RE**: case type validation
For `belongs_to` associations we should continue to validate presence of the associated object, 
not the foreign key `case_type_id` (see [rails guide](https://guides.rubyonrails.org/active_record_validations.html#presence)). However, the error should
 be added to the foreign key attribute for rendering in the view according to 
govuk-formbuilder (and rails?) standards.

#### Ticket

[CFP-54](CFP-54)

### DIFF
--- a/app/validators/claim/advocate_hardship_claim_validator.rb
+++ b/app/validators/claim/advocate_hardship_claim_validator.rb
@@ -39,7 +39,7 @@ class Claim::AdvocateHardshipClaimValidator < Claim::BaseClaimValidator
   # NOTE**: case_type is delegated to case_stage for hardship claims
   # and should not exist directly on the claim
   def validate_case_type_id
-    validate_absence(:case_type_id, 'present')
+    validates_belongs_to_attribute_absence(:case_type, :present)
   end
 
   def validate_case_stage_id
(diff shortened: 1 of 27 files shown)


app/validators/claim/advocate_hardship_claim_validator.rb
```diff
@@ -39,7 +39,7 @@ def self.fields_for_steps
   # NOTE**: case_type is delegated to case_stage for hardship claims
   # and should not exist directly on the claim
   def validate_case_type_id
-    validate_absence(:case_type_id, 'present')
+    validates_belongs_to_attribute_absence(:case_type, :present)
```
jsugarman: we have to validate absence of the `case_type_id` foreign key not the `case_type` object.

Explanation: this is because the claim `advocate_hardship_claim` (and `litigator_hardship_claim`) delegate `case_type` object to `case_stage` object, so the `case_type` object WILL exist, and this is correct. It MUST not have a `case_type_id` foreign however as this would override, or cause delegation problems/confusion.

naseberry (approved):
Looks good, a few validation errors persists but I should be able to continue updating these once merged into the base branch.
thanks @jsugarman

PR #4520, opened 2021-12-22
[CFP-327] Task to generate Canarytokens for documents and message attachments

jrmhaig:
#### What

Tasks to generate Canarytokens saved as documents ~and message~ attachments.

_For the moment, Canarytokens are not going to be added to message attachments._

#### Ticket

[Document and message canaries](https://dsdmoj.atlassian.net/browse/CFP-327)

#### Why

When logged in as an caseworker it is possible to download any evidence document and message attachments by hitting the endpoints `documents/<document-id>/download` and `messages/<message-id>/download_attachment`. An attacker who has gained unauthorized access to CCCD by logging in as a caseworker may use this in a script to download arbitrary documents. By having Canarytokens, not connected to any claims, placed at certain intervals (in terms of document id) could enable us to detect when this has been done.

Eg;

```bash
#!/usr/bin/env bash

# Cookies for an authenticated caseworker user session
COOKIES=...

for i in `seq 100 200`
do
  curl https://claim-crown-court-defence.service.gov.uk/documents/$i/download \
    -L -OJ -H "Cookie: $COOKIES"
done
```

#### How - Documents

New Canarytokens can be added to the `Document` model with;

```bash
rails canary:create_document_canary
```

This will generate a docx and pdf file based on a fake name and number for the document and converted preview document, such as; `john smith 123.docx` and `john smith 123.docx.pdf`.

These documents are created unattached to any claims and a list of all unattached documents can be found with

```bash
rails 'canary:find_unattached[documents]'
```

A scheduled job has been created to run the `canary:create_document_canary` task every two hours.

#### ~How - Messages~

~The intention is to add Canarytokens to messages in the same way as with documents. However, an instance of `Message` is required to have a claim and sender. There are two possible solutions to this;~

* ~Modify the `Message` class so that it does not require a claim and a sender. This is not desirable as it would affect validations on legitimate messages~
* ~Create a dummy user with a dummy claim solely for the purpose of attaching Canarytokens.~

--------

#### TODO (wip)

 - [x] Documents Rake task
 - [ ] ~Messages Rake task~
 - [x] Schedule Documents Rake task
 - [ ] ~Schedule Messages Rake task~
 - [x] Get some realistic looking base files to base the Canarytokens on


scheduled_tasks/document_canary_task.rb
```diff
@@ -0,0 +1,17 @@
+# https://github.com/ssoroka/scheduler_daemon for help
+class DocumentCanaryTask < Scheduler::SchedulerTask
+  every '120m'
```
jsugarman:
```suggestion
  every '1d', first_at: Chronic.parse('next 3:00 am')
```

jsugarman: I think we talked about making fewer of these @jrmhaig

jsugarman: i think we talked about making fewer of these

jsugarman (changes requested):
If we can just make 1 a day this can go in @jrmhaig

jsugarman (dismissed):
👍

jrmhaig:
(Blocker) There is a daily task that deletes unattached evidence documents (see `scheduled_tasks/document_cleaner_task.rb`) and this needs to be modified before this PR can be merged.

mpw5:
> (Blocker) There is a daily task that deletes unattached evidence documents (see `scheduled_tasks/document_cleaner_task.rb`) and this needs to be modified before this PR can be merged.

Would one alternative to changing `scheduled_tasks/document_cleaner_task.rb` be to schedule the `DocumentCanaryTask` to run immediately after the `DocumentCleanerTask`? As the cleaner runs at 04:00 daily, the Canarytoken could be generated at 04:01. That way there would always be a Canarytoken in the database.

If we want multiple tokens to exist simultaneously, then `DocumentCanaryTask` could be changed to batch create a few, or could run several times a day?

AnnaC-C:
Closed due to redundant code

PR #4521, opened 2021-12-23
Fix provider_delegator nil error on case_details page

jsugarman:
#### What
Fix provider_delegator nil error on case details page

#### Why
Long standing bug that causes a 500 error on production environments 
in certain scenarios:

To replicate:
  1. login as a provider that is a chamber with more than one user
  2. start a advocate final claim (others too)
  3. on case details page select a case type for a final/graduated fee
     ,such as a "Trial"
  4. hit "Save and continue" --> 500 error (NoMethod error)

This is because:
  1. Having more than one user in a provider means the case details page requires you to select an external user ("Choose an advocate").
  2. By selecting a case type that is "final/graduated" fee the base_claim model will instantiate basic fees. The saving process (before_validation) will attempt to set the `apply_vat` flag on the claim which will, in turn, atttempt to determine whether the external user associated with the claim is vat_registered or not.
  3. Since the user has not selected an external user yet this causes a `nil` `provider_delegator` to receive the message `vat_registered?`, raising the NoMethod error thereby cuasing a 500.

IMPORTANT: need to check impact on VAT setting. Specifically
if the user does not provider an external user AND does provide
a final/gradfee case type and submit THEN does it set the `apply_vat`
flag on the claim permanently (or is it updated when they then correct
the submission by select an external user)

#### How
Add safe navigator on provider_delegator. This means vat_registered will default to `nil/false` if the
claim creator does not select an external user for the claim. Need to check that once they select the user it updates this to use the external_users `vat_registered?` status

### DIFF
--- a/spec/models/claim/base_claim_spec.rb
+++ b/spec/models/claim/base_claim_spec.rb
@@ -772,10 +772,9 @@ RSpec.describe MockBaseClaim do
     subject(:registered) { claim.vat_registered? }
 
     let(:claim) { described_class.new }
-    let(:mock_provider_delegator) { provider_delegator }
 
     before do
-      allow(claim).to receive(:provider_delegator).and_return(mock_provider_delegator)
+      allow(claim).to receive(:provider_delegator).and_return(provider_delegator)
       allow(LogStuff).to receive(:error)
     end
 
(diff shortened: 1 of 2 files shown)


jsugarman:
Have sanity tested this on staging with:

1.  a user who is not registered for VAT
2. a user who is registered for VAT

In both instances i created an advocate final claim, triggered what would have caused the 500 error previously, but which this fixes. Then continued to create the claim with the above user and got the expected no VAT or VAT added results that are expected.

So it looks like this can be merged as a bug fix without regressions or other problems arising.

spec/models/claim/base_claim_spec.rb
```diff
@@ -772,10 +772,10 @@ def make_date_params(date_string)
     subject(:registered) { claim.vat_registered? }
 
     let(:claim) { described_class.new }
-    let(:mock_provider_delegator) { provider_delegator }
+    # let(:mock_provider_delegator) { provider_delegator }
```
jsugarman: remove

naseberry (approved):
LGTM, except for your comment

jrmhaig (approved):
This looks OK, subject to the removal of the commented line in `spec/models/claim/base_claim_spec.rb`.

I did, however, notice another bug related to this. See https://dsdmoj.atlassian.net/browse/CFP-375

PR #4523, opened 2021-12-23
fix deprecations when running cukes

jsugarman:
#### What
Fix some deprecations noticed when running cucumber

- Fix cucumber deprecation warning
- Fix ruby 2.7 kwarg deprecation warnings coming out in cucumber

#### Why
prepare for cucumber 8.0 and ruby 3.0


naseberry (approved):
LGTM

As discussed, kwarg deprecations not highlighted for other CCCD helpers, will be updated later

PR #4581, opened 2022-01-17
[CFP-348] Persist 100% traffic to live cluster cccd-production

jsugarman:
#### What
Persist 100% traffic to live cluster cccd-production

#### Why
100% traffic already directed to live cluster. This is to keep it 
that way.

#### Ticket

[CFP-348](https://dsdmoj.atlassian.net/browse/CFP-348)


jrmhaig (approved):
🎉

PR #4585, opened 2022-01-17
[CFP-373] Calculate weekly stats using 'transitioned at' date

jrmhaig:
#### What

Use 'transitioned at' date instead of 'originally submitted at' for counting intake stats in the weekly MI stats report.

#### Ticket

[MI weekly stats - intake generation](https://dsdmoj.atlassian.net/browse/CFP-373)

#### Why

Feedback from Jemma Woodhouse, LGFS supervisor, is that within the weekly MI stats report, they need intake stats to be based on the "transitioned at" column rather than "originally submitted at."

#### How

The 'transitioned at' date is added to the journeys query so that it can be used in `daily_report_count_query.rb`. For the time being, the results using `transitioned at` and `originally submitted at` are both included but the latter can be removed later.

--------

#### TODO (wip)

 - [X] Find 'transitioned at' date in the SQL query in `app/services/stats/management_information/concerns/journey_queryable.rb`
 - [x] Use the `transitioned_at` value from the journey query in the MI stats


AnnaC-C:
Closed due to being redundant

PR #4605, opened 2022-01-19
[CFP-54] Controller workaround for govuk_uk_date_field multiparameter assignment errors

jsugarman:
#### What
Handle `ActiveRecord::MultiparameterAssignmentErrors` for date
fields when supplying args such as 32/01/2021 or A/01/2021 or
1/13/2021.

Applies this to `certification` controller, which already errors in
production, plus the external_users/claims_controller form so that
we can progress case_details form step migration.

This should not have side affects since it only "cleans" params
that have an 'attribute_name(2i)|attribute_name(3i)' parameter
in the params of the request.

It handles deeply nested multiparameter dates using a recursive loop
over the entire params.

To apply to another controller we just need to add a `before_action`
such as below. The prepend was necessary for the claims controller
as other `before_actions` were beating it to punch and thereby encountering
the error.
```
  prepend_before_action :clean_multiparameter_dates, only: [:create, :update]
```

NOTE: Rails implicit parsing as Time or Time.zone.local bumps forward
certain dates automatically. This is default behaviour for rails but is
not necessarily desirable.
e.g.

31-09-2021 --> 01-10-2021
29-01-2021 --> 01-03-2021

We could prevent this behaviour here as well?!

#### Ticket

[CFP-54](https://dsdmoj.atlassian.net/browse/CFP-54)

#### Why
Currently entering invalid args such as 32/01/2021 or A/01/2021 or
1/13/2021 results in 500 errors.

### DIFF
--- a/features/step_definitions/certification_steps.rb
+++ b/features/step_definitions/certification_steps.rb
@@ -1,0 +1,37 @@
+Then('I should be on the certification page') do
+  expect(@certification_page).to be_displayed
+end
+
+Then('certified by should be set to current user name') do
+  expect(@certification_page.certified_by.value).to eql(@current_user.name)
+end
+
+Then('certification date should be set to today') do
+  dd, mm, yyyy = Time.current.strftime('%-d-%-m-%Y').split('-')
+  expect(@certification_page.certification_date.day.value).to eql(dd)
+  expect(@certification_page.certification_date.month.value).to eql(mm)
+  expect(@certification_page.certification_date.year.value).to eql(yyyy)
+end
+
+When('I check “I attended the main hearing”') do
+  @certification_page.attended_main_hearing.click
+end
+
+When('I fill in {string} as the certification date day') do |day|
+  @certification_page.certification_date.day.set(day)
+end
+
+When('I fill in todays date as the certification date') do
+  dd, mm, yyyy = Time.current.strftime('%-d-%-m-%Y').split('-')
+  @certification_page.certification_date.day.set(dd)
+  @certification_page.certification_date.month.set(mm)
+  @certification_page.certification_date.year.set(yyyy)
+end
+
+When('I click Certify and submit claim') do
+  allow(Aws::SNS::Client).to receive(:new).and_return Aws::SNS::Client.new(region: 'eu-west-1', stub_responses: true)
+  @certification_page.wait_until_certify_and_submit_claim_visible
+  patiently do
+    @certification_page.certify_and_submit_claim.click
+  end
+end
--- a/spec/requests/external_users/certification_spec.rb
+++ b/spec/requests/external_users/certification_spec.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Certification', type: :request do
+  let(:advocate) { create(:external_user, :advocate) }
+  let(:claim) { create(:advocate_final_claim) }
+  let(:today_parts) { %i[dd mm yyyy].zip(Time.current.strftime('%-d-%-m-%Y').split('-')).to_h }
+
+  describe 'POST #create' do
+    let(:certification_params) do
+      {
+        'claim_id' => claim.id,
+        'commit' => 'Certify and Submit Claim',
+        'certification' => {
+          'certification_type_id' => '1',
+          'certified_by' => 'Joe Bloggs',
+          'main_hearing' => 'true',
+          'certification_date(3i)' => today_parts[:dd],
+          'certification_date(2i)' => today_parts[:mm],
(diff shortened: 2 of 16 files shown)


features/step_definitions/certification_steps.rb
```diff
@@ -0,0 +1,37 @@
+Then('I should be on the certification page') do
+  expect(@certification_page).to be_displayed
+end
+
+Then('certified by should be set to current user name') do
+  expect(@certification_page.certified_by.value).to eql(@current_user.name)
+end
+
+Then('certification date should be set to today') do
+  dd, mm, yyyy = Time.current.strftime('%d-%-m-%Y').split('-')
```
jrmhaig: I presume this should be;

```suggestion
  dd, mm, yyyy = Time.current.strftime('%-d-%-m-%Y').split('-')
```

*Updated;* `%-d` and `%-m` to format day and month without leading zeros.

features/step_definitions/certification_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+When('I check “I attended the main hearing”') do
+  @certification_page.attended_main_hearing.click
+end
+
+When('I fill in {string} as the certification date day') do |day|
+  @certification_page.certification_date.day.set(day)
+end
+
+When('I fill in todays date as the certification date') do
+  dd, mm, yyyy = Time.current.strftime('%d-%-m-%Y').split('-')
```
jrmhaig: As above;

```suggestion
  dd, mm, yyyy = Time.current.strftime('%-d-%-m-%Y').split('-')
```

*Updated;* `%-d` and `%-m` to format day and month without leading zeros.

spec/requests/external_users/certification_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+              'certification_date(2i)' => 'JAN',
+              'certification_date(1i)' => '2021'
+            }
+          }
+        end
+
+        it 'invalid multiparameter date parts do not raise ActiveRecord::MultiparameterAssignmentErrors' do
+          post external_users_claim_certification_path(claim.id), params: certification_params
+          expect(response).to render_template(:new)
+        end
```
jrmhaig: For a test with this name I would expect something like;

```ruby
        it 'invalid multiparameter date parts do not raise ActiveRecord::MultiparameterAssignmentErrors' do
          expect { post external_users_claim_certification_path(claim.id), params: certification_params }
            .not_to raise_error(ActiveRecord::MultiparameterAssignmentErrors)
        end
```

Would it be better named something like "renders the new certification template" and have another test for the correct status code?

jrmhaig (commented):
I few minor comments. I haven't tested it yet.

features/step_definitions/certification_steps.rb
jsugarman: ooh yes. that would flicker on 1st-9th days otherwise i believe.

features/step_definitions/certification_steps.rb
jsugarman: Not important here as the leading zeros will be stripped but will update for consistency with other step

spec/requests/external_users/certification_spec.rb
jsugarman: yes, but generally better to have positive tests to avoid false positives... and rubocop says
```
WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives,
```
will tweak anyway

jrmhaig: I would just change the name of the test. What is important is that the page is rendered correctly, not that an exception is not raised.

spec/requests/external_users/certification_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        expect(response).to redirect_to new_user_session_path
+      end
+    end
+
+    context 'when user is signed in' do
+      before do
+        sign_in advocate.user
+      end
+
+      context 'with invalid parameters' do
```
jsugarman: should have a valid params test too really

imtiazAhmd (approved):
Sanity checked on `dev-lgfs`. Tried with combination of invalid inputs for the date fields. and page is showing relevant errors rather than `500`

features/step_definitions/certification_steps.rb
jsugarman: tweaked

features/step_definitions/certification_steps.rb
jsugarman: tweaked

spec/requests/external_users/certification_spec.rb
jsugarman: done

PR #4630, opened 2022-01-28
Add AccountSetter

jrmhaig:
Copied the AccountSetter class from laa-cccd-scripts

This has been copied into `lib/utils/` which, as a subdirectory of `lib/`,
means that it will not be autoloaded. To use;

```ruby
require 'utils/account_setter`

accounts = AccountSetter.new(['user1@example.com', 'user2.example.com'])
accounts.report
```

#### What

Copy the `AccountSetter` class from [laa-cccd-scripts](/ministryofjustice/laa-cccd-scripts)

#### Ticket

N/A

#### Why

laa-cccd-scripts contains some utilities for managing user accounts but to use them in an environment it is necessary to copy on to the kubernetes pod. It is copied into this repository for convenience.

#### How

Copy `account_setter.rb` into `lib/utils`. Being in a subdirectory it does not get autoloaded so to use it has to be required:

```ruby
require 'utils/account_setter'
```

### DIFF
--- a/spec/lib/utils/account_setter_spec.rb
+++ b/spec/lib/utils/account_setter_spec.rb
@@ -1,0 +1,128 @@
+require 'rails_helper'
+
+RSpec.describe AccountSetter do
+  subject(:instance) { described_class.new(emails) }
+
+  let(:emails) { [email] }
+  let(:email) { 'test@example.com' }
+  let(:user) { create(:user, id: 99, email: 'test@example.com', password: 'testing123') }
+
+  before { allow($stdout).to receive(:puts) }
+
+  describe '#report' do
+    subject(:report) { instance.report }
+
+    let(:external_user) { create(:external_user, user: user, provider: create(:provider, name: 'Test provider')) }
+
+    before { create_list(:claim, 5, external_user: external_user) }
+
+    it 'reports on the user' do
+      expect { report }.to output(
+        /User test@example\.com with id 99 have 5 claims for their provider, "Test provider"/
+      ).to_stdout
+    end
+
+    it { expect { report }.to output(/No deleted users found for email "test@example\.com\.deleted\..*"/).to_stdout }
+
+    context 'with an unknown user' do
+      let(:emails) { ['unknown@example.com'] }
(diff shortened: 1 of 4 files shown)


jsugarman (changes requested):
Good to go - i have approved on codeclimate -  but can you just exclude the file from rubocop for the stdout violation or just exclude the whole file generally. Alternatively you could tweak for complexity issues

jsugarman (approved):
👍 am approving as will be off monday and tuesday am, but see comment re rubocop

spec/lib/utils/account_setter_spec.rb
```diff
@@ -0,0 +1,128 @@
+require 'rails_helper'
+
+RSpec.describe AccountSetter do
+  subject(:account_setter) { described_class.new(emails) }
```
jsugarman: **nit** one of the benefits of described class is it is not tied to a particular name, so if the actual class names changes you can just change the `RSpec.describe` to the new/changed class name and the tests still pass. In that context i would make the named subject something more generic like `instance`
```
subject(:instance) { described_class.new(emails) }
```

spec/lib/utils/account_setter_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+      it { expect { un_soft_delete }.to output(/User email "unknown@example\.com\.delete\.%" not found/).to_stdout }
+    end
+  end
+
+  describe '#change_password' do
+    subject(:change_password) { account_setter.change_password }
+
+    it 'changes the password of the user' do
+      expect { change_password }.to change { user.reload.valid_password?('testing123') }.from(true).to(false)
```
jsugarman, line 119: nice way of testing that 👍

jsugarman (approved):
can go in as is or you can make/apply the nit/suggested tweak

spec/lib/utils/account_setter_spec.rb
jrmhaig: :+1: I'll change that.

PR #4631, opened 2022-01-28
An example and alternative controller specs to cover date handling

jsugarman:
#### What
An example of different controller specs to replace
or amend those existing

#### Why
In line with need to clear invalid date params


jsugarman:
feel free to discuss which is best approach and amend as you see fit. This can then be applied to the other TODOs in controller specs

jsugarman:
closing as was just example and handled in CFP-54-case-details-form branch

PR #4632, opened 2022-01-31
[CFP-64] FormBuilder migration - Offence details

imtiazAhmd:
#### What
Migrate external_users form step: transfer_details form

#### Ticket

[FormBuilder migration - Offence details](https://dsdmoj.atlassian.net/browse/CFP-64)

--------

#### TODO

- [x]  All components replaced with GovUK FormBuilder components
- [x]  Show correct error messages
- [x]  Tests passes as normal


jsugarman:
Offence classes for LGFS and scheme 9 AGFS are not rendering as expected/required

**old/production** 

Screenshot 2022-02-01 at 16 09 41

**current new:**
Screenshot 2022-02-01 at 16 12 15

app/controllers/offences_controller.rb
```diff
@@ -13,21 +13,19 @@ class OffencesController < ApplicationController
...
-
-      respond_to do |format|
-        format.json do
-          render json: @offences, each_serializer: FeeReform::OffenceSerializer
-        end
-      end
-    else
-      @offences = Offence.in_scheme_nine.where(offence_filter)
+    respond_to do |format|
+      format.json { render json: offences, each_serializer: FeeReform::OffenceSerializer }
```
jsugarman: name wise if we are using this serializer for all offences it should not be namespaced to `FeeReform`, which relates to AGFS fee scheme 10+ only

imtiazAhmd:
> Offence classes for LGFS and scheme 9 AGFS are not rendering as expected/required
> 
> **old/production**
> 
> Screenshot 2022-02-01 at 16 09 41
> 
> **current new:** Screenshot 2022-02-01 at 16 12 15

This should be resolved now

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
```diff
@@ -19,19 +19,16 @@
...
-    %a.fx-clear-search.hidden{ href: '#noop' } Clear search
+= f.hidden_field :offence_id, value: valid_offence ? f.object.offence.id : nil, class: 'fx-model'
+= f.hidden_field :fee_scheme, value: [@claim.fee_scheme.name, @claim.fee_scheme.version], class: 'fx-fee-scheme'
 
-
-    %input#offence.form-control.fx-input{ name: 'offence-search-input', type: 'input', autocomplete: 'off' }
-
-  = f.hidden_field :offence_id, value: valid_offence ? f.object.offence.id : nil, class: 'fx-model'
-  = f.hidden_field :fee_scheme, value: [@claim.fee_scheme.name, @claim.fee_scheme.version], class: 'fx-fee-scheme'
+= govuk_link_button_secondary('Clear search', '#noop', class: 'fx-clear-search hidden')
```
jrmhaig: It looks as though `fx-clear-search` was used in `app/webpack/javascripts/modules/Modules.OffenceSearchInput.js` but that is no longer the case. Does it still need to be here (and in `Modules.OffenceSearchView.js` on line 149)?

app/views/external_users/claims/offence_details/_default_fields.html.haml
```diff
@@ -1,25 +1,18 @@
...
-
-.offence-class-select.form-group
-  = render partial: 'external_users/claims/offence_details/offence_select', locals: { offences: @offences }
+  = f.govuk_collection_select :offence_category,
+    @offence_descriptions,
+    :description, :description,
+    form_group: { classes: 'fx-autocomplete-wrapper', id: 'cc-transfer-stage js-claim-offence-category-description' },
+    hint: { text: t('offence_category_hint', scope: locale_scope)},
+    label: { text: t('offence_category', scope: locale_scope)},
+    options: { include_blank: '', selected: selected_offence_category }
```
jrmhaig, line 14: Can this be indented for clarity.

```suggestion
  = f.govuk_collection_select :offence_category,
                              @offence_descriptions,
                              :description, :description,
                              form_group: { classes: 'fx-autocomplete-wrapper', id: 'cc-transfer-stage js-claim-offence-category-description' },
                              hint: { text: t('offence_category_hint', scope: locale_scope)},
                              label: { text: t('offence_category', scope: locale_scope)},
                              options: { include_blank: '', selected: selected_offence_category }
```

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
```diff
@@ -19,19 +19,16 @@
...
-.form-group.mod-search-input
-  %label.form-label{ for: 'offence' }
-    = 'Search for the offence'
-    %span.form-hint
-      = 'For example class, band, offence or act name'
+= f.govuk_text_field :offence_search_input,
+  class: 'fx-input',
+  form_group: { classes: 'mod-search-input' },
+  hint: { text: 'For example class, band, offence or act name' },
+  label: { text: 'Search for the offence' }
```
jrmhaig: Indentation;

```suggestion
= f.govuk_text_field :offence_search_input,
                     class: 'fx-input',
                     form_group: { classes: 'mod-search-input' },
                     hint: { text: 'For example class, band, offence or act name' },
                     label: { text: 'Search for the offence' }
```

app/views/external_users/claims/offence_details/_offence_select.html.haml
```diff
@@ -1,3 +1,6 @@
-= label_tag 'offence_class_description', t('.offence_class'), { class: 'form-label' }
-- selected_offence = @claim.try(:offence) ? @claim.offence.id : (offences.count == 1 ? offences.first : nil)
-= collection_select :offence_class, :description, offences, :id, :offence_class, options = { selected: selected_offence }, html_options = { class: 'form-control', disabled: offences.count == 1 }
+= f.govuk_collection_select :offence_class,
+  offences.map { |offence| [offence.id, offence.offence_class.description] },
+  :first,
+  :last,
+  form_group: { classes: 'js-offence-class-select'},
+  label: { text: t('.offence_class')}
```
jrmhaig: Indentation;

```suggestion
= f.govuk_collection_select :offence_class,
                            offences.map { |offence| [offence.id, offence.offence_class.description] },
                            :first,
                            :last,
                            form_group: { classes: 'js-offence-class-select'},
                            label: { text: t('.offence_class')}
```

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
jsugarman: it is still referenced at line 17 of that file - to set a handle to this button by the looks of it

jsugarman (commented):
While reviewing offence migration i notice that the "Clear selection" button does not appear to be doing anything in both master and on the offence migration branch on dev-lgfs. This will need looking into. may be another impact of previous migration work?!

Screenshot 2022-02-07 at 08 46 39

jsugarman:
The AGFS scheme 9 offences are still not working as expected. While creating a new claim or scheme 9 you get and  can select the appropriate  class options. but If you select one and progress to next page then go back to edit it it 1) gives a different description 2) does not populate the class list again so you can not change it



Screenshot 2022-02-07 at 08 54 47

jsugarman:
In addition the description appears to be wrong 

~~I had saved as class K (value in excess of 100k) but when editing it it says class F (value in excess of 100k) (which is not a class that exists)~~ my mistake the description is correct but should just be proceeded by
```
K: ...
```
Screenshot 2022-02-07 at 09 02 24

app/views/external_users/claims/offence_details/_default_fields.html.haml
```diff
@@ -1,25 +1,18 @@
 - locale_scope = 'external_users.claims.offence_details.fields'
-#cc-offence.form-group{ class: error_class?(@error_presenter, :offence_category) ? 'form-group-error' : '' }
-  %a#offence_category
-  %label.form-label-bold{ for: 'claim_offence_category_description' }
```
naseberry, line 4: `claim_offence_category_description` is this identifier still required in `features/page_objects/claim_form_page.rb`

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
naseberry: Are we keeping our custom style for `.fx-clear-search` or are we using the GOVUK defaults?

imtiazAhmd: we are not using custom style for it, but this class staying for JS selection. Custom style removed in 'Cleanup dead codes' commit

app/views/external_users/claims/offence_details/_default_fields.html.haml
imtiazAhmd: removed as suggested in `cleanup dead codes` commit

app/validators/claim/advocate_claim_validator.rb
```diff
@@ -45,7 +45,7 @@ def self.fields_for_steps
 
   def validate_offence
     return if fixed_fee_case?
-    error_message = @record.agfs_reform? ? 'new_blank' : 'blank'
+    error_message = @record.agfs_reform? ? :new_blank : 'blank'
```
jsugarman: should this not be `:blank` for it too pickup the error - 

```suggestion
    error_message = @record.agfs_reform? ? :new_blank : :blank
```

app/validators/claim/advocate_hardship_claim_validator.rb
```diff
@@ -48,7 +48,7 @@ def validate_case_stage_id
   end
 
   def validate_offence
-    error_message = @record.agfs_reform? ? 'new_blank' : 'blank'
+    error_message = @record.agfs_reform? ? :new_blank : 'blank'
```
jsugarman: [ditto](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4632/files#r800874658)

config/locales/en/models/claim/offence_details.yml
```diff
@@ -0,0 +1,22 @@
...
+          attributes:
+            offence:
+              blank: Choose an offence category
+              new_blank: Choose an offence
+
+        claim/advocate_interim_claim:
+          attributes:
+            offence:
+              blank: Choose an offence category
+              new_blank: Choose an offence
```
jsugarman: whole file should be named after the model `offence` to follow convention.

Also, since they are all the same error messages you can have just one set to go under `claim/base_caim` instead?

```
        claim/base_claim:
          attributes:
            offence:
              blank: Choose an offence category
              new_blank: Choose an offence
```

you could/should also arguably put these directly in `claim.yml`  anyway, as they are not related on the offence model but the claims foreign key attribute of offence

jsugarman (changes requested):
Functions great from what I can tell but a few issues related to error messages and their location in particular.

The error do appear to be displaying correctly for both LGFS, old-style and AGFS new style but, a) not sure how for the old style presence checker b) the message locations and keys need looking into.

jsugarman:
I am seeing a lot of left over error messages relating to offences in `error_messages.yml` and in validators. 

These need to all be removed if you believe they are unneeded or migrated to the new validation and form builder mechanism as covered by this guide:
https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/docs/govuk-formbuilder_migration.md

jsugarman:
Also, since the fix for the broken "Clear selection" has gone in to this branch it will need a regression feature test to cover it - It probably should have been done on a separate branch. I will be looking into that as a branch of this.

config/locales/en/models/claim/offence_details.yml
jsugarman: to clarify these should go in `claim.yml`. I also think, as a tangent, that the `transfer_details.yml` should actually be in `claim.yml` but that should be done on a separate branch.

naseberry: The idea here is to start separating each form step into its own `yml` file as to not have another monolith like the `en.yml` file

naseberry: > to clarify these should go in `claim.yml`. I also think, as a tangent, that the `transfer_details.yml` should actually be in `claim.yml` but that should be done on a separate branch.

The idea here is to start separating each form step into its own `yml` file as to not have another monolith like the `en.yml` file

jsugarman: the convention is to have it by model, and i would keep to that. If i was looking for errors on the claim object or another object i would expect to find them in there. The argument about en.yml does not hold up because that was literally everything, not just models. The file structure could/should reflect the yml structure so you could have all model errors in `models/claim/error.yml` or `errors/models/claim.yml` for example. Using an view related naming logic to define model error translations is likely to just get confusing

jsugarman (approved):
👍 LGTM

Have sanity check on hosted env again too. 

I note that i cannot find any reference in validators to the offence  `:invalid` error. indicating it may not be needed either after all - i'd still keep it in though.

jrmhaig (approved):

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
```diff
@@ -16,22 +16,19 @@
...
-  %label.form-label{ for: 'offence' }
-    = 'Search for the offence'
-    %span.form-hint
-      = 'For example class, band, offence or act name'
+= f.govuk_text_field :offence,
+  value: nil,
+  class: 'fx-input',
+  form_group: { classes: 'mod-search-input' },
+  hint: { text: 'For example class, band, offence or act name' },
+  label: { text: 'Search for the offence' }
```
naseberry: Minor: Could we move these text values to the relevant locale file

naseberry (approved):
💯

app/views/external_users/claims/offence_details/_fee_reform_fields.html.haml
imtiazAhmd: Done! 👍🏼

PR #4426, opened 2021-11-15
[CFP-54] FormBuilder migration - Case details

naseberry:
#### What

Migrate external_users form step: **case_details** form

#### Ticket

[FormBuilder migration - Case details](https://dsdmoj.atlassian.net/browse/CFP-54)

#### Why

- Update the service to use GOV.UK styles, components and patterns
- To improve and resolve the service's accessibility issues
- To reduce code complexity
- To improve maintainability

PR #4633, opened 2022-01-31
[CFP-54] Tidy some `Claim::AdvocateClaim` tests

jrmhaig:
#### What

Refactor some `Claim::AdvocateClaim` tests.

#### Ticket

[FormBuilder migration - Case details](https://dsdmoj.atlassian.net/browse/CFP-54)

#### Why

Clean up some of the tests that are being touched in #4426 (this is not a blocker to #4426 being merged into master).

#### How

* Some tests have names in the form "test X if Y". Change so that 'Y' is a context.
* Reduce unnecessary memoized variable following the splitting of tests into contexts.
* Use a named (and non-implicit) subject.
* Extract test setup into `before` and `let`.

### DIFF
--- a/spec/models/claim/advocate_claim_spec.rb
+++ b/spec/models/claim/advocate_claim_spec.rb
@@ -1046,19 +1046,35 @@ RSpec.describe Claim::AdvocateClaim, type: :model do
     end
   end
 
-  describe 'calculate_vat' do
-    it 'calaculates vat on submission if vat is applied' do
-      allow(VatRate).to receive(:vat_amount).and_return(10)
-      claim = create(:unpersisted_claim, :with_fixed_fee_case, total: 100)
-      claim.submit!
-      expect(claim.vat_amount).to eq 10
+  describe '#calculate_vat' do
+    subject { claim.vat_amount }
+
+    context 'when vat is applied' do
+      let(:claim) { create(:unpersisted_claim, :with_fixed_fee_case, total: 100) }
+
+      before do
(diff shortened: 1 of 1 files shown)


spec/models/claim/advocate_claim_spec.rb
```diff
@@ -23,31 +23,47 @@
     let(:same_provider_external_user) { create(:external_user, provider: provider) }
     let(:other_provider_external_user) { create(:external_user, provider: other_provider) }
 
-    it 'raises error message if no external user is specified' do
-      subject.external_user_id = nil
-      expect(subject).not_to be_valid
-      expect(subject.errors[:external_user_id]).to eq(['Choose an advocate'])
+    context 'with no external user' do
+      before do
+        subject.external_user_id = nil
```
jsugarman: **nit/extension-of-refactor**: as a refactor I would also use named subject, but i see this does not define one explictly

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -17,53 +17,36 @@
   specify { expect(subject.supplementary?).to be_falsey }
 
   describe 'validates external user and creator with same provider' do
+    subject(:claim) { described_class.create(external_user: external_user, creator: creator) }
```
jsugarman, line 20: 👍 ahh, just saw this on this commit so can ignore previous comment

jsugarman (approved):
👍

PR #4635, opened 2022-02-01
Move docker scan success alerting channel

jsugarman:
#### What
Move docker scan alerting channel.

Also remove unecessary output of step
status. It was just used during testing.

#### Why
To keep the alerts channel for failures only
and thereby avoid noise. It would however still
be of use to easily confirm that the scan is
functioning as expected.


jrmhaig (approved):
:+1:

naseberry (approved):

PR #4650, opened 2022-02-03
Fix JS caching issue

imtiazAhmd:
#### What
Improve web caching of Front End assets
#### Ticket

[Fix application assets caching issue](https://dsdmoj.atlassian.net/browse/CFP-393)

#### Why
Users are affected by stale caches upon production releases that contain Front End asset changes and cause issues where user journeys are affected.

#### Acceptance criteria
**Given** we release Front End asset changes
**When** users make a web request
**Then** they should immediately get the latest Front End assets

--------

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -175,7 +175,7 @@ module ApplicationHelper
     request.GET[:tab] == tab
   end
 
-  def app_javascript_include_tag(path = nil)
+  def app_javascript_include_tag(path)
     # create an array of all the files
     files = Dir.glob(path)
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -174,4 +174,15 @@ module ApplicationHelper
     tab = tab_present(path)
     request.GET[:tab] == tab
   end
+
+  def app_javascript_include_tag(path = nil)
+    # create an array of all the files
+    files = Dir.glob(path)
+
+    # sort array by last modified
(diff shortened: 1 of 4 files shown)


spec/support/jasmine-browser.json
```diff
@@ -1,8 +1,7 @@
 {
   "srcDir": ".",
   "srcFiles": [
-    "public/packs/js/application.bundle.js",
-    "public/packs-test/js/application.bundle.js"
+    "public/packs-test/js/application-bundle-*.js"
```
jsugarman, line 4: i thought you were looking to get rid of `packs-test`

naseberry: We are keeping it for now 👍

app/helpers/application_helper.rb
```diff
@@ -174,4 +174,15 @@ def tab_matches(path)
     tab = tab_present(path)
     request.GET[:tab] == tab
   end
+
+  def app_javascript_include_tag(path = nil)
```
jrmhaig: `path` should be a required parameter as it is required for `Dir.glob`.

```suggestion
  def app_javascript_include_tag(path)
```

jsugarman (approved):
👍 Looks good and works fine on dev-lgfs

Minor tweaks possible but can do after the fact

```
    # sort array by last modified
    file = files.sort { |a, b| File.mtime(b) <=> File.mtime(a) }.first

    # return the last modified file
    javascript_include_tag file.gsub(/^public/, ''), defer: true
```

PR #4653, opened 2022-02-03
Revert newer method to accommodate IE11

imtiazAhmd:
#### What
Our new Autocomplete wrapper uses `nodeList.forEach` and this is supported by all the major browsers other than Internet Explorer. This causes an error in IE even the user is not on a page that uses this component. 

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


naseberry (approved):
Hopefully once IE reaches it's EoL, we can revert this revert :)

jsugarman (approved):
👍 have you tested this does not break stuff

PR #4654, opened 2022-02-07
[CFP-295] Use kwargs instead of options hash

jrmhaig:
#### What

Some minor changes to ease the move to Ruby 3.

#### Ticket

#### Why

Prior to Ruby 3.0 it was possible to use the final positional argument to a function as a hash of keyword arguments. This has been removed in Ruby 3.0.

#### How

Update `app/services/stats/management_information/daily_report_count_generator.rb`, `app/services/stats/management_information/daily_report_generator.rb` and `app/services/stats/management_information_generator.rb` to handle arguments consistently and then change `app/services/stats/stats_report_generator.rb` so that the arguments are passed correctly as keyword arguments and not a hash.

Additionally, use the `...` syntax (added in Ruby 2.7) to pass through arguments in the `self.call` methods. See [this article](https://ryanbigg.com/2021/07/ruby-27s-new-triple-dot-syntax) for details.

### DIFF
--- a/app/services/stats/management_information/daily_report_count_generator.rb
+++ b/app/services/stats/management_information/daily_report_count_generator.rb
@@ -7,8 +7,8 @@ module Stats
     class DailyReportCountGenerator
       include StuffLogger
 
-      def self.call(**kwargs)
-        new(kwargs).call
+      def self.call(...)
+        new(...).call
       end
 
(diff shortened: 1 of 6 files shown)


jsugarman (approved):
👍

PR #4673, opened 2022-02-15
Disable and enable user buttons

jrmhaig:
#### What

Add options to enable and disable users in CCCD.

#### Ticket

TBD. Preparation work: [Review how we block users in CCCD](https://dsdmoj.atlassian.net/browse/CFP-379)

#### Why

We have had a number of requests to disable and then re-enable users in CCCD. The current process involves a dev logging in to the server and running commands from the Rails console. There needs to be a way for non-technical stakeholders can do this.

#### How

See [Review how we block users in CCCD](https://dsdmoj.atlassian.net/browse/CFP-379)

--------

#### TODO (wip)

 - [ ] Enable/disable user endpoints
   - [ ] Ensure access to these endpoints is suitably restricted
 - [ ] Add enable/disable buttons to external users 'show' pages
   - [ ] Ensure buttons are only visible to eligible users


jrmhaig:
Replaced by #4705

PR #4693, opened 2022-02-22
[CFP-295] Use kwargs instead of options hash

jrmhaig:
#### What

Some minor changes to ease the move to Ruby 3. See #4654 

#### Ticket

#### Why

Prior to Ruby 3.0 it was possible to use the final positional argument to a function as a hash of keyword arguments. This has been removed in Ruby 3.0.

#### How

To avoid issue changes are being made gradually. This PR updates files related to the generation of MI stats reports.


jsugarman (approved):
👍

PR #4694, opened 2022-02-22
Set surnames of test users

jrmhaig:
#### What

Fix one of the flickering tests.

#### Ticket

N/A

#### Why

One of the tests (spec/controllers/external_users/claims_controller_spec.rb:475) flickers because it is searching for the advocates surname and expecting nothing to be returned but the surname is randomly chosen and 'Smith' is used as one of the other test user's name. See #4452 for the previous partial fix.

#### How

The solution is to fix the name of all the test users to ensure that a result is not found.


jsugarman (approved):
👍

jsugarman (approved):
👍

PR #4720, opened 2022-03-02
[CFP-508] Add disableable users

jsugarman:
#### What
- Add disabled_at column to users
- Add disablable mixin to user
- Delegate external user disablable to user
- Modify provider management to use disable/enable

#### Ticket

[CFP-508](https://dsdmoj.atlassian.net/browse/CFP-508)
[CFP-490](https://dsdmoj.atlassian.net/browse/CFP-490)

#### Why
To allow for disabling users without soft deleting them, which
has implications for API submission and allows new accounts
with the old email address to be created.

#### How
Use mixin with a disabled_at attribute that can be checked
by the `active_for_authentication?` devise override.


spec/support/shared_examples/models/shared_examples_for_disablable.rb
```diff
@@ -0,0 +1,162 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples 'a disablable object' do
+  let(:factory_name) { described_class.name.underscore.to_sym }
```
jrmhaig: FYI, for a class with a namespace, such as `Claim::AdvocateClaim` this will result in `:"claim/advocate_claim"` and I don't think that FactoryBot would be able to deal with that unless we rename lots of factories. Probably not an immediate problem but may become an issue if we start using this for other classes.

spec/support/shared_examples/models/shared_examples_for_disablable.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    before do
+      create(factory_name, disabled_at: nil)
+      create(factory_name, disabled_at: Time.zone.now)
+    end
+
+    it 'returns records with disabled_at present' do
+      is_expected.to all(have_attributes(disabled_at: be_kind_of(Time)))
+    end
+  end
```
jrmhaig: I think it is clearer if the instances are explicitly matched. Eg,

```suggestion
  describe '.disabled' do
    subject { described_class.disabled }
    
    let!(:disabled_user) { create(factory_name, disabled_at: Time.zone.now) }

    before do
      create(factory_name, disabled_at: nil)
    end

    it 'returns records with disabled_at present' do
      is_expected.to match_array([disabled_instance])
    end
  end
```

spec/support/shared_examples/models/shared_examples_for_disablable.rb
jsugarman: ok, yes. can just use [demodulize](https://apidock.com/rails/v6.1.3.1/String/demodulize) here

spec/support/shared_examples/models/shared_examples_for_disablable.rb
jsugarman: have tweaked this. although not keen on the separation of setup!

features/enable_disable_users.feature
```diff
@@ -58,28 +52,23 @@ Scenario: Super admin can enable user
...
     Then I should be on the provider manager user change availability page
-    And the page should be accessible
-    And I should see 'Are you sure you want to enable John Doe?'
-    And I should see button 'Enable account'
 
     When I click the button 'Enable account'
     Then I should be on the provider manager user show page
-    And the page should be accessible
-    And I should see 'Live'
+    And I should see 'Live, Enable'
```
jsugarman: should be `Enabled`

app/models/user.rb
```diff
@@ -109,11 +110,17 @@ def unauthenticated_message
...
 
   def inactive_message
-    active? ? super : 'This account has been disabled.'
+    if active? && enabled?
+      super
+    elsif softly_deleted?
+      'This account has been deleted.'
+    elsif disabled?
+      'This account has been disabled.'
+    end
```
jrmhaig, line 123: Having different messages could provide a malicious user with information about which accounts to target. A generic message is better. Possibly something even more vague - e.g. 'Unknown user' - but this probably isn't for us to decide here.

```suggestion
    return 'This account has been disabled' unless active? && enabled?

    super
```

lib/utils/account_setter.rb
```diff
@@ -90,18 +82,54 @@ def un_soft_delete
...
+  def disable
+    emails.each do |email|
+      user = User.find_by(email: email)
+
+      if user&.enabled?
+        user.disable
+        puts "User with email \"#{user.email}\" disabled!".green
+      else
+        puts "Enabled user with email \"#{email}\" not found!".red
+      end
```
jrmhaig, line 103: To provide more informative user feedback, especially if there comes a time when a constraint or validation prevents enabling a user;

```suggestion
      if user.nil?
        puts "User with email \"#{email}\" not found!".red
      elsif user.enabled?
        puts "User with email \"#{user.email}\" already enabled!".amber
      elsif user.disable
        puts "User with email \"#{user.email}\" disabled!".green
      else
        puts "Unable to disable user with email \"#{user.email}\"!".red
      end
```

NB;
* It might be possible to do this better than a multi-part `if`. Maybe extract to a separate method and use guard clauses?
* I'm guessing that an `amber` colour works but I haven't checked.
* Something similar can be done for the `enable` method, below.

lib/utils/account_setter.rb
```diff
@@ -23,29 +24,23 @@ def initialize(emails)
   # Find active or inactve accounts with matching email and
   # report basic information.
   #
-  # rubocop:disable Rails/Output
   # rubocop:disable Metrics/AbcSize
-  # rubocop:disable Metrics/MethodLength
-  def report
+  def report(format: nil)
+    report = []
```
jrmhaig, line 29: Prefer not to use the same name as the method name. Maybe `report_rows`?

spec/requests/provider_management/external_users_spec.rb
```diff
@@ -128,33 +160,57 @@
...
-      context 'when external user belongs to a different provider' do
+      context 'when successfull response' do
+        before { enable_user }
+
+        it { expect(external_user.reload).to be_enabled }
+        it { expect(response).to redirect_to provider_management_provider_external_user_path }
+        it { expect(flash[:notice]).to eq('User successfully enabled') }
+      end
+
+      # TODO: is this behaviour needed? when would a external user ever be from another provider at this point
```
jrmhaig, line 171: This isn't so much behaviour that is needed but rather guarding against badly formed urls. The update is called with a PATCH request to;

```
/provider_management/providers/:provider_id/external_users/:id/update_availability
```

where `:id` is the id of an external user belonging to the provider `:provider_id`. The test checks that with a badly formed url, where the `:id` does not correspond to `:provider_id`, then the behaviour is predictable and, specifically, no update is made. If the behaviour is not predictable then there is a possibility of a security exploit.

This is related to [CFP-484](https://dsdmoj.atlassian.net/browse/CFP-484) which is regarding viewing the external user page.

app/models/user.rb
mpw5: To be fixed in https://dsdmoj.atlassian.net/browse/CFP-533.

mpw5 (approved):
Reviewed with @jrmhaig and @NenehB.

PR #4723, opened 2022-03-08
[CFP-454] Use more RESTful form based approach

jsugarman:
To apply a more conventional approach
with somewhat more RESTful verb/action
mappings

### DIFF
--- a/app/controllers/provider_management/external_users_controller.rb
+++ b/app/controllers/provider_management/external_users_controller.rb
@@ -2,7 +2,7 @@ class ProviderManagement::ExternalUsersController < ApplicationController
   include PasswordHelpers
 
   before_action :set_provider, except: %i[find search]
-  before_action :set_external_user, only: %i[show edit update change_password update_password]
+  before_action :set_external_user, only: %i[show edit update change_password update_password change_availability update_availability]
   before_action :external_user_by_email, only: %i[search]
 
   def show; end
--- a/app/controllers/provider_management/external_users_controller.rb
+++ b/app/controllers/provider_management/external_users_controller.rb
@@ -65,7 +65,7 @@ class ProviderManagement::ExternalUsersController < ApplicationController
     end
   end
 
-  def confirmation
+  def change_availability
     if @external_user.active?
       render :disable_confirmation
     else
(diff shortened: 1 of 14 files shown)


app/controllers/provider_management/external_users_controller.rb
```diff
@@ -65,34 +66,23 @@ def update_password
...
-    end
+  def change_availability
+    render @external_user.active? ? :disable_confirmation : :enable_confirmation
   end
 
-  def enable
-    if (@external_user.provider == @provider) && @external_user.softly_deleted? && @external_user.un_soft_delete
-      redirect_to_show_page(notice: I18n.t('provider_management.external_users.enable_confirmation.success_message'))
+  def update_availability
+    if external_user_params[:availability].eql?('true')
```
jrmhaig: You can use `ActiveRecord::Type::Boolean.new.cast` to test boolean params, which would make this;

```ruby
    ActiveRecord::Type::Boolean.new.cast(params[:availablability]) ? enable : disable
```

(I cannot make this a 'suggestion' as Github doesn't see every line as changed)

spec/requests/provider_management/external_users_spec.rb
```diff
@@ -1,14 +1,20 @@
...
+    end
 
     let(:external_user) do
       create :external_user, provider: provider, user: create(:user, email: 'bubbletea@example.com')
     end
+
     let(:provider) { create :provider }
 
+    let(:params) { { external_user: { availability: 'false' } } }
+
```
jrmhaig: **(nit)** If the context is to disable the user (and likewise below, enable the user) you do not need to define another memoized value here, and the parameters could go straight in line 7.

jrmhaig (approved):
A couple of things you may choose to change but otherwise OK.

app/controllers/provider_management/external_users_controller.rb
jsugarman: have used `ActiveModel::Type::Boolean.new.cast` instead following suggestions on SO.

spec/requests/provider_management/external_users_spec.rb
jsugarman: done

spec/requests/provider_management/external_users_spec.rb
```diff
@@ -1,12 +1,61 @@
...
+    end
+
+    context 'when logged in as super_admin with a disabled user' do
+      let(:super_admin) { create :super_admin }
+      let(:user) { super_admin.user }
+
+      let(:external_user) do
+        eu = create(:external_user, provider: provider, user: create(:user, email: 'bubbletea@example.com'))
+        eu.soft_delete
+        eu
```
jrmhaig: **(nit)** This could be;

```suggestion
        create(:external_user, provider: provider, user: create(:user, email: 'bubbletea@example.com'))
          .tap(&:soft_delete)
```

PR #4735, opened 2022-03-10
[CFP-454] change flash message

jsugarman:
Change flash message from deleted to disabled


jsugarman:
closing as cherry picked into main CFP-454 PR

PR #4736, opened 2022-03-14
[CFP-527]- Paginate provider index page

NenehB:
#### What

The provider index page is very slow to load. 

#### Ticket

[CFP-527-Spike- Paginate provider index page](https://dsdmoj.atlassian.net/browse/CFP-527)

#### Why

Due to the large number of providers.

#### How

By paginating the list the load time can be improved greatly.

--------

#### TODO (wip)

 - [x] paginate provider page
 - [X] paginate caseworkers page

### DIFF
--- a/app/controllers/case_workers/admin/case_workers_controller.rb
+++ b/app/controllers/case_workers/admin/case_workers_controller.rb
@@ -1,5 +1,6 @@
 class CaseWorkers::Admin::CaseWorkersController < CaseWorkers::Admin::ApplicationController
   include PasswordHelpers
+  include PaginationHelpers
 
   before_action :set_case_worker, only: %i[show edit update destroy change_password update_password]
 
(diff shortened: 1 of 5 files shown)


naseberry (commented):
👍

jrmhaig:
From a technical point of view I'm happy to approve this **but** it is changing the view for case workers. In particular, to find a particular provider in the list it may be necessary to move forward multiple pages, especially as the 'find provider by email' does not actually find providers but users. It would be useful to get some user input before putting this live.

jsugarman:
> From a technical point of view I'm happy to approve this **but** it is changing the view for case workers. In particular, to find a particular provider in the list it may be necessary to move forward multiple pages, especially as the 'find provider by email' does not actually find providers but users. It would be useful to get some user input before putting this live.

^^ 👍  I agree, or sign off from product

mpw5 (approved):
LGTM. Did we ever get confirmation from the business that they are happy with the change? I think we should wait until that is confirmed before releasing.

mpw5:
Following the switch from Kaminari to Pagy for pagination, this PR is no longer valid so I'm closing it (sorry Neneh).

PR #4743, opened 2022-03-16
Replace deprecated postgres circleci db image

jsugarman:
#### What
Replace all deprecated circleci container images with modern convenience images

#### Ticket

n/a

#### Why
Warnings raised on circleci UI plus we need to bump to postgres 13.3 which is current 
production version in use.

see https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034
for details.


naseberry (approved):
Great 👍

PR #4752, opened 2022-03-21
Ensure attributes have ErrorMessage::Key type

jrmhaig:
#### What

Use `Key.new` for remainder of regex match.

#### Ticket

#### Why

In Ruby 2.7 `Regexp.last_match(n)` returns a `ErrorMessage::Key` when the original string is an `ErrorMessage::Key`. In Ruby 3.0 the matches are returned as Strings. The last 'remainder' from the regex match in `parse` needs to be an instance of `ErrorMessage::Key` so in Ruby 3 it needs to be created explicitly (and it also works in Ruby 2.7).

#### How

```ruby
attribute = Key.new(Regexp.last_match(4))
```

This will ensure that the value is an instance of `ErrorMessage::Key`. This works correctly whether `Regexp.last_match(4)` is already a `ErrorMessage::Key` or is a string.

### DIFF
--- a/app/presenters/error_message/key.rb
+++ b/app/presenters/error_message/key.rb
@@ -60,7 +60,7 @@ module ErrorMessage
         model.slice!('_attributes')
         model = model.singularize
         all_model_indices[model] = Regexp.last_match(3)
-        attribute = Regexp.last_match(4)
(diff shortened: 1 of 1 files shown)


mpw5 (approved):
LGTM 👍

PR #4753, opened 2022-03-21
Replace jsugarman with mpw5 as dependabot reviewer

jrmhaig:
#### What

Replace jsugarman with mpw5 as dependabot reviewer

#### Ticket

N/A

#### Why

@jsugarman and @mpw5 have swapped teams.

#### How

Edit `.github/dependabot.yml`.


mpw5 (approved):
👍

PR #4764, opened 2022-03-22
[CFP-295] Ruby 3 prep; More use of kwargs

jrmhaig:
#### What

Update use of keyword arguments as required in Ruby 3.

#### Ticket

#### Why

Keyword arguments can no longer be passed to methods as a has as the final argument in Ruby 3. These changes prepare for this.

#### How

Hashes or arguments are passed as `method(**kwargs)` instead of just `method(kwargs)`.


mpw5 (approved):
LGTM 👍

PR #4769, opened 2022-03-23
[CFP-295] Ruby 3 prep; Even more use of kwargs

jrmhaig:
#### What

Update use of keyword arguments as required in Ruby 3.

#### Ticket

#### Why

Keyword arguments can no longer be passed to methods as a has as the final argument in Ruby 3. These changes prepare for this.

#### How

The subject in `spec/services/stats/management_information/daily_report_generator_spec.rb` need to be fixed as it causes all the tests to fail.


mpw5 (approved):
👍

PR #4770, opened 2022-03-25
[CFP-295] Fix `Remote` module for Ruby 3

jrmhaig:
#### What

Update `app/services/remote/*` to use keyword arguments correctly for Ruby 3.

#### Ticket

#### Why

Ruby 3 changes the behaviour of keyword arguments so that they can no longer be passed as a hash in the final positional argument. The `Remote` module was written to explicitly take hash as an argument for keyword arguments and so this is now inconsistent with correct usage.

#### How

Method definitions are updated to have the argument `**query` instead of `query = {}` so that keyword arguments are read in correctly as a hash into `query`. Calls to these methods are updated accordingly.

Originally, the `Remote::CaseWorker#all` and `Remote::Claim#all_by_status` methods would ignore an `api_key` argument if it is passed in in favour of the default value. This has been changed (4c849534bd5971671a9cbd5e857eaae0e8056ec8) so that the passed in value can override the default. This isn't currently used anywhere but it is more logical.

E.g.

```ruby
Remote::CaseWorker.all(user, api_key: 'non-standard-api-key', **query)

# see app/services/remote/case_worker.rb:9
# Old behaviour
super(api_key: user.api_key, **query)

# New behaviour
super(api_key: 'non-standard-api-key', **query)
```

(Possibly, the `api_key` should be passed in anyway as `user` is not used for anything else)

### DIFF
--- a/app/services/case_worker_service.rb
+++ b/app/services/case_worker_service.rb
@@ -7,6 +7,6 @@ class CaseWorkerService
   end
 
   def active
-    Remote::CaseWorker.all(current_user, criteria)
+    Remote::CaseWorker.all(current_user, **criteria)
   end
 end
--- a/app/services/claims/case_worker_claims.rb
+++ b/app/services/claims/case_worker_claims.rb
@@ -26,19 +26,19 @@ module Claims
     private
 
     def current_claims
-      Remote::Claim.user_allocations(current_user, criteria)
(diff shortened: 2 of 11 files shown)


mpw5 (approved):
LGTM 🏆 

(I might have to get you to explain the `Remote` module to me at some point 😅)

PR #4779, opened 2022-03-30
[CFP-533] - Wording on login page for a disabled user

NenehB:
#### What
Making the error message on login page generic 
#### Ticket

[Wording on login page for a disabled user needs to be a generic message instead of ‘this user is disabled’.](https://dsdmoj.atlassian.net/browse/CFP-533)

#### Why
Not providing information for security reasons 

#### How
Making message generic

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -114,13 +114,9 @@ class User < ApplicationRecord
   end
 
   def inactive_message
-    if active? && enabled?
-      super
-    elsif softly_deleted?
-      'This account has been deleted.'
-    elsif disabled?
-      'This account has been disabled.'
-    end
+    return 'Invalid Email or password.' unless active? && enabled?
+
+    super
   end
 
   def email_notification_of_message
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -114,9 +114,11 @@ class User < ApplicationRecord
   end
 
   def inactive_message
-    return 'Invalid Email or password.' unless active? && enabled?
-
-    super
+    if active? && enabled?
+      super
+    else
+      I18n.t('activerecord.attributes.user.disabled_and_deleted_message')
+    end
   end
 
   def email_notification_of_message
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -186,6 +186,7 @@ en:
         first_name: *first_name
         last_name: *last_name
         email_confirmation: *email_confirmation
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -114,13 +114,9 @@ def active_for_authentication?
...
   def inactive_message
-    if active? && enabled?
-      super
-    elsif softly_deleted?
-      'This account has been deleted.'
-    elsif disabled?
-      'This account has been disabled.'
-    end
+    return 'Invalid Email or password.' unless active? && enabled?
+
```
imtiazAhmd: Does this mean, even the a user has been disabled / deleted they will see the same message as 'Invalid Email or Password' ?

NenehB: Yes, for security reasons if we keep the account has been deleted/disabled, these messages are just saying the details you have put in is correct however the account has been deleted/disabled which we don't want if the user trying to login is not the actual user.

app/models/user.rb
```diff
@@ -114,13 +114,9 @@ def active_for_authentication?
...
 
   def inactive_message
-    if active? && enabled?
-      super
-    elsif softly_deleted?
-      'This account has been deleted.'
-    elsif disabled?
-      'This account has been disabled.'
-    end
+    return 'Invalid Email or password.' unless active? && enabled?
```
mpw5: Could this message go in a translation file rather than be hardcoded here? It would be a little neater and might make it easier to maintain in future.

mpw5 (commented):
Looks good. I've left a comment about moving the error text to a translation file. I'm happy to approve when that's done!

config/locales/en.yml
```diff
@@ -186,6 +186,7 @@ en:
         first_name: *first_name
         last_name: *last_name
         email_confirmation: *email_confirmation
+        disabled_and_deleted_message: 'Invalid Email or password.'
```
naseberry, line 189: Nit: content design issue, consider 'Invalid email or password'

naseberry (approved):
👍 looks good, aside from the nit

config/locales/en.yml
jrmhaig: The message that the devise gem gives, which will appear in the `active? && enabled?` case, has a capital 'E' so for consistency (for the moment) it will need to stay like this.

jrmhaig (approved):

PR #4780, opened 2022-03-30
CFP-532 Provider admin: Display disabled user status

mpw5:
#### What

Display the correct status of disabled provider users to provider admins

#### Ticket

[CFP-532](https://dsdmoj.atlassian.net/browse/CFP-532)

#### Why

When a provider admin views the list of users on the Manage users page,
there is currently nothing to indicate that a user has been disabled;
they appear with the Edit and Delete links as with an enabled user. This
may be confusing for admins.

#### How

* Update `app/views/external_users/admin/external_users/index.html.haml` to display a disabled user as `Inactive`.
* Add a feature test to confirm this works as expected.


imtiazAhmd (approved):
LGTM 👍🏼

jrmhaig (approved):

PR #4782, opened 2022-03-31
[CFP-534] [CFP-535] Show disabled user as inactive in superadmin view

jrmhaig:
#### What

Display the correct status of a disabled external user when viewed in the users list by the superadmin or caseworker with provider management access.

#### Ticket

* [Superadmin: Display when a user is disabled](https://dsdmoj.atlassian.net/browse/CFP-534)
* [Caseworker admin: Display when a user is disabled](https://dsdmoj.atlassian.net/browse/CFP-535)

#### Why

When the superadmin views a list of users for a provider there is no indication that a user is disabled.

#### How

Update `app/views/provider_management/external_users/index.html.haml` to show state inactive when a user is disabled.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -103,7 +103,7 @@ class Ability
   def can_administer_any_provider
     can %i[show index new create edit update], Provider
     can %i[index new create find search], ExternalUser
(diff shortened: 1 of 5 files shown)


naseberry (approved):
👍  Checked on local environment

PR #2995, opened 2019-09-25
Add fonts to Dockerfile to enable Evidence attachments to render properly

willmcb:
#### What
Adds fonts to the version of Alpine Linux used with docker to enable pdf previews of `.docx` documents to render correctly.

[See this Stack Overflow post](https://stackoverflow.com/questions/8626594/wicked-pdf-error-pdf-could-not-be-generated/58095049#58095049) for more information.

#### Ticket

[CBO-896: Evidence attachments not rendering properly](https://dsdmoj.atlassian.net/browse/CBO-886)

PR #4802, opened 2022-04-08
Update to Ruby 2.7.5

jrmhaig:
#### What

Upgrade to [Ruby 2.7.5](https://www.ruby-lang.org/en/news/2021/11/24/ruby-2-7-5-released/)

#### Ticket

N/A

#### Why

Snyk has reported high vulnerabilities with the docker image we are using, ruby:2.7.4-alpine3.13.

#### How

Upgrade Ruby to 2.7.5 and the docker image to ruby:2.7.5-alpine3.15.

PR #4803, opened 2022-04-08
Remove ttf-ubuntu-font-family from Dockerfile

jrmhaig:
Testing on an envionment will be done to test if the PDF conversion issue has
been resolved.

#### What

Remove the `ttf-ubuntu-font-family` package from the `Dockerfile`.

#### Ticket

N/A

#### Why

`ttf-ubuntu-font-family` was added in #2995 to fix PDF conversion of uploaded documents. This package has since been removed from Ubuntu and is not available in the latest alpine image.

#### How

Remove form `docker/Dockerfile`

--------

#### TODO

 - [ ] Test document conversion on an environment

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,7 +35,6 @@ RUN apk --no-cache add --virtual build-dependencies \
                   nodejs \
                   linux-headers \
                   runit \
-                  ttf-ubuntu-font-family \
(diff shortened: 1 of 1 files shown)


jrmhaig:
Reverting #4802 and will fix this issue later.

PR #4822, opened 2022-04-14
CFP-485 Add rake task to identify and disable inactive users

mpw5:
#### What

Add a facility to identify and disable inactive users

#### Ticket

[CFP-485](https://dsdmoj.atlassian.net/browse/CFP-485)

#### Why

We need a re-runnable (and potentially automatable) process to allow us  to identify and subsequently disable user records that have been inactive for over six months.

#### How

This change adds a rake file `inactive_users` which contains six tasks:

1) `inactive_users:list_providers` which generates a csv file of all inactive provider users (ie users which were last logged into more than six months ago or were created more than six months ago and have never been logged into, and are not already disabled). A csv file is used as this may potentially result in large volumes of data which will be hard to manipulate from the command line.

2) `inactive_users:list_caseworkers` as above but for inactive caseworker users.

3) `inactive_users:list_disabled_providers` which generates a csv file of all provider user records where the `disabled_at` field has been populated. A csv file is used as this may potentially result in large volumes of data which will be hard to manipulate from  the command line.

4) `inactive_users:list_disabled_caseworkers` as above but for disabled caseworker users.

5) `inactive_users:disable_providers` which populates the `disabled_at` field on all provider user records identified by (1) with the current date and time.

6) `inactive_users:disable_caseworkers` as above but for inactive caseworker users.

It also adds a helper class to provide methods to support these actions.

In combination, these should allow us to identify which users need to be disabled, disable them, and subsequently identify 
which have successfully been disabled and which have failed.

### DIFF
--- a/lib/tasks/inactive_users.rake
+++ b/lib/tasks/inactive_users.rake
@@ -1,0 +1,78 @@
+namespace :inactive_users do
+  desc 'List inactive users'
+  task list: :environment do
+    CSV.open(Rails.root.join('tmp/inactive_users.csv'), 'wb') do |csv|
+      csv << %w[
+        id
+        email
+        last_sign_in_at
+        current_sign_in_at
+        created_at
+        updated_at
+        persona_type
+      ]
+      inactive_users = User.where('disabled_at IS NULL AND (last_sign_in_at < :limit OR (last_sign_in_at IS NULL AND created_at < :limit))', limit: 6.months.ago.to_date)
+      inactive_users.each do |inactive_user|
+        csv << [
+          inactive_user.id,
+          inactive_user.email,
+          inactive_user.last_sign_in_at,
+          inactive_user.current_sign_in_at,
+          inactive_user.created_at,
+          inactive_user.updated_at,
+          inactive_user.persona_type
+        ]
+      end
+      puts "Listed #{inactive_users.count} inactive users in tmp/inactive_users.csv"
+    end
+  end
+
+  desc 'Disable inactive users'
+  task disable: :environment do
+    inactive_users = User.where('disabled_at IS NULL AND (last_sign_in_at < :limit OR (last_sign_in_at IS NULL AND created_at < :limit))', limit: 6.months.ago.to_date)
+    puts "There are #{inactive_users.count} inactive users to be disabled"
+
+    disabled_pre = User.where('disabled_at IS NOT NULL').count
+    error_count = 0
+
+    inactive_users.each do |inactive_user|
+      inactive_user.update!(disabled_at: DateTime.now)
+    rescue ActiveRecord::RecordInvalid
+      error_count += 1
+    end
+
+    disabled_post = User.where('disabled_at IS NOT NULL').count
+    puts "#{disabled_post - disabled_pre} inactive users have been disabled - run `rake inactive_users:list_disabled` for details"
+    puts "Errors occurred when disabling #{error_count} users - run `rake inactive_users:list` for details" if error_count > 0
+  end
+
+  desc 'List disabled users'
+  task list_disabled: :environment do
+    CSV.open(Rails.root.join('tmp/disabled_users.csv'), 'wb') do |csv|
+      csv << %w[
+        id
+        email
+        last_sign_in_at
+        current_sign_in_at
+        created_at
+        updated_at
+        persona_type
+        disabled_at
+      ]
+      disabled_users = User.where('disabled_at IS NOT NULL')
+      disabled_users.each do |disabled_user|
+        csv << [
+          disabled_user.id,
+          disabled_user.email,
+          disabled_user.last_sign_in_at,
+          disabled_user.current_sign_in_at,
+          disabled_user.created_at,
+          disabled_user.updated_at,
+          disabled_user.persona_type,
+          disabled_user.disabled_at
+        ]
+      end
+      puts "Listed #{disabled_users.count} disabled users in tmp/disabled_users.csv"
+    end
+  end
+end
--- a/lib/tasks/inactive_users.rake
+++ b/lib/tasks/inactive_users.rake
@@ -1,38 +1,28 @@
+require_relative 'rake_helpers/user_management.rb'
+
 namespace :inactive_users do
   desc 'List inactive users'
   task list: :environment do
-    CSV.open(Rails.root.join('tmp/inactive_users.csv'), 'wb') do |csv|
-      csv << %w[
-        id
-        email
-        last_sign_in_at
-        current_sign_in_at
-        created_at
-        updated_at
-        persona_type
-      ]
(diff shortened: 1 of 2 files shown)


lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
...
+      csv << %w[
+        id
+        email
+        last_sign_in_at
+        current_sign_in_at
+        created_at
+        updated_at
+        persona_type
+      ]
+      inactive_users = User.where('disabled_at IS NULL AND (last_sign_in_at < :limit OR (last_sign_in_at IS NULL AND created_at < :limit))', limit: 6.months.ago.to_date)
```
jrmhaig: I think this can be done without raw SQL as;

```suggestion
      inactive_users = User.where(last_sign_in_at: ..6.months.ago).or(User.where(last_sign_in_at: nil, created_at: ..6.months.ago)).where(disabled_at: nil)
```

lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
...
+          inactive_user.persona_type
+        ]
+      end
+      puts "Listed #{inactive_users.count} inactive users in tmp/inactive_users.csv"
+    end
+  end
+
+  desc 'Disable inactive users'
+  task disable: :environment do
+    inactive_users = User.where('disabled_at IS NULL AND (last_sign_in_at < :limit OR (last_sign_in_at IS NULL AND created_at < :limit))', limit: 6.months.ago.to_date)
```
jrmhaig: As above,

```suggestion
    inactive_users = User.where(last_sign_in_at: ..6.months.ago).or(User.where(last_sign_in_at: nil, created_at: ..6.months.ago)).where(disabled_at: nil)
```

Also, can this go into a helper in (e.g.) `lib/tasks/rake_helpers/user_management.rb`?

lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
...
+    puts "There are #{inactive_users.count} inactive users to be disabled"
+
+    disabled_pre = User.where('disabled_at IS NOT NULL').count
```
jrmhaig:
```suggestion
    disabled_pre = User.where.not(disabled_at: nil).count
```

lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
...
+    error_count = 0
+
+    inactive_users.each do |inactive_user|
+      inactive_user.update!(disabled_at: DateTime.now)
+    rescue ActiveRecord::RecordInvalid
+      error_count += 1
+    end
+
+    disabled_post = User.where('disabled_at IS NOT NULL').count
```
jrmhaig:
```suggestion
    disabled_post = User.where.not(disabled_at: nil).count
```

lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
...
+        id
+        email
+        last_sign_in_at
+        current_sign_in_at
+        created_at
+        updated_at
+        persona_type
+        disabled_at
+      ]
+      disabled_users = User.where('disabled_at IS NOT NULL')
```
jrmhaig:
```suggestion
      disabled_users = User.where.not(disabled_at: nil)
```

lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
+namespace :inactive_users do
+  desc 'List inactive users'
+  task list: :environment do
+    CSV.open(Rails.root.join('tmp/inactive_users.csv'), 'wb') do |csv|
```
jrmhaig: Maybe add a date or timestamp to the file name.

lib/tasks/inactive_users.rake
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    disabled_post = User.where('disabled_at IS NOT NULL').count
+    puts "#{disabled_post - disabled_pre} inactive users have been disabled - run `rake inactive_users:list_disabled` for details"
+    puts "Errors occurred when disabling #{error_count} users - run `rake inactive_users:list` for details" if error_count > 0
+  end
+
+  desc 'List disabled users'
+  task list_disabled: :environment do
+    CSV.open(Rails.root.join('tmp/disabled_users.csv'), 'wb') do |csv|
```
jrmhaig: This is (largely*) the same as how the CSV file is written in the `inactive_users` tasks. Could this be put into a helper?

*I see that the CSV file here has an extra column for `disabled_at`. Presumably, this column in the file created by the `inactive_users` tasks would always be blank.

mpw5:
Thanks for your comments @jrmhaig. I've updated to include your feedback (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4822/commits/fd04de20c84d779ad7cd7db0afd04bf7ab40b807).

By that point the the `list` and `list_disabled` tasks were very similar so I've also added a commit (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4822/commits/475c777b9458c5562c655b156cc9d09aa8e44291) which reduces them to a single task that takes an argument to drive which type of user is listed. I'm not completlely convinced that I'm not sacrificing comprehension for DRYness here; I'd be interested in any thoughts you have.

jrmhaig:
> By that point the the `list` and `list_disabled` tasks were very similar so I've also added a commit ([475c777](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/475c777b9458c5562c655b156cc9d09aa8e44291)) which reduces them to a single task that takes an argument to drive which type of user is listed. I'm not completlely convinced that I'm not sacrificing comprehension for DRYness here; I'd be interested in any thoughts you have.

My comment was more about putting the code for creating the CSV file into the helper module as a single method, which you had done with `create_csv_for`, rather than combining the tasks. I agree that having two separate tasks makes more sense. Sorry, I was not clear.

I will approve and leave it to you to decide whether or not to remove commit 475c777.

jrmhaig (approved):

mpw5:
> Sorry, I was not clear.

Don't worry, your comments were clear! I just got carried away with refactoring 😁 I'm going to remove that commit, I wasn't completely happy with it myself.

lib/tasks/rake_helpers/user_management.rb
```diff
@@ -34,4 +45,14 @@ def self.create_csv_for(users, filename)
       end
     end
   end
+
+  def self.disable(users)
+    error_count = 0
+    users.each do |user|
+      user.update!(disabled_at: DateTime.now)
```
jrmhaig: I missed this with the last review. Can we use this instead;

```suggestion
      user.disable
```

Also, in the queries above you could use `User.disabled.where(...)` instead of `User.where(...).where.not(disabled_at: nil)`.

(See `app/models/concerns/disablable.rb`)

jrmhaig (approved):
Approved again, although I have added a comment that you might consider.

lib/tasks/rake_helpers/user_management.rb
mpw5: Good spot. Thanks.

PR #4823, opened 2022-04-19
[CFP-370] FormBuilder migration - Defendants details

NenehB:
#### What

Migrate external_users form step: **defendants** form

#### Ticket

[FormBuilder migration - Defendants details](https://dsdmoj.atlassian.net/browse/CFP-370)

#### Why

- Update the service to use GOV.UK styles, components and patterns
- To improve and resolve the service's accessibility issues
- To reduce code complexity
- To improve maintainability

### DIFF
--- a/app/views/external_users/claims/defendants/_defendant_fields.html.haml
+++ b/app/views/external_users/claims/defendants/_defendant_fields.html.haml
@@ -12,18 +12,18 @@
     - @representation_order_count = 0
 
     = f.govuk_text_field :first_name,
-      form_group: { classes: ['first-name'] },
+      form_group: { classes: ['cc-first-name'] },
       label: { text: t('.first_name_html') },
       width: 'one-half'
       
     = f.govuk_text_field :last_name,
-      form_group: { classes: ['last-name'] },
+      form_group: { classes: ['cc-last-name'] },
       label: { text: t('.last_name_html') },
       width: 'one-half'
 
     = f.govuk_date_field :date_of_birth,
       date_of_birth: true,
-      form_group: { classes: ['dob'], id: "defendant_#{@defendant_count+1}_date_of_birth_group" },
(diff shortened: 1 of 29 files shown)


app/views/external_users/claims/defendants/_defendant_fields.html.haml
```diff
@@ -1,35 +1,39 @@
...
-          = f.check_box :order_for_judicial_apportionment
-          = f.label :order_for_judicial_apportionment, t('.order_for_judicial_apportionment_html'), class: 'judicial-apportionment-notice'
-          = validation_error_message(@error_presenter, :order_for_judicial_apportionment)
-
-      .form-group
-        = govuk_detail 'Help with judicial apportionment' do
-          = t('.order_for_judicial_apportionment_help')
+      = f.govuk_check_box :order_for_judicial_apportionment,
+        1,
+        0,
```
mpw5, line 30: It's not clear what `1` and `0` are here. Is there a better way of writing this that makes it more obvious? For example, something like:
```
= f.govuk_check_box :order_for_judicial_apportionment,
   value: 1,
   name: 0,
```
`value` and `name` are just guesses at what's intended, I'm not suggesting using them! I'm also not sure what the formbuilder allows/expects so maybe we're restricted by that.

mpw5: Answering my own question, I see this is taken directly from the formbuilder docs here: https://govuk-form-builder.netlify.app/form-elements/checkboxes/#input-slim-generating-a-single-checkbox.

If that's the recommended approach then I'm happy to live with it 😁

mpw5 (approved):
Looks good to me 🎉 

I had one question but managed to answer it myself by reading the docs 🙂

imtiazAhmd (approved):
LGTM 👍🏼

PR #4834, opened 2022-04-25
Remove json-schema gem

jrmhaig:
#### What

Remove the `json-schema` gem.

#### Ticket

N/A

#### Why

The json-schema gem was added by f56ec1b1fa4bc513e9a6a51e9cada70766a0ad36 but after c37df706c54967b3682b5c9018faad248cabc2fe it appears no longer to be used.

By reducing the number of 3rd party libraries installed we reduce the possibility of security vulnerabilities.

#### How

Remove from `Gemfile`.


jrmhaig:
My mistake. `JSON::Validator` is part of this gem.

Issue #2661, opened 2019-02-27
Consolidate API shared examples further


jsugarman:
The shared specs below could be combined:

RSpec.shared_examples "invalid API key validate endpoint"
RSpec.shared_examples "invalid API key create endpoint"

jsugarman:
@colinbruce Have created an issue for this, from your PR comment

mpw5:
Fixed by https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4835

PR #4835, opened 2022-04-26
Refactor rspec shared examples

mpw5:
#### What

Refactor rspec shared examples to reduce duplication

#### Ticket

Fixes issue https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/issues/2661.

#### Why

The `invalid API key validate endpoint` and `invalid API key create endpoint` rspec shared examples are essentially duplicates and can be consolidated to improve code quality.

#### How

* combine the two examples into a single shared example `invalid API key` which uses an `action` parameter to determine which endpoint is called. 
* update tests use the shared examples to pass in the appropriate `action` parameter

### DIFF
--- a/spec/api/v1/external_users/date_attended_spec.rb
+++ b/spec/api/v1/external_users/date_attended_spec.rb
@@ -63,7 +63,7 @@ RSpec.describe API::V1::ExternalUsers::DateAttended do
     end
 
     context 'when date_attended params are invalid' do
-      include_examples 'invalid API key create endpoint', exclude: :other_provider
+      include_examples 'invalid API key', exclude: :other_provider, action: :create
 
       context 'missing expected params' do
         it 'returns a JSON error array with required model attributes' do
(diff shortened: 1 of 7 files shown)


jrmhaig (approved):
Looks good. I'm happy for this to go in as is.

A couple of other things to consider, although there is no urgency to do it now;

* Can `post_to_create_endpoint` and `post_to_validate_endpoint` could be defined with `let` instead of defined as functions?
* Can `post_to_create_endpoint` and `post_to_validate_endpoint` be changed simply to `post_to_endpoint`, which would mean that you do not need to define it in the shared examples and the `action` option is not required? I do not think there are any cases where both the `create` and `validate` forms are defined at the same time.

mpw5:
Thanks @jrmhaig. Both your comments are fair. I've addressed the first point (using `let`s instead of `def`s). I'll merge this PR and look into the other another time.

jrmhaig (approved):

PR #4860, opened 2022-05-04
User list page

jrmhaig:
#### What

Add a page to view recently created users.

#### Ticket

N/A

#### Why

For security reasons we want to be able to monitor new users for possible unauthorized creation.

#### How

Create an index page in the `User` controller that is only visible to a superadmin user. The users are paginated and ordered by the 'created at' date in descending order.

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -23,6 +23,10 @@ RSpec.describe Ability do
     it { should_not be_able_to(:update, UserMessageStatus.new) }
     it { should_not be_able_to(:create, Document.new) }
     it { should_not be_able_to(:update_settings, User.new) }
+
+    context 'with user management' do
+      it { is_expected.not_to be_able_to(:index, User) }
+    end
   end
 
   context 'when a signed in user' do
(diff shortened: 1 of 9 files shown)


spec/models/ability_spec.rb
```diff
@@ -23,6 +23,10 @@
     it { should_not be_able_to(:update, UserMessageStatus.new) }
     it { should_not be_able_to(:create, Document.new) }
     it { should_not be_able_to(:update_settings, User.new) }
+
+    context 'with user management' do
+      it { is_expected.not_to be_able_to(:index, User) }
```
mpw5, line 28: Could we use `it { should_not be_able_to(:index, User) }` here, for consistency with other tests? (Same goes for the other new tests you've added.)

mpw5 (approved):
Great work 🎉 One non-blocking suggestion, but I'm happy for it to be released as it is.

spec/models/ability_spec.rb
jrmhaig: Thanks. We are going the other way so I should probably have changed all the instances of `should` and `should_not` to `is_expected.to` and `is_expected.not_to` while I'm here.

mpw5: Oh, I see. In that case, leave it for now. Your code is fine, we can fix the other tests in another PR.

PR #4862, opened 2022-05-05
Block crawlergo requests based on HTTP headers

mpw5:
#### What

Block requests from the potentially malicious browser crawler `crawlergo`.

#### How

Requests from the tool include the HTTP header `spider-name` with a value of `crawlergo`. This change updates ingress files so that requests that include that header return a 403 unauthorised response. This won't help us block other crawlers, or if users are able to configure the tool so that it uses a different `spider-name`, but at least will give us some protection.


pking-moj (approved):

jrmhaig (approved):
:+1: Looks good

PR #4889, opened 2022-05-18
Increase RestClient OpenTimeout from `5` to `10`

naseberry:
#### What

Increase RestClient OpenTimeout from `5` to `10`

#### Ticket

[n/a](#)

#### Why

Users are increasingly experiencing time outs connecting to server

### DIFF
--- a/config/initializers/remote_client.rb
+++ b/config/initializers/remote_client.rb
@@ -1,7 +1,13 @@
+# open_timeout value temporarily increased
+# to reduce caseworkers timeout errors
+#
+# Ticket CFP-587 has been created to further investigate
+# the root cause and therefore reduce the open_timeout
+#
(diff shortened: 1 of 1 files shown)


config/initializers/remote_client.rb
```diff
@@ -2,6 +2,6 @@
   client.base_url = Settings.remote_api_url
   client.api_key = Settings.remote_api_key
   client.logger = Rails.logger
-  client.open_timeout = 5
+  client.open_timeout = 10
```
jrmhaig, line 11: Maybe add a comment in the code to indicate why the timeout is set to 10, and that we ideally want it to be lower. You could reference https://dsdmoj.atlassian.net/browse/CFP-587

jrmhaig (approved):
Looks OK. Just one suggestion for a comment.

PR #4890, opened 2022-05-18
Refactor ZendeskSender specs

jrmhaig:
#### What

Refactor `ZendeskSender` specs.

#### Ticket

N/A

#### Why

Dependabot PR #4885 is failing because of one of the spec tests. The test in question is testing the request that the Zendesk API makes so it is not really necessary.

#### How

Refactor the tests to check that `ZendeskAPI::Ticket.create!` is called with the correct arguments when called either as `ZendeskSender.send!(...)` or `ZendeskSender.new(...).send!`


mpw5 (approved):
Nice!

PR #4892, opened 2022-05-20
Rename Slack bot for archive job alerts

jrmhaig:
#### What

Rename the Slack bot name used when reporting failures in the stale claim archiver task.

#### Ticket

N/A

#### Why

Alerts are being sent with the name 'Injectotron', which is the default Slack bot name that was set when the alerter was only used for Injection failures.

#### How

Add an optional argument to `SlackNotifier.new` for the bot name.


mpw5 (approved):
LGTM

PR #4893, opened 2022-05-20
[CFP-543] Update to Ruby 3.1.2

mpw5:
#### What

Bump the version of ruby used by CCCD from 3.0 to 3.1. 

#### Ticket

[CFP-543](https://dsdmoj.atlassian.net/browse/CFP-543)

#### Why

To keep us on the most recent version of the language.

#### How
* Update `Gemfile`, `.ruby-version`, `docker/Dockerfile` and `.circleci/config.yml` to bump us to the correct version and add dependencies.
* Fix issues caused by ruby's internal upgrade to v4.0 of the `psych` library.
* Fix an issue caused by a bug in the `rspec-mock` library.


jrmhaig:
There are a couple of other places that need updating:

* `.rubocop.yml` needs `TargetRubyVersion: 3.1`, and this will need various fixes due to Ruby 3.1 changes;
  - Arguments like `func(arg: arg)` can be written as `func(arg:)` ([`Style/HashSyntax`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax)).
  - A block passed to a function, such as `&block`,  can be written simply as `&` ([`Naming/BlockForwarding`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/BlockForwarding))
* `.github/actions/delete-ecr-images/Dockerfile`
  - I'm not sure that this file actually needs a Ruby docker image so maybe it can be changed to `alpine3.15`

jrmhaig:
Do we need to do some clean up (after this PR) to resolve some of the Psych 4.0 issues? Is there anything we can do? Using a function called `unsafe_load_file` feels like a red flag.

mpw5:
> There are a couple of other places that need updating:
> 
> * `.rubocop.yml` needs `TargetRubyVersion: 3.1`, and this will need various fixes due to Ruby 3.1 changes;
>   
>   * Arguments like `func(arg: arg)` can be written as `func(arg:)` ([`Style/HashSyntax`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax)).
>   * A block passed to a function, such as `&block`,  can be written simply as `&` ([`Naming/BlockForwarding`](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/BlockForwarding))
> * `.github/actions/delete-ecr-images/Dockerfile`
>   
>   * I'm not sure that this file actually needs a Ruby docker image so maybe it can be changed to `alpine3.15`

Thanks, I'll update. I didn't realise we had two `Dockerfiles` 🤦

mpw5:
> Do we need to do some clean up (after this PR) to resolve some of the Psych 4.0 issues? Is there anything we can do? Using a function called `unsafe_load_file` feels like a red flag.

Yeah, I don't entirely like it either. The consensus seems to be that `unsafe_load_file` is ok when used to load internal files, it just shouldn't be used for parsing user input. But your point is a good one.

There is an alternative we can use instead in `spec/requests/geckoboard/widgets_spec.rb`, where we would pass a list of valid classes to `load_file` instead of calling `unsafe_load_file`. Something like:

```
  def self.widgets
    permitted_classes = [
      Stats::ClaimPercentageAuthorisedGenerator, 
      Stats::ClaimCompletionReporterGenerator, 
      Stats::ClaimCreationSourceDataGenerator
      [...]
    ]

    @widgets ||= YAML.load_file('spec/fixtures/geckoboard_api_widgets.yaml', permitted_classes: permitted_classes)
  end
```

It's a little clunky as there are a dozen widgets that need permitting but removes the need for `unsafe_load_file`. What do you think?

mpw5:
> I'm not sure that this file actually needs a Ruby docker image so maybe it can be changed to alpine3.15

This `Dockerfile` needs ruby to run `.github/actions/delete-ecr-images/entrypoint`, so I've kept it as `ruby:3.1.2-alpine3.15`.

jrmhaig:
> It's a little clunky as there are a dozen widgets that need permitting but removes the need for `unsafe_load_file`. What do you think?

Looking at it again I see that it is in the specs and it is related to Geckoboard, which we might want to review anyway. I'd say leave it as it is.

jrmhaig (approved):
🎉

PR #4894, opened 2022-05-20
Refactor of error message presenter

jrmhaig:
#### What

Separate out concerns in the error message presenter

#### Ticket

N/A

#### Why

The error message presenter is complex and it is difficult to understand what is going on with various concerns in one place.

#### How

* Move the responsibility for avoiding duplication of error details into `ErrorMessage::DetailCollection`
  - Also update to check for duplication with all existing errors, not just the first
* Use the abstract factory pattern for creating instances of `ErrorMessage::Detail`
  - Currently this does not add much but ultimately the factory could return a null object in certain circumstances to drop the error message (see 3f7fe2615aa0c5a6b99b863c42d65039b95bf0c1 in [cfp-77-graduated-fees-mw](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/tree/cfp-77-graduated-fees-mw))

### DIFF
--- a/app/presenters/error_message/detail_collection.rb
+++ b/app/presenters/error_message/detail_collection.rb
@@ -9,11 +9,10 @@ module ErrorMessage
     end
 
     def []=(fieldname, error_detail)
-      if @detail_collection.key?(fieldname)
-        @detail_collection[fieldname] << error_detail
-      else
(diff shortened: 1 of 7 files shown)


jrmhaig:
Closing for now

PR #4896, opened 2022-05-23
Extract sequencer to separate class

jrmhaig:
#### What

Separate out concerns in the error message presenter.

#### Ticket

N/A

#### Why

The error message presenter is complex and it is difficult to understand what is going on with various concerns in one place.

#### How

Extract the 'sequencer' to a separate class.

The sequencer calculates a sequence number for an error message so that all errors can be displayed in the correct order.


mpw5 (approved):
LGTM

PR #4912, opened 2022-05-25
[CFP-577] Generate GOVUK formbuilder compatible errors

mpw5:
#### What

Unblock work on the GOVUK formbuilder migration by amending error validation/presentation to ensure that linkable errors are generated.

#### Ticket

[CFP-577](https://dsdmoj.atlassian.net/browse/CFP-577)

#### Why

The current method of error generation produces certain errors with a different attribute format to the form items generated by the GOVUK formbuilder. This means that error summary links do not correctly link to the form item.

#### How

* Amend `BaseSubModelValidator` to extend existing error generation for `has_many` associations to `has_one` associations.
* The above change produces duplicate errors, so amend `ErrorMessage::Presenter` to hide old-style errors.

By itself this change has no impact on the application. Developers will need to make changes when implementing migrations of specific forms in order for this to have an effect:

* Update line 50 of `BaseSubModelValidator` to add the name of the assocation (eg `graduated_fee`) to the array in `validate_association_for`.
* Update line 74 of `BaseSubModelValidator` to add the name of the assocation (eg `graduated_fee`) to the array in `associated_error_attribute`.
* Update line 39 of `ErrorMessage::Presenter` to add the name of the assocation (eg `graduated_fee.`) to the array in `govuk_has_one_associations` (the `.` is important here!).
* Amend the `config/locales/en/error_messages/claim.yml` file to rename keys for error message translations. See https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/fd947aa0bf5f5ad29b5dbb6a6eb2602ed8ac93ef for an example of how this should be done. Essentially the translations need to be renamed to match the attributes now being generated, and indentation removed so that they are no longer nested. 
* The tests on lines 146 and 248 of `spec/presenters/error_message/presenter_spec.rb` have been marked as pending. They will pass for now but will start to fail when the graduated fees form is migrated. At that point the `pending` command can be removed and the tests will pass again.


spec/presenters/error_message/presenter_spec.rb
```diff
@@ -143,6 +143,16 @@
       it { is_expected.to eq('Foo bar') }
     end
 
+    xcontext 'with a nested error that is generated by BaseSubModelValidator' do
```
imtiazAhmd: needs to remove extra 'x' from 'context'

spec/presenters/error_message/presenter_spec.rb
```diff
@@ -235,6 +245,21 @@
       it { is_expected.to eq(expected_error_messages) }
     end
 
+    xcontext 'with a nested error that is generated by BaseSubModelValidator' do
```
imtiazAhmd, line 248: needs to remove extra 'x' from 'context'

spec/presenters/error_message/presenter_spec.rb
mpw5: That's deliberate, adding `x` before the test skips it. Those two tests don't actually do anything until this code is used to migrate a form, so I've disabled them for now (maybe I should have removed them completely 🤷 ). 

My thinking is that once the `graduated_fee` form (for example) has been migrated we can remove the `x` and the test will run and pass. (If another form is migrated first then the content of the test will need to be tweaked a bit).

I mention this in the PR description but I can add comments to the tests for more clarity if it would be useful?

imtiazAhmd: Ah, thanks and apologies for not reading the description properly. I hope we don't forget to update this file during the migrated form implementation.

mpw5: No worries. I agree there's a small risk that it gets missed. I'll be checking your PRs with eagle eyes 😀

imtiazAhmd (approved):

spec/presenters/error_message/presenter_spec.rb
jrmhaig: Are they expected to fail at the moment? If you add `pending` instead of the `x` then they will report as a pass if they fail and as a failure if they pass. This means that when the behaviour is fixed the tests get highlighted so that they can be updated.

app/presenters/error_message/presenter.rb
```diff
@@ -26,12 +26,19 @@ def generate_messages
...
         add_error_detail(attribute, message)
       end
     end
 
+    def error_is_for_govuk_has_one_association?(attribute)
+      govuk_has_one_associations.any? { |association| attribute.to_s.include?(association) }
+    end
+
+    def govuk_has_one_associations
+      @govuk_has_one_associations ||= %w[] # add associations to the array as they are migrated - eg ['graduated_fee.', 'fixed_fee.']
```
jrmhaig: **(nit)**
```suggestion
      @govuk_has_one_associations ||= %w[] # add associations to the array as they are migrated - eg %w[graduated_fee. fixed_fee.]
```

spec/presenters/error_message/presenter_spec.rb
mpw5: I did not know about `pending`. That does exactly what I need. Every day's a school day!

app/presenters/error_message/presenter.rb
mpw5: 👍

jrmhaig (approved):
:+1:

PR #4914, opened 2022-05-26
Fix some Rubocop errors

jrmhaig:
#### What

Fix some Rubocop failures.

#### Ticket

N/A

#### Why

Improving code quality

#### How

* `Lint/DuplicateBranch`: This cop failed in three places:
  - In one case, in `app/helpers/external_users/claim_types_helper.rb`, the only use of the method was removed in 847f26a97844257ef95dc77b3575adc70ed904b2 so it can be removed completely
  - In the other two cases one option in an `if` or `case` statement has been combined with the default option


jrmhaig:
Closing because there is clearly something funny going on as I cannot see any change in the logic but tests are failing.

PR #4942, opened 2022-06-13
Validate case type for interim claims

jrmhaig:
#### What

Validate the case type for interim claims.

#### Ticket

[Zendesk tickets - Sprint Elastigirl](https://dsdmoj.atlassian.net/browse/CFP-612)

#### Why

When creating a claim via the API it is possible to choose any case type. For interim claims only 'trial' and 'retrial' are valid and interim claims that have been created with certain case types can prevent the claim summary from loading with the error ``undefined method `fixed_fees_total'`` in `app/views/external_users/claims/_final_claim_cost_summary.html.haml`. This is because the value of the flag `claim.fixed_fee_case?`is based on the case type and not the claim type.

#### How

Add a validator to check that the case type for an interim claim is included in the list given by `Claim::InterimClaim#eligible_case_types`. Specifically, these are case types with the `interim` role.


mpw5 (approved):
LGTM

PR #4961, opened 2022-06-24
Some transfer brain cleanup

jrmhaig:
#### What

Refactor of the Transfer Brain.

#### Ticket

N/A

#### Why

To make it more understandable.

#### How

* The data item object now includes `ActiveModel::Model`, which makes it simpler for creating instances from CSV rows without having a temporary intermediate representation as a `Struct`.
* Move data item and data item collection into a namespace.
  * Note that Github is not showing the renamed files. `app/models/claim/transfer_brain_data_item.rb` and `app/models/claim/transfer_brain_data_item_collection.rb` are moved to `app/models/claim/transfer_brain/data_item.rb` and `app/models/claim/transfer_brain/data_item_collection.rb` respectively, and their spec files have been moved likewise.

Most of the commit are minor changes. The exception to this is 12005c4f6fdf68f8bcf33e91c29864c0724866dc which modifies the `DataItem` collection to take the data directly from a row from the CSV file instead of a temporary `Struct`.

**Question:** ac1f50265b0b2e048c27501eb1c4dd307ce15d9f moves the delegatable concerns inside the `Claim::TransferBrain` namespace. Is this the right thing to do or should the concerns be kept separate?


app/models/claim/transfer_brain_data_item_collection.rb
```diff
@@ -73,10 +73,8 @@ def wildcard_mapping_for(detail)
       )
     end
 
-    def read_csv
-      file = Rails.root.join('config', 'transfer_brain_data_items.csv')
-      csv_content = File.read(file)
-      CSV.parse(csv_content, headers: true)
+    def csv
+      @csv ||= CSV.parse(File.read(Rails.root.join('config', 'transfer_brain_data_items.csv')), headers: true)
```
naseberry: Not a blocker, but found the previous implementation easier to read.

jrmhaig: That is fair. Clarity is probably more important than compactness here.

mpw5: I agree with Peter that the original is easier to read. I think if you want to keep the format but include memoisation, you should be able to do something like: 

```
    def csv
      @csv ||= begin
        Rails.root.join('config', 'transfer_brain_data_items.csv')
        csv_content = File.read(file)
        CSV.parse(csv_content, headers: true)
      end
    end
```

(I've not tested that it works...)

mpw5 (approved):
LGTM. I think it's a definite improvement on the current code. I agree with the comment on the readability of the `csv` change, but it's not a blocker and would be fine to release in its current state.

app/models/claim/transfer_brain_data_item_collection.rb
jrmhaig: I've updated with along the lines of @mpw5's suggestion, and rebased.

PR #4979, opened 2022-06-29
Remove Deterimation#to_s

jrmhaig:
#### What

Remove the `Determination#to_s` method.

#### Ticket

Follow up to PR #4977 

#### Why

`Determination#to_s` appears to have been added for debugging in #1890 but never removed. Rubocop 1.31, updated in #4977, flags this. Instead of fixing the coding style it makes more sense to remove it if it is not being used.

#### To do

- [ ] Deploy on an environment and test with a few claims to confirm that this is only used for debugging.

#### How

Delete the method and update `.rubocop_todo.yml`.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -6,12 +6,6 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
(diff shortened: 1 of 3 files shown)


mpw5 (approved):
👍

PR #4984, opened 2022-07-04
Quickfix: remove case conclusion value

naseberry:
#### What

Fix Transfer details Javascript to remove value from hidden input

#### Ticket

n/a

#### Why

When signed in as an LGFS user and completing a Transfer fee claim, while editing the transfer details form step and toggling the visual state of the input `How did the case conclude?`, the validation is off.

If you edit the transfer details of an LGFS transfer claim and change from ‘new’ litigator type to ‘original’ you get a ‘Do not enter a case conclusion’ error when you try to save and continue.


imtiazAhmd (approved):
LGTM 👍🏼

PR #4985, opened 2022-07-04
[CFP-581] lgfs ecnp transfer claim spike

jrmhaig:
#### What

Map 'Elected Case Not Proceeded' transfer claims to the equivalent graduated fees for LGFS fee scheme 10.

When a litigator creates a transfer claim for fee scheme 10 and an 'elected case' is selected then the claim is treated instead as a 'cracked' trial. For fee scheme 9, an 'elected case not proceeded' claim is a fixed fee while a 'cracked' trial is a graduated fee, based on pages of prosecuting evidence.

#### Ticket

[Transfer fees - Implementing the 'elected case not proceeded' fee change](https://dsdmoj.atlassian.net/browse/CFP-639)

#### Why

The 'Elected Case Not Proceeded' case type does not exist in LGFS fee scheme 10 and the corresponding claims map to graduated fees.

#### How

The 'transfer brain' is used in transfer claims to select the correct 'bill scenario' to use when querying the Fee Calculator. The 'transfer brain' is therefore modified so that the 'bill scenario' corresponding to cracked trials are selected for an elected case.


spec/models/claim/transfer_claim_spec.rb
```diff
@@ -128,5 +130,39 @@
...
+      FeeScheme.find_or_create_by(name: 'LGFS', version: 10, start_date: Settings.lgfs_scheme_10_clair_release_date.beginning_of_day)
+    end
+
+    context 'when case is not elected' do
+      let(:options) { { elected_case: false } }
+
+      it { is_expected.to be_truthy }
+    end
+
+    context 'when fee scheme is 8 and case is elected' do
```
mpw5: Nitpick: I think this should be 9, not 8.

app/models/claim/transfer_brain/data_item_collection.rb
```diff
@@ -54,7 +54,9 @@ def valid_case_conclusion_ids(litigator_type, elected_case, transfer_stage_id)
       end
 
       def data_item_for(detail)
-        seek = DataItem.new(detail.slice(:litigator_type, :elected_case, :transfer_stage_id, :case_conclusion))
+        seek = DataItem.new(detail.slice(:litigator_type, :elected_case, :transfer_stage_id, :case_conclusion, :claim))
+        # TODO: Log this (or remove completely) instead of raising an error
```
mpw5, line 58: Do we want to address this TODO now? (Personally I'm happy to leave it for now, I don't think it's a blocker to releasing this).

mpw5 (approved):
Looks good to me. Excellent work 🏆 

I've left a couple of nitpicky comments, not anything that would block release though.

PR #5006, opened 2022-07-14
CFP-624 Spike validation of ECNP LGFS Final Fee claims for CLAIR

mpw5:
#### What

Spike the addition of validation to prevent a user creating an LGFS Final Fee for an Elected Case Not Proceeded claim under the new CLAIR fee scheme.

Note that this PR includes changes to add the new fee scheme. This is out of scope for this spike but was necessary to allow the work to be done. All relevant changes are in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/commit/676d173af92ef095c35d6b1ba50d3412bcc5e0a8.

#### Ticket

[CFP-624](https://dsdmoj.atlassian.net/browse/CFP-624)

#### How

* Adds an extra validation method to `app/validators/representation_order_validator.rb` to validate the representation order date for ECNP cases against the start of the CLAIR fee scheme.
* Adds wording to translation files to give an example error message. The actual wording for this error message is TBC.
* Adds tests to `spec/validators/representation_order_validator_spec.rb` to prove this works correctly for user-entered claims.
* Adds tests to `spec/api/v1/external_users/representation_order_spec.rb` to prove this works correctly for claims entered  via the API.


app/models/fee/base_fee_type.rb
```diff
@@ -25,7 +25,7 @@ def initialize(message = DEFAULT_MSG)
   end
 
   class BaseFeeType < ApplicationRecord
-    ROLES = %w[lgfs agfs agfs_scheme_9 agfs_scheme_10 agfs_scheme_12].freeze
+    ROLES = %w[lgfs agfs agfs_scheme_9 agfs_scheme_10 agfs_scheme_12 lgfs_scheme_9 lgfs_scheme_13].freeze
```
naseberry, line 28: 👍

app/validators/representation_order_validator.rb
```diff
@@ -70,6 +72,14 @@ def validate_maat_reference
     validate_matt_reference_uniqueness(:maat_reference, :unique)
   end
 
+  def validate_against_elected_case_not_proceeded_case_type
+    case_type = claim.try(:case_type)
+    return unless case_type&.name == 'Elected cases not proceeded' # could also check case_type.id == 9 here
```
naseberry, line 77: Nit: equality operator usage `==` or `eql?`. We have been using `eql?` lately

naseberry (approved):
LGTM

app/validators/representation_order_validator.rb
mpw5: `==` is preferred here as `case_type&.name` can be `nil`. In that scenario calling `.eql?` on it will result in a failure, whereas `==` will still work and give us the correct result.

mpw5:
Closing as this has been implemented in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5075. This spike is no longer needed.

PR #5018, opened 2022-07-21
CFP-627 Add LGFS fee scheme 10 (CLAIR)

mpw5:
#### What

Creates a new LGFS fee scheme 10 for the CLAIR changes to be introduced in October 2022.

This was previously spiked in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5006. This pull request takes that prototype, corrects some issues (particularly with fee scheme numbering) and adds extra testing .

**A temporary commit has been added to this PR to set the start date for the scheme to 01/06/2022 to allow testing. This causes tests to fail, but they will all pass when that commit is removed. This commit must be removed before merging the PR.**

#### Ticket

[CFP-627](https://dsdmoj.atlassian.net/browse/CFP-627)

#### How
* Creates a new seeder in `db/seeds/lgfs_scheme_10.rb` and adds it to `db/seeds.rb`, so that databases are automatically seeded with the new fee scheme;
* Creates a new rake task `lib/tasks/lgfs_scheme_ten.rake` which runs code in a new seed schema `db/seeds/schemas/add_lgfs_fee_scheme_10.rb` which controls the bulk of the process to create and rollback the fee scheme;
* Adds a start date for the fee scheme to `config/settings.yml`;
* Adds details of the new fee scheme to `app/models/fee_scheme.rb`;
* Adds details of the new fee scheme, and scheme 9 for completeness, to `app/models/fee/base_fee_type.rb`;
* Updates LGFS records in `lib/assets/data/fee_types.csv` to link fee types to the appropriate LGFS fee schemes;
* Updates unit tests and factories;
* Updates feature tests.


app/models/fee_scheme.rb
```diff
@@ -15,6 +15,7 @@ class FeeScheme < ApplicationRecord
   scope :ten, -> { where(version: FeeScheme::TEN) }
   scope :eleven, -> { where(version: FeeScheme::ELEVEN) }
   scope :twelve, -> { where(version: FeeScheme::TWELVE) }
+  scope :lgfs_ten, -> { lgfs.where(version: FeeScheme::TEN) }
```
naseberry: Do we need a scope to differentiate between schemes 9 & 10 for AGFS and LGFS?

mpw5: Good question. The `lgfs_ten` scope I added isn't being used anywhere. I thought we might need it but we don't - I just forgot to remove it. I'll delete it from the commit.

Where the numbered scopes are used, they're combined with the `agfs` and `lgfs` scopes, so we do `FeeScheme.lgfs.nine` or `FeeScheme.agfs.nine` or `FeeScheme.lgfs.ten` or `FeeScheme.agfs.ten`.

features/fee_calculator/litigator/interim_fee_calculator.feature
```diff
@@ -177,3 +177,182 @@ Feature: litigator completes interim fee page using calculator
     And the interim fee should have its price_calculated value set to true
 
     And I eject the VCR cassette
+
+  @fee_calc_vcr
```
naseberry, line 179: Nit: Could we define the feature flag `@fee_calc_vcr` once on line #1

mpw5: Fixed 👍

jrmhaig (approved):
:+1:

naseberry (approved):

PR #5049, opened 2022-08-09
[Spike] Allow use of future dates when creating claims

mpw5:
#### What

Add a mechanism to temporarily allow the use of future dates when creating claims.

#### Why

There are scenarios where being able to create a future-dated claim would be useful - for example when testing a new fee scheme. At the moment this requires temporary commits/pull requests against both the CCCD and Fee Calculator repos to amend the start date of the scheme, which is risky as there is the potential of accidentally deploying the change to production and also takes more maintenance due to the need to rebase and redeploy frequently.

A better option might be to allow an environment to be temporarily configured to permit claims to be created using future dates. 

#### How

Fee schemes are driven by the representation order date, however many other dates are also validated against being in the future so it is necessary to relax the validation for all such dates.

This pull request take a two-step approach:

1) Adding a new method `validate_not_in_future` to `app/validators/base_validator.rb` and replacing all other calls to `validate_on_or_before(Date.today...` with a call to that method. This also requires some small changes to `config/locales/en/models/claim.yml` to ensure consistency of errors.

2) Adding a clause to that new method to skip the validation if the `ALLOW_FUTURE_DATES` environment variable is true. This can be configured in the `app-config.yaml` file for each environment individually (or in `env.development` for local testing), allowing us to switch the functionality on as required. This PR sets the variable to `true` for `dev` and `false` for all other environments. Additional validation is included to prevent this functioning on production.

### DIFF
--- a/app/validators/base_validator.rb
+++ b/app/validators/base_validator.rb
@@ -157,6 +157,10 @@ class BaseValidator < ActiveModel::Validator
     compare_date_with_attribute(date, attribute, message, :>=)
   end
 
+  def validate_not_in_future(attribute)
+    validate_on_or_before(Time.zone.today, attribute, :check_not_in_future)
+  end
+
   def validate_has_role(object, role_or_roles, error_message_key, error_message)
     return if object.nil?
 
--- a/app/validators/base_validator.rb
+++ b/app/validators/base_validator.rb
@@ -158,7 +158,7 @@ class BaseValidator < ActiveModel::Validator
   end
 
   def validate_not_in_future(attribute)
-    validate_on_or_before(Time.zone.today, attribute, :check_not_in_future)
+    validate_on_or_before(Time.zone.today, attribute, :check_not_in_future) unless allow_future_dates
   end
 
   def validate_has_role(object, role_or_roles, error_message_key, error_message)
@@ -237,5 +237,11 @@ class BaseValidator < ActiveModel::Validator
   def looks_like_a_case_number?(attribute)
     return if attr_blank?(attribute)
     @record.__send__(attribute).match?(CASE_NUMBER_OR_URN_PATTERN)
+  end
+
+  def allow_future_dates
+    return false if ENV.fetch('ENV', nil).eql?('production')
+
+    ActiveRecord::Type::Boolean.new.cast(ENV.fetch('ALLOW_FUTURE_DATES', nil))
   end
 end
--- a/spec/validators/representation_order_validator_spec.rb
+++ b/spec/validators/representation_order_validator_spec.rb
@@ -176,6 +176,47 @@ RSpec.describe RepresentationOrderValidator, type: :validator do
         end
       end
     end
+
+    context 'when future dates are allowed' do
+      let(:reporder) { build(:representation_order, defendant:, representation_order_date: rep_order_date) }
+
+      before do
+        allow(ENV).to receive(:fetch).and_call_original
+        allow(ENV).to receive(:fetch).with('ALLOW_FUTURE_DATES', nil).and_return('true')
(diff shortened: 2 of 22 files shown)


spec/validators/representation_order_validator_spec.rb
```diff
@@ -176,6 +176,45 @@
         end
       end
     end
+
+    context 'when future dates are allowed' do
+      let(:reporder) { build(:representation_order, defendant:, representation_order_date: rep_order_date) }
+
+      before do
+        allow(ENV).to receive(:fetch).with('ALLOW_FUTURE_DATES', nil).and_return('true')
+        allow(ENV).to receive(:fetch).with('ENV', nil).and_return('dev')
```
jrmhaig, line 186: To ensure that other environment variables are not affected;

```suggestion
        allow(ENV).to receive(:fetch).and_call_original
        allow(ENV).to receive(:fetch).with('ALLOW_FUTURE_DATES', nil).and_return('true')
        allow(ENV).to receive(:fetch).with('ENV', nil).and_return('dev')
```

spec/validators/representation_order_validator_spec.rb
```diff
@@ -176,6 +176,45 @@
...
+
+      context 'when a past date is entered' do
+        let(:rep_order_date) { 2.days.ago }
+
+        it { expect(reporder).to be_valid }
+      end
+
+      context 'when on the the production environment' do
+        before do
+          allow(ENV).to receive(:fetch).with('ENV', nil).and_return('production')
```
jrmhaig, line 204: As above;

```suggestion
          allow(ENV).to receive(:fetch).and_call_original
          allow(ENV).to receive(:fetch).with('ENV', nil).and_return('production')
```

app/validators/base_validator.rb
```diff
@@ -234,4 +238,8 @@ def looks_like_a_case_number?(attribute)
     return if attr_blank?(attribute)
     @record.__send__(attribute).match?(CASE_NUMBER_OR_URN_PATTERN)
   end
+
+  def allow_future_dates
+    ENV.fetch('ALLOW_FUTURE_DATES', nil).eql?('true') && !ENV.fetch('ENV', nil).eql?('production')
```
jrmhaig: A couple of suggestions that are matters of opinion, so I don't have objections if you decide against them;

```suggestion
    return false if ENV.fetch('ENV', nil).eql?('production')

    ActiveModel::Type::Boolean.new.cast(ENV.fetch('ALLOW_FUTURE_DATES', false))
```

1. Using a guard clause to exclude the 'production' case up-front.
2. Rails has a helper to parse boolean values. It is really intended for http parameters but I believe it can be used here too.

spec/validators/representation_order_validator_spec.rb
mpw5: 👍

spec/validators/representation_order_validator_spec.rb
mpw5: 👍

app/validators/base_validator.rb
mpw5: Both good points. I've added the guard. I found that CCCD defines its own [helper](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/lib/extensions/string_extension.rb) for parsing boolean strings, so I've used that instead as it's a bit cleaner.

jrmhaig: Hmmm, I'd prefer to get away from those monkey-patched custom helpers, although they are fairly deeply embedded. I think this one was put in before the standard Rails way of doing it existed.

jrmhaig (approved):
:+1:

PR #5065, opened 2022-08-11
Add additional tests for creation of LGFS Scheme 10 claims via API

mpw5:
#### What

Adds additional tests to `spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb` to ensure that creation of LGFS Scheme 10 claims works correctly through the API.

Requires a rewrite of the spec file to move the tests into a `shared_example` which can then be reused to test LGFS Schemes 9 and 10.

#### Why

To increase our confidence that the LGFS CLAIR changes are working correctly.

### DIFF
--- a/spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
+++ b/spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
@@ -50,6 +50,220 @@ RSpec::Matchers.define :be_valid_api_lgfs_claim do |expected|
   end
 end
 
+RSpec.shared_examples 'LGFS API claim creation' do
+  context 'graduated fee claim' do
+    let(:case_type) { CaseType.find_by(fee_type_code: 'GRTRL') } # Trial
+
+    specify 'Case management system creates a valid graduated fee claim' do
+      post ClaimApiEndpoints.for(:final).create, claim_params.merge(offence_id: offence.id, actual_trial_length: 10), format: :json
+      expect(last_response.status).to eq 201
+
+      claim = Claim::BaseClaim.find_by(uuid: last_response_uuid)
+
+      post endpoint(:defendants), defendant_params.merge(claim_id: claim.uuid), format: :json
+      expect(last_response.status).to eq 201
+
+      defendant = Defendant.find_by(uuid: last_response_uuid)
+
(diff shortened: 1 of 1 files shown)


spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
```diff
@@ -226,229 +441,24 @@ def results(claim, expected)
...
-        post endpoint(:representation_orders), representation_order_params.merge(defendant_id: defendant.uuid), format: :json
-        expect(last_response.status).to eq 201
-
-        post endpoint(:fees), graduated_fee_params.merge(claim_id: claim.uuid, fee_type_id: graduated_fee_type.id, date: representation_order_date), format: :json
-        expect(last_response.status).to eq 201
-
-        expect(claim).to be_valid_api_lgfs_claim(fee_scheme: ['LGFS', 9], offence:, total: 5142.87, vat_amount: 1028.57)
-      end
-    end
+    it_behaves_like 'LGFS API claim creation'
```
jrmhaig: A couple of things you may consider (up to you);

1) I think, if you want, you can do this:

```ruby
    it_behaves_like 'LGFS API claim creation' do
      let(:representation_order_date) { Date.new(2022, 9, 30).as_json }
      let(:fee_scheme) { ['LGFS', 10] }
      let(:date_param) { '2022-10-01' }

      around do |example|
        travel_to(Settings.lgfs_scheme_10_clair_release_date + 1.month) do
          example.run
        end
      end
```

(and likewise for the scheme 9 version, above)


2) Also, instead of `around` I prefer something like this if possible;

```ruby
    before { travel_to(Settings.lgfs_scheme_10_clair_release_date + 1.month) }
    after { travel_back }
```

I see that `around` is used elsewhere and it may not be as simple to split into a `before` and `after` part.

jrmhaig (approved):
Just a couple of suggestions you may consider but no blockers.

spec/api/v1/external_users/claims/integration/litigator_claim_creation_spec.rb
mpw5: Both good points, thanks. I've updated.

(The `after` block isn't actually needed - when `travel_to` is called in a block, `travel_back` is implicitly called when the block completes.)

PR #5075, opened 2022-08-22
[CFP-626] Implement AGFS Fee Scheme 13

jrmhaig:
#### What

Implement the new AGFS fee scheme 13 for CLAIR.

#### Ticket

[Creating AGFS Fee Scheme 13 in CCCD](https://dsdmoj.atlassian.net/browse/CFP-626)

#### Why

Updating Legal Aid fees following CLAIR.

#### How

* Add Rake task to create AGFS Fee Scheme 13
* Add `agfs_scheme_13` as a new 'role' for `/api/advocate_categories`
* Add `FeeScheme.agfs_scheme_13?`
* Validation for earliest rep order date for elected case not proceeded cases

The error that is displayed for a scheme 13 rep order with an elected case not proceeded is;

Screenshot 2022-09-01 at 09 14 00

although this may change.


spec/services/claims/fee_calculator/graduated_price_spec.rb
```diff
@@ -527,6 +527,32 @@
             it_returns 'a successful fee calculator response', amount: 8585.00
           end
         end
+
+        context 'scheme 12' do
+          let(:offence_band) { create(:offence_band, description: '1.1') }
+          let(:offence) { create(:offence, :with_fee_scheme_eleven, offence_band:) }
```
mpw5: Should this use the `:with_fee_scheme_twelve` trait, instead of `:with_fee_scheme_eleven`?

spec/services/claims/fee_calculator/graduated_price_spec.rb
```diff
@@ -527,6 +527,32 @@
...
+          context 'trial' do
+            let(:case_type) { create(:case_type, :trial) }
+
+            it_returns 'a successful fee calculator response', amount: 8585.00
+          end
+        end
+
+        context 'scheme 13' do
+          let(:offence_band) { create(:offence_band, description: '1.1') }
+          let(:offence) { create(:offence, :with_fee_scheme_eleven, offence_band:) }
```
mpw5: Should this use the `:with_fee_scheme_thirteen` trait, instead of `:with_fee_scheme_eleven`?

app/validators/claim/base_claim_validator.rb
```diff
@@ -372,4 +372,12 @@ def has_higher_rate_mileage?
   def increased_travel?
     @record.expenses.find { |x| x.calculated_distance && x.distance && (x.distance > x.calculated_distance) }
   end
+
+  def validate_earliest_representation_order_date
+    return unless @record.case_type&.name == 'Elected cases not proceeded'
+    return unless @record.earliest_representation_order_date
+    return if @record.earliest_representation_order_date < Settings.agfs_scheme_13_clair_release_date
```
mpw5, line 380: I wonder if it's worth noting in a comment that although this uses the `agfs_scheme_13_clair_release_date` for validation, it actually affects both AGFS and LGFS claims? It's possible that someone in future might assume from the code that we're only validating AGFS claims.

features/fee_calculator/advocate/fixed_fee_calculator.feature
```diff
@@ -85,21 +85,40 @@ Feature: Advocate completes fixed fee page using calculator
...
+      | fixed   | Number of defendants uplift | 38.80  | Number of additional defendants | true |
 
     And I eject the VCR cassette
 
     Then I click "Continue" in the claim form
     And I am on the miscellaneous fees page
 
     And all the fixed fees should have their price_calculated values set to true
+
+  Scenario: I create attempt to create a post-CLAIR elected cases not proceeded claim
```
mpw5, line 107: The ECNP validation also affects LGFS claims. I think we should also add this test to `/features/fee_calculator/litigator/fixed_fee_calculator.feature`.

(Not a blocker for this PR, we can do it as a separate change if necessary.)

mpw5 (commented):
I've left some comments in the code. Nothing major, I think this is good to go in its current state, but feel free to consider them.

A couple of other thoughts around testing:

1. Is it worth updating `spec/api/v1/external_users/claims/integration/advocate_claim_creation_spec.rb` to add tests for scheme 13? (This is the one I missed in my original PR and updated separately in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5065). It feels like there's some value in assuring that we can create claims for the new fee scheme through the API.

2. I think we also need to consider adding some additional feature tests. For the LGFS changes I replicated the existing tests for the previous fee scheme in `features/claims/litigator` and `features/fee_calculator/litigator`. We could do the same here for the advocate tests. I wouldn't object to this not being done as part of this PR though, I think we could create a separate ticket for it, which might tie in with the work Angela is doing on https://dsdmoj.atlassian.net/browse/CFP-678. Happy to discuss.

app/validators/claim/base_claim_validator.rb
jrmhaig: Good point. LGFS hadn't occurred to me at the time. Strictly it should probably have `if agfs ... elsif lgfs ...` but just having a comment is probably more sensible.

mpw5 (approved):
LGTM 🎉

config/settings.yml
```diff
@@ -103,6 +103,7 @@ agfs_fee_reform_release_date: <%= Date.new(2018, 4, 1) %>
 agfs_scheme_11_release_date: <%= Date.new(2018, 12, 31) %>
 clar_release_date: <%= Date.new(2020, 9, 17) %>
 lgfs_scheme_10_clair_release_date: <%= Date.new(2022, 9, 30) %>
+agfs_scheme_13_clair_release_date: 2022-09-30
```
naseberry, line 106: Would it be ideal to note the reason behind the different date format?

mpw5: The plan is to change all the dates to the string format Joe has used in this commit. We just don't want to do it as part of this change, to reduce the risk of changing too many things at once and breaking something. I guess we could add a comment to that effect, but it would (hopefully) be short-lived (See https://dsdmoj.atlassian.net/browse/CFP-712).

spec/factories/fee_schemes.rb
```diff
@@ -29,6 +29,12 @@
       version { 12 }
     end
 
+    trait :agfs_thirteen do
```
naseberry, line 32: As we now have a new agfs scheme, do we need to apply an end date to the previous agfs scheme (`trait :agfs_twelve`)?

config/settings.yml
jrmhaig: That is correct. I intend to change all the dates but as a separate PR. It is possible that it could have some side effects on some tests, but (I'm fairly sure) not on the production code.

features/claims/advocate/scheme_thirteen/advocate_fixed_fee_claim_submit.feature
```diff
@@ -0,0 +1,53 @@
...
+    Then I click "Continue" in the claim form
+    And I should be in the 'Miscellaneous fees' form page
+
+    Then I click the link 'Back'
+    And I should be in the 'Fixed fees' form page
+
+    When I uncheck the govuk checkbox "Appeals to the crown court against conviction"
+    Then I click "Continue" in the claim form
+    And I should be in the 'Fixed fees' form page
+    And I should see govuk error summary with 'Total value claimed must be greater than £0.00'
```
mpw5, line 52: Interesting that this feature is called `advocate_fixed_fee_claim_submit.feature` but the claim doesn't actually get submitted at the end 🤦

features/claims/advocate/scheme_thirteen/advocate_trial_claim_edit_submit.feature
```diff
@@ -0,0 +1,58 @@
...
+      | miscellaneous | Unused materials (over 3 hours)  | 45.30  | Number of hours | true |
+      | miscellaneous | Paper heavy case                 | 45.30  | Number of hours | true |
+      | miscellaneous | Deferred sentence hearings       | 201.00 | Number of days  | true |
+
+    When I click "Continue" in the claim form
+    And I click the link 'Back'
+    Then I should be in the 'Miscellaneous fees' form page
+    And I should not see 'This claim should be eligible for unused materials fees (up to 3 hours)'
+
+    And I eject the VCR cassette
```
mpw5, line 58: Another one that doesn't actually submit the claim at the end...

mpw5 (approved):
LGTM

I've left comments against a couple of feature tests where the name suggests that the claim should be submitted at the end but the test doesn't go that far. I don't think it should be a blocker at this point (maybe the tests could be renamed, or we can review them when Angela has finished https://dsdmoj.atlassian.net/browse/CFP-678 and make changes if necessary).

PR #5076, opened 2022-08-23
Allow future rep order dates on api-sandbox

jrmhaig:
#### What

Allow future rep order dates on api-sandbox.

#### Ticket

N/A

#### Why

This will allow software vendors to test in advance of the new fee scheme change.

#### How

Set `ALLOW_FUTURE_DATES` to true in `.k8s/live/api-sandbox/app-config.yaml`.


mpw5 (approved):
👍

PR #5102, opened 2022-08-31
Update net-scp version

jrmhaig:
#### What

Update the `net-scp` and `net-ssh` gems.

#### Ticket

N/A

#### Why

`net-ssh` version 7 removes some insecure cyphers but `net-scp` version 3 pins `net-ssh` to < 7. Dependabot will not update `net-scp` as the latest version is a release candidate, `4.0.0.rc1`. These gems are only installed in development and are only used in;

* `script/db_dump.rb`
* `script/db_static_data_upload.rb`
* `script/db_upload.rb`

so there is no great risk in using a release candidate but it will allow the more secure version of `net-ssh` to be used.

#### How

Update `Gemfile`;

```ruby
  gem 'net-ssh', '~> 7.0'
  gem 'net-scp', '>= 4.0.0.rc1'
```


mpw5 (approved):
👍

PR #5103, opened 2022-09-01
Allow creation of future-dated warrant fees

mpw5:
While testing https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5075 I discovered I could not create a future-dated warrant fee despite having the `ALLOW_FUTURE_DATES` flag set to `true`:

image

This change adds an extra check to the validation in `Fee::WarrantFeeValidator` so that the relevant validation is bypassed when the flag is true.


jrmhaig (approved):
:+1:

PR #5104, opened 2022-09-05
Fix flakey tests

jrmhaig:
#### What

Fix some flakey tests.

#### Ticket

N/A

#### Why

Following 25e69d546ec1ab6769fcee9534e8d879e77f4cc4 the `@data_items` instance variable is not created in the class initializer so it will be `nil` unless other tests are run first.

#### How

The subject is changed to the (private) method instead of the instance variable.


mpw5 (approved):
LGTM

PR #5112, opened 2022-09-08
Remove naseberry as a dependabot reviewer

naseberry:
And whether we shall meet again I know not.
Therefore our everlasting farewell take:
For ever, and for ever, farewell, CCCD!
If we do meet again, why, we shall smile;
If not, why then, this parting was well made.


mpw5 (approved):
For ever and for ever farewell, Naseberry!
If we do meet again, we'll smile indeed;
If not, 'tis true this parting was well made.

PR #5117, opened 2022-09-09
Fix ordering of fee scheme seeding

mpw5:
#### What

The ordering of the seeding of AGFS fee scheme 13 and LGFS fee scheme 10 causes failures when seeding offence data in new databases due to dependencies on the `offence_id`.

This PR fixes the ordering so that offences are seeded successfully. We may want to change this functionality in future to prevent this problem occurring again. It also fixes a small typo in the AGFS fee scheme 13 seeder.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -27,11 +27,11 @@ SEED_FILES = %w[
   scheme_10
   scheme_11
   scheme_12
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
:+1:

PR #5136, opened 2022-09-16
[CFP-712] Fix format of dates in config/settings.yml

mpw5:
#### What

Several date parameters are defined in `config/settings.yml`. These are created in erb blocks using `Date.new(year, month, day)` which formats the date by default as `yyyy-mm-dd`. This is how the `config` gem expects dates to be formatted.

There is no particular reason why the dates need to be generated using `erb`. Some dates have been converted to simple strings in earlier commits. This PR converts remaining dates in the settings file to `yyyy-mm-dd` strings. It also removes some unnecessary old code for the release of a change to reject/refuse messaging which took place several years ago and can now be removed.

#### Ticket

[CFP-712](https://dsdmoj.atlassian.net/browse/CFP-712)


config/settings.yml
```diff
@@ -97,11 +97,11 @@ maintenance_mode_enabled?: <%= ENV.fetch('MAINTENANCE_MODE', nil)&.downcase&.eql
 feature_flags_enabled?: true
 active_features: []
 
-reject_refuse_messaging_released_at: <%= DateTime.new(2018, 4, 24, 23, 05, 0).change(offset: '+1:00') %>
+reject_refuse_messaging_released_at: 2018-04-24
```
jrmhaig: Is the time significant here?

Also, is it possible that we can dispense with the `reject_refuse_messaging_released_at` method in `lib/release.rb` after this change? It looks as though it is just catering for the ambiguous type of that setting. I suspect that this fix exists in a lot of places.

Also also, is the `reject_refuse_messaging_released?` method in that same class redundant now, as we are a long way past the release date?

For reference, `lib/release.rb` was created in a09775ab119bbfc9f887e665eaacb0b489577e2e, from #2327.

mpw5: All good questions! I was just looking at that one again now - I'm going to add the time back in for safety, although I'm not sure it's entirely relevant at this point. There's also some tidying up of tests and other code that can be done, so I'll add an extra commit with that.

jrmhaig (approved):
Looks good.

PR #5152, opened 2022-09-26
[CFP-642] Update ECNP final fee error message

mpw5:
#### What

The error displayed when trying to claim a Final ECNP fee on a claim that falls in the CLAIR fee schemes (LGFS 10 or AGFS 13) currently has a temporary wording.

This change now updates that to the the correct wording.

#### Ticket

[CFP-642](https://dsdmoj.atlassian.net/browse/CFP-642)

#### How

Updates the temporary message `Check combination of representation order date and case type, on the previous page` to the new message `The representation order date and case type cannot be combined` in translation and tests.


jrmhaig (approved):
🎉

PR #5163, opened 2022-09-29
[CFP-574] Investigate sessions in CCCD

NenehB:
What
Currently, users can stay signed in to CCCD without their session expiring and being logged out.

Ticket

[CFP-754 - Investigating sessions in CCCD](https://dsdmoj.atlassian.net/browse/CFP-574)

Why
This poses as a security risk as it increases the likelihood that hostile attacks can gain access to accounts that are already logged in to our production environments. 

How
Logging users out after 3 hours of inactivity.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -40,6 +40,7 @@ class User < ApplicationRecord
           :recoverable,
           :rememberable,
           :trackable,
+          :timeoutable,
           :validatable,
           :lockable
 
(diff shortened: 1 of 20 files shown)


app/models/user.rb
```diff
@@ -40,6 +40,7 @@ class User < ApplicationRecord
           :recoverable,
           :rememberable,
           :trackable,
+          :timeoutable,
```
jrmhaig, line 43: Is the timeout limit configured somewhere?

mpw5: It's configured in the `devise` initializer: [`config/initializers/devise.rb`](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/36155ab6a470e89a3e515619bd53b3e266f3a84d/config/initializers/devise.rb#L153).

mpw5 (approved):
Great work 🥇

app/models/user.rb
NenehB: yes in the devise.rb file

PR #5164, opened 2022-09-30
Small cleanup of API smoke test runner

jrmhaig:
#### What

Some refactoring of `spec/support/api`

#### Ticket

N/A

#### Why

General housekeeping

#### How

1. Change control of debugging to an attribute managed by the `Debuggable` module.
2. Handle API response in test client so that a consistent data structure is returned. `post_to_endpoint` was returning an instance of `RestClient::Response` while `get_dropdown_endpoint` was returning a JSON string of the response body. Now the response body converted to a hash is returned in both cases.
3. Reorganise test classes. Advocate and litigator tests are in separate directories and class inheritance is used to avoid duplicating common code.
4. Add the smoke tests to the development build pipeline. The tests will run after `build-app-container` and failure will not block deploying to an evironment. 

(2) is useful for a possible future piece of work to migrate from Rest Client to Faraday (see [this Jira ticket](https://dsdmoj.atlassian.net/browse/CFP-491)) as the tests depended on the API response objected having a string representation, which Rest Client has but Faraday doesn't.


.circleci/config.yml
```diff
@@ -70,8 +70,13 @@ executors:
...
     resource_class: medium
     working_directory: /usr/src/app
+    parameters:
+      tag:
+        description: Image tag to use for test
+        default: app-latest
+        type: string
     docker:
-      - image: ${ECR_ENDPOINT}/laa-get-paid/cccd:app-latest
+      - image: ${ECR_ENDPOINT}/laa-get-paid/cccd:<<parameters.tag>>
```
mpw5, line 79: Nice! I like that approach 👍

mpw5 (approved):
Nice! 

One suggestion: would it be worth adding something to the testing section of the readme (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/docs/testing.md) to document this, as it was something that we'd both missed before? Possibly with an explanation of how the debug function works (your [commit](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5164/commits/65055a49a641d28466d8921317ede04422c726bf) message for it is really clear).

jrmhaig:
> One suggestion: would it be worth adding something to the testing section of the readme (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/docs/testing.md) to document this, as it was something that we'd both missed before?

Thanks. I've added something now.

mpw5:
> > One suggestion: would it be worth adding something to the testing section of the readme (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/docs/testing.md) to document this, as it was something that we'd both missed before?
> 
> Thanks. I've added something now.

Thanks. Happy for it to be released.

PR #5165, opened 2022-10-03
CFP-579 Update ingress controllers to v1.2

mpw5:
#### What

Update ingress controllers to v1.2

#### Ticket

[CFP-579](https://dsdmoj.atlassian.net/browse/CFP-579)

#### Why

To comply with Cloud Platform requirements, remain in service, and benefit from improved security and logging.

#### How

Updates kubernetes configuration for all five environments to use v1.2 of the ingress controller. This has been applied using the zero-downtime guidance provided by the Cloud Platform team: https://user-guide.cloud-platform.service.justice.gov.uk/documentation/networking/Switch-ingress-to-v1-ingress-controller.html#introduction. This PR tidies up the configuration in code so that future deployments will be consistent.

PR #5166, opened 2022-10-04
Revert "CFP-579 Update ingress controllers to v1.2"

mpw5:
Reverts ministryofjustice/Claim-for-Crown-Court-Defence#5165 due to issues experienced by API users on the new ingress.


jrmhaig (approved):
:cry:

PR #5184, opened 2022-10-05
[CFP-720] Allow KC as advocate category

jrmhaig:
#### What

Allow the API to accept KC as an advocate category.

#### Ticket

[Allow KC as advocate category](https://dsdmoj.atlassian.net/browse/CFP-720)

#### Why

It is possible that vendor software will start creating claims with an advocate category 'KC'. These claims should not fail and they should be dealt with correctly.

#### How

When a claim is submitted with an advocate category of 'KC' it is changed to 'QC' before creating the claim instance. This is a temporary measure until the data is changed internally.

The 'KC' option now appears in the API Swagger documentation so it may result in it appearing, alongside 'QC', in some vendor software. Is this a problem?

Screenshot 2022-10-06 at 09 13 10

### DIFF
--- a/app/interfaces/api/helpers/resource_helper.rb
+++ b/app/interfaces/api/helpers/resource_helper.rb
@@ -38,6 +38,7 @@ module API::Helpers
     end
 
     def validate_resource(klass)
+      params['advocate_category'] = 'QC' if params['advocate_category'] == 'KC'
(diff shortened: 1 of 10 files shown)


mpw5 (approved):
LGTM

mpw5 (approved):
👍

PR #5202, opened 2022-10-14
Use Singleton class for Remote::HttpClient

jrmhaig:
#### What

Use the `Singleton` class for `Remote::HttpClient`.

~Note, this will probably conflict with #5200, which should have priority.~ #5200 still needs more work and it is not a priority at the moment.

A further point of discussion is whether we should be using singletons at all. Some consider it to be an anti-pattern.

#### Ticket

N/A

#### Why

`Remote::HttpClient` is a singleton class but it is using a custom implementation of the singleton pattern instead of the built-in Singleton module.

#### How

* The `Singleton` module deals with making `new` private and defining `instance`
* It is now not necessary to test the singleton `instance` method
* The class variables are changed to instance variables
* The configure method sets the singleton's instance variables instead of class variables


mpw5 (approved):
👍

PR #5204, opened 2022-10-17
[CTSKF-10] Update delegated authority limits

jrmhaig:
#### What

Update delegated authority limits;

* £25,000 -> £30,000
* £100,000 -> £115,000
* £150,000 -> £175,000

#### Ticket

[Implement delegated authority changes in CCCD](https://dsdmoj.atlassian.net/browse/CTSKF-10)

#### Why

After the increase in fees for CLAIR the authority limits for case workers are being amended accordingly.

#### How

Update the filter in the case worker admins' allocation page.


app/services/claims/value_bands.rb
```diff
@@ -3,10 +3,10 @@ class ValueBands
...
 
     VALUE_BANDS = {
-      10 => Struct::ValueBandDefinition.new(10, 'less than £25,000', 0.0, 25_000.0),
-      20 => Struct::ValueBandDefinition.new(20, '£25,001 - £100,000', 25_000.01, 100_000.0),
-      30 => Struct::ValueBandDefinition.new(30, '£100,000 - £150,000', 100_000.01, 150_000.0),
-      40 => Struct::ValueBandDefinition.new(40, 'more than £150,000', 150_000.01, 99_999_999.99)
+      10 => Struct::ValueBandDefinition.new(10, 'less than £30,000', 0.0, 30_000.0),
+      20 => Struct::ValueBandDefinition.new(20, '£30,001 - £115,000', 30_000.01, 115_000.0),
+      30 => Struct::ValueBandDefinition.new(30, '£115,001 - £175,000', 115_000.01, 175_000.0),
+      40 => Struct::ValueBandDefinition.new(40, 'more than £175,000', 175_000.01, 99_999_999.99)
```
mpw5: Nitpick: Looking at this again, I think it would make sense to change the descriptions to match the actual thresholds. It just looks a bit wrong at the moment. 

```suggestion
      10 => Struct::ValueBandDefinition.new(10, '£30,000 or less', 0.0, 30_000.0),
      20 => Struct::ValueBandDefinition.new(20, '£30,000.01 - £115,000', 30_000.01, 115_000.0),
      30 => Struct::ValueBandDefinition.new(30, '£115,000.01 - £175,000', 115_000.01, 175_000.0),
      40 => Struct::ValueBandDefinition.new(40, 'more than £175,000', 175_000.01, 99_999_999.99)
```

Although as this doesn't seem to be used and we might remove it, maybe it's irrelevant 🤷

mpw5:
Moving to Ready for Review as we appear to be planning to release tomorrow (20/10/2022).

config/locales/en.yml
```diff
@@ -2425,10 +2425,10 @@ en:
         filter_value_bands:
           all: All
           value: 'Value'
-          tier1: less than £25,000
-          tier2: £25,000 - £100,000
-          tier3: £100,000 - £150,000
-          tier4: more than £150,000
+          tier1: less than £30,000
```
Obsiye, line 2428: sorry to nitpick, i'm assuming this should be `£30,000 or less` or are we preferring this way for users readability?

spec/models/claims/totals_spec.rb
```diff
@@ -236,45 +236,30 @@
...
 
-        create :disbursement, claim: claim, net_amount: 25_000.0, vat_amount: 5_000.0
-
-        claim.reload
-        expect(claim.total).to eq 25_025.0
-        expect(claim.vat_amount).to eq 5_000.0
-        expect(claim.value_band_id).to eq 20
+        it { expect { new_fee }.to change(claim, :total).by(30_000) }
+        it { expect { new_fee }.to change(claim, :vat_amount).by(5_000) }
+        it { expect { new_fee }.to change { claim.reload.value_band_id }.from(10).to(20) }
```
Obsiye, line 244: for clarification, the auth limit is affected by vat?

like here, Disbursement is 30,000 then VAT of 5,000 is added then the auth limit is checked?

Would this need to be in line with CCR/CCLF?

mpw5: Yes, the authority limit applies to the total value of the claim, including VAT. In CCCD `total` does not include VAT (it's the sum of all fees net of VAT) so we need to specify `vat_amount` separately.

Paul and I discussed this in a meeting with Nick Poulter and people from Finance and I think we agreed that this is the correct approach and consistent across all the apps. This PR is not changing the way it behaves, just the thresholds, so if it's working correctly now then it should continue to work correctly after the increase.

config/locales/en.yml
mpw5: I think your nitpick is technically correct, but `less than £xx,xxx` is the current wording which caseworkers are used to so it seems sensible to leave it as it is. Also the chances of receiving a claim that is exactly £30,000.00 is so small that we probably don't need to worry about it.

spec/models/claims/totals_spec.rb
Obsiye: ohh makes sense thank you. i'm getting myself confused XD

Obsiye (approved):

PR #5211, opened 2022-10-19
Allow Claim for Payment as code owner

jrmhaig:
#### What

Update the code owners to list both Claim for Payment and CLAIR Taskforce as code owners.

#### Ticket

N/A

#### Why

With the previous configuration, the second line making the CLAIR Taskforce the code owner of the whole repository took ownership away from the Claim for Payment team on the previous line.

#### How

Listing both teams on the same line should make both code owners.


mpw5 (approved):
Let's give this a try 😀

PR #5215, opened 2022-10-24
[CTSKF-28] Update ModSec rules to reduce false positives when sending messages

mpw5:
#### What
Disable ModSecurity rule 921110 for `/messages`.

#### Ticket

[CTSKF-28](https://dsdmoj.atlassian.net/browse/CTSKF-28)

#### Why

Caseworkers and providers sending messages on claims are sometimes experiencing failures. This occurs because requests to `/messages` are being blocked by the WAF as false positives because they contain the text 'post' (or other HTTP method names) in combination with a line feed or carriage return: 

e.g. `post 16.5.22\x0d`

This violates rule [921110 - HTTP Request Smuggling](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/a216353c97dd6ef767a6db4dbf9b724627811c9b/rules/REQUEST-921-PROTOCOL-ATTACK.conf#L62).

#### How

Disable rule `921110` using the `ctl:ruleRemoveById` command only for requests where the `REQUEST_URI` contains `/messages`. This is applied to ModSecurity Phase 2 which processes the request body.

```
      SecRule REQUEST_URI "@contains /messages" \
        "id:1000,phase:2,pass,nolog,ctl:ruleRemoveById=921110"
```

See these links for more information:
https://coreruleset.org/docs/concepts/false_positives_tuning/#example-5-ctlruleremovebyid
https://malware.expert/modsecurity/processing-phases-modsecurity/

#### Testing

Before, attempting to send a message that contains `post` and a carriage return results in a 403 error and the message is not posted:

image

After, the same message returns a 200 status and is posted successfully:

image

### DIFF
--- a/.k8s/live/api-sandbox/ingress.yaml
+++ b/.k8s/live/api-sandbox/ingress.yaml
@@ -12,6 +12,8 @@ metadata:
       SecAction "id:900200,phase:1,nolog,pass,t:none,setvar:tx.allowed_methods=GET HEAD POST OPTIONS PUT PATCH DELETE"
       SecRuleRemoveById 200002
       SecRuleRemoveById 200003
+      SecRule REQUEST_URI "@contains /messages" \
+        "id:1000,phase:2,pass,nolog,ctl:ruleRemoveById=921110"
     nginx.ingress.kubernetes.io/server-snippet: |
       deny 116.204.211.188;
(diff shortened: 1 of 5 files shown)


jrmhaig (approved):
:+1:

ivanELEC (approved):

PR #5223, opened 2022-10-31
Remove 'pending' from timezone sensitive test

mpw5:
#### What

Reverts the change made in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4773/files to allow a timezone sensitive test to continue to pass during BST. Now we are back in GMT the temporary change can be removed, as it is now causing the test to fail.

This will need to be re-added in March 2023 if we do not fix this issue before.


jrmhaig (approved):
:+1:

PR #5243, opened 2022-11-11
Fix cucmber_opts warning message

mpw5:
#### What

Running any rake task currently results in the following message being displayed:

```
WARNING: consider using an array rather than a space-delimited string with cucumber_opts to avoid undesired behavior.
```

This changes the parameters passed to `cucumber_opts` in `cucumber.rake` from strings to an array of strings to prevent this message being displayed.

(I've never used this rake task. If no one else does then a better alternative might be to simply delete it.)


jrmhaig (approved):
I don't use this task either but it looks like a sensible change.

PR #5268, opened 2022-11-29
[CTSKF-70] Add main hearing date to case details form

jrmhaig:
#### What

Add the main hearing date field to the case details form.

#### Ticket

[CCCD: Add main hearing date to 'Case details' form](https://dsdmoj.atlassian.net/browse/CTSKF-70)

#### Why

The main hearing date will be used for determining the fee scheme in addition to the representation order date.

#### How

Add an extra field to the case details form.

Screenshot 2022-11-29 at 12 18 06

--------

#### TODO (wip)

 - [ ] Validations
 - [ ] Styling; By default, the label for a [date field in form builder](https://govuk-form-builder.netlify.app/form-elements/date-input/) uses `govuk-fieldset__legend`, which is bold. This is different from the [text input,](https://govuk-form-builder.netlify.app/form-elements/text-input) which uses `govuk-label`. This results in an inconsistent form, as shown above. This is also how date fields appear on other pages, such as the defendants page. Is this an issue?
 - [x] Hide behind feature flag
 - [x] Update feature tests


features/step_definitions/common_steps.rb
```diff
@@ -210,3 +210,7 @@
   overwrite_constant :ENABLED, false, Rack::NoPopups
 end
 
+Given(/^the main hearing date feature is enabled$/) do
+  ENV['MAIN_HEARING_DATE_ENABLED'] = 'true'
```
Obsiye: i'm curious, will this revert back to the original value of MAIN_HEARING_DATE_ENABLED after the tests are run?

app/controllers/external_users/claims_controller.rb
```diff
@@ -289,6 +289,7 @@ def claim_params
       :case_number,
       :case_type_id,
       :case_stage_id,
+      :main_hearing_date,
```
mpw5: This is a duplicate of line 302. I think it was previously added in Neneh's PR to display the main hearing date on the summary page.

mpw5 (approved):
LGTM. One minor comment

features/step_definitions/common_steps.rb
jrmhaig: As it turns out, yes this is persistent. I have changed it so that it gets set at the start of the suite instead of individually on a feature.

NenehB (approved):
LGTM

PR #5269, opened 2022-11-29
[CTSKF-68] Add main_hearing_date to AGFS API endpoints

mpw5:
#### What

Adds a `main_hearing_date` parameter to the following AGFS endpoints, hidden behind a feature flag until ready for release to production:

`/api/external_users/claims/advocates/final`
`/api/external_users/claims/advocates/interim`
`/api/external_users/claims/advocates/supplementary`
`/api/external_users/claims/advocates/hardship`
`/api/external_users/claims/advocates/final/validate`
`/api/external_users/claims/advocates/interim/validate`
`/api/external_users/claims/advocates/supplementary/validate`
`/api/external_users/claims/advocates/hardship/validate`

Also clears up rubocop offences in the affected files, and refactors the tests for clarity.

#### Ticket

[CTSKF-68](https://dsdmoj.atlassian.net/browse/CTSKF-68)

#### Why

To allow the inclusion of `main_hearing_date` in API claims, so that providers are paid correctly for their work.

#### Note

Some tests for scenarios where the feature flag is enabled currently fail, due to the way the API endpoints are initialised before the tests run, making it difficult to stub. These have been marked as `skip` for now; `pending` was not suitable as some shared examples are currently passing. When the feature flag is removed, these tests will need to be enabled and the tests that are currently running (for scenarios where the feature flag is disabled) removed.

#### Evidence

Tested in `dev` with the feature flag set to false (so `main_hearing_date` is not available):
image

Tested in `dev-lgfs` with the feature flag set to true (so `main_hearing_date` is available):
image


spec/api/v1/external_users/claims/advocates/final_claim_spec.rb
```diff
@@ -41,35 +41,73 @@
...
+        ].each do |error|
+          expect(last_response.status).to eq(400)
+          expect(last_response.body).to include(error)
+        end
+      end
+    end
+  end
+
+  context 'when CLAIR contingency functionality is enabled',
+          skip: 'Skipped pending removal of the main_hearing_date feature flag' do
```
jrmhaig, line 67: Can `pending` be used instead of `skip`? A pending test will report as a failure if it starts passing so it alerts when the feature has been fixed (or turned on, in this case). I don't know if `pending` can be used as metadata on the `context`, like `skip` here.

mpw5: I'd have liked to use `pending` but there are issues with it:

* It's not possible to use `pending` on individual failing tests as they are mostly in shared examples (actually, all of them are following my refactor). Marking them as pending causes failures when they run in the context where the feature flag is disabled.

* It's possible to mark a whole `context` as pending, similar to how I used `skip` in the PR, but as not all the specs in that context are failing, using `pending` leads to failures where the tests are passing (if that makes sense 🤯).

If you've got any thoughts on other approaches I'd love to hear them 🙂

Obsiye (approved):

spec/api/v1/external_users/claims/advocates/final_claim_spec.rb
jrmhaig: OK. This should be quite short lived so it is probably not worth it.

jrmhaig (approved):

PR #5296, opened 2022-12-14
[CTSKF-127] Update ecnp signposting text

NenehB:
What

When CLAIR phase one was introduced, signposting is used to alert providers to the fact that elected cases not proceeded final fees can no longer be claimed. This text needs to be updated to also include claims that fall under the CLAIR contingency rules

Ticket

[CTSKF-127- Update ecnp signposting text](https://dsdmoj.atlassian.net/browse/CTSKF-127)

Why

So that providers can claim the correct fee

TODO (wip)

 - [x] Testing


jrmhaig (approved):
Looks good

PR #5300, opened 2022-12-20
Fix wrapping of 'Judicial apportionment' text on claim summary screen

mpw5:
#### What

Fixes the incorrect wrapping of the text 'Judicial apportionment' on the summary screen of a submitted claim.

#### Why

In the Defendants section of the summary screen for a submitted claim, the text 'Judicial apportionment' wraps inelegantly, with the last 't' appearing on a new line:

image

After this change the text appears correctly:

image

#### How

Adjusts the width of the key in the `app-card--defendant` element to allow the text to appear correctly.

### DIFF
--- a/app/webpack/stylesheets/components/_card.scss
+++ b/app/webpack/stylesheets/components/_card.scss
@@ -1,4 +1,7 @@
 .app-card--defendant {
+  .govuk-summary-list__key {
+    width: 35%;
+  }
   margin-top: $gutter-half;
   margin-bottom: $gutter-half;
   padding: $gutter-half $gutter-one-third $gutter-half $gutter-half;


app/webpack/stylesheets/components/_card.scss
```diff
@@ -1,6 +1,7 @@
 .app-card--defendant {
+  width: 445px;
```
pking-moj: This hard coding of width bothers me somewhat as when your scaling the page for different screen sizes it doesnt help. If its just the word wrapping thats the problem could we use `overflow-wrap: inherit;` to potentially fix this? The other thing is the Key width is 30% on the card `.govuk-summary-list__key` so we could potentially override this to be 35% if you wanted to.

jrmhaig:
I'm also bothered by hard-coded values.

Also, I see that the defendants are displayed differently on the summary page for providers and case workers. The cards shown above are from the caseworkers' view, while providers see them with full width. We could remove the problem by getting rid of the cards and, at the same time, bring the two summary pages a little closer together.

I'll check to see if there is anything in the history to explain why it is done like this.

app/webpack/stylesheets/components/_card.scss
mpw5: Ta. I tried amending `overflow-wrap` initially, but couldn't get it to have any effect. Overriding the key width worked nicely though.

mpw5:
> Also, I see that the defendants are displayed differently on the summary page for providers and case workers. The cards shown above are from the caseworkers' view, while providers see them with full width. We could remove the problem by getting rid of the cards and, at the same time, bring the two summary pages a little closer together.

My screenshots in the PR description are both from the provider's view, but _after_ the claim has been submitted. You're right, though, that the provider sees the defendant details in a different format _before_ submitting. 

Some consistency would be nice but would be more work than I was planning to get into today, I just found the cosmetic issue jarring 😅 

It feels like this is something that would be worth getting a designer's eye on. I'll put a ticket in the backlog.

pking-moj (approved):

PR #5301, opened 2022-12-20
Allow CircleCI to correctly split cucumber tests

mpw5:
#### Why

CircleCI is configured to split cucumber tests by timings to allow for more efficient parallelisation:

`circleci tests split --split-by=timings --timings-type=filename`

However this does not work correctly. All tests result in the message:

image

#### How

This is occurring because the test results are output using the `json` formatter, which is not supported by CircleCI for splitting by timings (and additionally is now deprecated).

To fix the problem:

* switch from using the `json` formatter to using the `junit` formatter, which is required by CircleCI for splitting on timings: https://circleci.com/docs/use-the-circleci-cli-to-split-tests/#junit-xml-reports;
* add the `fileattribute=true` flag to include filenames in the xml output, which is needed for splitting accurately;
* write the xml output to a directory rather than a file (a requirement of the `junit` formatter).

#### Proof

This has the effect of dropping the run time of the feature test suite by approximately two minutes from ~15 minutes to ~13 minutes, which isn't huge but brings it closer to the runtime of the rpsec suite and saves us a few credits.

Before:

image

After:

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -408,8 +408,8 @@ jobs:
 
   cucumber-tests:
     executor: test-executor
-    resource_class: medium
-    parallelism: 4
+    resource_class: small
+    parallelism: 6
     steps:
       - checkout
       - build-base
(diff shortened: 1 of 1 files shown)


Obsiye (approved):

jrmhaig (approved):
🥇

PR #5309, opened 2022-12-21
CTSKF-128 Update ECNP Final claim validation

mpw5:
#### What

Update validation of ECNP Final claims so that CLAIR contingency claims are validated in the same way as CLAIR claims

#### Ticket

[CTSKF-128](https://dsdmoj.atlassian.net/browse/CTSKF-128)

#### Why

Claims that fall under CLAIR contingency rules (where the representation order date is on or after 17 September 2020 with a main hearing date on or after 31 October 2022) need to have the same validation for Elected Cases Not Proceeded (ECNP) as claims that fall under CLAIR rules (where the representation order date is on or after 30 September 2022).

#### How

* amend the logic in `Claim::BaseClaimValidator` to also include the main hearing date in the validation.
* update the wording of the error displayed when the validation fails 
* use feature flags to switch between the different versions.

### DIFF
--- a/app/validators/claim/base_claim_validator.rb
+++ b/app/validators/claim/base_claim_validator.rb
@@ -376,9 +376,26 @@ class Claim::BaseClaimValidator < BaseValidator
   def validate_earliest_representation_order_date
     return unless @record.case_type&.name == 'Elected cases not proceeded'
     return unless @record.earliest_representation_order_date
-    # This applies to both agfs fee scheme 13 and lgfs fee scheme 10 but the dates are the same
-    return if @record.earliest_representation_order_date < Settings.agfs_scheme_13_clair_release_date
+    return if allow_elected_case_not_proceeded?
 
-    @record.errors.add(:earliest_representation_order_date, 'invalid for elected case not proceeded')
+    if main_hearing_date_enabled?
+      @record.errors.add(:earliest_representation_order_date,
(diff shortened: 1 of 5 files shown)


pking-moj (approved):

NenehB (approved):
LGTM

Obsiye (approved):

app/validators/claim/base_claim_validator.rb
```diff
@@ -376,9 +376,26 @@ def increased_travel?
   def validate_earliest_representation_order_date
     return unless @record.case_type&.name == 'Elected cases not proceeded'
     return unless @record.earliest_representation_order_date
+    return if allow_elected_case_not_proceeded?
+
+    if main_hearing_date_enabled?
+      @record.errors.add(:earliest_representation_order_date,
+                         'invalid for elected case not proceeded and main hearing date')
+    else
+      @record.errors.add(:earliest_representation_order_date, 'invalid for elected case not proceeded')
```
jrmhaig, line 385: This is not blocking merging as it pre-dates this PR (and occurs in various other places in the validators) but we could put these messages in translations files. I see that someone had the good intention of having a generic `error_message_for` method in the `BaseValidator` class but it is not actually used anywhere. Maybe something to consider resurrecting?

jrmhaig (approved):
One comment but not a blocker.

PR #5287, opened 2022-12-13
[CTSKF-151] Create factory to fetch Fee Scheme

jrmhaig:
#### What

Create factories for fetching the fee scheme based on rep order and main hearing date.

#### Ticket

[CCCD: Use main hearing date when finding fee scheme for claim](https://dsdmoj.atlassian.net/browse/CTSKF-151)

#### Why

Currently the fee scheme for a claim is fetched based solely on the representation order date with a database query searching between a start and end date. With the addition of the main hearing date as an additional parameter this query becomes more complex and requires additional database fields.

#### How

Encapsulate the logic of determining the fee scheme in `FeeSchemeFactory::AGFS` and `FeeSchemeFactory::LGFS`.

--------

#### TODO (wip)

 - [x] Address the code smells and duplication that sonarcloud is reporting
 - [x] Fix broken tests
 - [x] Tests are failing because the Transfer Brain now depends on the fee scheme and, for most of the tests, there are not fee schemes in the database. This can be solved either by calling `seed_fee_schemes` in the `before` of all these or to seed the fee schemes at the very start of all tests so that they are always there.

[CTSKF-151]: https://dsdmoj.atlassian.net/browse/CTSKF-151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5314, opened 2022-12-28
[CTSKF-151] Seed fee schemes once at the start of all tests

jrmhaig:
#### What

Seed fee schemes at the start of all rspec tests instead of selectively in individual test files.

#### Ticket

[CCCD: Use main hearing date when finding fee scheme for claim](https://dsdmoj.atlassian.net/browse/CTSKF-151)

#### Why

Some tests require the fee schemes in the database and so they are added using the `seed_fee_schemes` when necessary. However, they are always the same and their presence should not affect other tests so they can be added at the start of the suite. This has the following benefits:

* Changes that introduce a dependency on the fee scheme do not automatically cause test failures. (See #5287)
* Start up and tear down times is reduced.

#### How

Add `seed_fee_schemes` to the `config.before(:suite)` block in `rspec_helper.rb` and add `fee_schemes` to the list of excluded tables in the `clean_database` helper.

#### Note

These changes are included in #5287. If that PR is merged then this can be closed. Alternatively, this PR can be merged independently.

### DIFF
--- a/spec/api/v1/dropdown_data_spec.rb
+++ b/spec/api/v1/dropdown_data_spec.rb
@@ -71,7 +71,6 @@ RSpec.describe API::V1::DropdownData do
     end
 
     before do
-      seed_fee_schemes
       seed_case_types
       seed_case_stages
       create_list(:court, 2)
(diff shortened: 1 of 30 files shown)


mpw5:
Closed following release of https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5287

PR #5316, opened 2022-12-29
[CTSKF-182] Remove dead code in the Transfer Brain

jrmhaig:
#### What

Remove some dead code

#### Ticket

[CCCD - Transfer Brain code cleanup](https://dsdmoj.atlassian.net/browse/CTSKF-182)

#### Why

The "Transfer Brain" is already complicated and this is compounded by code that is no longer required.

* `Claim::TransferBrain::DataItem#to_h` is only used in two places;
  * the private method `Claim::TransferBrain::DataItemCollection#collection_hash`
  * a unit test
* `Claim::TransferBrain::DataItemCollection#collection_hash` is only used in the methods `#to_h`, which is never used, and `#to_json`
* `Claim::TransferBrain::DataItemCollection#to_h` is only used in the class `Claim::TransferBrain.data_attributes`
* `Claim::TransferBrain.data_attributes` is only used in a unit test

#### How

Cut it out

### DIFF
--- a/app/models/claim/transfer_brain.rb
+++ b/app/models/claim/transfer_brain.rb
@@ -73,10 +73,6 @@ module Claim
       CASE_CONCLUSIONS[detail.case_conclusion_id]
     end
 
-    def self.data_attributes
-      TransferBrain::DataItemCollection.instance.to_json.chomp
-    end
-
     #
     # only new litigators that transfered onto unelected cases at specific stages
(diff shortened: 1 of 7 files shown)


mpw5 (approved):
LGTM. Providing all the tests still pass and the transfer functionality has been tested on an environment, I think it's good to go.

(Also I love `brain_dead_code` as a branch name 🙂)

PR #5381, opened 2023-02-15
Update Ruby to 3.1.3 (from 3.1.2)

jrmhaig:
#### What

Update to the latest Ruby 3.1 version. Also update the Alpine image to 3.17.

#### Ticket

N/A

#### Why

Keep up to date with the latest security releases. See https://www.ruby-lang.org/en/news/2022/11/24/ruby-3-1-3-released/


mpw5 (commented):
I think the ruby version also needs updating in the Gemfile. 

It might be worth using an approach like this to reduce the need to update in multiple places: https://andycroll.com/ruby/read-ruby-version-in-your-gemfile/.

jrmhaig:
Closing for now as it seems that there are other things that need to be fixed first.

PR #5382, opened 2023-02-15
Remove guard and related gems

jrmhaig:
#### What

Remove the `guard` gem.

#### Ticket

N/A

#### Why

`guard` allows tests to be run automatically when code is changed but this feature is not used by any of the developers. Several extra gems are included as dependencies, some of which do not appear to have been maintained for a while. In particular, `eventmachine` has not had a release in 5 years and the recommended use is to download directly from Github.

#### How

Remove `guard*` gems from `Gemfile` and remove the files `Guardfile` and `bin/guard`.


mpw5 (approved):
LGTM

PR #4722, opened 2022-03-08
[CTSKF-233] Replace webpacker with jsbundling-rails

imtiazAhmd:
#### What
From Rails 7, webpacker will be replaced as default. This gives us the opportunity to explore the newest option available for the asset bundler using jsbundling-rails. This is a lighter integration with ruby/rails compared to webpacker and the freedom of using any cutting-edge asset bundler such as esbuild, snowpack, rollup, parcel. 

#### Ticket

[Upgrade Asset bundler in preparation for  Ruby 3 / Rails 7](https://dsdmoj.atlassian.net/browse/CTSKF-233)

--------

#### TODO (wip)

 - [x] Retire `webpack-concat-plugin`
 - [x] Retire `webpacker` and introduce `jsbundling-rails`
 - [x] Trial use of `webpack 5`
 - [x] Separate debounce JS code block to a new file
 - [x] Update commands to build in CI
 - [x] Update `Jasmine` for debounce changes
 - [x] Solve issues with `Bloodhound`
 - [ ] Alternative ways of bundling to improve performance using `esbuild` / `snowpack`. To be done at a later date.

PR #5384, opened 2023-02-21
[CTSKF-233] Use html <template> in place of JQuery template

jrmhaig:
#### What

A couple of changes to update some old JQuery to more modern javascript:

#### Ticket

[CCCD: Replace Asset bundler (Pre Rails 7)](https://dsdmoj.atlassian.net/browse/CTSKF-233)

#### Why

Keeping the codebase up to latest coding standards and (slowly) moving away from JQuery.

This will help to simplify #4722

#### How

* `$.extend` can be replaced by the new(ish) spread syntax
* `$.template` can be replaced using the `<template>` HTML tag


mpw5 (approved):
👍

PR #1780, opened 2016-09-14
Twitter Type Ahead ( Select Boxes )

cleargif:
https://www.pivotaltracker.com/n/projects/1520095/stories/126680583

https://govuk-accessibility-sandbox.herokuapp.com/patterns/autocomplete/03/

PR #5391, opened 2023-02-27
[CTSKF-233] Remove obsolete Autocomplete javascript module

jrmhaig:
#### What

Remove `app/webpack/javascripts/modules/Modules.Autocomplete.js` together with some other related things.

#### Ticket

[CCCD: Replace Asset bundler (Pre Rails 7)](https://dsdmoj.atlassian.net/browse/CTSKF-233)

#### Why

`Autocomplete` was added in #1780 but it appears now not to be used anymore, replaced by `AutocompleteWrapper`. It used a [fork](https://github.com/UKHomeOffice/typeahead-aria) of a [third-party library](https://github.com/twitter/typeahead.js) neither of which is supported anymore. There is a more recently updated fork [here](https://github.com/corejavascript/typeahead.js) but if we are not using it then it is better just to remove it.

This will help to simplify #4722

#### How

Remove:

* `app/webpack/javascripts/modules/Modules.Autocomplete.js`
* `app/webpack/javascripts/vendor/typeahead-aria.js`
* `spec/javascripts/Modules.Autocomplete_spec.js`

and remove references elsewhere.


mpw5 (approved):
👍

PR #5400, opened 2023-03-01
Check updates daily

jrmhaig:
#### What

Set dependabot to check daily instead of weekly.

#### Ticket

N/A

#### Why

We seem only to get a limited number of dependabot PRs each Wednesday and some libraries are not getting updated. By changing the schedule to daily should work though more of them.

#### How

Change the `interval` setting for each package ecosystem.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: weekly
+    interval: daily
     day: wednesday
     time: "03:00"
     timezone: Europe/London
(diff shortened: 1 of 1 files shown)


mpw5 (approved):

.github/dependabot.yml
```diff
@@ -7,7 +7,7 @@ updates:
     day: wednesday
     time: "03:00"
     timezone: Europe/London
-  open-pull-requests-limit: 5
+  # open-pull-requests-limit: 5
```
pking-moj: [open-pull-requests-limit](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit) - The default is 5 so your change wouldnt do anything

jrmhaig: Hmmm, I didn't know that. I'm fairly sure we used to get spammed with an (apparently) unlimited number.

PR #5480, opened 2023-03-14
Small improvements to API claim tests

mpw5:
#### What

Makes some small improvements to improve consistency between the tests for AGFS and LGFS claims in `spec/api/v1/external_users/claims/`, improve code quality, and fix some potential issues.

#### Why

To improve test coverage, consistency and adherence to code quality standards.

#### How

* Align the tests for LGFS claims with the AGFS tests by including the `malformed or not iso8601 compliant dates` and `optional parameter validation` shared examples.
* Remove duplication of the `post_to_validate_endpoint` definition in each claim spec by moving it to the relevant shared examples.
* Fix hard-coded dates in tests which would have eventually led to failures.
* Fix various rubocop offences


spec/support/shared_examples_for_api.rb
```diff
@@ -47,6 +47,10 @@
 end
 
 RSpec.shared_examples 'malformed or not iso8601 compliant dates' do |options|
+  subject(:post_to_validate_endpoint) do
+    post options.fetch(:relative_endpoint), valid_params, format: :json
```
jrmhaig: Is there a reason why `options[:relative_endpoint]` cannot be used here

spec/api/v1/external_users/claims/final_claim_spec.rb
```diff
@@ -30,7 +28,7 @@
     }
   end
 
-  after(:all) { clean_database }
+  after { clean_database }
```
jrmhaig: I believe that `after(:all)` was done in places like this for efficiency. I would be in favour of avoiding `after(:all)` where possible but it is worth checking how long the tests are now taking to run in CircleCI.

spec/support/shared_examples_for_api.rb
mpw5: `options.fetch` was being used in other places in this file already. I've changed them all to `options[]` now 🙂

spec/api/v1/external_users/claims/final_claim_spec.rb
mpw5: Fair point. Leaving it as `after(:all)` shaves about a minute off the tests which I think is enough to make it worthwhile. I've also changed it to `after(:all)` AGFS tests so that they're quicker too (and consistent).

jrmhaig (approved):

PR #5491, opened 2023-03-17
Fix open redirect vulnerability in ManagementInformationController

mpw5:
#### What

Snyk has a [flagged](https://app.snyk.io/org/legal-aid-agency/project/3c84d1a7-ff1f-4508-87d5-c74b0b4f23df#issue-da65fc4d-0e11-4f22-9e92-e5a13c15b921) an open redirect vulnerability in `CaseWorkers::Admin::ManagementInformationController` due to the use of unsanitized request parameters in a `redirect_to` method call.

image

That controller already implements strong parameters but they are not always used in the code. This commit replaces direct access to the parameters (`params[:report_type]`) with use of the strong parameters (`report_params[:report_type]`).

(This isn't strictly necessary as that class uses a `before` hook which validates the `:report_type` parameter before it is used, which mitigates the vulnerability, but this change is more idiomatic and keeps Snyk happy).

### DIFF
--- a/app/controllers/case_workers/admin/management_information_controller.rb
+++ b/app/controllers/case_workers/admin/management_information_controller.rb
@@ -17,7 +17,7 @@ class CaseWorkers::Admin::ManagementInformationController < CaseWorkers::Admin::
 
   def download
     log_download_start
-    record = Stats::StatsReport.most_recent_by_type(params[:report_type])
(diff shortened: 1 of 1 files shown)


mpw5:
Hmm. Brakeman is reporting this as a vulnerability, even with the change in place. We already have this as ignored in `brakeman.ignore` so I'm going to close this and ignore it in Snyk too.

PR #5523, opened 2023-03-30
CTSKF-299 Send feedback to Zendesk temporarily

Obsiye:
#### What

Send feedback to Zendesk temporarily

#### Ticket

[CTSKF-299](https://dsdmoj.atlassian.net/browse/CTSKF-299)

#### Why

We no longer have access to SurveyMonkey, but still use it as the link for feedback on CCCD. This means users are getting errors when trying to provide us with valuable feedback! 

The solution was to temporarily use Zendesk until we have regained SurveyMonkey access.

#### How

- Send feedback to Zendesk similar to how the bugreport was doing it
- Update Production and Staging with ZENDESK_FEEDBACK env
- Add newline to feedback and bugreports descriptions after each question and answer
- Update tests (unit and feature tests)

[CTSKF-299]: https://dsdmoj.atlassian.net/browse/CTSKF-299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

#### Attachments

Zendesk feedback example


app/models/feedback.rb
```diff
@@ -60,13 +72,13 @@ def save_feedback
...
+    @response_message = success
   rescue ZendeskAPI::Error::ClientError => e
-    @response_message = 'Unable to submit fault report'
+    @response_message = failure
     LogStuff.error(class: self.class.name, action: 'save', error_class: e.class.name, error: e.to_s) do
-      'Bug report submisson failed!'
+      log_message
     end
     false
   end
```
mpw5: I wonder if it would be cleaner to do something like
```
  def send_to_zendesk(type)
    ZendeskSender.send!(self)
    @response_message = "#{type} submitted"
  rescue ZendeskAPI::Error::ClientError => e
    @response_message = "Unable to submit #{type}"
    LogStuff.error(class: self.class.name, action: 'save', error_class: e.class.name, error: e.to_s) do
      "#{type} submission failed!"
    end
    false
  end
```

And then we'd just need to call `send_to_zendesk('Feedback')` and `send_to_zendesk('Fault')` on lines 55 and 62, rather than have to include the full message wordings. (Tests would probably need to be updated as a result, as the wordings of the confirmation/failure messages would be slightly different to what they are now)

Obsiye: that's clever. yep that works thank you

Obsiye: updated it :) . I used `feedback_type` as the parameter name only because `type` is an existing attr_accessor / variable name

mpw5: Ah, I wonder if we can use that `type` variable to simplify things more? So instead of passing in the type, we can just do 

```
  def send_to_zendesk
    feedback_type = type.humanize
    ZendeskSender.send!(self)
    @response_message = success
    @response_message = "#{feedback_type} submitted"
  rescue ZendeskAPI::Error::ClientError => e
    @response_message = failure
    @response_message = "Unable to submit #{feedback_type.downcase}"
    LogStuff.error(class: self.class.name, action: 'save', error_class: e.class.name, error: e.to_s) do
      log_message
      "#{feedback_type} submisson failed!"
    end
    false
  end
```

and then we don't need to pass in any arguments at all when we call the method

app/models/feedback.rb
```diff
@@ -45,12 +45,24 @@ def subject
...
 
   def description
-    feedback_type_attributes.map { |t| "#{t}: #{send(t)}" }.join(' - ')
+    feedback_type_attributes.map { |t| "#{t}: #{send(t)}" }.join("\n")
   end
 
   private
 
   def save_feedback
+    if Settings.zendesk_feedback_enabled?
```
jrmhaig: You could potentially clean this up a bit by having for the `save` method, above:

```
def save
  return unless valid?
  return send_to_survey_monkey if feedback? && !Settings.zendesk_feedback_enabled?
  send_to_zendesk
end
```

or even, replace the `feedback?` method with

```
def feedback?
  is?(:feedback) && !Settings.zendesk_feedback_enabled?
end
```

Additionally, if you do this then the `zendesk_feedback_enabled?` setting could be reversed.

Obsiye: great idea thank you

jrmhaig (approved):

ivanELEC (approved):

PR #5551, opened 2023-04-06
Amend logging of API requests

jrmhaig:
#### What

Add the inputs to the response log for the API.

#### Ticket

[CCCD - Zendesk #311681 - unable to submit claims via Leap](https://dsdmoj.atlassian.net/browse/CTSKF-296)

#### Why

Formerly, the parameters for a request were only being logged as `env['rack.request.form_hash']`. This only shows fields in a form submission but requests from vendor software has the data as JSON in the body, so the logs do not show anything in the production environment.

#### How

During the processing of a request `env['api.request.input']` gets set to the input values. As this is blank at the start of the request processing it needs to be logged together with details of the response at the end.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -2173,7 +2173,6 @@ Rails/SkipsModelValidations:
 # SupportedStyles: strict, flexible
 Rails/TimeZone:
   Exclude:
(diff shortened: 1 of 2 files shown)


Obsiye (approved):
LGTM. Would we soon after be looking to scrub the logs of personal data?

PR #5599, opened 2023-04-21
Docker improvements

mpw5:
#### What

Based on some recent L&D on Docker best practices, this applies some improvements to the CCCD Dockerfile.

#### Why

To improve the quality of the Dockerfile and reduce the size of the resulting image. A smaller image is good as it reduces the attack surface of the application and should lead to quicker builds and pod start-up.

Building the image locally, it appears that these changes have reduced the size of the final image from **~1.7gb** to **~800mb**.

#### How

* replace deprecated `MAINTAINER` field with individual `LABEL` fields for Owner, Service name and Contact;
* move to a multi-stage build process so that build packages are excluded from the final image, resulting in a smaller image size;
* fix (small) [vulnerability](https://sonarcloud.io/organizations/ministryofjustice/rules?open=docker%3AS6470&rule_key=docker%3AS6470) introduced through the use of a wildcard `COPY Gemfile* ./` by explicitly copying `Gemfile` and `Gemfile.lock`;
* using named directories for copying files to (rather than relying on `.`) for better control over what is included in the build;
* delete unnecessary files (ie `node_modules`, and residue from `bundler`), resulting in a smaller image size;
* improve the `bundle` command by adding parallelisation and retries;
* run `apk --update-cache upgrade` before `apk add` to ensure we're installing the latest versions of libraries;
* add more files and directories that aren't needed for the build to `.dockerignore.yaml` to keep unnecessary files out of the build.

I've also added a commit which greatly reduces the amount of comments in the file, as I found they were overly verbose and made the content of the file harder to parse when I was working on it. I'm happy for pushback though as other peoples mileage may vary - I can easily remove that commit.

#### Resources

https://docs.docker.com/build/building/multi-stage/
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
https://ledermann.dev/blog/2018/04/19/dockerize-rails-the-lean-way/
https://lipanski.com/posts/dockerfile-ruby-best-practices

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -41,7 +41,7 @@ COPY Gemfile Gemfile.lock $APP_HOME/
 RUN gem install bundler -v $(cat Gemfile.lock | tail -1 | tr -d " ") \
 && bundle config without test development devunicorn \
 && bundle config build.nokogiri --use-system-libraries \
-&& bundle install \
+&& bundle install --jobs 4 --retry 3 \
 && rm -rf /usr/local/bundle/cache/*.gem \
 && find /usr/local/bundle/gems/ -name "*.c" -delete \
 && find /usr/local/bundle/gems/ -name "*.o" -delete
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,20 +11,21 @@ WORKDIR $APP_HOME
 # build dependencies:
 # -virtual: create virtual package for later deletion
 # - build-base for alpine fundamentals
+# - git for installing gems referred to use as git:// uri
 # - ruby-dev/libc-dev for compiling raindrops, at least
 # - libxml2-dev/libxslt-dev for nokogiri, at least
 # - postgres-dev for pg/activerecord gems
-# - git for installing gems referred to use as git:// uri
 # - yarn for js dependency management
 #
-RUN apk --no-cache add \
+RUN apk --update-cache upgrade \
+ && apk --no-cache add \
                    build-base \
+                   git \
+                   gmp=6.2.1-r2 \
                    libxml2-dev \
                    libxslt-dev \
                    postgresql-dev \
-                   git \
-                   yarn \
-                   gmp=6.2.1-r2
+                   yarn
 
 ######################
 # DEPENDENCIES START #
@@ -56,8 +57,6 @@ RUN SECRET_KEY_BASE=a-real-secret-key-is-not-needed-here \
 RAILS_ENV=production \
 bundle exec rails assets:precompile 2> /dev/null
 
-# tidy up installation
-RUN apk --update-cache upgrade
 RUN rm -rf node_modules
 
 ####################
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,88 +1,71 @@
...
-                  runit \
-                  ttf-freefont \
-                  libreoffice \
-                  redis \
-                  postgresql-client
-
-# add non-root user and group with alpine first available uid, 1000
-RUN addgroup -g 1000 -S appgroup \
-&& adduser -u 1000 -S appuser -G appgroup
+ENV APP_HOME /app
```
jrmhaig: Is there any particular reason why this `APP_HOME` is different from the setting in stage 2, `/usr/src/app`?

docker/Dockerfile
```diff
@@ -1,88 +1,71 @@
...
+COPY . $APP_HOME/
 
-# precompile assets, silently (as verbose)
 RUN SECRET_KEY_BASE=a-real-secret-key-is-not-needed-here \
 RAILS_ENV=production \
 bundle exec rails assets:precompile 2> /dev/null
 
-# tidy up installation
-RUN apk --update-cache upgrade && apk del build-dependencies
+RUN rm -rf node_modules
```
jrmhaig, line 39: Is it possible to use a `.dockerignore` file to avoid these files, as well as the files in `/usr/local/bundle/gems/` above, being copied rather than deleting them?

I see that `.dockerignore` is itself listed in `.dockerignore` and this makes sense for a one step process. Perhaps it can be removed so that it is copied into the stage 1 image and then added with:

```
RUN echo .dockerimage >> .dockerimage
```

so that it is not copied into the stage 2 image.

jrmhaig (approved):
Looks like a great improvement. I have added a couple of comment but neither are blockers so I'm happy to approve.

docker/Dockerfile
```diff
@@ -1,88 +1,71 @@
...
+                   libreoffice \
+                   linux-headers \
+                   nodejs \
+                   postgresql-client \
+                   redis \
+                   runit \
+                   ttf-freefont
+
+RUN addgroup -g 1000 -S appgroup \
+&& adduser -u 1000 -S appuser -G appgroup
```
Obsiye, line 65: Maybe also reduce this users permissions to the files it needs to write to like log, tmp, db etc:
```
# non-root/appuser should own only what they need to
RUN chown -R appuser:appgroup log tmp db
```

This is a line we have in CDUI's dockerfile at the moment.

i'm guessing the line `--chown=appuser:appgroup ./app` means that the appuser has ownership of all the files under app

docker/Dockerfile
mpw5: Not especially. Stage 2 has to use `/usr/src/app` as the working directory, otherwise subsequent deploy steps fail. I think I was just trying to simplify the first stage, but it probably makes sense for them to be consistent. I'll change this one.

docker/Dockerfile
mpw5: Sadly the `.dockerignore` file only works on the first step of a build, copying files from the filesystem into the image. The`.dockerignore` has no effect when files are copied between images on subsequent steps 😢. Some discussion of it here: https://github.com/moby/moby/issues/33923.

docker/Dockerfile
mpw5: We originally had the same line `RUN chown -R appuser:appgroup log tmp db` in CCCD. I removed it after reading that explicitly running `chown` has a big impact on resulting image size and replaced it with the command you highlighted, which sets permissions when copying the files across. You're right that the drawback of this approach is that it's more permissive but seems to be the recommended approach (eg https://blog.mornati.net/docker-images-and-files-chown).

Obsiye: ah ok makes sense, i'm happy with this. thanks for the article link too

PR #5605, opened 2023-04-24
CTSFK-278: Display multi-line text for additional claim information

ivanELEC:
#### What
[Display multi-line text for additional claim information
](https://dsdmoj.atlassian.net/browse/CTSKF-278)

#### Why

When adding additional claim information onto a claim in CCCD, a user can utilise newlines in the text box to make the information more legible, however when anyone goes to the claim’s summary view to review the information, all newlines are removed, making the information hard to read (particularly as they’re not replaced with anything)

#### How

Use [simple_format](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format) function on text where additional information is being displayed in view mode

--------

Note: I did not add this in as a helper to apply to all text as simple_format wraps text in a <p> tag. In some cases e.g. when rendering lists, this can lead to unnecessary whitespace.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -144,7 +144,7 @@ module ApplicationHelper
   end
 
   def format_multiline(html_string)
-    simple_format(html_string, {}, sanitize: false )
+    simple_format(html_string, {}, sanitize: false)
   end
 
   private
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -143,6 +143,10 @@ module ApplicationHelper
      on_home_page?].all?
   end
 
+  def format_multiline(html_string)
+    simple_format(html_string, {}, sanitize: false )
+  end
+
   private
 
   def on_home_page?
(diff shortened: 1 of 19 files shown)


pking-moj (approved):

jrmhaig:
Be aware that `simple_format` allows for some html tags, although it appears to do this by permitting certain tags and stripping out anything else so it is relatively safe. This is noted in the 'sanitization' section at the top of the documentation link in the description.

Also, the WAF is blocking any text with `<script>` in it.

jrmhaig:
The additional information for travel expenses is in this file:

* `app/views/external_users/claims/expenses/_summary.html.haml`

Should this need to be fixed too as part of this ticket?

ivanELEC:
> Be aware that `simple_format` allows for some html tags, although it appears to do this by permitting certain tags and stripping out anything else so it is relatively safe. This is noted in the 'sanitization' section at the top of the documentation link in the description.
> 
> Also, the WAF is blocking any text with `<script>` in it.

Looking at the source for this function, I should be able to set sanitize: false in the options parameter and still get the paragraph splits

PR #5618, opened 2023-04-27
[CTSKF-359] AGFS Fee Scheme 15

mpw5:
#### What

Adds AGFS fee scheme 15

#### Ticket

#### Why

From the 17 April 2023, claims with a representation order on or after 17 April 2023 are eligible for an additional £62 fixed fee. From this date, the advocate role of QC will be replaced with KC. Creating this fee scheme will allow us to implement those changes. This will be done in separate PRs.

#### How

* Creates a new AGFS fee scheme 15 which applies to claims with a representation orders dated on or after 17/04/2023 and is identical to AGFS fee scheme 14 in all other respects.
* Sets the end date of AGFS fee scheme 14 to 16/04/2023.


spec/factories/fee_schemes.rb
```diff
@@ -41,6 +41,12 @@
       version { 14 }
```
Obsiye, line 41: nitpicky, should the end date for agfs_fourteen be the beginning of `fee scheme 15 - 1.day `

Obsiye: maybe that doesn't matter

Obsiye (commented):
awesome , code lgtm 💯 . i just need to test this. can i test it without the fee calculator fee scheme 15 in?

spec/factories/fee_schemes.rb
mpw5: Nope, you're right, I missed that. Looks like scheme 13 was missed too. I'll fix it.

ivanELEC (approved):

mpw5:
> awesome , code lgtm 💯 . i just need to test this. can i test it without the fee calculator fee scheme 15 in?

Yes, you can test locally without updating fee calculator, the two will work fine separately. Or, this is currently deployed to https://dev-lgfs.claim-crown-court-defence.service.justice.gov.uk/ if you'd rather test there (that talks to the dev version of fee calculator, which does currently have fee scheme 15 deployed, so will be a more realistic test).

Obsiye (approved):

PR #5670, opened 2023-05-15
[CTSKF-394] Support chronological assessment of expenses by Date

SamKyprianou:
#### What
Provides a solution to the sorting issue with the expenses table in CCCD’s summary page that does not allow caseworkers to organise the list  - the date sort function did sort on dates (it treated them as numbers).

#### Ticket

[CTSKF-394](https://dsdmoj.atlassian.net/browse/CTSKF-394)

#### Why
It will allow the Billing Caseworker to be able to sort the expense tables by date so that they can more easily match the claim to the evidence they are assessing.

#### How
I refactored and added code to the Module.ExpenseDataTable.js file to convert string dates into a sortable JS date format.

```javascript
function parseDate(dateString, _format) {
  // Split the date string into day, month, and year parts
  var parts = dateString.split('/');

  // Extract the numeric values for day, month, and year
  var day = parseInt(parts[0], 10);
  var month = parseInt(parts[1], 10);
  var year = parseInt(parts[2], 10);

  // Create a new JavaScript Date object using the extracted values
  // Note: Months in JavaScript Date objects are zero-based, so we subtract 1 from the month value
  return new Date(year, month - 1, day);
}
```

this would allow the date column to be sorted correctly as JS breaks the date obj down to millisecond when sorting.

[CTSKF-394]: https://dsdmoj.atlassian.net/browse/CTSKF-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/webpack/javascripts/modules/Modules.ExpensesDataTable.js
+++ b/app/webpack/javascripts/modules/Modules.ExpensesDataTable.js
@@ -71,20 +71,21 @@ moj.Modules.ExpensesDataTable = {
   },
 
   setOrder: function () {
-    const order = this.dataTable.order()
-    const columnIndex = order[0][0]
-    const direction = order[0][1]
+    const order = this.dataTable.order();
+    const columnIndex = order[0][0];
+    const direction = order[0][1];
+
     // this check is to ensure only type of expense (index 0)
     // and reason for travel (index 1) have secondary order column
     // set to date of expense (index 3) as the current configuration
     // for columnDefs does not support setting sorting direction
     if (columnIndex === 0 || columnIndex === 1) {
-      this.dataTable.order([columnIndex, direction], [3, 'asc'])
+      this.dataTable.order([columnIndex, direction], [3, 'desc']);
+    } else if (columnIndex === 3) {
+      // Sort the date column as a date object
+      this.dataTable.order([columnIndex, direction]);
     }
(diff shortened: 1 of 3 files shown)


app/webpack/javascripts/modules/Modules.ExpensesDataTable.js
```diff
@@ -33,7 +31,21 @@ moj.Modules.ExpensesDataTable = {
...
+          if (!data || isNaN(parseDate(data, 'dd/mm/yyyy'))) {
+            return ''
+          }
+
+          // Convert the valid date string to a sortable format (e.g., '2023-04-29')
+          return parseDate(data, 'dd/mm/yyyy')
+        }
+
+        return data
+      }
```
jrmhaig: I don't know what the best practice is in Javascript, but can this be made into a named function?

SamKyprianou: I've extracted out the logic under the render and created a named function `renderDate` - looks a lot more readable @jrmhaig

jrmhaig (approved):
:+1:

ivanELEC (approved):

PR #5674, opened 2023-05-16
[CTSKF-318] Additional Preparation Fee and enable KC injection to CCR

mpw5:
#### What

Adds a new miscellaneous fee for additional preparation

As this will be released at the same time as Fee Scheme 15 is deployed to CCR, this PR also enables the `CAN_INJECT_KC` feature flag, so that the use of advocate categories is consistent across the two services.

#### Ticket

[CTSKF-318](https://dsdmoj.atlassian.net/browse/CTSKF-318)

#### Why 

From the 17 April 2023, AGFS final, hardship and supplementary claims with a representation order on or after 17 April 2023 AND where the case is a cracked trial, trial, cracked retrial or retrial, are eligible for an additional £62 fixed fee.

#### How

Adds the new miscellaneous fee to AGFS Fee Scheme 15, restricted to the claim types and case types listed above, and restricted to one fee per claim.

[CTSKF-318]: https://dsdmoj.atlassian.net/browse/CTSKF-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.k8s/live/production/app-config.yaml
+++ b/.k8s/live/production/app-config.yaml
@@ -22,4 +22,4 @@ data:
   LAA_FEE_CALCULATOR_HOST: https://laa-fee-calculator.service.justice.gov.uk/api/v1
   ALLOW_FUTURE_DATES: 'false'
   ZENDESK_FEEDBACK: 'true'
-  CAN_INJECT_KC: 'false'
+  CAN_INJECT_KC: 'true'
--- a/.k8s/live/api-sandbox/app-config.yaml
+++ b/.k8s/live/api-sandbox/app-config.yaml
@@ -19,4 +19,4 @@ data:
   SETTINGS__AWS__S3__REGION: 'eu-west-2'
   LAA_FEE_CALCULATOR_HOST: https://laa-fee-calculator.service.justice.gov.uk/api/v1
   ALLOW_FUTURE_DATES: 'true'
-  CAN_INJECT_KC: 'false'
+  CAN_INJECT_KC: 'true'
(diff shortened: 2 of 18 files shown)


jrmhaig (approved):
👍

Obsiye (approved):

ivanELEC (approved):

.k8s/live/production/app-config.yaml
```diff
@@ -22,4 +22,4 @@ data:
   LAA_FEE_CALCULATOR_HOST: https://laa-fee-calculator.service.justice.gov.uk/api/v1
   ALLOW_FUTURE_DATES: 'false'
   ZENDESK_FEEDBACK: 'true'
-  CAN_INJECT_KC: 'false'
+  CAN_INJECT_KC: 'true'
```
jrmhaig, line 25: This will start sending claims with 'KC' as the advocate category to CCR. As long as CCR is ready for this, this is OK.

FWIW, we have not yet had any claims with a 'KC' advocate category so there are not likely to be problems if the changes do not happen with CCCD and CCR at exactly the same time.

jrmhaig (approved):

.k8s/live/production/app-config.yaml
ivanELEC: CCR is not ready for this in production yet - therefore, this would need to happen after our release

ivanELEC (approved):

.k8s/live/production/app-config.yaml
mpw5: Thanks both. This will only be deployed to production after the CCR release tonight, so there should be no issues there 🤞

PR #5680, opened 2023-05-18
Fix QC/KC as appropriate for fee scheme

jrmhaig:
#### What

Update the claim cleaner to change;

* QC advocate category to KC for fee scheme 15 and later
* KC advocate category to QC for fee scheme 14 and earlier

#### Ticket

[CCCD QC to KC](https://dsdmoj.atlassian.net/browse/CTSKF-337)

#### Why

If a claim is created in one fee scheme, a QC or KC advocate category selected, and then the details are changed so that the fee scheme is different then it is possible that the advocate category is invalid.

#### How

Create a new method `fix_advocate_categories` to be called by all the advocate claim cleaners.

### DIFF
--- a/app/models/claim/advocate_interim_claim.rb
+++ b/app/models/claim/advocate_interim_claim.rb
@@ -85,5 +85,11 @@ module Claim
     def fee_scheme_factory
       FeeSchemeFactory::AGFS
     end
+
+    private
+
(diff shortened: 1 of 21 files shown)


mpw5 (approved):
Nice work 👏

PR #5683, opened 2023-05-19
Revert "Upgrade Ruby to 3.2.2"

mpw5:
Reverts ministryofjustice/Claim-for-Crown-Court-Defence#5315

This is causing failures with the `scheduler_daemon` gem, which does not appear to be able to support ruby 3.2


jrmhaig (approved):
(Not a blocker) Is it possible to keep the `exists?` -> `exist?` change in so that we don't need to remember it next time (or, more likely, forget about it and then spend half an hour rediscovering it)? I cannot remember whether it was already deprecated in 3.1.

PR #5690, opened 2023-05-23
Stop rewriting KC to QC

jrmhaig:
The advocate category is now fixed in the cleaner so it does not need to be fixed here.

#### What

Revert the changes that allowed API users to submit claims with a 'KC' advocate category, which would be silently changed to 'QC'

#### Ticket

[CCCD QC to KC](https://dsdmoj.atlassian.net/browse/CTSKF-337)

#### Why

Before the 'KC' advocate category was introduced a change was made so that third party vendor software could be modified to allow this advocate category to be used without causing problems for CCCD (#5184). This was done by accepting both 'QC' and 'KC' (as an exceptional case) and then silently modify the value before creating the claim. Now 'KC' is permitted (not as an exceptional case) and is the correct value for AGFS fee scheme 15 and later so the modification of the value is no longer required. Additionally, the cleaners, which are executed before the validation step on the creation or updating of a claim, now fixes this value based on the fee scheme.

#### How

Removal of the code to modify the value in two places in `app/interfaces/api/helpers/resource_helper.rb`. Tests have been updated as appropriate.


mpw5 (approved):
Nice 👍

PR #5699, opened 2023-05-25
CTSKF-417 Internalise calls to V2 Caseworker API and Reduce 5xx errors

Obsiye:
#### What

Fix the intermittent 5xx errors on the CaseWorkers API calls.

#### Ticket

[CCCD: Reduce Time Taken to Run Web Calls Within Its Container](https://dsdmoj.atlassian.net/browse/CTSKF-417)

#### Why

This is so that Caseworker's are be able to access CCCD consistently so that they can process claims and complete their work within time standards

#### How

- Update the selector label for the Sidekiq worker kubernetes deployments
- Replace the CaseWorker API call URL with the Kubernetes Service URL `cccd-app-service`
- Add extra headers to the Caseworker RestClient call to bypass HTTPS requirements as it is now an internal k8s cluster call. Cloud Platforms K8s cluster does not provide TLS/HTTPS internally so a https internal call will fail.
- Replace `GRAPE_SWAGGER_ROOT_URL` with `CASEWORKER_API_URL` where the call is not related to the swagger docs but instead to the caseworker api calls. This is because the swagger root fails to call the k8s service due to CORS and thus the documentation page breaks.

#### Additional Information

[Confluence documentation on the journey to this solution](https://dsdmoj.atlassian.net/wiki/spaces/CT/pages/4415979656/DRAFT+Kubernetes+5xx+errors+SOLVED)

Co-Author: @mpw5

### DIFF
--- a/.k8s/live/api-sandbox/app-config.yaml
+++ b/.k8s/live/api-sandbox/app-config.yaml
@@ -9,6 +9,7 @@ data:
   RAILS_ENV: 'production'
   RAILS_SERVE_STATIC_FILES: only-presence-required-to-enable
   GRAPE_SWAGGER_ROOT_URL: 'https://api-sandbox.claim-crown-court-defence.service.justice.gov.uk'
+  CASEWORKER_API_URL: 'cccd-app-service'
   GA_TRACKER_ID: 'UA-37377084-48'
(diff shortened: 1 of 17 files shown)


jrmhaig (approved):
:+1:

PR #5714, opened 2023-05-30
CTSKF-415 Amend Instructed Advocate to Trial Advocate

AlexRyanShep:
#### What 

Updated the wording in en.yml, claim_state_transition_reason.en.yml to use the new approved wording for Instructed Advocate (changed to Trial Advocate)

Updated state_transition_message_builder_spec.rb and and refuse_claim.feature to test new functionality

Updated advocate_admin_trial_claim_edit_submit.feature; test not actually changed, but wording updated from advocate to trial to make clear for future.

#### Ticket

[CTSKF-415 CCCD: Amend Instructed Advocate to Trial Advocate ](https://dsdmoj.atlassian.net/browse/CTSKF-415)

#### Why

Current wording has caused confusion between caseworkers and advocates. 

#### How

Updated yml files, unit and feature tests as above

### DIFF
--- a/config/locales/claim_state_transition_reason.en.yml
+++ b/config/locales/claim_state_transition_reason.en.yml
@@ -51,8 +51,8 @@ en:
         long: We rejected your claim because you didn’t attach the final invoice. Please submit the claim again with the final invoice clearly stating the breakdown of costs and defendant’s name. We cannot accept estimated costs.
     refused_advocate_claims:
       wrong_ia:
-        short: &wrong_ia_short Not Trial Advocate
-        long: &wrong_ia_long We refused your claim because court records show a different advocate was the Trial Advocate for the case. Please submit the claim again with the correct trial advocate details or include evidence to support why you think the court records are wrong.
+        short: &wrong_ia_short Not trial advocate
+        long: &wrong_ia_long We refused your claim because court records show a different advocate was the trial advocate for the case. Please submit the claim again with the correct trial advocate details or include evidence to support why you think the court records are wrong.
       duplicate_claim:
         short: *duplicate_claim_short
         long: *duplicate_claim_long
@@ -117,8 +117,8 @@ en:
         long: *blank
     refused_advocate_hardship_claims:
       wrong_ia:
-        short: &wrong_ia_short Not Trial Advocate
-        long: &wrong_ia_long We refused your claim because court records show a different advocate was the Trial Advocate for the case. Please submit the claim again with the correct trial advocate details or include evidence to support why you think the court records are wrong.
+        short: &wrong_ia_short Not trial advocate
+        long: &wrong_ia_long We refused your claim because court records show a different advocate was the trial advocate for the case. Please submit the claim again with the correct trial advocate details or include evidence to support why you think the court records are wrong.
     refused_litigator_hardship_claims:
       duplicate_claim:
         short: *duplicate_claim_short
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -949,10 +949,10 @@ en:
           section_help_text: 'You must enter the following before adding %{section}:'
       case_details:
         advocate_fields:
-          advocate: Instructed advocate
-          advocate_hint: The instructed advocate for the case
+          advocate: Trial Advocate
+          advocate_hint: The trial advocate for the case
           help_missing_advocates_summary_body_html: You can add more advocates in settings under the <a class="govuk-link" href="%{link}">%{link_text}</a> section
-          help_missing_advocates_summary_heading: Help with missing instructed advocates
+          help_missing_advocates_summary_heading: Help with missing trial advocates
```
pking-moj, line 955: We have used Trial upto now - any reason we have gone for a lower case t?

config/locales/claim_state_transition_reason.en.yml
```diff
@@ -51,8 +51,8 @@ en:
         long: We rejected your claim because you didn’t attach the final invoice. Please submit the claim again with the final invoice clearly stating the breakdown of costs and defendant’s name. We cannot accept estimated costs.
     refused_advocate_claims:
       wrong_ia:
-        short: &wrong_ia_short Wrong Instructed Advocate
-        long: &wrong_ia_long We refused your claim because court records show a different advocate was instructed for the case. Please submit the claim again with the correct advocate details or include evidence to support why you think the court records are wrong.
+        short: &wrong_ia_short Not Trial Advocate
+        long: &wrong_ia_long We refused your claim because court records show a different advocate was the Trial Advocate for the case. Please submit the claim again with the correct trial advocate details or include evidence to support why you think the court records are wrong.
```
pking-moj: Do we need to be consistent with how we are showing `Trial` in these messages?

config/locales/en.yml
AlexRyanShep: I was following the capitalisation as it existed before when the wording was Instructed, but will double-check this with Ewan

AlexRyanShep: Confirmed with Ewan that for consistency, trial advocate should not be capitalised here.

config/locales/claim_state_transition_reason.en.yml
AlexRyanShep: Confirmed with Ewan that for consistency, trial advocate should not be capitalised here. New commit removes this instance of capitalisation and some others.

Obsiye (approved):

jrmhaig (approved):
:+1:

jrmhaig (approved):
Approved again.

FYI, when making changes you might consider fixing some of the simpler stylistic issues. There are several in `spec/services/claims/state_transition_message_builder_spec.rb` most of which would be a bit involved to fix but an easy one is that `context '#call' do` on line 6 should be `describe '#call' do`.

The behaviour of `describe` and `context` is identical but they are intended to be used for different purposes;

* `describe` is used to group tests of the same functionality. In this case, the tests are for the `call` instance method.
* `context` allows for common state to be set up for a group of tests. In this file there are two contexts; when a claim is refused and when it is rejected.

I am approving this PR as this suggested change is not a blocker and it is up to you whether you want to add it. If you want we could pair to look at some of the other issues in this file.

Obsiye (approved):

ivanELEC (approved):

PR #5742, opened 2023-06-06
CTSKF-279 Update Misc Fee Unit

AlexRyanShep:
#### What

Changes made to feeQuantityChange function:

- $els set to refer to the Misc Fees form as a whole (#misc-fees), rather than just the fee quantity field (.js-fee-quantity) 
- changed triggers for els.on from 'change keyup' to 'click keyup'

#### Ticket

[CTSKF 279 CCCD: Show correct misc fee units as soon as misc fee type is selected](https://dsdmoj.atlassian.net/browse/CTSKF-279)

#### Why

When users selected an option in the Misc Fees form the label was not updating until they entered something into the Quantity field

#### How

The label update happens as part of the feeQuantityChange function, which was only triggered by a change keyup listener on the Quantity field. By changing this to a click/keyup listener for the entire form (leaving change in created an infinite loop) this check will happen more often. It is creating some unnecessary requests but these seem to be low impact.

### DIFF
--- a/app/webpack/javascripts/application.js
+++ b/app/webpack/javascripts/application.js
@@ -105,7 +105,7 @@ if (!String.prototype.supplant) {
   $('#misc-fees').on('cocoon:after-insert', function (e, insertedItem) {
     const $insertedItem = $(insertedItem)
     moj.Modules.FeeCalculator.UnitPrice.miscFeeTypesSelectChange($insertedItem.find('.fx-misc-fee-calculation'))
-    moj.Modules.FeeCalculator.UnitPrice.feeQuantityChange($insertedItem.find('.js-fee-calculator-quantity'))
+    moj.Modules.FeeCalculator.UnitPrice.miscFeeFormRefresh($insertedItem.find('.js-fee-calculator-quantity'))
     moj.Modules.FeeCalculator.UnitPrice.feeRateChange($insertedItem.find('.js-fee-calculator-rate'))
     moj.Modules.FeeTypeCtrl.miscFeeTypesSelectChange($insertedItem)
     moj.Modules.FeeTypeCtrl.miscFeeTypesRadioChange($insertedItem)
--- a/app/webpack/javascripts/modules/external_users/claims/fee_calculator/FeeCalculator.UnitPrice.js
+++ b/app/webpack/javascripts/modules/external_users/claims/fee_calculator/FeeCalculator.UnitPrice.js
@@ -97,9 +97,9 @@
     // needs to be usable by cocoon:after-insert so can bind to one or many elements
     feeQuantityChange: function ($el) {
       const self = this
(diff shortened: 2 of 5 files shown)


pking-moj (approved):

app/webpack/javascripts/modules/external_users/claims/fee_calculator/FeeCalculator.UnitPrice.js
```diff
@@ -97,9 +97,15 @@
     // needs to be usable by cocoon:after-insert so can bind to one or many elements
     feeQuantityChange: function ($el) {
       const self = this
-      const $els = $el || $('.js-fee-quantity')
+      const $miscEls = $el || $('#misc-fees')
+      const $fixedEls = $el || $('.js-fee-quantity')
```
jrmhaig: If this is being used in two (or more) different places then can it be made more general by using a new class on the wrapping element instead of using the `misc-fees` id? This can be added both to the `misc-fees` element and the similar element in the other pages where this is used. It would be good to use a name beginning with `js-` so that it can be seen that it relates to Javascript.

If the two places where this happens need to be handled differently then perhaps there needs to be a new function instead of having a single function that does both.

AlexRyanShep: With the first part, we can't use one shared wrapping element as we need slightly different behaviours on the two forms. The first form we've not been working on is more automated and needs the ability to update on the original change behaviour as far as we can see. The misc fees form creates an infinite loop if we do that.

As to the second part, good point :-) Just pushing in a new commit now that splits the two into two different functions.

app/webpack/javascripts/application.js
```diff
@@ -105,7 +105,7 @@ if (!String.prototype.supplant) {
   $('#misc-fees').on('cocoon:after-insert', function (e, insertedItem) {
     const $insertedItem = $(insertedItem)
     moj.Modules.FeeCalculator.UnitPrice.miscFeeTypesSelectChange($insertedItem.find('.fx-misc-fee-calculation'))
-    moj.Modules.FeeCalculator.UnitPrice.feeQuantityChange($insertedItem.find('.js-fee-calculator-quantity'))
+    moj.Modules.FeeCalculator.UnitPrice.miscFeeFormRefresh($insertedItem.find('.js-fee-calculator-quantity'))
```
jrmhaig, line 108: I'm not sure. Aren't both the old and the new lines required here?

AlexRyanShep: This cocoon section is specifically targeting the Misc Fees form (#misc-fees), so we only need our new function here. The old function (feeQuantityChange) is used in the fixed fee form on the previous page, elsewhere in code we haven't changed.

Obsiye (approved):

PR #5787, opened 2023-06-19
Minor layout improvements to caseworker expense table

mpw5:
#### What

When a caseworker views a submitted claim with expense data, there are a couple of small issues with the expenses table that is displayed:

1) no margin is rendered below the table, meaning that the 'Additional claim information' section appears directly below the table with no spacing. 

2) Numeric cells are left-justified, while GDS guidance states that they (and their column headers) should be right justified for readability.

Both these are caused by the use of the datatables library to sort the data in the expenses table. This has the affect of overriding govuk table styling.

#### Why

To improve the appearance of the service.

#### How

Adds custom styling to `app/webpack/stylesheets/components/_table.scss` to explicitly set the `margin-bottom` and `text-align` properties for the `expenses-data-table` class.

#### Before

image

#### After

image

### DIFF
--- a/app/webpack/stylesheets/components/_table.scss
+++ b/app/webpack/stylesheets/components/_table.scss
@@ -166,6 +166,11 @@ table.dataTable {
 
   &.expenses-data-table {
     margin-bottom: govuk-spacing(6);
+
+    .govuk-table__header--numeric,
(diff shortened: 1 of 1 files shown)


jrmhaig (approved):
:+1:

PR #5838, opened 2023-07-05
Fix n+1 query in CaseWorkerClaim model

mpw5:
#### What

Similar to https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5735. 

Fixes an inefficient approach to creating `UserMessageStatus` records in the `generate_message_statuses` method in `CaseWorkerClaim`.

#### Why

To make database calls more efficient and reduce page load times.

#### How

Use `Message` and `User` objects rather than their ids when creating new `UserMessageStatus` records, as using ids requires additional database calls.


SamKyprianou (approved):

Obsiye (approved):

spec/models/case_worker_claim_spec.rb
```diff
@@ -12,6 +12,44 @@
...
+  describe 'generate_message_statuses' do
+    before do
+      messages = create_list(:message, 3)
+      claim.messages << messages
+    end
+
+    it { expect { case_worker_claim }.to change(UserMessageStatus, :count).by(3) }
+  end
+
+  describe 'set_claim_allocated!' do
```
jrmhaig: (nit)

```suggestion
  describe '#set_claim_allocated!' do
```

jrmhaig (approved):
:+1:

PR #5867, opened 2023-07-11
CTSKF-373 Fix failing state transition validations

mpw5:
#### What

Fix failing state transition validations

#### Tickets

[CTSKF-373](https://dsdmoj.atlassian.net/browse/CTSKF-373)
[CFP-223](https://dsdmoj.atlassian.net/browse/CFP-223)
[CFP-257](https://dsdmoj.atlassian.net/browse/CFP-257)

#### Why

The automated overnight stale claim archiver fails each night with 80 claims that cannot be archived. These 80 claims all fail due to validation errors.

Claims should not be validated when transitioning state. Code in the `StateMachine` model disables validation, but needs to be called by the claim models being validated. This is done for all advocate claims and the `LitgatorClaim` model, but not for litigator interim, hardship or transfer claims.

All 80 claims currently affected by this issue are litigator interim claims.

#### How
This commit updates the validation rules on the litigator `InterimClaim`, `LitigatorHardshipClaim` and `TransferClaim` models, bringing them inline with the `LitigatorClaim` model.

[CTSKF-373]: https://dsdmoj.atlassian.net/browse/CTSKF-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CFP-223]: https://dsdmoj.atlassian.net/browse/CFP-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CFP-257]: https://dsdmoj.atlassian.net/browse/CFP-257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/claim/litigator_hardship_claim.rb
```diff
@@ -2,8 +2,9 @@ module Claim
   class LitigatorHardshipClaim < BaseClaim
     route_key_name 'litigators_hardship_claim'
 
-    validates_with ::Claim::LitigatorHardshipClaimValidator
-    validates_with ::Claim::LitigatorSupplierNumberValidator
+    validates_with ::Claim::LitigatorHardshipClaimValidator,
+                   unless: proc { |c| c.disable_for_state_transition.eql?(:all) }
+    validates_with ::Claim::LitigatorSupplierNumberValidator, if: proc { |c| c.draft? }
```
jrmhaig, line 7: I guess that this validator isn't relevant for this work but it is worth making consistent with the `Claim::LitigatorClaim` class.

mpw5: Yes, I added this for consistency. I can't see any reason why we should be validating supplier numbers differently on hardship or interim claims so it made sense to bring them in line.

I think they should all the litigator claim models should be consistent now (unless I'm missing something obvious, which is quite possible 😀 )

jrmhaig (approved):

PR #5874, opened 2023-07-12
N+1 fix for external user claim summary page - eager loading attachments and blobs

SamKyprianou:
**What**
The CCCD application is currently experiencing performance issues and potential errors due to the presence of 'N+1' queries in the ClaimsController. These queries are causing poor database performance and negatively impacting user experience. The purpose of this PR is to address and optimise the 'N+1' queries in the ClaimsController, improving application performance and resolving potential errors.

**Why**
The 'N+1' queries identified in the ClaimsController are leading to poor performance and potential errors for users. These queries retrieve unnecessary data from the database, resulting in additional round trips and slowing down the application. By optimising these queries, we aim to enhance the application's performance, reduce response times, and eliminate errors caused by inefficient database access.

**How**
To address the 'N+1' queries, the following steps were implemented:

- Identified the specific areas in the ClaimsController where the 'N+1' queries were occurring: ExternalUsers::ClaimsController#Summary.
- Analysed the code and database access patterns to understand the underlying cause of the 'N+1' queries.
- Implemented efficient database access strategies, such as eager loading or proper association handling, to optimise the queries.
- Modified the code to ensure that the required data is loaded in a single query or through optimised database joins, reducing unnecessary round trips.

### DIFF
--- a/app/services/s3_zip_downloader.rb
+++ b/app/services/s3_zip_downloader.rb
@@ -7,7 +7,7 @@ class S3ZipDownloader
 
   def generate!
     "./tmp/#{@claim.case_number}-#{SecureRandom.uuid}-documents.zip".tap do |bundle|
-      build_zip_file @@claim.documents.includes(:document_blob, :converted_preview_document_attachment), bundle
+      build_zip_file @claim.documents.includes(:document_blob, :converted_preview_document_attachment), bundle
     end
   end
 
--- a/app/services/s3_zip_downloader.rb
+++ b/app/services/s3_zip_downloader.rb
@@ -14,6 +14,7 @@ class S3ZipDownloader
   private
 
   def build_zip_file(documents, bundle)
+    documents.includes(:document_blob, :converted_preview_document_attachment) # Eager load associations
     Dir.mktmpdir("#{@claim.case_number}-") do |tmp_dir|
       Zip::File.open(bundle, Zip::File::CREATE) do |zip_file|
         documents.map(&:document).each_with_index do |document, i|
--- a/app/services/s3_zip_downloader.rb
+++ b/app/services/s3_zip_downloader.rb
@@ -7,14 +7,13 @@ class S3ZipDownloader
 
   def generate!
     "./tmp/#{@claim.case_number}-#{SecureRandom.uuid}-documents.zip".tap do |bundle|
-      build_zip_file @claim.documents, bundle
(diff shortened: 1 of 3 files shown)


mpw5 (approved):
Looks good. I'm happy to approve this, but I wonder if this might also be a good opportunity to fix a couple of similar N+1s.

I did a quick check of the code and there a two other places where documents are being iterated over and we would benefit from some eager loading:

* https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/b0b040e9b9fc1a517c7991e220f14ac1a00a7ea5/app/views/external_users/claims/supporting_documents/_new.html.haml#L8

and

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/f2e8e7f4d13e52d1dc24027f39776af3b879bd99/app/services/s3_zip_downloader.rb#L19

It's a bit of scope creep but I think it's essentially the same fix in multiple places and might save us having to tackle those separately in future.

app/services/s3_zip_downloader.rb
```diff
@@ -14,6 +14,7 @@ def generate!
   private
 
   def build_zip_file(documents, bundle)
+    documents.includes(:document_blob, :converted_preview_document_attachment) # Eager load associations
```
mpw5: It might be better to do the eager loading when `documents` are first queried on line 10 - eg:

```
build_zip_file @claim.documents.includes(:document_blob, :converted_preview_document_attachment), bundle
```

I think your approach will result in two database calls, whereas this will do it all in a single call.

mpw5 (changes requested):
I've suggested an improvement to the change to `app/services/s3_zip_downloader.rb`, but looks good apart from that 👍

app/services/s3_zip_downloader.rb
```diff
@@ -7,7 +7,7 @@ def initialize(claim)
 
   def generate!
     "./tmp/#{@claim.case_number}-#{SecureRandom.uuid}-documents.zip".tap do |bundle|
-      build_zip_file @claim.documents, bundle
+      build_zip_file @@claim.documents.includes(:document_blob, :converted_preview_document_attachment), bundle
```
mpw5: One last change, sorry 😬 - I don't think the additional `@` is necessary here. It changes `claim` from an instance variable to a class variable, which isn't needed here (and I think is making your tests fail 😀)

mpw5 (approved):
🎉

PR #5884, opened 2023-07-14
[CTSKF-297] Task to fix LGFS offences in database

jrmhaig:
#### What

Remove the duplicate set of offence records in the database for LGFS fee scheme 10.

#### Ticket

[CCCD - Fix offence records](https://dsdmoj.atlassian.net/browse/CTSKF-297)

#### Why

The `Offence` model contains details of offences linked linked to fee schemes and this is a one-to-many relation. For example, a sample offence may be linked to both AGFS fee scheme 9 and LGFS fee scheme 9.

```
pry(main)> o = Offence.where.not(offence_class_id: nil).sample
pry(main)> o.fee_schemes.pluck(:name, :version)
=> [["AGFS", 9], ["LGFS", 9]]
```

With the creation of LGFS fee scheme 10 the entire list of LGFS fee scheme 9 offences was duplicated but for simplicity it is better to simply add the new fee scheme to the one-to-many relation on the existing records. This reduces unnecessary duplication.

#### How

A Rake task is created to perform the update in a three step process;

1. Add LGFS fee scheme 10 to all offence records that currently apply to LGFS fee scheme 9. This results in the list of offences as being 'common' to both LGFS fee schemes.
2. Find all LGFS fee scheme 10 claims and update the offence to the equivalent common offence, as updated in step 1. The correct offence can be found by the unique code. The fee scheme 9 (now common) offence has a unique code in an alphanumeric format with a single underscore, such as `OKPDOE_B`, while the fee scheme 10 offence has the same unique code appended with `~10`, such as `OKPDOE_B~10`.
3. Remove all offences that were created new for LGFS fee scheme 10.

Note that due to an oversight when creating LGFS fee scheme 10 the new offences were never used. `app/services/claims/fetch_eligible_offences.rb` is used to get a list of the correct offences but unless `claim.agfs_reform?` is true then the default set is always returned. To test the task completely, therefore, at least one claim should be changed to use an LGFS fee scheme 10 offence and this can be done with;

```ruby
claim = Claim::LitigatorClaim.last
offence = Offence.find_by(unique_code: "#{claim.offence.unique_code}~10")
claim.offence = offence
claim.save
```

#### Result

The `db:lgfs_scheme_ten_clean:status` task shows the current status. Prior to updating the results are;

```
% bundle exec rails db:lgfs_scheme_ten_clean:status
Offences linked to any LGFS fee scheme:     782
Offences linked only to LGFS scheme 9:      391
Offences linked only to LGFS scheme 10:     391
Offences linked to LGFS scheme 9 and 10:    0

LGFS scheme 10 offences attached to claims: 1
  Offence: Unlawful wounding
    Lesser offences involving violence or damage and less serious drug offences
    Unique code:                  UNLAWW_C~10
    Scheme 9 offence exists:      Yes
    Claim ids:
      1474802
```

The task to clean up the offences is `db:lgfs_scheme_ten_clean:merge` which will do a dry-run unless the `true` argument is added;

```
% bundle exec rails 'db:lgfs_scheme_ten_clean:merge[true]'
This will merge LGFS fee scheme 10 offences into LGFS fee scheme 9. Are you sure?: [no/yes] yes
working...
Offence: ABOCUT_C~10
  Abandonment of children under two
  Lesser offences involving violence or damage and less seriou
  Claims to update: 0
    [UPDATE] Add fee scheme 'LGFS 10' to offence ABOCUT_C
      [UPDATE] Before: AGFS 9, LGFS 9
      [UPDATE] After: AGFS 9, LGFS 9, LGFS 10
    [REMOVE] Offence ABOCUT_C~10
(etc)
```

After the update;

```
% bundle exec rails db:lgfs_scheme_ten_clean:status
Offences linked to any LGFS fee scheme:     391
Offences linked only to LGFS scheme 9:      0
Offences linked only to LGFS scheme 10:     0
Offences linked to LGFS scheme 9 and 10:    391

LGFS scheme 10 offences attached to claims: 0
```

#### Rollback

The scheme 10 offences can be recreated using the rollback task; `db:lgfs_scheme_ten_clean:rollback`. Note that this will not update the existing scheme 10 claims to use the new offence although this can be done by uncommented the relevant seconds from `db/seeds/schemas/clean_lgfs_fee_scheme_10.rb`. This can take a long time - over an hour for an anonymised copy of the production database.

### DIFF
--- a/db/seeds/schemas/clean_lgfs_fee_scheme_10.rb
+++ b/db/seeds/schemas/clean_lgfs_fee_scheme_10.rb
@@ -7,8 +7,9 @@ module Seeds
       class MissingSchemeNineOffence < StandardError; end
       class SchemeTenOffenceExists < StandardError; end
 
-      def initialize(pretend: false)
+      def initialize(pretend: false, quiet: false)
         @pretend = pretend
+        @quiet = quiet
       end
 
       def status
@@ -37,28 +38,28 @@ module Seeds
 
       def up
         lgfs_scheme_ten_only.each do |offence|
-          puts "Offence: #{offence.unique_code}"
-          puts "  #{offence.description[0, 60]}"
-          puts "  #{offence.offence_class.description[0, 60]}"
+          puts "Offence: #{offence.unique_code}" unless @quiet
+          puts "  #{offence.description[0, 60]}" unless @quiet
+          puts "  #{offence.offence_class.description[0, 60]}" unless @quiet
           Offence.transaction do
             new_offence = scheme_nine_offence_for(offence)
             update_claims(offence.claims, new_offence)
             add_scheme_ten_to(new_offence)
             remove_redundant(offence)
           end
-          puts "-----"
+          puts "-----" unless @quiet
         end
       end
 
       def down
         all_lgfs_offences.each do |offence|
-          puts "Offence: #{offence.unique_code}"
-          puts "  #{offence.description[0, 60]}"
-          puts "  #{offence.offence_class.description[0, 60]}"
+          puts "Offence: #{offence.unique_code}" unless @quiet
+          puts "  #{offence.description[0, 60]}" unless @quiet
+          puts "  #{offence.offence_class.description[0, 60]}" unless @quiet
           Offence.transaction do
             create_scheme_ten_offence_for(offence)
           end
-          puts "-----"
+          puts "-----" unless @quiet
         end
       end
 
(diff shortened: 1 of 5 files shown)


db/seeds/schemas/clean_lgfs_fee_scheme_10.rb
```diff
@@ -0,0 +1,163 @@
...
+        if offence.unique_code.match(/~10/)
+          puts "    [NOT-DUPLICATING] Offence #{offence.unique_code}".yellow
+          return
+        end
+        begin
+          new_offence = offence.dup
+          new_offence.id = 7999 + offence.id # lib/tasks/lgfs_scheme_ten.rake starts adding ids at 8000
+          new_offence.unique_code = new_offence.unique_code + '~10'
+          new_offence.fee_schemes = [fee_scheme_ten]
+          # claims = offence.claims.select { |claim| claim.fee_scheme == fee_scheme_ten }
```
mpw5, line 95: Can this and the other comments in this file be removed?

lib/tasks/lgfs_scheme_ten_clean.rake
```diff
@@ -0,0 +1,48 @@
...
+      cleaner.status
+    end
+
+    desc 'Merge LGFS scheme 10 offences into LGFS scheme 9 offences'
+    task :merge, [:not_pretend] => :environment do |_task, args|
+
+      # seed['true'] should seed, otherwise pretend
+      args.with_defaults(not_pretend: 'false')
+      not_pretend = !args.not_pretend.to_s.downcase.eql?('false')
+      pretend = !not_pretend
```
mpw5, line 17: I don't think it makes any functional difference, but something like the following might be easier for a reader to parse, as it contains less negation:

```
      pretend = args.not_pretend.to_s.downcase.eql?('false')
      not_pretend = !pretend
```

lib/tasks/lgfs_scheme_ten_clean.rake
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+
+    desc 'Revert LGFS fee scheme 10 offences'
+    task :rollback, [:not_pretend] => :environment do |_task, args|
+
+      # rollback['true'] should rollback, otherwise pretend
+      args.with_defaults(not_pretend: 'false')
+      not_pretend = !args.not_pretend.to_s.downcase.eql?('false')
+      pretend = !not_pretend
```
mpw5, line 36: As above

db/seeds/schemas/clean_lgfs_fee_scheme_10.rb
```diff
@@ -0,0 +1,163 @@
...
+        end
+      end
+
+      def display_fee_schemes(*fee_schemes)
+        fee_schemes.map { |fs| "#{fs.name} #{fs.version}" }.join(', ')
+      end
+
+      def remove_redundant(offence)
+        if pretending?
+          puts "    [WOULD-REMOVEE] Offence #{offence.unique_code}".yellow
```
mpw5: Should be `WOULD-REMOVE` not `WOULD-REMOVEE`?

mpw5 (approved):
I've left some nit-picky comments, but other than those this LGTM 👍

db/seeds/schemas/clean_lgfs_fee_scheme_10.rb
jrmhaig: I was in two minds about this. The commented sections are what are required to set the offences of fee scheme 10 claims but it is very slow if there are lots of claims.

mpw5: Ah, sorry, yes, you mentioned this in the PR description. I think it's fine to leave it in there if there's a chance we might need to roll back - we can cross the performance bridge when we come to it.

mpw5 (approved):
👍

db/seeds/schemas/clean_lgfs_fee_scheme_10.rb
```diff
@@ -0,0 +1,164 @@
...
+          new_offence.fee_schemes = [fee_scheme_ten]
+          # claims = offence.claims.select { |claim| claim.fee_scheme == fee_scheme_ten }
+          raise SchemeTenOffenceExists unless new_offence.valid?
+          if pretending?
+            puts "    [WOULD-CREATE] Offence #{new_offence.id}/#{new_offence.unique_code}".yellow unless @quiet
+            puts "    [WOULD-REMOVE] Fee scheme 10 from offence #{offence.unique_code}".yellow unless @quiet
+            # puts "    [WOULD-UPDATE] Move #{claims.count} fee scheme 10 claims (out of #{offence.claims.count}) to new offence".yellow unless @quiet
+          else
+            puts "    [CREATE] Offence #{new_offence.id}/#{new_offence.unique_code}".green unless @quiet
+            new_offence.save
```
Obsiye: would `new_offence.save!` make a difference? raising an exception if it fails?

PR #5887, opened 2023-07-17
CTSKF-399 Disable false positive in ModSecurity firewall

Obsiye:
#### What

Disable the ModSecurity rule [PHP Functions: Variable Function Prevent Bypass](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf) for the /messages endpoint  

Enable access to our ModSecurity Audit Logs

#### Ticket

[CTSKF-399](https://dsdmoj.atlassian.net/browse/CTSKF-399)

#### Why

There are some false positives in ModSecurity which is preventing users from sending messages in the chat feature.
Disabling this rule will fix this problem.

Enabling access to our audit logs will allow us more observability over what ModSecurity is blocking and why.

#### How

Update the ingress, SecRule, to remove the rule for requests sent to the /messages endpoint that include a body

Update the ingress, SecDefaultAction, to give us access to [ModSecurity audit logs](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/networking/modsecurity.html#accessing-your-modsec-logs).

[CTSKF-399]: https://dsdmoj.atlassian.net/browse/CTSKF-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- [X] Dev
- [ ] Api sandbox
- [x] Dev lgfs
- [x] Staging
- [x] Production

### DIFF
--- a/.k8s/live/dev/ingress.yaml
+++ b/.k8s/live/dev/ingress.yaml
@@ -7,6 +7,7 @@ metadata:
     nginx.ingress.kubernetes.io/enable-modsecurity: "true"
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
+      SecDefaultAction "phase:2,pass,log,tag:github_team=crime-billing-online"
       SecAction "id:900110,phase:1,nolog,pass,t:none,setvar:tx.inbound_anomaly_score_threshold=6"
       SecAction "id:900200,phase:1,nolog,pass,t:none,setvar:tx.allowed_methods=GET HEAD POST OPTIONS PUT PATCH DELETE"
       SecRuleRemoveById 200002
--- a/.k8s/live/dev/ingress.yaml
+++ b/.k8s/live/dev/ingress.yaml
@@ -11,8 +11,7 @@ metadata:
       SecAction "id:900200,phase:1,nolog,pass,t:none,setvar:tx.allowed_methods=GET HEAD POST OPTIONS PUT PATCH DELETE"
       SecRuleRemoveById 200002
       SecRuleRemoveById 200003
-      SecRule REQUEST_URI "@contains /messages" \
-        "id:1000,phase:2,pass,nolog,ctl:ruleRemoveById=921110"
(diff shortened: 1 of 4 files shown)


.k8s/live/dev/ingress.yaml
```diff
@@ -7,12 +7,12 @@ metadata:
...
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
+      SecDefaultAction "phase:2,pass,log,tag:github_team=crime-billing-online"
       SecAction "id:900110,phase:1,nolog,pass,t:none,setvar:tx.inbound_anomaly_score_threshold=6"
       SecAction "id:900200,phase:1,nolog,pass,t:none,setvar:tx.allowed_methods=GET HEAD POST OPTIONS PUT PATCH DELETE"
       SecRuleRemoveById 200002
       SecRuleRemoveById 200003
-      SecRule REQUEST_URI "@contains /messages" \
-        "id:1000,phase:2,pass,nolog,ctl:ruleRemoveById=921110"
+      SecRule REQUEST_URI "@contains /messages" "id:1000,phase:2,pass,nolog,ctl:ruleRemoveById=921110,ctl:ruleRemoveById=933210"
```
mpw5, line 15: Not a blocker, but is there any reason why this is on one line while the others are split over two?

mpw5 (approved):
LGTM. I've left one nit-picky comment, but not a blocker. 

I also wonder if it might be worth adding the line 

```
SecRule REQUEST_URI "@contains /messages" \ 
  "id:1000,phase:2,pass,nolog,ctl:ruleRemoveById=921110,ctl:ruleRemoveById=933210"
```
to `.k8s/live/api-sandbox/ingress.yaml` so that all our environments are consistent. I'm not sure why didn't add the last rule change to that environment, but now might be a good time to realign it.

.k8s/live/dev/ingress.yaml
Obsiye: i don't believe so. i think the newline is for readability. i'll standardise it

mpw5 (approved):

PR #5899, opened 2023-07-20
[CTSKF-267] Unsafe use of parameter as class name

AlexRyanShep:
#### What
`#pricer` method in `ExternalUsers::Fees::PricesController` was unsafely using a class name passed to it without validation, and calling the `#call` method of that class.

#### Ticket

[CTSKF-267 Unsafe use of parameter as class name](https://dsdmoj.atlassian.net/browse/CTSKF-267)

#### Why
This was poor practice and could have created an injection vulnerability in the future

#### How
`#pricer` now checks the Price Type input to it against a hash of allowable types and either returns the correct class directly, or returns a new Null Object class which has an overridden `#call` that returns a response indicating an error.

New rspec test created to test this, and cleaned up a few of the other tests.


mpw5 (approved):
👍

PR #5925, opened 2023-07-28
Enable rerunning of intermittently failing cucumber tests in CircleCI

mpw5:
#### What

At the moment, if a cucumber test fails when run in CircleCI, all feature tests must be rerun. This takes ~10 minutes and costs credits.

CircleCI has [introduced](https://discuss.circleci.com/t/product-launch-rerun-failed-tests-circleci-tests-run/47775) the ability to only rerun failed tests, which will save time and money when experiencing flaky tests which fail intermittently.

This commit implements the necessary change to enable that functionality following guidance here https://circleci.com/docs/rerun-failed-tests/.

Note: this only works for tests that are intermittently failing. If a test fails because of issues in code and is resolved by a subsequent commit, the whole test pipeline will run as usual.

#### Why

To save money and developer time.

#### How

 - Amends the `run-cucumber` job to include the `circleci tests run` command, which is necessary to enable the rerunning of tests
 - Adds a 'halt' step to the `cucumber-tests` job to prevent unnecessary use of test containers when there are no tests to run. As we run our feature tests in parallel across six containers, if we subsequently re-run a single failed test, all six containers will still start up, install dependencies, create a database, precompile assets etc, even though there are no tests to run on five of them. This allows us to fail fast, saving additional credits.

#### Evidence

See this run for example: https://app.circleci.com/pipelines/github/ministryofjustice/Claim-for-Crown-Court-Defence/14326

1) In the initial test run, one (deliberately flaky) test failed:
image

2) I am now able to select `Rerun failed tests` from the `Rerun` dropdown:
image

3) This caused the single failing test to be re-run, which this time passed:
image

4) The other containers stopped quickly as there were no tests for them to run:
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,10 +218,6 @@ commands:
       - install-chrome
       - precompile-assets
       - run:
-          name: Make cucumber test storage dir
-          command: |
-            mkdir -p /tmp/test-results/cucumber
-      - run:
           name: Run cucumber tests
           command: |
             circleci tests glob "features/**/*.feature" \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,13 +224,13 @@ commands:
       - run:
           name: Run cucumber tests
           command: |
-            TESTFILES=$(circleci tests glob "features/**/*.feature" | circleci tests split --split-by=timings --timings-type=filename)
-
-            RUBYOPT='-W0' bundle exec cucumber \
-            --format pretty \
-            --format junit,fileattribute=true \
-            --out /tmp/test-results/cucumber \
-            -- ${TESTFILES}
+            circleci tests glob "features/**/*.feature" \
+              | circleci tests run --command="xargs bundle exec cucumber \
+                                                      --format pretty \
+                                                      --format junit,fileattribute=true \
+                                                      --out /tmp/test-results/cucumber" \
+                                   --split-by=timings \
+                                   --timings-type=filename
       - store_artifacts:
           path: tmp/capybara
       - store_test_results:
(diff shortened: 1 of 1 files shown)


Obsiye (approved):

.circleci/config.yml
```diff
@@ -432,6 +428,13 @@ jobs:
     parallelism: 6
     steps:
       - checkout
+      - run:
+          name: Halt if no cucumber tests to run
+          command: |
+            mkdir -p ./tmp && \
+            >./tmp/tests.txt && \
```
Obsiye, line 435: i'm curious, what does `>` do in `>./tmp/tests.txt` ?

.circleci/config.yml
```diff
@@ -432,6 +428,13 @@ jobs:
...
+            circleci tests glob "features/**/*.feature" | circleci tests run --command ">./tmp/tests.txt xargs echo" --split-by=timings
```
Obsiye, line 436: not an issue, just that xargs defaults to `echo` so maybe `echo` isn't needed here? 

i'm learning as I read this PR .

mpw5: Both good questions @Obsiye. To be honest, I didn't write that command myself, I just copied it from the CircleCI guidance: https://circleci.com/docs/rerun-failed-tests/#known-limitations 😀 

I think you're right about `echo` not being needed. I'll test it when I'm back from leave.

My shell scripting is basic, but as I understand it, the `>` in `>./tmp/tests.txt` is a redirect, so that command taken as a whole creates the `./tmp` directory, then creates the file `tests.txt` in it, and redirects the output of the following command into that file. The third line runs the code which splits the tests between parallel containers, so we end up with a file in `./tmp/tests.txt` which contains the names of all of the tests which would be run in that container.

The next line `[ -s tmp/tests.txt ] || circleci-agent step halt` checks to see if there are any tests to be run, and if not, stops.

Hopefully that helps (but possibly didn't make any sense at all!) We can catch up on it in a couple of weeks if you want.

PR #5988, opened 2023-08-22
More miscellaneous rubocop fixes

mpw5:
#### What

Fixes a number of rubocop offences.

#### Why

To improve the overall quality of our code.


spec/services/claims/fee_calculator/unit_price_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.shared_examples 'a successful daily attendance fee calculation' do
+RSpec.shared_context 'a successful daily attendance fee calculation' do
```
jrmhaig: These are shared examples rather than shared contexts but the tests are using the custom `it_returns`, which is probably why there are Rubocop offences flagged as it appears that there are no tests in the block. The correct solution is probably either to add `it_returns` in the configuration as a recognised test or to use something that is more standard. I think that `it_returns` is just an alias for `it_behaves_like`.

mpw5: > These are shared examples rather than shared contexts but the tests are using the custom `it_returns`, which is probably why there are Rubocop offences flagged as it appears that there are no tests in the block. The correct solution is probably either to add `it_returns` in the configuration as a recognised test or to use something that is more standard. I think that `it_returns` is just an alias for `it_behaves_like`.

Doh. Good spot. I'll fix it.

mpw5: @jrmhaig I've fixed the issue you spotted by adding `it_returns` as an accepted rspec example in `.rubocop.yml`. This also had the rather nice side effect of removing 170 offences of another cop 🎉 

image

Can you re-review when you have time? No urgency. Thanks.

jrmhaig (approved):

PR #6013, opened 2023-08-31
Prepare AWS access for move to IAM

jrmhaig:
#### What

Allow AWS access to use access keys and secrets if they exist and to use IAM if they do not.

#### Ticket

[How will our systems continue to operate under short lived credentials?](https://dsdmoj.atlassian.net/browse/CTSKF-574)

#### Why

The configuration of our environments is being changed to use IAM instead of credentials. To allow this change to be made gradually both methods will be allowed temporarily.

#### How

At each place where AWS is accessed the presence of an access key is tested. If it is present then it will be used.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -45,7 +45,7 @@ Rails.application.configure do
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :amazon
+  config.active_storage.service = Settings.aws&.s3&.access ? :amazon : :amazon_with_iam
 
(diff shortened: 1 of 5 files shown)


config/environments/production.rb
```diff
@@ -45,7 +45,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :amazon
+  config.active_storage.service = Settings.aws&.s3&.access ? :amazon_wthout_iam : :amazon
```
mpw5: Typo: `amazon_wthout_iam` should be `amazon_without_iam `

mpw5 (approved):

PR #6016, opened 2023-09-01
Use latest version of Chrome for tests

jrmhaig:
#### What

Use the latest version of Chrome for tests.

#### Ticket

N/A

#### Why

#6012 fixed the version of Chrome because there was no version of Chromedriver released for Chrome 116 and this was causing the pipelines to fail. Ideally we would want tests to run against the latest release.

#### How

The Chrome version is now unrestricted.

If this PR is periodically rebased then it will be possible to see when the latest version can be used again as the tests will run successfully. The quickest way to see this is to check the 'Install Chromedriver' step of the 'other tests' pipeline.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,10 +128,7 @@ commands:
 
   install-chrome:
     steps:
-      - browser-tools/install-chrome:
-          # There is no version of Chromedriver for the current latest version of Chrome (116)
-          chrome-version: 114.0.5735.90
-          replace-existing: true
+      - browser-tools/install-chrome
       - browser-tools/install-chromedriver
   precompile-assets:
     description: >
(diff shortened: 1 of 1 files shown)


Obsiye (approved):

Obsiye:
sorry i accidentally approved

Obsiye (changes requested):
accidentally approved this without looking at the failing tests

Obsiye:
In CDUI, i resolved a similar issue https://github.com/ministryofjustice/laa-court-data-ui/commit/1eb4c9b692330057a1d4daf761568dedd6e47fc0  by upgrading `gem 'selenium-webdriver', '~> 4.14'` to 4.14

Obsiye (approved):

PR #6042, opened 2023-09-11
Rubocop: Rails/I18nLocaleTexts - move all outstanding text to translations

mpw5:
#### What

Fixes all outstanding `Rails/I18nLocaleTexts` offences in CCCD by moving text from controllers, helpers and models into locale files. (There's additional work that could be done to structure these locale files better but this PR is already big enough 😬).

#### Why

To improve our overall code quality and maintainability.

### DIFF
--- a/config/locales/devise.en.yml
+++ b/config/locales/devise.en.yml
@@ -92,9 +92,10 @@ en:
         page_heading: Sign in
         password: Password
         sign_in: Sign in
-      signed_in: ""
-      signed_out: ""
-      already_signed_out: ""
+      user:
+        signed_in: Signed in successfully
+        signed_out: You have signed out
+        already_signed_out: You have signed out
     shared:
       error_messages:
         error_summary_title: There is a problem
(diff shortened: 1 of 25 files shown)


config/locales/devise.en.yml
```diff
@@ -92,9 +92,10 @@ en:
         page_heading: Sign in
         password: Password
         sign_in: Sign in
-      signed_in: ""
-      signed_out: ""
-      already_signed_out: ""
+      user:
+        signed_in: Signed in successfully
```
jrmhaig, line 96: Do you have any idea why these messages were blanked? Does this affect behaviour in any way.

For reference see #647, although that doesn't really say much.

jrmhaig (approved):
:+1:

config/locales/devise.en.yml
mpw5: I saw that helpful commit too 😁 

The wordings that were in this file before that commit are the default devise messages. Blanking them out didn't actually have any affect - devise will still default to using them as they are also configured inside the gem. CCCD was then using a customised message for signing out set in the `SessionsController`.

I've moved that into the locale file now. Technically we could leave the `signed_in` message blank (or remove it completely) as we're just using the default, but I added it in as I thought it would be more consistent.

I've tested locally comparing against staging and the messages are appearing the same in both environments.

PR #6048, opened 2023-09-13
Remove audit tasks

jrmhaig:
#### What

Remove auditing rake tasks.

#### Ticket

N/A

#### Why

These tasks are not used and one of them doesn't appear to run reliably. For instance, it fails on staging, although I haven't tried it on production as I don't really know what it is doing.

#### How

Remove app/services/vat_auditor.rb and lib/tasks/audit.rake


mpw5 (approved):
🎉

PR #6058, opened 2023-09-15
Update .nvmrc from 14.17.0 to 14.20.0

AlexRyanShep:
#### What
Update the version of node referenced in .nvmrc from 14.17.0 to 14.20.0

#### Ticket

No associated ticket

#### Why

Current node version in this file (14.17.0) is incompatible with other gems we use (Selenium webdriver). Selenium needs at least node 14.20.0 as a minimum.

#### How

Updated file will now make `nvm install` as in development setup instructions install and activate the correct node version.


jrmhaig (approved):
:+1:

PR #6067, opened 2023-09-19
Further Rubocop fixes

mpw5:
#### What

Fixes offences for a number of rubocop cops:

`Style/WordArray`
`Style/SymbolProc`
`Style/StringLiterals`
`Style/SoleNestedConditional`
`Style/SingleLineMethods`
`Style/Semicolon`
`Style/RedundantSort`
`Style/SafeNavigation`

Also deletes the `spikes` directory, which contained code with quality issues. This directory hasn't been touched in many years and we're unlikely to ever do anything with the code, so it feels like it can be safely deleted.

#### Why

To improve overall code quality


app/validators/expense_validator.rb
```diff
@@ -73,8 +73,8 @@ def validate_calculated_distance
   def validate_mileage_rate_id
     if @record.car_travel? || @record.bike_travel?
       validate_presence(:mileage_rate_id, :blank)
-      unless @record.mileage_rate_id.nil?
-        add_error(:mileage_rate_id, :invalid) if car_travel_missing_milage_rates || bike_travel_missing_milage_rates
+      if !@record.mileage_rate_id.nil? && (car_travel_missing_milage_rates || bike_travel_missing_milage_rates)
```
jrmhaig: It is possible that `@record.mileage_rate_id.present?` could be used instead of `!@record.mileage_rate_id.nil?`, although I am not sure if it is an exact equivalent.

spec/controllers/application_controller_spec.rb
```diff
@@ -57,7 +57,10 @@
     end
 
     context 'given a user with a different role' do
-      before { user = build(:user); sign_in user }
+      before do
+        user = build(:user)
+        sign_in user
+      end
```
jrmhaig: Possibly consider something like;

```ruby
  describe '#after_sign_in_path_for' do
    before { sign_in user }

    context 'given a ??? user' do
      let(:user) { build(:???_user) }

      ...
    end

    ...
  end
```

(Not a blocker)

jrmhaig (approved):
Looks good. I have made a couple of comments that you may or may not decide to act on.

PR #6079, opened 2023-09-22
Replace unicorn with puma

jrmhaig:
#### What

Use puma instead of unicorn

#### Ticket

[SPIKE: Replace unicorn web server with puma](https://dsdmoj.atlassian.net/browse/CTSKF-487)

#### Why

For some time we have been considering moving CCCD to puma, not least as we are using puma in development but unicorn in production. Also there are various posts online that indicate that puma is preferable; E.g. https://deliveroo.engineering/2016/12/21/unicorn-vs-puma-rails-server-benchmarks.html

#### How

* Change the command to start the server in `Procfile` and `docker/docker-entrypoint.sh`.
* Remove configuration for `unicorn-worker-killer`. This was to mitigate memory leaks in unicorn.
* Remove unicorn related gems from `Gemfile`.


jrmhaig:
This deploys correctly but the smoke tests are not running for some reason.

PR #6084, opened 2023-09-25
CTSKF-541 Fee Scheme usage MI report

AlexRyanShep:
#### What
Creation of a new MI report showing a summarised view of total figures, claim types and case types across the different fee schemes in CCCD.

#### Ticket

[CTSKF-541](https://dsdmoj.atlassian.net/browse/CTSKF-541)

#### Why
The data from this report will allow our product managers and business analysts to have an easier view of how the claims we get are split between the different fee schemes, to help identify when fee schemes may be coming to end of life. 

#### How
A new report generator class (FeeSchemeUsageGenerator.rb) has been created, along with associated rspec tests. Other files have been configured to include this in the existing list of MI reports

[CTSKF-541]: https://dsdmoj.atlassian.net/browse/CTSKF-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/factories/case_stages.rb
```diff
@@ -10,63 +10,63 @@
       description { 'Pre PTPH' }
       unique_code { 'PREPTPH' }
       position { 10 }
-      case_type_id { create(:case_type, :discontinuance).id }
+      case_type { create(:case_type, :discontinuance) }
```
mpw5, line 16: I think you can fix some rubocop offences by changing the way you set up the `case_type` in these factories. Because there is an association between the `case_stage` and `case_type` models, you can use `association` instead of `create`:

```suggestion
      case_type { association(:case_type, :discontinuance) }
```

This applies to all the traits in this factory.

spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Elected cases not proceeded').first)
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Guilty plea').first)
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Hearing subsequent to sentence').first)
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Retrial').first)
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Trial').first)
+      # This is a hacky, brute-force way of generating the following two test claims, but it's the only way I could
+      # get around errors generated by the factories
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Trial').first)
+      Claim::BaseClaim.last.update_attribute(:case_type_id, 6)
+      Claim::BaseClaim.last.update_attribute(:case_type, CaseType.where(name: 'Cracked Trial').first)
```
mpw5: Using `Claim::BaseClaim.last` to update the previous claim after it's been created feels a little risky - for example if another developer one day starts changing the order in which these are created we might end up with unpredictable results. 

You could try using a block to ensure that the claim that is updated is definitely the one which was most recently created:

```suggestion
      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Trial').first) do |claim|
        claim.update_attribute(:case_type_id, 6)
        claim.update_attribute(:case_type, CaseType.where(name: 'Cracked Trial').first)
      end
```

spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+      create(:advocate_claim, :submitted, case_type: CaseType.where(name: 'Trial').first)
+      Claim::BaseClaim.last.update_attribute(:case_type_id, 7)
+      Claim::BaseClaim.last.update_attribute(:case_type, CaseType.where(name: 'Cracked before retrial').first)
```
mpw5: Ditto

app/services/stats/fee_scheme_usage_generator.rb
```diff
@@ -0,0 +1,151 @@
+require 'csv'
+
+module Stats
+  class FeeSchemeUsageGenerator
+    ORDERED_FEE_SCHEMES =
+      # Slightly wordy, but gets the intended ordering of FeeSchemes
```
mpw5: I'd probably remove this comment. You're right, and it's good that you've acknowledged it, but from a long-term point of view I don't think it adds much value.

app/services/stats/fee_scheme_usage_generator.rb
```diff
@@ -0,0 +1,151 @@
...
+      FeeScheme.where(name: 'AGFS')
+               .map { |scheme| "#{scheme.name} #{scheme.version}" }
+               .sort_by { |x| x[/\d+/].to_i } +
+      FeeScheme.where(name: 'LGFS')
+               .map { |scheme| "#{scheme.name} #{scheme.version}" }
+               .sort_by { |x| x[/\d+/].to_i }
+    CLAIM_TYPES =
+      # TODO: Low priority, but this (and the headers) could be improved by being made not static
```
mpw5: I think you could probably remove this comment - I don't think the code urgently needs changing and if we're not going to fix it now then it's unlikely we're ever going to, so the comment can probably go 🙂

spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+      'Elected cases not proceeded',
+      'Guilty plea',
+      'Hearing subsequent to sentence',
+      'Retrial',
+      'Trial'
+    ]
+  end
+
+  describe '#call' do
+    subject(:call) { described_class.new.call }
```
mpw5, line 41: The `.new` here is redundant - `described_class.call` will do the same thing.

This is because of the block of code in `app/services/stats/fee_scheme_usage_generator.rb` which we discussed last week, which allows you to call `call` directly on the class, and handles the instantiation of the `new` object for you. Not a blocker if you don't use it, just a nice to have.

```    
def self.call(...)
    new(...).call
end
```

app/services/stats/fee_scheme_usage_generator.rb
```diff
@@ -0,0 +1,152 @@
+require 'csv'
+
+module Stats
+  class FeeSchemeUsageGenerator
+    ORDERED_FEE_SCHEMES =
```
mpw5: I think it might be better if this was defined in a method rather than as a constant, as it's being dynamically derived. You could do something like:
```
    def ordered_fee_schemes
      @ordered_fee_schemes ||= FeeScheme.where(name: 'AGFS')
      .map { |scheme| "#{scheme.name} #{scheme.version}" }
      .sort_by { |x| x[/\d+/].to_i } +
      FeeScheme.where(name: 'LGFS')
      .map { |scheme| "#{scheme.name} #{scheme.version}" }
      .sort_by { |x| x[/\d+/].to_i }
    end
```

This would also make the code here easier to test, by adding something like this to the spec:

```
  describe '#ordered_fee_schemes' do
    it { expect(described_class.new.ordered_fee_schemes).to eq(fee_scheme_array) }
  end
```

spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+
+    let(:total_claims_array) do
+      ['0', '0', '0', '0', '0', '0', '0', '1', '0', nil,
+       '0', '0', '0', '0', '0', '0', '0', '1', '0', nil,
+       '0', '0', '0', '0', '0', '0', '0', '0', '0', nil,
+       '0', '0', '0', '0', '0', '0', '0', '0', '0', nil,
+       '0', '0', '0', '0', '0', '0', '0', '0', '0', nil,
+       '15', '0', '0', '0', '1', '0', '0', '4', '0', nil]
+    end
+    # TODO: Would it be better to actually have this static for testing purposes?
```
mpw5: Potentially yes, especially as this code is the same as the code used in the service class to create the fee scheme array, so this test is a little tautological. The only drawback to changing it to a hard-coded array is that it would need updating when we add new fee schemes.

Also - you don't actually use `fee_scheme_array` in a test! See my comment on the service class about how you could do that (or if you choose not to, you can just remove this).

app/services/stats/fee_scheme_usage_generator.rb
```diff
@@ -0,0 +1,154 @@
...
+
+    private
+
+    def ordered_fee_schemes
+      @ordered_fee_schemes ||= FeeScheme.where(name: 'AGFS')
+                                        .map { |scheme| "#{scheme.name} #{scheme.version}" }
+                                        .sort_by { |x| x[/\d+/].to_i } +
+                               FeeScheme.where(name: 'LGFS')
+                                        .map { |scheme| "#{scheme.name} #{scheme.version}" }
+                                        .sort_by { |x| x[/\d+/].to_i }
```
jrmhaig: You can sort in the database query, which would also enable you to avoid two queries;

```suggestion
      @ordered_fee_schemes ||= FeeScheme.where(name: ['AGFS', 'LGFS']).order(:name, :version)
                                        .map { |scheme| "#{scheme.name} #{scheme.version}" }
```

AlexRyanShep: Thanks Joe! I didn't realise that you could put multiple conditions to sort by into the order method like that. I've made a new commit using this code (with the array tweaked to use %w[] to make rubocop happy)

jrmhaig (approved):

PR #6103, opened 2023-09-29
Remove the orig_scheme_filters partial

jrmhaig:
This partial appears to be part of an incomplete refactor and does not seem to serve any useful purpose. The result is slightly different from the scheme_filters partial so the this will change the display of the reallocation page.

#### Ticket

#### Before

Screenshot 2023-09-29 at 13 08 21

#### After

Screenshot 2023-09-29 at 13 08 50

This is consistent with the allocation page.

--------

#### TODO (wip)

 - [ ] The list of claims is updated automatically when AGFS or LGFS is selected but the two forms do this in different ways;
   * In the reallocation form the page is reloaded via app/webpack/javascripts/modules/Modules.ReAllocationFilterSubmit.js
   * In the allocation form the table is updated via an AJAX request in app/webpack/javascripts/modules/Modules.AllocationDataTable.js using app/webpack/javascripts/modules/Modules.AllocationScheme.js


jrmhaig:
I naively thought this would be a simple cleanup.

PR #6104, opened 2023-09-29
Fix rubocop offences in base_claim_presenter_spec

mpw5:
#### What

Fixes nearly all code quality issues identified by rubocop in `spec/presenters/claim/base_claim_presenter_spec.rb`.

#### Why

To improve the quality of code in CCCD.


spec/presenters/claim/base_claim_presenter_spec.rb
```diff
@@ -323,20 +328,27 @@
...
-    claim.case_workers << build(:case_worker, user: build(:user, first_name: 'Alexander', last_name: 'Bell'))
-    claim.case_workers << build(:case_worker, user: build(:user, first_name: 'Louis', last_name: 'Pasteur'))
-    expect(subject.case_worker_names).to eq('Alexander Bell, Louis Pasteur')
+  describe '#case_worker_names' do
+    before do
+      claim.case_workers << build(:case_worker, user: build(:user, first_name: 'Alexander', last_name: 'Bell'))
+      claim.case_workers << build(:case_worker, user: build(:user, first_name: 'Louis', last_name: 'Pasteur'))
+    end
+
+    it { expect(presenter.case_worker_names).to eq('Alexander Bell, Louis Pasteur') }
```
jrmhaig: `claim.case_workers` appears to be empty when `claim` is first created but it is possible that the factory method could change resulting in this test breaking. It might be worth explicitly setting it to ensure that this does not happen.

```suggestion
  describe '#case_worker_names' do
    before do
      claim.case_workers = [
          build(:case_worker, user: build(:user, first_name: 'Alexander', last_name: 'Bell')),
          build(:case_worker, user: build(:user, first_name: 'Louis', last_name: 'Pasteur'))
        ]
    end

    it { expect(presenter.case_worker_names).to eq('Alexander Bell, Louis Pasteur') }
```

There are other examples of this throughout the file.

jrmhaig (approved):
Great cleanup. Approved, with one suggestion that you may like to consider.

PR #6127, opened 2023-10-06
Remove diskspace check

jrmhaig:
#### What

Remove the alert for disk space from Prometheus

#### Ticket

N/A

#### Why

The container_fs_usage_bytes metric is no longer available in k8s so this alert is just reporting false positives. It is possible that this metric can be re-enabled via a plugin but 

#### How

Remove the `DiskSpace-Threshold-Reached` alert from each of the Prometheus configuration files.


jsugarman (approved):
LGTM 👍

Obsiye (approved):

PR #6139, opened 2023-10-11
Set AllowedPatterns for Naming/VariableNumber cop

mpw5:
#### What

The rubocop `Naming/VariableNumber` cop flags variables which end in a number. The CCCD codebase uses this format a lot, especially in tests relating to fee schemes - eg `agfs_scheme_9`.

One option would be to rename all such instances (eg to `agfs_scheme_nine`), however this would be a big change and it feels like that in these cases our current usage is valid.

Instead this commit adds an `AllowedPattern` regex to exclude all variable names that end in `_scheme_<number>` from the cop.


jrmhaig (approved):
:+1:

PR #6148, opened 2023-10-12
CTSKF-241 Zeitwerk: miscellaneous namespace fixes

mpw5:
#### What

As a pre-requisite for upgrading to rails 7, CCCD needs to move from the `classic` autoloader to `zeitwerk`. This requires us to update the CCCD codebase to be compatible with `zeitwerk`.

Zeitwerk requires namespaces for classes and modules to match the file structure of the code. In certain places in CCCD this is not the case.

For example the `SearchResultHelpers` module is located at `app/interfaces/api/helpers/search_result_helpers.rb`. It is defined simply as `module SearchResultHelpers` however under `zeitwerk` this need to be changed to reflect the filepath, ie:

```
module API
  module Helpers
    module SearchResultHelpers
```

#### Ticket

[CTSKF-241](https://dsdmoj.atlassian.net/browse/CTSKF-241) (formerly CFP-179)

#### Why

* To ensure future compatibility with the rails framework.

#### How

Fixes five instances of incorrect namespacing and updates all occurrences in code. 

[CTSKF-241]: https://dsdmoj.atlassian.net/browse/CTSKF-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/caching/caching.rb
```diff
@@ -1,17 +1,19 @@
 class Caching
-  cattr_accessor :backend
+  class Caching
```
jrmhaig, line 2: I think it would make more sense to move this to `lib/caching.rb` instead of nesting the namespace.

jrmhaig (commented):
One comment but overall the naming and layout looks much clearer now.

lib/caching/caching.rb
mpw5: As discussed, I've left it as is for now and will investigate how we can refactor this later.

jrmhaig (approved):

PR #6155, opened 2023-10-16
Add reporter's email address to Zendesk ticket

jrmhaig:
#### What

Add the reports email address to Zendesk tickets

#### Ticket

N/A

#### Why

Users reporting bugs in CCCD were not getting email responses from Zendesk. This is because they were not added as the requester, who must be a Zendesk user, nor in the list of CC email addresses. Addresses can be added in the Zendesk interface but it is more convenient if they are added automatically.

#### How

When a ticket is raised using the [raise a fault](https://claim-crown-court-defence.service.gov.uk/feedback/new?type=bug_report) form there is an optional field for email address. This is currently added to the ticket in the description. This PR will additionally add it to the `email_ccs` list.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -201,7 +201,6 @@ Metrics/MethodLength:
     - 'app/services/message_queue/message_template.rb'
     - 'app/services/reports/provisional_assessments_by_dates.rb'
     - 'app/services/stats/management_information_generator.rb'
-    - 'app/services/zendesk_sender.rb'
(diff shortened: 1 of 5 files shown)


mpw5 (approved):
Excellent work!

My only thought was that maybe we could filter out invalid email addresses using a regex, so that if a user enters an incorrect/incomplete address then it isn't populated in the CC field - it might reduce the chance of us thinking we're emailing somebody but our replies not going anywhere.

I don't think it's really necessary though and overall a big improvement on what we've got at the moment, so happy for it go as it is 👍

PR #6171, opened 2023-10-20
Ctskf 589 Super Admin Data Visualisation

AlexRyanShep:
#### What

Create and add some charts to the CCCD SuperAdmin page to give a visualisation of some key stats for the claims in the system.

#### Ticket

#### Why

- Useful for BAs/PM
- POC for wider use of charts
- Alex's apprenticeship UI/Front-end work

#### How
- Add Chartkick Gem to CCCD
- Create new page off Super Admin landing page with some charts pulling data from the Claims model.
- Associated rspec request tests and one cucumber feature test


app/controllers/super_admins/stats_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    def show
+      @date_err = false
+      @total_claims = empty_results_hash
+      @total_values = empty_results_hash
+      @six_month_breakdown = empty_linechart_array
+      @chart_colours = %w[#ffdd00	#00703c #5694ca #912b88 #f47738 #85994b #003078 #f499be #b52c17 #ea7361 #66ff66]
+      process_dates
+      set_times
+      retrieve_data
```
jrmhaig: You might consider using `before_action` for these. See https://guides.rubyonrails.org/action_controller_overview.html#filters

app/controllers/super_admins/stats_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+      @to_str = @to.strftime('%d %b')
+    end
+
+    def process_dates
+      @from_input = "#{params['date_from(3i)']}-#{params['date_from(2i)']}-#{params['date_from(1i)']}"
+      @to_input = "#{params['date_to(3i)']}-#{params['date_to(2i)']}-#{params['date_to(1i)']}"
+    end
+
+    def set_provided_dates
+      @from = ActiveSupport::TimeZone['UTC'].parse(@from_input)
```
jrmhaig: Is there a reason not to use;

```ruby
Time.zone.parse(@from_input)
```

or

```ruby
Time.zone.new(
  params['date_from(3i)'].to_i,
  params['date_from(2i)'].to_i,
  params['date_from(1i)'].to_i
)
```

Also, these can be memoized with something like this:

```ruby
def from
  @from ||= Time.zone.new(
    params['date_from(3i)'].to_i,
    params['date_from(2i)'].to_i,
    params['date_from(1i)'].to_i
  )
rescue NoMethodError, ArgumentError
  Time.zone.now.at_beginning_of_month
end
```

app/controllers/super_admins/stats_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+      output
+    end
+
+    def retrieve_data
+      Claim::BaseClaim.active.non_draft.where(last_submitted_at: @from..@to).find_each do |claim|
+        fee_scheme = "#{claim.fee_scheme.name} #{claim.fee_scheme.version}"
+        @total_claims[fee_scheme] += 1
+        @total_values[fee_scheme] += (claim.total.to_f + claim.vat_amount.to_f)
+      end
```
jrmhaig: There are methods on the enumerable and hash classes that allow this to be simplified. For example;

```ruby
@total_claims = Claim::BaseClaim.active.non_draft
                  .map(&:fee_scheme)
                  .tally
                  .transform_keys { |fee_scheme| "#{fee_scheme.name} #{fee_scheme.version}" }
```

* `map` picks out the fee scheme for each claim
* `tally` reduces the array to a count of identical items as a hash
* `transform_keys` modifies the keys of the hash from instances of the FeeScheme model to a string

It is possible to do something similar for `@total_values` and the `@six_month_breakdown`.

By doing this you can avoid having to use `empty_linechart_array`, although it will not include entries for fee schemes that are not included in the selection, where the total is zero.

jrmhaig (commented):
Thanks for this. I have added a few suggestions of things you might consider.

More generally, it would be good to extract the logic of building the stats data into a separate class or a service so that the controller can be simpler. By separating the concerns in this way the tests will be clearer and it will be easier to extend in the future.

app/controllers/super_admins/stats_controller.rb
AlexRyanShep: As discussed, some parts of the page rely on these being in the order they are

app/controllers/super_admins/stats_controller.rb
AlexRyanShep: Used a modified version of this as discussed

app/controllers/super_admins/stats_controller.rb
AlexRyanShep: Used a modified version of this as discussed

app/services/stats/graphs/six_month_period.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      def last_six_months
+        output = []
+        6.times do |offset|
+          month = (@to - offset.month).strftime('%b')
+          output << month
+        end
+        output.reverse
+      end
```
jrmhaig, line 31: I wasn't aware this before but it turns out that `Date::MONTHNAMES` and `Date::ABBR_MONTHNAMES` provide arrays of month names in full form and short (3 letter) form.

This Stack Overflow question indicates how to use it for what we want here: https://stackoverflow.com/questions/38510701/get-last-x-month-names-in-rails/38510938

app/services/stats/graphs/six_month_period.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def empty_linechart_array
+        output = []
+        ordered_fee_schemes.each do |scheme|
+          output.append({ name: scheme, data: Hash.new(0) })
+          last_six_months.each do |month|
+            output[-1][:data][month] = 0
+          end
```
jrmhaig, line 39: If you have two arrays, e.g.;

```ruby
months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun']
zeroes = [0, 0, 0, 0, 0, 0] # Can be created with; Array.new(6) { 0 }
```

They can be made into a hash using `zip` and `to_h`:

```ruby
months
  .zip(zeroes)    # Pair up array elements => [['Jan', 0], ['Feb', 0], ['Mar', 0], ...]
  .to_h           # Convert pairs into key/value pairs => { 'Jan' => 0, 'Feb' => 0, 'Mar' => 0, ... }
```

With this you may be able to create the `output` array with just a `map`. Let me know if you want more pointers.

spec/services/stats/graphs/six_month_period_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+  describe '#call' do
+    subject(:data_hash) { graph_data.call }
+
+    let(:mon_name) do
+      output = []
+      6.times do |offset|
+        month = (Time.current.end_of_month - offset.month).strftime('%b')
+        output << month
+      end
+      output.reverse
```
jrmhaig, line 15: See above for a simpler way to get a list of months.
Also, it might be good to use `travel_to` to get to a specific date and then explicitly listing the months instead of calculating them. This will make it easier to have tests that are known to run the same way every time, as well as allowing the testing of edge cases as appropriate.

jrmhaig (commented):
Thanks. It looks good and has reduced the amount of code in the controller. I've added a few suggestions for improvements. We can get together to go into them more in detail if you want.

In general, if you find you have this pattern;

```ruby
output = ... # Set the initial value

array.each do |element|
  # Do something to output based on element
end

return output
```

there is a good chance that you can do it in a better way using methods from the [enumerable model.](https://ruby-doc.org/3.2.2/Enumerable.html) It might be worth exploring this module in L&D some time.

app/services/stats/graphs/six_month_period.rb
AlexRyanShep: Implemented to use something similar to the stack overflow response (though simplified a bit) during our pairing :-)

app/services/stats/graphs/six_month_period.rb
AlexRyanShep: Now using last_six_months.index_with { 0 }, which achieves the same thing

spec/services/stats/graphs/six_month_period_spec.rb
AlexRyanShep: Tests have been re-written to use a static date, and a date_array that's explicitly typed strings

mpw5 (approved):
🎉

PR #6191, opened 2023-10-27
Refactor determination javascript

jrmhaig:
#### What

Rewrite the javascript for calculating determinations to follow the [GovUK Javascript style guide.](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/js.md)

#### Ticket

N/A

#### Why

Some motivation is written in `docs/javascript_restructure.md`

In summary, the Javascript used in CCCD is written in inconsistent styles and have dependencies that are starting to make it difficult to keep the application up to date.

#### How

The determination calculation on the claim summary page for case workers was in `app/webpack/javascripts/modules/case_worker/claims/DeterminationCalculator.js`. The new version is now in `app/webpack/javascripts/modules/determination.js` and tests are in `app/webpack/javascripts/modules/determination_spec.mjs`.

The GovUK `govuk-frontend` module is enabled by adding the class `govuk-frontend-supported` to the `<body>`. This convention is for the new `Determination` class. The class is initialised in `app/webpack/packs/govuk-frontend.js` in the same way as `govuk-frontend` modules are initialised by `require('govuk-frontend').initAll()`.


app/webpack/javascripts/modules/determination.mjs
```diff
@@ -0,0 +1,99 @@
...
+      apply_vat: this.ajaxVat,
+      net_amount: parseFloat(netAmount)
+    })
+    const response = await fetch(this.vatUrl + '?' + params.toString())
+
+    return await response.json()
+  }
+
+  cleanNumber = (element) => {
+    element.value = element.value.replaceAll(/[^\d.]/g, '').replace(/^(\d*\.\d\d).*$/, '$1')
```
jrmhaig: There is a better way to do this by splitting the string on the first `.` to get the integer and mantissa parts and then remove all of the non-numeric characters from each part. As it is, a string like `12.3.4` does not get fixed.

AnnaC-C:
This PR needs a ticket/Epic in JIRA so we can revisit this to modernise our codebase

PR #6200, opened 2023-10-31
CTSKF-585 Fee Scheme Usage Test Fixes

AlexRyanShep:
What
Improvements to RSpec tests for Fee Scheme Usage Generator

Ticket
[CTSKF-585](https://dsdmoj.atlassian.net/browse/CTSKF-585)

Why
Previous tests blocked S28 work

How
Better tests using CSV searching methods

[CTSKF-585]: https://dsdmoj.atlassian.net/browse/CTSKF-585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/stats/fee_scheme_usage_generator_spec.rb
+++ b/spec/services/stats/fee_scheme_usage_generator_spec.rb
@@ -55,12 +55,10 @@ RSpec.describe Stats::FeeSchemeUsageGenerator do
         'AGFS 13',
         'AGFS 14',
         'AGFS 15',
-        'AGFS 16',
         'LGFS 9',
         'LGFS 10'
       ]
     end
-    let(:block_length) { fee_scheme_array.count + 1 }
 
     before do
       seed_case_types
@@ -100,7 +98,6 @@ RSpec.describe Stats::FeeSchemeUsageGenerator do
       create(:litigator_claim, :submitted, case_type: CaseType.where(name: 'Trial').first)
       create(:interim_claim, :interim_warrant_fee, :submitted, case_type: CaseType.where(name: 'Trial').first)
       create(:transfer_claim, :with_transfer_detail, :submitted)
-      create(:fee_scheme, name: 'AGFS', start_date: Time.current, version: 16)
 
       # Included in MI report, past claims
       travel_to(4.months.ago.beginning_of_day) do
@@ -125,7 +122,7 @@ RSpec.describe Stats::FeeSchemeUsageGenerator do
 
     context 'when generating the most recent month' do
       it 'returns rows containing the correct fee schemes' do
-        expect(csv['Fee scheme'][(block_length * 5)..-2]).to eq(fee_scheme_array)
+        expect(csv['Fee scheme'].uniq.compact).to eq(fee_scheme_array)
       end
 
       it 'returns the correct data for AGFS 9' do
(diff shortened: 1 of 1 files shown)


spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -59,10 +55,12 @@ def amend_claim(claim, case_type_id, case_type_name)
         'AGFS 13',
         'AGFS 14',
         'AGFS 15',
+        'AGFS 16',
         'LGFS 9',
         'LGFS 10'
       ]
     end
+    let(:block_length) { fee_scheme_array.count + 1 }
```
mpw5: It might be better to define this in the context block where it's being used (as it's only called in one place). It's good practice not to define too much at the top level unless they're really needed, otherwise you risk having things leak between tests. You could move it into the block `context 'when generating the most recent month' do`.

spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -116,35 +115,35 @@ def amend_claim(claim, case_type_id, case_type_name)
...
     context 'when generating the most recent month' do
       it 'returns rows containing the correct fee schemes' do
-        expect(csv['Fee scheme'][50...59]).to eq(fee_scheme_array)
-      end
-
-      it 'returns rows containing the correct total value of claims' do
-        expect(csv['Total value of claims'][50...60]).to eq([
-                                                              '380.0', '0', '0', '0', '0.0', '0', '0', '75.02', '0', nil
-                                                            ])
+        expect(csv['Fee scheme'][(block_length * 5)..-2]).to eq(fee_scheme_array)
```
mpw5: This is a bit opaque. It's not clear what `[(block_length * 5)..-2]` really means. I kind of get that it's finding the last block of fee scheme names. But it's not that easy to interpret without context.

Could you could do something like `expect(csv['Fee scheme'].uniq.compact).to eq(fee_scheme_array)`? It produces (I think) the same result and is a bit more idiomatic.

spec/services/stats/fee_scheme_usage_generator_spec.rb
```diff
@@ -59,10 +55,12 @@ def amend_claim(claim, case_type_id, case_type_name)
         'AGFS 13',
         'AGFS 14',
         'AGFS 15',
+        'AGFS 16',
```
mpw5: It might be better not to include the new fee scheme at this point. If you remove this line and line 103, you can treat this PR as a generic improvement to the tests. You could then add a new commit to your fee scheme PR to add fee scheme 16 to these tests. That way you don't need to create the dummy fee scheme on line 103, and all the changes relating to fee scheme 16 are encapsulated in the other PR.

mpw5 (changes requested):
I've left some suggested improvements. Happy to discuss.

spec/services/stats/fee_scheme_usage_generator_spec.rb
AlexRyanShep: Good idea, thanks Mark! That's much neater and still tests the same thing. I've changed it to this

spec/services/stats/fee_scheme_usage_generator_spec.rb
AlexRyanShep: I've gone with your suggestion of using .uniq.compact for that fee_scheme test instead which was the only place left using this, so I've removed this helper entirely.

spec/services/stats/fee_scheme_usage_generator_spec.rb
AlexRyanShep: Good point, all the other tests still work without AGFS 16, they just now can have it appear without breaking. I'll remove these bits, make the other changes and update the PR in a moment

mpw5 (approved):
👍

PR #6201, opened 2023-10-31
Claims by fee scheme

jrmhaig:
#### What

Add a way to get the list of all claims for a fee scheme.

#### Ticket

[Add data visualisation to CCCD Superadmin page](https://dsdmoj.atlassian.net/browse/CTSKF-589)

#### Why

This is to help #6171 which is adding a dashboard to the Super Admin section to display the number of claims for each fee scheme.

#### How

The fee scheme of a claim is not stored as references in the database. Instead, it is calculated based on the date of the claim. As a result, fetching a list of claims for a fee scheme involves iterating over all claims and looking up the fee scheme of each individually. The `FeeScheme#claims` method find the list as follows;

* Fetch the list of all representation order with dates within the range of the fee scheme. Pre-fetching is included for;
  * The defendants of these representation orders
  * The claims of these defendants
  * All defendants for these claims
  * All representation orders for these defendants
* This list of representation orders is converted into a unique list of claims
* The list of claims is filtered based on the date range for the representation order date and the main hearing date, as appropriate

### DIFF
--- a/app/models/fee_scheme.rb
+++ b/app/models/fee_scheme.rb
@@ -50,4 +50,8 @@ class FeeScheme < ApplicationRecord
   def agfs_scheme_15?
     agfs? && version.eql?(15)
   end
+
+  def claims
+    Claim::BaseClaim.all.select { |claim| claim.fee_scheme == self }
+  end
 end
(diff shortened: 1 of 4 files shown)


mpw5 (approved):
🏆

PR #6222, opened 2023-11-10
Update read me to include yarn and assets in the development set up

Katy600:
#### What
This ticket updates the read me to include installing Yarn, building Yarn and  running Asset Package Manager
#### Ticket

N/A

#### Why
Asset Package Manager involves processing and compressing the asset files so that they can be served more efficiently in a production environment. If it is not installed then the server is not able to run. 

Yarn is a package manager for JavaScript dependencies. Also needed as a part of building frontend assets. 

#### How

These extra set up instructions have been included in the readme. 

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -14,6 +14,14 @@
 
 ## Setting up development environment
 
+Install Postgres on your local machine. If you would like to use a different database, you may need to update the config/database.yml file
+
+- Install yarn
+
+```
+yarn install
+```
+
 - Install dependencies
 
 ```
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -14,6 +14,12 @@
 
 ## Setting up development environment
 
+- Prerequisite for development environment
+
+```
+yarn install
```
Obsiye, line 22: could add this here:

Install Postgres on your local machine. If you would like to use a different database, you may need to update the `config/database.yml` file.

docs/development.md
```diff
@@ -14,6 +14,12 @@
 
 ## Setting up development environment
 
+- Prerequisite for development environment
```
mpw5: You could possibly reword this to `Install yarn` to be more consistent with the other items on the page, which call out what the step actually is (`Install dependencies`, `Install node` etc). Not a blocker though.

mpw5 (approved):
Looks good. I've left a nitpicky comment and I notice that Abdi left some feedback too, but I don't think either of them are enough to block releasing this, so go ahead and merge if you don't want to fix them 🙂

PR #6246, opened 2023-11-17
Rubocop fixes for advocate_claim_spec

mpw5:
#### What

General refactor of `spec/models/claim/advocate_claim_spec.rb` that fixes (almost) all rubocop offences.

#### Why

To improve code quality.


spec/models/claim/advocate_claim_spec.rb
```diff
@@ -53,17 +53,20 @@
   end
 
   describe 'validate external user is advocate role' do
-    let(:claim) { build(:unpersisted_claim, :with_fixed_fee_case) }
+    subject { claim.external_user.is?(:advocate) }
```
jrmhaig, line 61: I presume that it is OK that the claim that is being tested is now created by a different factory.

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -926,36 +926,36 @@
...
-      [100, 200, 399, 2].each do |value|
+      [100, 200, 399, 399.99, 2].each do |value|
         claims << create(:submitted_claim, total: value)
       end
 
       claims
     end
 
-    let(:greater_than_400) do
+    let(:claims_with_total_greater_than_or_equal_to_400_pounds) do
```
jrmhaig, line 968: Not a blocker but a possible improvement. This could be done simpler using `each_with_object`.

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -87,36 +87,36 @@
...
-      @fft2 = create(:fixed_fee_type, :lgfs)
-    end
+    let(:basic_fee_type_schemes_nine_and_ten) { create(:basic_fee_type, roles: %w[agfs agfs_scheme_9 agfs_scheme_10]) }
+    let(:basic_fee_type_no_scheme) { create(:basic_fee_type) }
+    let(:basic_fee_type_scheme_nine) { create(:basic_fee_type, roles: %w[agfs agfs_scheme_9]) }
+    let(:basic_fee_type_scheme_ten) { create(:basic_fee_type, roles: %w[agfs agfs_scheme_10]) }
 
-    after(:all) do
-      clean_database
+    before do
```
jrmhaig, line 103: I much prefer not to have the `before(:all)` and `after(:all)`. However, I understand that this was done to speed up the tests. Have you checked how much this change has affected how long this file takes?

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -87,36 +87,36 @@
...
-      @mft1 = create(:misc_fee_type, :agfs_scheme_9)
-      @mft2 = create(:misc_fee_type, :lgfs)
-      @mft3 = create(:misc_fee_type, :agfs_scheme_10)
-      @fft1 = create(:fixed_fee_type)
-      @fft2 = create(:fixed_fee_type, :lgfs)
-    end
+    let(:basic_fee_type_schemes_nine_and_ten) { create(:basic_fee_type, roles: %w[agfs agfs_scheme_9 agfs_scheme_10]) }
+    let(:basic_fee_type_no_scheme) { create(:basic_fee_type) }
+    let(:basic_fee_type_scheme_nine) { create(:basic_fee_type, roles: %w[agfs agfs_scheme_9]) }
+    let(:basic_fee_type_scheme_ten) { create(:basic_fee_type, roles: %w[agfs agfs_scheme_10]) }
```
jrmhaig, line 101: I cannot see these claims being created before the tests are run so I'm a little suspicious about the tests if they are still passing. Am I missing something?

jrmhaig (commented):
A few comments. Some are not so important.

spec/models/claim/advocate_claim_spec.rb
mpw5: I think so. The `advocate_claim` and `unpersisted_claim` factories both create a claim with an `external_user` with the `advocate` role, which is all this test needs.

spec/models/claim/advocate_claim_spec.rb
mpw5: Good spot. I've pushed an extra commit to improve it (and also for `claims_with_total_under_400_pounds`, which I realised wasn't actually being used anywhere, so I've added an additional test for that scope too).

spec/models/claim/advocate_claim_spec.rb
mpw5: I've made that mistake before so I did check this as I was working on it. This change doesn't affect the time taken to run the tests - the affected tests still run in about ~1 second. But your comment prompted me to check the time for running the whole file after all these changes into account, and it is about 6 or 7 seconds than slower than running them against master (typically 44s vs 37s) 😢 

Having done some digging, it looks like [this](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/6246/commits/8a040867e589bab6caa60aae3448e4b49ad4d585) commit is slowing things down.

I suspect it's because it moves a lot of expectations out of single examples into multiple `it` blocks, which means  setups are being run more often. I'll see if I can find a way of speeding things up a bit, but other than introduce more `before(:all)` blocks (or reverting the changes) I'm not sure there's much we can do.

spec/models/claim/advocate_claim_spec.rb
mpw5: This is creating fee types rather than claims. It sets up a number of fee types which apply to various fee schemes. The tests below then call `eligible_basic_fee_types` on a `claim` (which is created just before those `let`s) for a given fee scheme and checks that the correct fee types are returned.

Originally the setup was all done in the single `before(:all)` block but I've changed it so that the four fee types which are explicitly checked for in the tests are created in `let` statements (with clearer names, eg `@bft1` is now `basic_fee_type_schemes_nine_and_ten`), while the six which aren't referenced are created in the `before` (arguably we could delete them all together - they just exist to prove that, for example, an LGFS fee type isn't included on an AGFS claim).

spec/models/claim/advocate_claim_spec.rb
jrmhaig: The solution, which would not be quick or easy, might be to refactor large parts of the test suite so that they are more targeted. I suspect that a lot of time is spent doing setup that is not required in specific tests.

In the short term, I would prefer tests that are understandable over tests that are quick. If the test suite does get too time consuming, however, we may need to revise this.

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -134,36 +141,37 @@
...
-        expect(service).to have_received(:call)
       end
+
+      it { expect(service).to have_received(:call) }
     end
   end
 
   describe '#eligible_advocate_categories' do
-    let(:categories) { double(:mocked_categories_result) }
+    let(:categories) { ['Junior alone', 'Leading junior', 'Led junior', 'QC'] }
```
Obsiye, line 169: i'm guessing QC is the same as KC in cccd? so doesn't matter here?

spec/models/claim/advocate_claim_spec.rb
```diff
@@ -729,26 +718,30 @@
...
       end
 
-      it 'updates total when expense/fee destroyed' do
-        expense = claim.expenses.first
-        fee = claim.fees.first
-        expense.destroy
-        fee.destroy
-        claim.reload
-        expect(claim.total).to eq(146.00)
+      context 'when removing fees' do
```
Obsiye: `when removing fees and expenses`. Is there a difference between fees and expenses?

Obsiye (approved):
LGTM! 🚀

spec/models/claim/advocate_claim_spec.rb
mpw5: It's not so much that QC and KC are the same, it's that the claim that's being created in this spec will pre-date the introduction of KC (so it's testing what the eligible advocate categories were a couple of years ago rather than now).

spec/models/claim/advocate_claim_spec.rb
mpw5: Not a big difference, but it can't hurt to be more explicit so I've updated so that the context accurately describes what the test is doing 👍

PR #6258, opened 2023-11-21
CTSKF-492 Remove line break from Miscellaneous fees page.

Katy600:
#### What
Remove the line break and increase the spacing on the advocate and the litigators miscellaneous fee page
#### Before (with the line break)

Screenshot 2023-11-21 at 12 14 20

#### After -  on the advocate miscellaneous fees page
Screenshot 2023-11-28 at 11 38 54

#### After - on the litigator miscellaneous fees page
Screenshot 2023-11-28 at 09 18 50

This also includes adding Brewfile.lock to gitignore but this is unrelated.

#### Ticket

[CCCD - Remove line breaks to avoid visually grouping the ‘add another fee’ button with the warrant fees section](https://dsdmoj.atlassian.net/browse/CTSKF-492)

#### Why
There is a line break and dividing line on the miscellaneous fees page that is not in the right place. It comes before the ‘add another fee’ button which visually groups the button with the warrant fees section below it. 

#### How

Delete it.

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ yarn-debug.log*
 
 /app/assets/builds/*
 !/app/assets/builds/.keep
+
+# ignore Brewfile changes
+Brewfile.lock.json
--- a/app/views/external_users/claims/interim_claim_info/_fields.html.haml
+++ b/app/views/external_users/claims/interim_claim_info/_fields.html.haml
@@ -1,5 +1,5 @@
 - if claim.requires_interim_claim_info?
-  #interim-claim-info
+  #interim-claim-info.interim-claim
     %h2.govuk-heading-m
       = t('.warrant_fee')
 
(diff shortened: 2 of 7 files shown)


mpw5 (commented):
The change looks good but I think there's a couple of additional changes needed. 

* There are actually two versions of the misc fees page, one for advocates (which you've changed) and one for litigators. The line break will need removing from that version too (you can find it at `app/views/external_users/claims/misc_fees/litigators/_misc_fee_fields.html.haml`)
* The [Figma](https://www.figma.com/file/ji0NwMxvla3H2x8XhXT4ZN/CLAIR-Taskforce-s.28-file?type=design&node-id=1136-5140&mode=design) design suggests that there should be some additional spacing between the `Add another fee` button and the `Warrant fees` section. You could have a go at doing that, maybe by using `govuk-spacing` from the [design system](https://design-system.service.gov.uk/styles/spacing/).

mpw5 (approved):

PR #6263, opened 2023-11-23
[CTSKF-268] Migrate scheduled tasks to Sidekiq Scheduler

jrmhaig:
#### What

Replace `scheduler-daemon` with `sidekiq-scheduler` for running background tasks.

#### Ticket

[CCCD: migrate scheduler daemon to sidekiq scheduler](https://dsdmoj.atlassian.net/browse/CTSKF-268)

#### Why

The `scheduler-daemon` gem is unsupported and preventing some other gems from being updated. It is now blocking an update of Ruby to version 3.2.

As we already use Sidekiq for background tasks, `sidekiq-scheduler` is a sensible choice for a replacement.

#### How

All the scheduled tasks are defined as classes in the `scheduled_tasks` folder. With some modifications, these are moved into the `lib/schedule` folder. The schedule is now defined in the Sidekiq configuration file, `config/sidekiq.yml` instead of in the task classes themselves.

The six tasks for creating the MI reports are almost identical. This duplication is required by `scheduler-daemon` so that the schedule can be defined correctly. With `sidekiq-scheduler` the same task can be used multiple times so these six classes are reduced to just one.

#### To do

- [x] Add `sidekiq-scheduler`
- [x] Configure `sidekiq-scheduler`
- [x] Migrate `agfs_management_information_generation_task.rb`
- [x] Migrate `agfs_management_information_v2_generation_task.rb`
- [x] Migrate `document_cleaner_task.rb`
- [x] Migrate `lgfs_management_information_generation_task.rb`
- [x] Migrate `lgfs_management_information_v2_generation_task.rb`
- [x] Migrate `management_information_generation_task.rb`
- [x] Migrate `management_information_v2_generation_task.rb`
- [x] Migrate `performance_platform_channel_task.rb`
- [x] Migrate `performance_platform_cost_per_transaction_task.rb`
- [x] Migrate `poll_injection_responses_task.rb`
- [x] Remove tasks that are no longer required


docs/development.md
```diff
@@ -113,16 +113,9 @@ bundle exec sidekiq
...
 
-To process scheduler daemon jobs in development run the scheduler in a terminal and set the relevant scheduler "task" to run regularly (see [scheduled_tasks](../scheduled_tasks) directory)
-
-see [schedular_daemon](https://github.com/ssoroka/scheduler_daemon) for details.
-
-```
-# run scheduler daemon in console mode
-bundle exec scheduler_daemon run
-```
+TODO
```
mpw5: Is it worth updating this section now while it's still fresh in our minds? Alternatively it might be worth writing a ticket so that we don't forget to document it.

config/sidekiq.yml
```diff
@@ -14,3 +14,35 @@ development:
 production:
   :verbose: true
   :concurrency: 5
+:scheduler:
+  :schedule:
+    poll_injection_responses:
+      cron: '0 * * * * *'
```
mpw5, line 20: More for my own understanding than anything else - I think I've only ever seen a cron schedule defined using five fields, but you're using six here. What's the purpose of the leading 0 on these?

mpw5 (approved):
A couple of comments but no blockers 🎉

config/sidekiq.yml
jrmhaig: This confused me at first so perhaps this can go in the documentation too.

Sidekiq Scheduler (more specifically, the [fugit](https://github.com/floraison/fugit) gem) allows for seconds to be specified in the crontab so this is indicating that the job runs at 0 seconds of every minute. I don't think it is required but all the examples have 6 values.

docs/development.md
jrmhaig: Thanks. I did intend to come back to this but forgot.

mpw5 (approved):
👍

PR #6267, opened 2023-11-24
[L&D] Randomise jasmine tests

mpw5:
#### What

CCCD contains a suite of jasmine tests which unit test the javascript code. These run in the same order each time, which hides some issues with the tests. This PR enables randomised running of the tests and fixes those issues.

#### Ticket

[CTSKF-445](https://dsdmoj.atlassian.net/browse/CTSKF-445)

#### Why

It is desirable for tests to run in a random order as each test should be isolated and independent of other tests. Running the tests in a random order causes tests to start failing, highlighting why this is a problem.

#### How

* Sets the `random` attribute in `spec/support/jasmine-browser.json` to `true`. 
* Fixes all failures that occur after enabling random running
* Fix a couple of other minor issues with tests (typos etc).

There are still a number of improvements that could be made to the jasmine tests (eg ensuring that all code is covered by tests; improving consistency between tests; removing jQuery; rewriting the tests in a more modern style as the javascript is modernised - see https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/6191 for example) but are out of scope for this PR.

[CTSKF-445]: https://dsdmoj.atlassian.net/browse/CTSKF-445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/javascripts/modules-sidebar_spec.js
+++ b/spec/javascripts/modules-sidebar_spec.js
@@ -232,10 +232,10 @@ describe('Modules.SideBar.js', function () {
         expect($el.find('.total-grandTotal')[0].innerHTML).toBe('£333.00')
       })
 
-      it('should call `sanitzeFeeToFloat`', function () {
-        spyOn(moj.Modules.SideBar, 'sanitzeFeeToFloat')
+      it('should call `sanitizeFeeToFloat`', function () {
+        spyOn(moj.Modules.SideBar, 'sanitizeFeeToFloat')
         moj.Modules.SideBar.render()
-        expect(moj.Modules.SideBar.sanitzeFeeToFloat).toHaveBeenCalled()
+        expect(moj.Modules.SideBar.sanitizeFeeToFloat).toHaveBeenCalled()
       })
     })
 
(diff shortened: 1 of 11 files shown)


spec/javascripts/modules-sidebar_spec.js
```diff
@@ -112,7 +114,7 @@ describe('Modules.SideBar.js', function () {
     })
 
     it('should have a `phantomBlockList` property defined', function () {
-      expect(moj.Modules.SideBar.phantomBlockList).toEqual(['fixedFees', 'gradFees', 'miscFees', 'warrantFees', 'interimFees', 'transferFees', 'disbursements', 'expenses'])
+      expect(moj.Modules.SideBar.phantomBlockList).toEqual(['fixedFees', 'gradFees', 'miscFees', 'warrantFees', 'interimFees', 'transferFees', 'hardshipFees', 'disbursements', 'expenses'])
```
Katy600, line 117: @mpw5 this is just me attempting to code review so please ignore if not important. Is it possible to change the setup so that you don't have to make this change twice? Or is that an unnecessary refactor and a faff?

Katy600 (approved):
I think it all makes sense to me! I really appreciate all the explanations on each commit. I just made one comment but mostly for me.

spec/javascripts/modules-sidebar_spec.js
mpw5: Thanks Katy. It's a good question. I guess it could be optimised slightly to remove the repetition and I think if that list of fees was being used in more places then I probably would pull it out into a variable so that we could re-use it, but as it's only being used once in the setup and once in the expectation it doesn't feel too bad.

PR #6303, opened 2023-12-04
Ctskf 535 reduce duplication in claim presenters

AlexRyanShep:
#### What
Remove some duplicated methods from presenter classes

#### Ticket

[CTSKF-535](https://dsdmoj.atlassian.net/browse/CTSKF-535)

#### Why
Guard against potential issues in the future if these methods were updated in base_claim_presenter, but then unintentionally overwritten in child classes by the duplicated methods. 

[CTSKF-535]: https://dsdmoj.atlassian.net/browse/CTSKF-535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/presenters/claim/advocate_claim_presenter.rb
+++ b/app/presenters/claim/advocate_claim_presenter.rb
@@ -56,14 +56,6 @@ class Claim::AdvocateClaimPresenter < Claim::BaseClaimPresenter
     raw_fixed_fees_total + raw_basic_fees_total + raw_misc_fees_total
   end
 
-  def misc_fees_vat
-    h.number_to_currency(raw_misc_fees_vat)
-  end
-
-  def misc_fees_gross
-    h.number_to_currency(raw_misc_fees_gross)
-  end
-
   def summary_sections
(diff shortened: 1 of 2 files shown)


mpw5 (approved):
LGTM.

I think there is more that could be done to reduce duplication across the presenter classes - for example there are methods that are duplicated between the `AdvocateClaimPresenter` and `AdvocateHardshipClaimPresenter` classes (eg `raw_basic_fees_total`, `raw_basic_fees_vat`, `raw_basic_fees_gross` etc) which could potentially be moved into the `BaseClaimPresenter`.

Similarly there are methods like `raw_warrant_fees_vat`, `raw_warrant_fees_gross` etc which are duplicated across `LitigatorClaimPresenter` and `AdvocateInterimClaimPresenter`.

But I think what you've got here is good in terms of the scope of the ticket, so it's probably wise to merge this and add another ticket to cover that work separately.

PR #6307, opened 2023-12-05
[CTSDK-268] Increase number of SQS messages to poll to 10

jrmhaig:
#### What

Increase the number of messages to pull from the SQS queue to 10.

#### Ticket

[CCCD: migrate scheduler daemon to sidekiq scheduler](https://dsdmoj.atlassian.net/browse/CTSKF-268)

#### Why

The number of messages on the SQS queue has started increasing after migrating the polling task from `scheduler-daemon` to `sidekiq-scheduler`. It appears that `scheduler-daemon` was running on all pods, and so 5 times in total, whereas `sidekiq` is only running on the worker pod. This would account for the messages getting ahead of the polling.

#### How

Set the `POLL_MESSAGE_COUNT` variable to 10 instead of 1.


mpw5 (approved):
👍

PR #6313, opened 2023-12-06
CTSKF-487 Migrate to puma

Obsiye:
#### What

Replace Unicorn web server with Puma

#### Ticket

[CTSKF-487](https://dsdmoj.atlassian.net/browse/CTSKF-487)

#### Why

Puma is more up to date, more performant, more secure and more frequently maintained. It's the standard choice for web servers in rails applications.

**After migrating to Puma the smoke-test-dev job failed with 2 issues:**
- The schema isn't loaded in the `smoke-test-dev` CircleCI job / `smoke-test-executor` executor
- The database migrations aren't finished running on the `smoke-test-executor` before the `smoke-test` CircleCI command is run.

**TLDR:** The CircleCI `smoke-test-executor` didn't complete running db migrations or puma server before `smoke-test` command was ran causing the `./runtests.sh` script to fail.

#### How

- Update the Gemfile to include Puma
- Update the docker entry point file to run the puma server
- Remove unicorn code

**Migrating to puma revealed a smoke-test-dev CircleCI job issue:**

It seems like there's a data race (thanks @jrmhaig for that great catch : D ). The smoke tests are failing to complete migrations before the `bundle exec rake api:smoke_test` which then causes it to fail with no error message just a kill signal to CircleCI's job.

We think the data race was because puma is much faster than unicorn. Unicorn's slowness gave time to allow the database to be properly migrated 🤞🏾 

**Solution:**

Before running the smoke tests recreate the docker/docker-entrypoint.sh commands for smoke-test-executor:
```
circleci config
      ......
      - run: bundle exec rails db:schema:load
      - run:
          name: Run app server
          command: bundle exec puma -p 3000
          background: true
      ......
```

- remove entrypoint for the smoke-test-executor

co-authors and pairing help:
@jrmhaig and @mpw5 

[CTSKF-487]: https://dsdmoj.atlassian.net/browse/CTSKF-487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config.ru
```diff
@@ -1,6 +1,6 @@
 # This file is used by Rack-based servers to start the application.
 
-if %w(production devunicorn).include?(ENV['RAILS_ENV']) && ENV['ENV']
+if %w(production).include?(ENV['RAILS_ENV']) && ENV['ENV']
```
mpw5: Can this whole block be removed? I think this was configured so that it would run against environment using `unicorn`, but as we have now removed `unicorn` I suspect it is not needed.

If we delete this we can probably also remove `lib/host_memory` and its tests.

Gemfile
```diff
@@ -62,13 +62,14 @@ gem 'laa-fee-calculator-client', '~> 1.4'
 gem 'active_storage_validations'
 gem 'faraday', '~> 1.10'
 gem 'faraday_middleware', '~> 1.2'
+gem 'puma'
 
-group :production, :devunicorn do
+group :production do
   gem 'unicorn-rails', '2.2.1'
   gem 'unicorn-worker-killer', '~> 0.4.5'
```
mpw5: Can these gems be removed now?

docker/docker-entrypoint.sh
```diff
@@ -19,6 +19,9 @@ else
   printf '\e[33mINFO: Using remote redis-server specified in REDIS_URL\e[0m\n'
 fi
 
-printf '\e[33mINFO: Launching unicorn\e[0m\n'
+printf '\e[33mINFO: Starting scheduler_daemon daemon\e[0m\n'
+bundle exec scheduler_daemon start
+
```
mpw5: `scheduler_daemon` has been removed now, you might need to rebase to pick up those changes, which should hopefully make these lines disappear.

docker/docker-entrypoint.sh
```diff
@@ -19,6 +19,9 @@ else
...
+printf '\e[33mINFO: Launching puma\e[0m\n'
```
mpw5, line 24: I think technically this line should go on line 26, so it's immediately before the command to start `puma`, but that's an existing thing so not a priority

mpw5 (commented):
Assuming there's no issues with running the smoke tests against the `cccd_test` environment, then I think I'm OK with your fix.

I've spotted a few other bits of tidying up that should probably done before releasing this though.

jrmhaig (approved):
Thanks for sorting this out. It would be good to get this in soon so that there is a bit of time to monitor it before everyone disappears for Christmas.

mpw5 (approved):
👍

PR #6317, opened 2023-12-07
Increase poll message count for production

jrmhaig:
#### What

Increase the number of messages taken from the SQS queue with each poll.

#### Ticket

N/A

#### Why

The `SETTINGS__AWS__POLL_MESSAGE_COUNT` setting determines how many messages are taken from the SQS queue for injection responses with each poll. Throughout the day this queue builds up, although never to an unmanageable level. 
##### How

By increasing the number of messages taken in production we should be able to silence the occasional alerts that the queue is getting too large.


mpw5 (approved):

mpw5:
Depending on how well this works, another option we have is to increase the threshold for the alert. I think I chose 10 as a fairly arbitrary limit, so if we're finding we legitimately have more than 10 in the queue at times then there's no reason we can't increase it.

PR #6330, opened 2023-12-11
[CTSKF-234] Modify install and update scripts in line with templates in Rails

jrmhaig:
See https://github.com/rails/rails/pull/34084

After CCCD was created the setup and update script templates in Rails were changed in line with [Style/MixinUsage](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MixinUsage) to avoid including modules in the global namespace. By bringing our scripts more into line with the templates there will be fewer conflicts when upgrading.

#### What

Small modifications to the setup and update scripts.

#### Ticket

[Upgrade CCCD app framework to Rails 7](https://dsdmoj.atlassian.net/browse/CTSKF-234)

#### Why

The Rails upgrade script, for moving between major versions, compares certain files against the Rails templates. Some differences result from changes made to the templates after CCCD was originally set up. By aligning our versions of the scripts more closely to the templates, the upgrading process can be simpler.

#### How

See changes to the Rails templates;

* https://github.com/rails/rails/pull/34084
* https://github.com/rails/rails/pull/29176
* https://github.com/rails/rails/pull/29303

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,9 +1,8 @@
 #!/usr/bin/env ruby
-require 'pathname'
 require 'fileutils'
 
 # path to your application root.
-APP_ROOT = Pathname.new File.expand_path('..', __dir__)
+APP_ROOT = File.expand_path('..', __dir__)
 
 def system!(*args)
   system(*args) || abort("\n== Command #{args} failed ==")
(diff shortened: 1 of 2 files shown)


mpw5 (approved):
👍

PR #6331, opened 2023-12-11
[CTSKF-591] Fix anonymised db dump

mpw5:
#### What

Fixes the anonymised database dump task

#### Ticket

[CTSKF-591](https://dsdmoj.atlassian.net/browse/CTSKF-591)

#### Why

Moving to short-lived AWS credentials has broken the anonymised database dump process as the job relies on long-lived credentials to access the S3 bucket used to store the dumps. This task is useful when debugging production issues and it is valuable to restore it to aid with resolving incidents.

#### How

* Moves to using service accounts to access s3 buckets by creating a new `dump.yaml` config file for each CCCD environment, which removes references to the old AWS S3 keys/secrets and instead instructs the job to use the relevant service account for authentication.

* This also necessitates changes to `.k8s/live/scripts/job.sh` as `dump.yaml` files are now stored in environment directories instead of the `jobs` directory.

#### Notes

* ideally we would continue using a single `dump.yml` file, with the `serviceAccountName` populated dynamically based on the environment where the dump job is run, however this is not possible with our current setup. This is something to consider if we move to `helm` to manage our kubernetes configuration.

* this fix slightly changes the way the dump/restore process is run. It is now not possible to run rake tasks locally that interact with the S3 bucket. This includes commands to list/copy/delete dumps. Instead a developer needs to log on to a kubernetes pod to run those commands. Documentation has been updated to reflect this. We might want to look at future improvements to this.

[CTSKF-591]: https://dsdmoj.atlassian.net/browse/CTSKF-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.k8s/live/api-sandbox/dump.yaml
+++ b/.k8s/live/api-sandbox/dump.yaml
@@ -1,0 +1,46 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: db-dump
+spec:
+  template:
+    metadata:
+      name: db-dump
+    spec:
+      serviceAccountName: cccd-api-sandbox-service
+      restartPolicy: OnFailure
+      BackoffLimit: 1
+      containers:
+        - name: cccd-job
+          imagePullPolicy: Always
+          image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-get-paid/cccd:set-me
+          securityContext:
+            capabilities:
+              drop:
+              - ALL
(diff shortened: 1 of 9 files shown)


Obsiye (approved):
LGTM

PR #6337, opened 2023-12-12
CTSKF-533 Re-enable Survey Monkey

AlexRyanShep:
#### What
Re-enable survey monkey feedback and update to new details
#### Ticket

[CTSKF-533](https://dsdmoj.atlassian.net/browse/CTSKF-533)

#### Why
Separate feedback out from Zendesk bug reports, re-align with User Research preferred feedback option

#### How

Some tweaks to allow easier switching between zendesk and survey monkey, update feature flags, update staging and production environment secrets with new bearer token

[CTSKF-533]: https://dsdmoj.atlassian.net/browse/CTSKF-533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


features/support/stub_feedback_requests.rb
```diff
@@ -2,30 +2,42 @@
...
-def successful_zendesk_body
-  { item: {} }.to_json
+def successful_feedback_body
+  if Settings.zendesk_feedback_enabled?
+    # Zendesk
+    { item: {} }.to_json
+  else
+    # Survey Monkey
+    { id: '123' }.to_json
+  end
```
jrmhaig: For the sake of clarity I would suggest going back to `successful_zendesk_body` (etc) and creating a new `successful_survey_monkey_body` function. Whilst what you have here is good for production code, it is better to have clarity rather than efficiency in test code.

Potentially there could be a new test for changing the `zendesk_feedback_enabled` setting.

AlexRyanShep: My latest commit should solve this hopefully. I've taken the logic out of these bits and added extra tests/an extra step so that we now have tests for work SurveyMonkey and Zendesk sides, and are specifying the zendesk_feedback_enabled? setting in the test rather than using the one actually set for the app.

jrmhaig (approved):

PR #6338, opened 2023-12-12
(WIP) Use govuk-components gem

jrmhaig:
#### Ticket

#### Why

One of the issues with upgrading to Rails 7 is that the loading of modules in `lib/govuk_component/railties.rb` was deprecated in 6.1 and now no longer works. Instead of attempting to rewrite this, I am first trying to remove these custom helpers completely and replace with the `govuk-components` gem.

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


jrmhaig:
Closing for now.

PR #6392, opened 2024-01-03
Ctskf 632 virus scan

jrmhaig:
#### Ticket

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

### DIFF
--- a/config/initializers/clamby.rb
+++ b/config/initializers/clamby.rb
@@ -1,0 +1,16 @@
+Clamby.configure({
+  # check: false,
+  # daemonize: true,
+  config_file: Rails.root.join('config', 'clamd.container.conf'),
+  # error_clamscan_missing: true,
+  # error_clamscan_client_error: false,
+  # error_file_missing: true,
+  # error_file_virus: false,
+  # fdpass: false,
+  stream: true,
+  # reload: false,
+  # output_level: 'medium',
+  # executable_path_clamscan: 'clamscan',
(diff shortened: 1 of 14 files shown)


config/initializers/clamby.rb
```diff
@@ -0,0 +1,16 @@
...
+  # error_clamscan_client_error: false,
+  # error_file_missing: true,
+  # error_file_virus: false,
+  # fdpass: false,
+  stream: true,
+  # reload: false,
+  # output_level: 'medium',
+  # executable_path_clamscan: 'clamscan',
+  # executable_path_clamdscan: 'clamdscan',
+  # executable_path_freshclam: 'freshclam',
```
mpw5, line 15: Would we want to enable any of these options when we implement this? I notice on Paul's [PR](https://github.com/ministryofjustice/laa-claim-non-standard-magistrate-fee-backend/commit/89a00f72ae515be56c29fa37606d145b0807dbdc#diff-875aa3f8c8f46ba5b72b86a3b131bd83faee982a0c5d6d2d2f3b2319c9781fb6) that several of them are set to true. Not necessarily something for now, it can probably be decided on when we do the implementation.

jrmhaig: I copied this from the [README of clamby](https://github.com/kobaltz/clamby?tab=readme-ov-file#configuration) and I believe that they are all the defaults except the two that I have changed. I left them in for reference in case we decide there are any we need to change.

PR #2649, opened 2019-02-13
Supplementary fees claim type

jsugarman:
#### What
Allow Advocates to claim stand-alone/supplementary fees.

*As an advocate provider* I want to be able to claim special prep, wasted prep, deferred sentence hearing (plus uplift), Confiscation hearings (half day AND whole day, plus uplifts), on its own
*So* that I can bill a valid claim without erroneously claiming a final/graduated fee

#### Why
This is because such bills can validly be claimed before or after the final claim is submitted (i.e. during a case or after it has concluded). Previously, advocates were using a workaround whereby they would submit a final claim with 1p basic fee plus the actual misc fee they were trying to claim.

#### Ticket

[CBO-566 - implement AGFS supplementary claim](https://dsdmoj.atlassian.net/browse/CBO-566)

#### How
Add an entirely new claim type with its own flows, validation, api endpoints.

--------

#### TODO (wip)
- [x] model (change name)
- [x] spec model
- [x] factory for model
- [x] validators
- [x] spec validators
- [x] presenters
- [x] spec presenters
- [x] Fee calc
- [x] external user views 
- [x] case worker flow
  - [x] allocation 
  - [x] accordion 
  - [x] assessment
  - [x] messaging
- [x] CYC page (external user and case worker)
- [x] claim Summary accordion  case workers
- [x] FE changes ** checkboxes
    - [x] cleaners
    - [x] spec cleaners
- [x] misc fee checkboxes
- [x] remove case type
- [x] remove offence
- [ ] API - moved to separate ticket
- [ ] DI - moved to separate ticket

#### Separate tickets
- [x] standard appearance fee (uplift is a misc fee but its "parent" is not. need to understand how to handle separately) 
- [x] centralise claim type list (mixin) for provider, base claim, clam csv presenter, context mapper??, claim type
- [x] centralise provider delegation

#### Refactors
- [x] refactor page headings (different styling of translation keys all over the shop)

PR #6402, opened 2024-01-05
Remove bury method

jrmhaig:
#### What

Remove `Extension::HashExtensions.bury`.

#### Ticket

N/A

#### Why

`Extensions::HashExtensions.bury` was added in #2327 and its only use was removed in #2649.


mpw5 (approved):
Nice! ⭐

PR #6418, opened 2024-01-09
Set up Sentry Crons to monitor the uptime and performance of scheduled jobs in sidekick

Katy600:
What
To test on staging only. Follow documentations setup in [Sentry- Set Up Crons ](https://docs.sentry.io/platforms/ruby/guides/sidekiq/crons/l)as one approach to monitoring scheduled tasks which run in the background.

Ticket
[CCCD: Investigate monitoring/alerting on scheduled tasks](https://dsdmoj.atlassian.net/browse/CTSKF-410)

Why
A recent [incident](https://dsdmoj.atlassian.net/browse/CTSKF-404) caused these to silently fail. It might be useful to be alerted to such failures automatically

How
By implementing Sentry Crons

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -7,5 +7,31 @@ if Rails.env.eql?('production') && ENV['SENTRY_DSN'].present?
 
     # Send 5% of transactions for performance monitoring
     config.traces_sample_rate = 0.05
-  end
+    config.enabled_patches += [:sidekiq_cron]
+   end
+
+   # Create a config from a crontab schedule (every 10 minutes)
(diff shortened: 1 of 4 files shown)


config/initializers/sentry.rb
```diff
@@ -7,5 +7,31 @@
 
     # Send 5% of transactions for performance monitoring
     config.traces_sample_rate = 0.05
-  end
+    config.enabled_patches += [:sidekiq_cron]
```
jrmhaig, line 10: We are using `sidekiq-scheduler` so this needs to be;
```suggestion
    config.enabled_patches += [:sidekiq_scheduler]
```

PR #6512, opened 2024-02-08
Ensure users' email addresses are unique in tests

jrmhaig:
#### What

Create email addresses in test users with unique email addresses.

#### Ticket

N/A

#### Why

It is possible for `Faker::Internet.email` to generate the same email address twice in a test and this can cause the test to fail. Adding `unique` ensures that this does not happen.

#### How

Use the `unique` feature of the Faker gem.


mpw5 (approved):
🎉

PR #6520, opened 2024-02-12
CTSKF-681 Remove govuk-frontend toolkit

Obsiye:
#### What

Remove GovUK Frontend Toolkit

#### Ticket

[CTSKF-675](https://dsdmoj.atlassian.net/browse/CTSKF-675)

#### Why

`govuk_frontend_toolkit` has been deprecated and it is no longer being maintained. We are directed to using `govuk-frontend` instead.

#### How

- Remove `govuk_frontend_toolkit`
- Remove `govuk-elements-sass`
- Replace govuk_frontend_toolkit scss variables with `govuk-frontend` scss alternatives
- Update offence `'Select and continue'` anchor tag to use `govuk_link_button` because the anchor tag styling is that of a link. To keep things unchanged, I used the existing `govuk_link_button` to style it as a button.
  - Also, retain the class names `'button offence-item-button set-selection'` in order to not break the JavaScript.
- Use `govuk-breadcrumbs` html to fix sub-navigation styling
   - Fix the sub navigation styling as it uses a more than symbol `>` for breadcrumbs when it should be disabled and a `'|'` like separator should be used. (border-left css)

**TODO:**

- [x] Fix issue with error message notification banner (especially on caseworker page)
- [x] Testing of error summary list

[CTSKF-675]: https://dsdmoj.atlassian.net/browse/CTSKF-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/webpack/stylesheets/cccd-measurements.scss
```diff
@@ -1,2 +1,8 @@
 // CCCD measurements
-$gutter-one-sixth: $gutter / 6;
+
+$gutter-one-sixth: $govuk-gutter / 6;
+
+$gutter-one-quarter: $govuk-gutter / 4;
+$gutter-one-third: $govuk-gutter / 3;
+$gutter-half: $govuk-gutter / 2;
+$gutter-two-thirds: $govuk-gutter - $gutter-one-third;
```
jrmhaig, line 8: These variables come from `govuk_frontend_toolkit` so ideally they should be removed. This is possibly beyond the scope of this work. Could a new ticket be written?

jrmhaig (approved):
:tada: This PR results in better alignment with the design system, such as;

* [Use of the GDS Transport font](https://design-system.service.gov.uk/styles/typography/#font)
* [Correct colour palette](https://design-system.service.gov.uk/styles/colour/)

app/webpack/stylesheets/cccd-measurements.scss
Obsiye: Ah `$gutter` comes from `govuk_frontend_toolkit` but `$govuk-gutter` comes from `govuk-frontend` npm package.

It's just that the proportions like *-one-quarter and *-two-thirds doesn't exist in `govuk-frontend` and is customised via cccd-measurements.scss file.

I think these values are being used to customise portions of cccd.

PR #6549, opened 2024-02-20
CTSKF-710 feedback model refactor

AlexRyanShep:
#### What
Refactor the Feedback model/controller, ZendeskSender and SurveyMonkey sender to remove responsibility for error handling and decision of what sending service to use from the model. 

#### Ticket

[CTSKF-710](https://dsdmoj.atlassian.net/browse/CTSKF-710)

#### Why

- Make the Feedback model conform to SOLID principles and follow the model > services design pattern. 
- Simplify the process for adding new sending services and/or switching between zendesk and surveymonkey for feedback. 

#### How

Model, Controller and Service files refactored to move logic around and use dependency inversion to provide the sender class for each piece of feedback.

[CTSKF-710]: https://dsdmoj.atlassian.net/browse/CTSKF-710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -19,11 +19,11 @@ class Feedback
   validates :type, inclusion: { in: FEEDBACK_TYPES.keys.map(&:to_s) }
   validates :event, :outcome, presence: true, if: :bug_report?
 
-  def initialize(sender = nil, attributes = {})
+  def initialize(attributes = {})
     attributes.each do |key, value|
       instance_variable_set(:"@#{key}", value)
     end
-    @sender = SENDING_SERVICES.include?(sender) ? sender : nil
+    @sender = SENDING_SERVICES.include?(@sender) ? @sender : nil
 
     @reason.compact_blank! if @reason.present?
   end
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -45,7 +45,7 @@ class Feedback
 
     resp = @sender.call(self)
     @response_message = resp[:response_message]
-    resp[:success?]
+    resp[:success]
   end
 
   def subject
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -41,7 +41,7 @@ class Feedback
   end
 
   def save
-    return unless valid? || @sender.nil?
+    return unless valid? && !@sender.nil?
 
     resp = @sender.call(self)
     @response_message = resp[:response_message]
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -14,10 +14,11 @@ class Feedback
   validates :type, inclusion: { in: FEEDBACK_TYPES.keys.map(&:to_s) }
   validates :event, :outcome, presence: true, if: :bug_report?
 
-  def initialize(attributes = {})
+  def initialize(attributes = {}, sender = nil)
     attributes.each do |key, value|
       instance_variable_set(:"@#{key}", value)
     end
+    @sender = sender
 
     @reason.compact_blank! if @reason.present?
   end
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -7,6 +7,11 @@ class Feedback
     bug_report: %i[case_number event outcome email]
   }.freeze
 
+  SENDING_SERVICES = [
+    SurveyMonkeySender,
+    ZendeskSender
+  ].freeze
+
   attr_accessor :email, :referrer, :user_agent, :type,
                 :event, :outcome, :case_number,
                 :task, :rating, :comment, :reason, :other_reason, :response_message
@@ -14,11 +19,11 @@ class Feedback
   validates :type, inclusion: { in: FEEDBACK_TYPES.keys.map(&:to_s) }
   validates :event, :outcome, presence: true, if: :bug_report?
 
-  def initialize(attributes = {}, sender = nil)
+  def initialize(sender = nil, attributes = {})
     attributes.each do |key, value|
       instance_variable_set(:"@#{key}", value)
     end
-    @sender = sender
+    @sender = SENDING_SERVICES.include?(sender) ? sender : nil
 
     @reason.compact_blank! if @reason.present?
   end
@@ -36,9 +41,11 @@ class Feedback
   end
 
   def save
-    return unless valid?
-    return send_to_survey_monkey if feedback? && !Settings.zendesk_feedback_enabled?
-    send_to_zendesk
+    return unless valid? || @sender.nil?
+
+    resp = @sender.call(self)
+    @response_message = resp[:response_message]
+    resp[:success?]
   end
 
   def subject
@@ -56,28 +63,6 @@ class Feedback
   end
 
   private
-
-  def send_to_survey_monkey
-    response = SurveyMonkeySender.call(self)
-    @response_message = if response[:success]
-                          'Feedback submitted'
-                        else
-                          "Unable to submit feedback [#{response[:error_code]}]"
-                        end
-    response[:success]
-  end
-
-  def send_to_zendesk
-    feedback_type = type.humanize
-    ZendeskSender.send!(self)
-    @response_message = "#{feedback_type.titleize} submitted"
-  rescue ZendeskAPI::Error::ClientError => e
-    @response_message = "Unable to submit #{feedback_type.downcase}"
-    LogStuff.error(class: self.class.name, action: 'save', error_class: e.class.name, error: e.to_s) do
-      "#{feedback_type.titleize} submisson failed!"
-    end
-    false
-  end
 
   def feedback_type_attributes
     FEEDBACK_TYPES[type.to_sym]
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -2,38 +2,19 @@ class Feedback
   include ActiveModel::Model
   include ActiveModel::Validations
 
-  FEEDBACK_TYPES = {
-    feedback: %i[task rating comment reason other_reason],
-    bug_report: %i[case_number event outcome email]
-  }.freeze
-
-  SENDING_SERVICES = [
-    SurveyMonkeySender,
-    ZendeskSender
-  ].freeze
-
   attr_accessor :email, :referrer, :user_agent, :type,
                 :event, :outcome, :case_number,
                 :task, :rating, :comment, :reason, :other_reason, :response_message
 
-  validates :type, inclusion: { in: FEEDBACK_TYPES.keys.map(&:to_s) }
-  validates :event, :outcome, presence: true, if: :bug_report?
+  validates :type, inclusion: { in: %w[feedback bug_report] }
+  validates :event, :outcome, presence: true, if: -> { is?(:bug_report) }
 
   def initialize(attributes = {})
     attributes.each do |key, value|
       instance_variable_set(:"@#{key}", value)
     end
-    @sender = SENDING_SERVICES.include?(@sender) ? @sender : nil
 
     @reason.compact_blank! if @reason.present?
-  end
-
-  def feedback?
-    is?(:feedback)
-  end
-
-  def bug_report?
-    is?(:bug_report)
   end
 
   def is?(type)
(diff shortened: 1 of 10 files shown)


spec/requests/feedback_spec.rb
```diff
@@ -45,6 +45,32 @@
       }
     end
 
+    context 'when feedback_controller is determining the sender class to use' do
```
Katy600: Do you need to have this context? Could you just have - `when Zendesk feedback enabled` 'it uses the Zendesk sender'. 

I just found this in a google search `_“describe” is to wrap a set of tests against one functionality while “context” is to wrap a set of tests against one functionality under the same state. `_ 
Perhaps this could be a `describe sender method' with your two different contexts inside?

spec/requests/feedback_spec.rb
```diff
@@ -45,6 +45,32 @@
...
+      context 'with Zendesk feedback enabled' do
+        before do
+          allow(Settings).to receive(:zendesk_feedback_enabled?).and_return(true)
+          allow(Feedback).to receive(:new).and_call_original
+          post_feedback
+        end
+
+        it 'Uses the Zendesk sender' do
+          expect(Feedback).to have_received(:new).with(anything, ZendeskSender)
+        end
```
Katy600: This is mostly for my own understanding as am not confident in how best to test. Is it possible to create 2 factorybot feedback users - one who has enabled zendesk feedback and one who has not enabled it and then test their interaction with the service? Or is this unnecessary and not actually possible? By testing just the Feedback class is it more like a controller test than a request test?

app/models/feedback.rb
```diff
@@ -7,18 +7,23 @@ class Feedback
...
+
   attr_accessor :email, :referrer, :user_agent, :type,
                 :event, :outcome, :case_number,
                 :task, :rating, :comment, :reason, :other_reason, :response_message
 
   validates :type, inclusion: { in: FEEDBACK_TYPES.keys.map(&:to_s) }
   validates :event, :outcome, presence: true, if: :bug_report?
 
-  def initialize(attributes = {}, sender = nil)
+  def initialize(sender = nil, attributes = {})
```
Katy600: Is it possible to send a hash as an argument?

`def initialize(sender: nil, attributes: {})`
And then it doesn't matter about the order. Or perhaps not necessary.

Katy600 (commented):
Hi @AlexRyanShep, I enjoyed reading your code as it felt really clear. Obviously take all my comments lightly as I am just trying to improve my code reviews as well as how I read code. And please correct any misconception I have.

spec/requests/feedback_spec.rb
AlexRyanShep: I don't think that creating two different users would work here, the toggle between Zendesk and SurveyMonkey for Feedback is set at the application level in the environment variables in `.k8s/live/production/app-config.yaml`, not at the user level. That's why these tests have the in the before block, `allow(Settings).to receive(:zendesk_feedback_enabled?).and_return(true)` (or false, depending on the test)

spec/requests/feedback_spec.rb
AlexRyanShep: Makes sense to me 👍  I'll do some re-writing to remove the unnecessary contexts

app/models/feedback.rb
AlexRyanShep: Agh, can't believe that I didn't see that! The attributes one is even already a hash, I'd just need to do some additional steps in the Feedback initialize to make sure the validation is being done. I'll work that into a new commit and push it up.

spec/services/zendesk_sender_spec.rb
```diff
@@ -33,14 +34,17 @@
...
-      expect(ZendeskAPI::Ticket).to have_received(:create!).with(ZENDESK_CLIENT, hash_including(bug_report_zendesk_payload))
+      expect(ZendeskAPI::Ticket)
+        .to have_received(:create!)
+        .with(ZENDESK_CLIENT, hash_including(bug_report_zendesk_payload))
     end
 
-    context 'with a reporter email' do
-      let(:ticket_payload) do
+    context 'with a valid ticket and reporter email' do
+      let(:bug_report_ticket_payload) do
```
Katy600: Is this the same bug_report_ticket_payload as the one above? Are you able to reuse it somehow?

AlexRyanShep:
> Hi @AlexRyanShep, I enjoyed reading your code as it felt really clear. Obviously take all my comments lightly as I am just trying to improve my code reviews as well as how I read code. And please correct any misconception I have.

Thanks @Katy600! Good spot with the Hash thing, I got too caught up in the original way that I solved that I think to recognise there was a simpler one!

I've just pushed a new commit that addresses the points you made :-)

app/services/zendesk_sender.rb
```diff
@@ -1,17 +1,28 @@
...
-    def send!(ticket_payload)
-      ZendeskSender.new(ticket_payload).send!
-    end
+  def self.call(ticket_payload)
+    new(ticket_payload).call
   end
 
   def initialize(ticket_payload)
     @ticket_payload = ticket_payload
+    @return_hash = {}
```
Katy600: Does '@return_hash' need to be an instance variable? Could it just be a local variable inside the call method? And then it could just be passed to catch_error. I am asking this to check my understanding of when to use an instance variable as I am not entirely sure myself.

spec/requests/feedback_spec.rb
AlexRyanShep: This has been tweaked in the new commit

app/models/feedback.rb
AlexRyanShep: New commit now submits :sender as a part of the existing attributes hash rather than on its own. Validation only required slight tweaking, as a non-existent :sender attribute is nil, which can still be checked against the list of acceptable senders

spec/services/zendesk_sender_spec.rb
AlexRyanShep: This is code that was already in this test file that I've just tweaked. The difference between them is that the first instance of the payload has `nil` for the email address, while this one has it populated. I assume this is to make sure that it can run with and without an email without raising an error.

@jrmhaig git blame shows you as the original author of most of this file, am I understanding the purpose of that right?

app/services/zendesk_sender.rb
AlexRyanShep: My logic was just that having it be an instance variable is a more elegant solution than having to have it be passed to the catch_error, and then catch_error would need to return it back so that call could return it.

That said, looking at it again has made me think of what I think is a nicer way of doing it. I've amended it onto the most recent commit, what do you think?

spec/requests/feedback_spec.rb
Katy600: Oh I see, thanks Alex, that makes sense.

spec/requests/feedback_spec.rb
Katy600: Oh great!

spec/services/zendesk_sender_spec.rb
Katy600: Oh I see, that makes sense. Thanks Alex. I always forget about git blame.

app/services/zendesk_sender.rb
Katy600: Oh I like that @AlexRyanShep. I would be interested in other peoples thoughts but it feels more readable having it as a whole hash inside the method - where you can see all the values at once. And more encapsulated perhaps!?!?

app/models/feedback.rb
```diff
@@ -18,6 +23,7 @@ def initialize(attributes = {})
     attributes.each do |key, value|
       instance_variable_set(:"@#{key}", value)
     end
+    @sender = SENDING_SERVICES.include?(@sender) ? @sender : nil
```
jrmhaig: The open-closed principle means that the functionality of the class can be extended without needing to modify the class itself. Therefore, you don't need (or want) something here to restrict the sending services that can be used. The only requirement, in this case, is that the sender has a `call` method that accepts an instance of this class (as on line 46).

In the feedback controller you are already correctly ensuring that only valid senders are allowed based on the `type` parameter.

app/services/zendesk_sender.rb
```diff
@@ -1,17 +1,24 @@
 class ZendeskSender
   attr_accessor :ticket_payload
 
-  class << self
-    def send!(ticket_payload)
-      ZendeskSender.new(ticket_payload).send!
-    end
+  def self.call(ticket_payload)
+    new(ticket_payload).call
```
jrmhaig: As of Ruby 2.7 the following shorthand syntax works when the argument list of a method is only passed on verbatim;

```suggestion
  def self.call(...)
    new(...).call
```

This has the advantage that it will still work if the argument list for `new` changes.

app/services/survey_monkey_sender.rb
```diff
@@ -9,7 +9,18 @@ def initialize(feedback)
...
+
+    return_hash[:success] = resp[:success]
+
+    return_hash[:response_message] = if resp[:success]
+                                       'Feedback submitted'
+                                     else
+                                       "Unable to submit feedback [#{resp[:error_code]}]"
+                                     end
+
+    return_hash
```
jrmhaig: It might be cleaner (matter of opinion) to separate things out so that the `call` method just contains the returned hash:

```suggestion
    {
      success: success?,
      response_message: message
    }
  end
  
  private
  def submission_response
    @submission_response ||= response.submit
  end
  
  def success?
    @success ||= submission_response[:success]
  end
  
  def message
    return 'Feedback submitted' if success?
    
    "Unable to submit feedback [#{error_code}]"
  end
  
  def error_code
    @error_code ||= submission_response['error_code']
  end
```

By using memoized attributes (with `||=`) we can make sure that the submission is only sent once and the same return values are used each time a value is used.

jrmhaig (commented):
Thanks, @AlexRyanShep. This looks great. I have made a few comments and I'd be happy to discuss them if you want.

Ideally it would be good if the `type` parameter were not required by feedback model at all. This would require some more refactoring so it is not necessary for this PR, unless you think it can be done simply. Some things to think about regarding this;

* The list of fields stored in the `FEEDBACK_TYPES` array are proper to the sender so instead of `FEEDBACK_TYPES[type.to_sym]` you could have `@sender.feedback_fields`.
* Following the above point, if the sender classes have the list of fields that are applicable then, in the case of Zendesk, the `subject`, `description` and `reporter_email` methods can also go in the sender class.

app/services/survey_monkey_sender.rb
AlexRyanShep: Looks good to me, thanks!

app/services/zendesk_sender.rb
AlexRyanShep: This is handy, thanks!

app/models/feedback.rb
AlexRyanShep: Implemented this recommendation during our call

AlexRyanShep:
@jrmhaig my latest commit implements most of the recommendations that you made, including moving the #subject, #description etc methods into the ZendeskSender class, I managed to figure out what was stopping them working shortly after you had to leave.

The only remaining thing I've not done is removal of the Type parameter from the Feedback model. I've looked more into how this is used and they're very closely linked. Type is not only used to determine which sender is used, but it is also used to determine which of the feedback templates are rendered when the page is loaded (within whatever haml file generates the link,  `new_feedback_path(type: 'bug_report')` for example).

The only way I can think of right now to decouple Feedback from the type parameter would be to create a new BugReport class that inherits from Feedback, but then that makes me start thinking about problems with duplication of code  in the controller for that new class and so on. Let's discuss tomorrow, but that last little bit there is feeling like more of a rabbit hole than it might be worth to me .

jrmhaig (approved):
If you haven't done already it would be good to put this on staging and try submitting some bug reports and feedback. Otherwise it looks good to go.

AlexRyanShep:
> If you haven't done already it would be good to put this on staging and try submitting some bug reports and feedback. Otherwise it looks good to go.

Thanks Joe! I've done a bit as I've gone, but I'll push it into those environments and do a last round of testing before I merge

AlexRyanShep:
I've done some testing, staging is sending feedback and bug reports through to SurveyMonkey and Zendesk fine. Merging into master now

PR #6562, opened 2024-02-23
[CTSKF-649] Prepare for Rails 7 upgrade

jrmhaig:
#### What

* Refactor some tests
* Add two attributes to the base claim model

#### Ticket

[CCCD - Investigate Claim::BaseClaim#force_validation](https://dsdmoj.atlassian.net/browse/CTSKF-649)

#### Why

Upgrading to Rails results in some errors due to internal changes.

* Some test objects in test objects now need to be reloaded between the test setup and the test themselves
* On the determination section of the summary page the `refuse_reason_text` and `reject_reason_text` fields cannot be added to the form using the form helpers as the attribute does not exist on the claim

#### How

* Add `claim.reload` to the end of the setup stage of some tests
* Create new requests specs to tests functionality that is currently covered by validation specs that will need to be set to pending
* Add `refused_reason_text` and `reject_reason_text` as attributes on `Claim::BaseClaim`


spec/requests/external_users/advocates/claim_journeys_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+              'representation_order_date(2i)' => '3',
+              'representation_order_date(1i)' => '2023',
+              'maat_reference' => '9876542'
+            }
+          }
+        end
+
+        it do
+          submit_defendants_update
+          expect(response).to be_successful # ???
```
mpw5: Can the ??? be removed?

mpw5 (approved):
Nice work 👍 

One little comment about a comment.

spec/requests/external_users/advocates/claim_journeys_spec.rb
jrmhaig: Probably. I had forgotten about that. I'm still not convinced that an error should come back with a 'successful' status.

PR #3651, opened 2020-12-21
Remove disc evidence coversheet

kmahern:
#### What
Remove wicked gem and disc_evidence_coversheet from CCCD codebase

#### Ticket
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=222&projectKey=CBO&modal=detail&selectedIssue=CBO-1628

#### Why
This code is redundant since Secure File Exchange

#### How
Removed gem and all disc evidence coversheet mvc code.
--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2

PR #6563, opened 2024-02-23
Remove GovUk::DateAccessor

jrmhaig:
#### What

Remove the `GovUk::DateAccessor` class.

#### Ticket

N/A

#### Why

The only use place where this class used was removed in #3651

#### How

Remove;

* `lib/gov_uk/date_accessor.rb`
* `spec/lib/gov_uk/date_accessor_spec.rb`


mpw5 (approved):
🥇

PR #6569, opened 2024-02-26
[CTSKF-628] Update govuk framework to v5

jrmhaig:
#### What

Update to use version 5 of the GovUK framework

#### Ticket

[CCCD - Upgrade govuk-frontend gem to v5](https://dsdmoj.atlassian.net/browse/CTSKF-628)

#### Why

To keep up to date with the official GovUK design standards.

#### How

* Update govuk_design_system_formbuilder gem from 4.1 to 5.2.
* Update govuk-frontend npm package from 4.7 to 5.2.
* Change import path from govuk-frontend/govuk to govuk-frontend/dist/govuk.
* Update favicons and app icons.
* Add the `govuk-frontend-supported` class to the body tag. This allows the govuk-frontend javascript to connect to the correct elements.
* Use `import { initAll } from 'govuk-frontend'` in `app/webpack/packs/govuk-frontend.js`. This file still includes a `require.context` - does this need to be changed?
* Change of the logo SVG in `app/views/layouts/_govuk_logo.html.haml` and remove the fallback (used for IE 8). The `GOV.UK` text is now included in the image.
* Update step definitions for Cucumber tests to allow for the change of formatting of tags (see below).
* Remove metatag for `X-UA-Compatible`, which was used for IE support.
* Replace deprecated `govuk-warning-text__assistive` class with `govuk-visually-hidden`.
* Remove deprecated `govuk-button--disabled` class.

This is deployed on dev-lgfs

##### The style for tags has changed from block-capitals to normal. CCCD uses tags for claim status.

| Before | After
|---|---|
| Screenshot 2024-02-28 at 10 06 39 | Screenshot 2024-02-28 at 10 06 53 |

##### Assets path

An extra path has been added to the assets pipeline in `config/initializers/assets.rb` for `node_modules/govuk-frontend/dist/govuk/assets`. The `images` subdirectory is already included. Are both required or can it be avoided?

##### Disabled buttons

The `govuk-button--disabled` class (I think) was only used on the 'Provisional assessment by date' MI report. The link is displayed as disabled but it not actually disabled, and this issue exists in production. The link already has the `disabled` tag and removing the `govuk-button--disabled` does not have any effect.

--------

#### TODO (wip)

 - [x] The 'menu' button for mobile does not work correctly in the Swagger docs. Why is this and how can it be fixed? This is a current issue in production.
 - [x] Check that icons are being used correctly. See app/views/layouts/application.html.haml and app/views/grape_swagger_rails/application/index.html.haml


app/webpack/packs/govuk-frontend.js
```diff
@@ -1,3 +1,4 @@
-require.context('govuk-frontend/govuk/assets', true)
+import { initAll } from 'govuk-frontend'
 
-require('govuk-frontend').initAll()
+require.context('govuk-frontend/dist/govuk/assets', true)
```
mpw5, line 3: I don't _think_ this is needed now - locally, the app looks and behaves as expected without it, so it feels safe to remove.

It looks like we could potentially move the rest of the contents of this file into `app/webpack/packs/application.js`, replacing line 80, and then delete this file. That's what other teams seem to be doing, and possibly simplifies things a little, but given the length of `application.js` maybe we're better of keeping them separate.

config/initializers/assets.rb
```diff
@@ -11,5 +11,6 @@
 # Rails.application.config.assets.precompile += %w( adp_swagger_application.js *.png)
 # Rails.application.config.assets.precompile += %w( application.test.js ) if Rails.env.test?
 # Rails.application.config.assets.precompile += %w( pdf.css )
-Rails.application.config.assets.paths << Rails.root.join('node_modules', 'govuk-frontend', 'govuk', 'assets', 'images')
+Rails.application.config.assets.paths << Rails.root.join('node_modules', 'govuk-frontend', 'dist', 'govuk', 'assets')
+Rails.application.config.assets.paths << Rails.root.join('node_modules', 'govuk-frontend', 'dist', 'govuk', 'assets', 'images')
```
mpw5, line 15: You asked about the possibility of removing this line. Testing locally, the app breaks without it. There may be a way to configure it so that adding the `/assets` directory is sufficient to include the `/images` subdirectory but I haven't been able to work out what it is and I have no objection to leaving this as it is.

mpw5 (approved):
Nice work 🥇 

I've left comments relating to a couple of questions you asked in the PR description, happy to discuss them, but I think this is pretty much good to go as is. 

We should probably do some thorough testing before releasing though 😀 I'll have a look on Monday.

Obsiye (approved):
LGTM. I tested it and it looks good to me there too

PR #6583, opened 2024-02-29
Fix K8s warnings

mpw5:
#### What

Update kubernetes securityContexts to reduce warnings

#### Why

The Cloud Platform upgrade of the kubernetes cluster to v1.26 has resulted in warnings of a missing security context being displayed when deploying the application, eg:

```
Warning: would violate PodSecurity "restricted:latest": seccompProfile (pod or container "cccd-app" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
Warning: would violate PodSecurity "restricted:latest": seccompProfile (pod or container "cccd-worker" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

#### How

Defines `seccompProfile` in all kubernetes config files with a value of `RuntimeDefault`, which allows us to default to the container's security profile.

#### Screenshots

Before:

image

After: 

image

### DIFF
--- a/.k8s/live/api-sandbox/deployment-worker.yaml
+++ b/.k8s/live/api-sandbox/deployment-worker.yaml
@@ -47,6 +47,8 @@ spec:
               - ALL
             runAsNonRoot: true
             allowPrivilegeEscalation: false
+            seccompProfile:
+              type: RuntimeDefault
 
           # configMapRef: non-secret env vars defined in `app-config.yaml`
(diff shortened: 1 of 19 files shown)


Katy600 (approved):
@mpw5 you will probably want another set of eyes but it all makes sense to me so I have approved. :)

jrmhaig (approved):

PR #6611, opened 2024-03-12
Ctskf 690 cccd   set up access to maat api

Katy600:
#### What
This is a combination of  [Maat integration](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/6438/files) which was a test ticket for implementing maat combined with [CCCD - Set up access to MAAT API from deployed environments](https://dsdmoj.atlassian.net/browse/CTSKF-690)  which was to get access to Access to the Maat Api with in cognito and create ab interface for making requests to the API.

#### Ticket

[CCCD - Set up access to MAAT API from deployed environments](l[ink](https://dsdmoj.atlassian.net/browse/CTSKF-690))

#### Why
This is a user centred ticket and is intended to automate something that caseworkers currently have to manually input. This will speed up the process as well as remove potential human error. 

#### How
Auto filling using  Maat API

### DIFF
--- a/app/views/shared/_representation_order_details.html.haml
+++ b/app/views/shared/_representation_order_details.html.haml
@@ -1,0 +1,11 @@
+= govuk_table do |table|
+  - table.with_caption(size: 's', text: 'Details from MAAT (beta)')
+
+  - table.with_body do |body|
+    - body.with_row do |row|
+      - row.with_cell(header: true, text: 'Case number')
+      - row.with_cell(text: representation_order.maat_details[:case_number])
+
+    - body.with_row do |row|
+      - row.with_cell(header: true, text: 'Date')
+      - row.with_cell(text: representation_order.maat_details[:representation_order_date])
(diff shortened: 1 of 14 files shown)


app/views/shared/_representation_order_details.html.haml
```diff
@@ -0,0 +1,11 @@
+= govuk_table do |table|
+  - table.with_caption(size: 's', text: 'Details from MAAT (beta)')
```
vloruncg, line 2: Will this still be in beta once released?

vloruncg (approved):
LGTM

app/views/shared/_representation_order_details.html.haml
jrmhaig: This will not actually be displayed yet as this PR is just to get the connection code in the codebase, so that it can be used and built on. I think, however, that when we get to the point of displaying information it should be indicated that it is beta or experimental for a period of time.

PR #6615, opened 2024-03-13
[CTSKF-653] Upgrade to Rails 7

jrmhaig:
#### What

Upgrade Rails to version 7.0

#### Ticket

[CCCD - Upgrade to Rails 7](https://dsdmoj.atlassian.net/browse/CTSKF-653)

#### Why

Support for Rails 6.1, which we are currently using, will end soon. To ensure that we keep up to date with security and bug fixes we need to upgrade.

#### How

Update the pinned version in `Gemfile`, run the update script and then fix any immediate issues. Further changes to enable all the 7.0 default configuration options will be managed in subsequent PRs.

In some places configuration files have been changed to reflect changes in the default templates whilst in other places the current configuration is carried over.

Some things to note;

* The 'restart server' step has been restored in `bin/setup`.
* `server_timing` has been enabled in `config/environments/development.rb`. This is a new feature in Rails 7.0.
* The `active_support.report_deprecations` setting replaces various other `active_support` configuration options regarding deprecations.
* `eager_load` in `config/environments/test.rb` is set accoring to the setting of the `CI` environment variable, so that the whole of Rails is not loaded when running individual tests locally.
* `consider_all_requests_local` is set to true in `config/environments/test.rb`. This was explicitly set to false and may need to change back again.
* No changes to the `filter_parameters` list. In the template changes were made such as `:password` changing to `passw`, which would also match `passwd`, for example.


mpw5 (approved):
🎉

PR #6655, opened 2024-03-26
Ctskf 655 cccd button to generates button tag

Katy600:
#### What
The default behaviour for Rails 7 is that the button_to helper renders a button element. This needs to be set in config/initializers/new_framework_defaults_7_0.rb as part of the upgrade process.

Starting from Rails 7.0 button_to renders a form tag with patch HTTP verb if a persisted Active Record object is used to build button URL so this method now needs to be explicitly set in the button_to method. 

#### Ticket
[CCCD - button_to_generates_button_tag](https://dsdmoj.atlassian.net/browse/CTSKF-655)

#### Why
Otherwise button_to helper will no longer render a button element preventing users from submitting forms. 

#### How
- Uncomment the button_to_generates_button_tag in the new framework config. 
- Update the method parameter that are passed to the button_to, as they no longer requires the value to be a string. Instead, in order to keep the current behaviour, explicitly pass methods, by replacing "patch" with :patch for example.


Katy600:
This branch has been deployed

PR #6657, opened 2024-03-26
WIP: Set the serializer to hybrid

Katy600:
#### What

Rails 7 changes the cookie serializer from ‘marshal’ to ‘json’ and to aid migration there is a ‘hybrid’ option. Until the load_defaults option is updated to 7.0 these are set in config/initializers/new_framework_defaults_7_0.rb, where further details can be found. The process is;

#### Ticket

[CCCD - Update cookie serializer](https://dsdmoj.atlassian.net/browse/CTSKF-656)

#### Why

Rails 7 introduced a change in the default cookie serializer from 'marshal' to 'json'. This is a security measure to mitigate certain vulnerabilities associated with the 'marshal' serializer. To aid in the migration process, Rails 7 offers a 'hybrid' option, which allows applications to gradually migrate from 'marshal' to 'json' serialization. The 'hybrid' option enables Rails to deserialize cookies using both 'marshal' and 'json', allowing existing cookies serialized with 'marshal' to be read and migrated over time to the 'json' format. This change helps ensure better security practices while providing backward compatibility for existing applications.

#### How

- Set the serializer to hybrid and let the application run for a period of time, until users' cookies are re-written

- Waiting to set the serializer to json

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -18,6 +18,9 @@
       },
       "user_input": null,
       "confidence": "Medium",
+      "cwe_id": [
+        915
+      ],
       "note": ""
     },
     {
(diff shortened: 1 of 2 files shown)


mpw5 (approved):
LGTM 👍

PR #5511, opened 2023-03-27
Mark timezone sensitive test as 'pending'

mpw5:
#### What

Interim fix to prevent a test failing while we are in BST, to avoid the pipeline being blocked until we can resolve this properly.

#### Ticket

[CTSKF-321](https://dsdmoj.atlassian.net/browse/CTSKF-321)

[CTSKF-321]: https://dsdmoj.atlassian.net/browse/CTSKF-321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6674, opened 2024-04-02
Mark timezone sensitive test as 'pending'

jrmhaig:
#### What

Interim fix to prevent a test failing while we are in BST, to avoid the pipeline being blocked until we can resolve this properly.

See #5511 and #6196


AlexRyanShep (approved):
LGTM

vloruncg (approved):
👍

PR #6712, opened 2024-04-11
[CTSKF-657] Raise on open redirects

jrmhaig:
#### What

Configure to raise exceptions on open redirects.

#### Ticket

[CCCD - Various post-Rails 7 upgrade configuration settings](https://dsdmoj.atlassian.net/browse/CTSKF-657)

#### Why

Bring the configuration into line with the Rails 7.0 defaults.

#### How

Set `Rails.application.config.action_controller.raise_on_open_redirects = true` in `config/initializers/new_framework_defaults_7_0.rb`.

There were a few tests that required modification. This is because a method used to generate urls for redirection was being mocked with a url including a domain that differed from the host domain used by the tests, and this is precisely what the setting prevents.


AlexRyanShep (approved):
LGTM

PR #6713, opened 2024-04-11
[CTSKF 658] Update cache format version, step 1

jrmhaig:
#### What

Set the cache serialization format to version 6.1.

#### Ticket

[CCCD - Update cache format version](https://dsdmoj.atlassian.net/browse/CTSKF-658)

#### Why

See https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#new-activesupport-cache-serialization-format

Rails 7.0 introduces a a faster and more compact serialization. Rails 6.1 is incompatible with this new serialization and so there is no rollback step.

The first step is to fix the serialization format to version 6.1 and then deploy this. Once all Rails instances have been updated the format version can be changed to 7.0.

#### How

Set `config.active_support.cache_format_version` to `6.1` in `config/application.rb`.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,7 +17,12 @@ require_relative '../lib/govuk_component'
 module AdvocateDefencePayments
   class Application < Rails::Application
     config.load_defaults 6.1
+
+    # These two options can be removed after load_defaults is updated to 7.0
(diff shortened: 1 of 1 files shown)


AlexRyanShep (approved):
LGTM

PR #6727, opened 2024-04-15
Remove adp_text_field

jrmhaig:
#### What

Remove `adp_text_field` helper.

#### Ticket

N/A

#### Why

The `adp_text_field` helper has been replaced by govuk_text_field and is no longer used.

#### How

Remove `adp_text_field` in `app/form_builders/adp_form_builder.rb`, `app/form_builders/adp_text_field.rb` and all related tests.

In `app/webpack/javascripts/modules/external_users/claims/BlockHelpers.js` there was a note indicating that `location_wrapper` was added by `adp_text_field`. This was done by `app/form_builders/adp_text_field.rb` on line 145 based on the field and the comment was originally added as it was not possible to search for this text. It is now explicitly  set in `app/views/external_users/claims/expenses/_expense_fields.html.haml`.


mpw5 (approved):
👍

PR #6754, opened 2024-04-19
CTSKF-483 Replace shell spinner gem

mpw5:
#### What

The `shell-spinner` gem is used to provide visual feedback that rake tasks and scripts are running. `shell-spinner` is not actively supported and uses the `colorize` gem which we would like to remove from CCCD.

This replaces `shell-spinner` with the `tty-spinner` gem, which has comparable functionality, is more widely used and supported, and does not depend on `colorize`, and updates usage within the CCCD code.

While testing this, I found that some tasks were broken due to the rails 7 upgrade. These are fixed in this PR. Additionally, two redundant scripts are deleted.

#### Ticket

[CTSKF-483](https://dsdmoj.atlassian.net/browse/CTSKF-483)

#### Evidence

Before:

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/assets/28729201/a77e63d6-699d-44ad-b285-020bda705911

After:

https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/assets/28729201/c3b088f3-ed6e-4e21-8655-4dad703792e0

[CTSKF-483]: https://dsdmoj.atlassian.net/browse/CTSKF-483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrmhaig (approved):
Very nice. I like the look of the `tty-*` gems.

PR #6759, opened 2024-04-22
[CTSKF-488] Use Faraday instead of RestClient

jrmhaig:
#### What

Replace `rest-client` with `faraday`.

#### Ticket

#### Why

`rest_client` is being removed to reduce required dependencies.

#### How

`RestClient` is replaced with `Faraday` with some modifications of arguments and responses, as appropriate.

With `RestClient`, a failed requests raises an exception and this causes the tests to fail. With `Faraday`, however, an exception is not raised and the response can be checked for the http status and for `success?`. This changes how tests are executed;

* When a misc fee is created in for example, `spec/support/api/claims/advocate_claim_test/final.rb`, a failure does not stop the test with an exception and attempting to find `id` from the response fails. To remedy this the test is halted if `@client.failure` is `true` at this point.
* When a request fails the test should display a list of errors. However, the exception from `RestClient` prevents this. When switching to `Faraday` these errors are seen but the error appears as `#<Faraday::Response:0x0000000163312860>`. This is replaced by the https status.

### DIFF
--- a/spec/support/api/api_test_client.rb
+++ b/spec/support/api/api_test_client.rb
@@ -77,13 +77,13 @@ class ApiTestClient
   # don't raise exceptions but, instead, return the
   # response for analysis.
   #
-  def get_dropdown_endpoint(resource, api_key, **kwargs)
+  def get_dropdown_endpoint(resource, **kwargs)
     path = "api/#{resource}"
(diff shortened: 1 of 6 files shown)


mpw5 (approved):
👍

PR #6760, opened 2024-04-22
fake log payload

AlexRyanShep:
Do not approve or merge this

For dev only, to allow experimentation with what LogStuff sends to Kibana


AlexRyanShep:
No longer needed, closing PR and deleting branch

PR #6761, opened 2024-04-22
CTSKF-484 Remove colorize gem

mpw5:
#### What

CCCD uses the `colorize` gem to display coloured text in the terminal (for example when running rake tasks and scripts). `colorize` is released under the GPL licence, which has implications for the licensing of CCCD.

This removes `colorize` and the single place where it is referenced (`lib/tasks/brakeman.rake`). Most of the usage was removed when files were deleted in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/6753.

CCCD already uses the `awesome_print` gem which provides coloured output and automatically works with code in the format `"string".color`, which covers all remaining usage in CCCD, so there is no requirement to add a replacement.

#### Ticket

[CTSKF-484](https://dsdmoj.atlassian.net/browse/CTSKF-484)

#### Why

To remove an unneeded dependency and improve our licence compliance.

#### How

Remove the gem and references to it.

[CTSKF-484]: https://dsdmoj.atlassian.net/browse/CTSKF-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrmhaig (approved):
:+1:

PR #6766, opened 2024-04-23
CTSKF-305 API Request Logging

AlexRyanShep:
#### What
Enabling logging of API requests to STDOUT

#### Ticket

[CTSKF-305](https://dsdmoj.atlassian.net/browse/CTSKF-305)

#### Why
To help us debug issues with the API and notice them when they occur

#### How
Rewrote existing API logging to use LogStuff, added new helper to log errors during API exception handling, wrote RSpec unit tests for logging functionality

[CTSKF-305]: https://dsdmoj.atlassian.net/browse/CTSKF-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/interfaces/api/logger.rb
+++ b/app/interfaces/api/logger.rb
@@ -43,13 +43,13 @@ module API
               e)
     end
 
-    def log_api(level, type, data, error = nil)
+    def log_api(level, type, message, error = nil)
       LogStuff.send(
         level.to_sym,
         type:,
         error: error ? "#{error.class} - #{error.message}" : 'false'
       ) do
-        "API Log: #{data.to_json}"
+          message
       end
     end
   end
--- a/app/interfaces/api/logger.rb
+++ b/app/interfaces/api/logger.rb
@@ -1,27 +1,21 @@
 module API
   class Logger < Grape::Middleware::Base
     def before
-      log_api_request(env['REQUEST_METHOD'], env['PATH_INFO'], env['rack.request.form_hash'])
+      log_api(:info,
+              'api-request',
+              { method: env['REQUEST_METHOD'],
+                path: env['PATH_INFO'],
+                data: env['rack.request.form_hash'] })
     end
 
     def after
-      log_api_response
+      log_api(:info,
+              'api-response',
+              { inputs: input_params, status: response_status, response_body: })
       @app_response # this must return @app_response or nil
     end
 
     private
-
-    def log_api_request(method, path, data)
-      log_api(:info,
-              'api-request',
-              { method:, path:, data: })
-    end
-
-    def log_api_response
-      log_api(:info,
-              'api-response',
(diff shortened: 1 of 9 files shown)


app/interfaces/api/logger.rb
```diff
@@ -12,17 +12,24 @@ def after
     private
 
     def log_api_request(method, path, data)
-      log_api('api-request', method:, path:, data:)
+      log_api(:info,
+              'api-request',
+              { method:, path:, data: })
```
Katy600: Hi Alex, do the params need to be passed to the hash in the log_api method? Or does that not make any difference? 
 
 `{ method: method, path: path, data: data}`

AlexRyanShep: Hi Katy, this is known as Shorthand Hash Syntax. When you're declaring a Hash and you have a variable which is the same name as the key you don't need to explicitly write it down. `method:` by itself is the same thing as `method: method`

app/interfaces/api/logger.rb
```diff
@@ -35,13 +42,21 @@ def response_body
...
-      Rails.logger.info api_request.to_json
+    rescue JSON::ParserError => e
+      log_api(:error,
+              'response-body',
+              "Error parsing API response body: \n#{@app_response[2].first}",
+              e)
+    end
+
+    def log_api(level, type, data, error = nil)
+      LogStuff.send(
```
Katy600, line 92: Is it better to pass a hash of arguments to make it more readable and remove any reliance on the order. Or perhaps that method is not going to be reused outside this context anyway and the params are very clear - so maybe not necessary. It was just a thought.

`(level: level, type: type, data: data, error = nil)`

AlexRyanShep: I think that might be a bit overkill here? Due to how Grape Middleware works, this is unlikely to be exported to any other areas of the application. It's just an abstraction used in this specific file. Happy to change if you feel strongly about it but not sure it's needed

app/interfaces/api/logger.rb
Katy600: Thanks @AlexRyanShep That is really helpful :)

app/interfaces/api/logger.rb
Katy600: That makes sense @AlexRyanShep. This is mostly me just trying to be thorough in a code review and making sure I understand it fully.

Katy600 (commented):
Hi @AlexRyanShep , 
Thanks you for all your feedback on my comments.  I have learnt a lot looking at your code and it would be great to talk this through to learn more. I have no other feed back unfortunately :)

mpw5 (approved):
Nice work. This all looks good to me.

PR #6770, opened 2024-04-24
Index fees table on uuid to fix slow query

mpw5:
#### What

Sentry is reporting that the most [time-consuming query](https://ministryofjustice.sentry.io/performance/?project=5436820&statsPeriod=90d&utc=true) in CCCD is

```
SELECT "fees".* FROM "fees" WHERE "fees"."uuid" = $1 ORDER BY "fees"."id" ASC LIMIT $2
```

which takes on average ~400ms to complete.

image

This PR attempts to make that query more efficient by indexing the `fees` table on the `uuid` field.

#### Why

To improve the overall performance of CCCD.

#### How

The slow query originates in the following line in `app/interfaces/api/v1/external_users/date_attended.rb`:

```
::Fee::BaseFee.find_by(uuid: params[:attended_item_id])
```

which is executed when a provider uses the API to add a 'date attended' to a fee.

This code looks up the fee record by `uuid`, which is an unindexed field in the `fees` table. In production this table contains over 2.4 million records, and this query is taking increasingly long to run.

The speed of lookup can be improved by adding an index to the `uuid` field.

To prevent this having an impact on users by locking the table when adding the index, the `algorithm` attribute is set to `:concurrently`. This necessitates disabling transactions for the migration. See https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in for more details.

#### Proof

Tested on an anonymised copy of production running locally.

##### Before

image

##### After

image

There is always a concern when adding indexes to tables that it may have an adverse affect on inserts into that table. In this instance this does not appear to be an issue:

##### Before

image

##### After 

image

### DIFF
--- a/db/migrate/20240419173039_add_index_to_fees_on_uuid.rb
+++ b/db/migrate/20240419173039_add_index_to_fees_on_uuid.rb
@@ -1,0 +1,7 @@
+class AddIndexToFeesOnUuid < ActiveRecord::Migration[7.0]
+  self.disable_ddl_transaction!
+
+  def change
+     add_index :fees, :uuid, unique: true, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,8 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


jrmhaig (approved):
🚀

PR #6771, opened 2024-04-24
[CTSKF-488] Remove GoogleAnalytics::API

jrmhaig:
#### What

Remove use `rest-client` gem.

#### Ticket

#### Why

`rest_client` is being removed to reduce required dependencies.

#### How

`GoogleAnalytics::API` was added in #1980 and used `RestClient.get` to submit data to Google Analytics. After #4265 it is no longer being used so it can be deleted.


mpw5 (approved):
Nice!

PR #6776, opened 2024-04-25
[CTSKF-488] Remove rest-client gem

jrmhaig:
#### What

Remove use `rest-client` gem.

#### Ticket

#### Why

`rest-client` is being removed to reduce required dependencies.

#### How

Following various PRs to replace use of `rest-client` with `faraday` it is now removed from `Gemfile` and `Gemfile.lock`.

VCR files used in some cucumber tests that query the internal API have been recreated to update the details of requests as performed by `faraday` instead of `rest-client`.


mpw5 (approved):
👍

PR #6792, opened 2024-04-30
Fix XSS javascript vulnerability in `Modules.AllocationDataTable.js`

mpw5:
#### What

`app/webpack/javascripts/modules/Modules.AllocationDataTable.js` uses the `.html` jquery method to build a message to display when a caseworker admin allocates claims to a caseworker. The way we use this method introduces a security vulnerability which needs to be addressed.

#### Why

Snyk has [identified](https://app.snyk.io/org/legal-aid-agency/project/3c84d1a7-ff1f-4508-87d5-c74b0b4f23df#issue-ab2a726a-84ba-4490-beb0-f76fddad4f97) the use of this method as a cross-site scripting vulnerability:

```
Unsanitized input from data from a remote resource flows into html,
where it is used to dynamically construct the HTML page on client side.
This may result in a DOM Based Cross-Site Scripting attack (DOMXSS)
```

More about DOM based XSS vulnerabilities: https://learn.snyk.io/lesson/dom-based-xss/

#### How
Fixed by replacing the `.html` method with `.text`, which is a more secure alternative that escapes special characters passed into it, removing the XSS vulnerability while not affecting the message we display to users.

### DIFF
--- a/app/webpack/javascripts/modules/Modules.AllocationDataTable.js
+++ b/app/webpack/javascripts/modules/Modules.AllocationDataTable.js
@@ -378,7 +378,7 @@ moj.Modules.AllocationDataTable = {
       }).done(function (result) {
         self.ui.$notificationMsg.removeClass('govuk-!-display-none govuk-notification-banner--error')
         self.ui.$notificationMsg.addClass('govuk-notification-banner--success')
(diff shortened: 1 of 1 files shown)


Katy600 (approved):
That makes sense :)

PR #6793, opened 2024-04-30
[CTSKF-682] Enable content security policy

jrmhaig:
#### What

Enable content security policy in 'report only' mode.

#### Ticket

#### Why

The Content Security Policy ensures that unauthorised scripts are not executed. In 'report only' mode then possible violations will be reported but not blocked. This can be monitored to ensure that the policy has been configured correctly before being enabled fully.

#### How

The recommended settings are enabled together with:

```ruby
config.content_security_policy_report_only = true
```

A new endpoint, `/csp_report`, is created to capture reports and post them to Slack in the `laa-cccd-alerts` channel.

The `rack-livereload` gem, which is used when running in a development environment, results in CSP violations. This will need to be resolved before the policy can be enabled.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -4,22 +4,22 @@
 # See the Securing Rails Applications Guide for more information:
 # https://guides.rubyonrails.org/security.html#content-security-policy-header
 
-# Rails.application.configure do
-#   config.content_security_policy do |policy|
-#     policy.default_src :self, :https
-#     policy.font_src    :self, :https, :data
-#     policy.img_src     :self, :https, :data
-#     policy.object_src  :none
-#     policy.script_src  :self, :https
-#     policy.style_src   :self, :https
-#     # Specify URI for violation reports
-#     # policy.report_uri "/csp-violation-report-endpoint"
-#   end
-#
-#   # Generate session nonces for permitted importmap and inline scripts
-#   config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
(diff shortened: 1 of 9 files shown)


Katy600 (commented):
Thanks for talking this through earlier @jrmhaig :) Let me know if you want me to approve it. I imagine you will want another pair of eyes though.

mpw5 (approved):
Nice! I especially like the slack integration.

What's the issue with `rack-livereload`?

config/initializers/content_security_policy.rb
```diff
@@ -4,22 +4,22 @@
...
-#   # config.content_security_policy_report_only = true
-# end
+Rails.application.configure do
+  config.content_security_policy do |policy|
+    policy.default_src :self, :https
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :https, :data
+    policy.object_src  :none
+    policy.script_src  :self, :https
+    policy.style_src   :self, :https
```
mpw5, line 14: I wonder if the use of `:https` is too permissive, which is why you're seeing javascript still running.

If I'm reading it right, I think https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy says that `:self` will allow our resources, while `:https` allows resources from any `https` origin, so I wonder if we need to remove those.

For comparison, the Civil Apply CSP (which is the only application I've worked on with this set up) generally only allows `:self` with some specific exceptions: https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/config/initializers/content_security_policy.rb

PR #6815, opened 2024-05-08
[CTSKF-682] Relax Content Security Policy

jrmhaig:
#### What

Permit inline script without nonce.

#### Ticket

#### Why

Much of the custom Javascript in CCCD is triggering alerts from the CSP as inline script with a nonce. This may be due to JQuery and so this issue may be resolved when the code is refactored.

#### How

Enable `unsafe-inline` for the `script-src` policy. Also, for the moment alerts are not reporting to Slack and are written to the log file in the pod instead.

### DIFF
--- a/app/controllers/csp_reports_controller.rb
+++ b/app/controllers/csp_reports_controller.rb
@@ -3,13 +3,14 @@ class CspReportsController < ApplicationController
   skip_forgery_protection
 
   def create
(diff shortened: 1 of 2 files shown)


Katy600 (approved):
That makes sense @jrmhaig. You will probably want another set of eyes on it but I have approved it anyway - as it is a small change.

PR #6832, opened 2024-05-14
CTSKF-305 API Logging bugfix

AlexRyanShep:
#### What
Bugfix for where logger.rb was incorrectly categorising 201 response codes as errors.

#### Ticket

[CTSKF-305](https://dsdmoj.atlassian.net/browse/CTSKF-305)

[CTSKF-305]: https://dsdmoj.atlassian.net/browse/CTSKF-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/interfaces/api/logger.rb
+++ b/app/interfaces/api/logger.rb
@@ -13,12 +13,13 @@ module API
     end
 
     def after
-      if response_status == '200'
+      return if response_status.nil?
+
+      if response_status.between?(200, 399)
         log_api(:info, 'api-response',
-                { request_id:, path:, status: response_status,
-                  claim_id: response_param('claim_id'), case_number: response_param('case_number'),
-                  id: response_param('id') })
-      elsif response_status
+                { request_id:, path:, status: response_status, claim_id: response_param('claim_id'),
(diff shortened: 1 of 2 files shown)


app/interfaces/api/logger.rb
```diff
@@ -13,12 +13,13 @@ def before
     end
 
     def after
-      if response_status == '200'
+      return if response_status.nil?
+
+      if response_status.between?(200, 399)
```
jrmhaig, line 18: 3xx statuses are redirects. I wouldn't expect to see many of these, but do we want to treat them identically to 2xx statuses?

AlexRyanShep: For logging purposes here I think we can yeah. The main difference between log_api and log_error here is that log_api is just expecting some data items to display as the message, while log_error also wants to be given an error object. Since 3xx responses aren't generating errors I can't think of anything to do handling them differently here.

jrmhaig (approved):
:+1:

PR #6981, opened 2024-06-18
Add edit and delete links

Katy600:
#### What
Allow provider admin to edit inactive users

#### Ticket

[CCCD - Admin unable to edit inactivate accounts](https://dsdmoj.atlassian.net/browse/CTSKF-822)

#### Why
An account is disabled when it is still live, so that claims can still be created for this user, but the claims are created by another user.

It appears that when an account is disabled then the admin of the user’s provider is also unable to edit this account. As the user still needs to have claims it should be possible for the admin to edit the details, as appropriate.

#### How

Only hide links for inactive users.

### DIFF
--- a/app/views/external_users/admin/external_users/index.html.haml
+++ b/app/views/external_users/admin/external_users/index.html.haml
@@ -39,8 +39,9 @@
 
         = govuk_table_td('data-label': t('.actions')) do
           - if advocate.active? && advocate.enabled?
-            .app-link-group
-              = govuk_link_to t('.edit_html', context: advocate.name), edit_external_users_admin_external_user_path(advocate)
-              = govuk_link_to t('.delete_html', context: advocate.name), external_users_admin_external_user_path(advocate), method: :delete, data: { confirm: t('.confirmation') }
+            = render partial: 'edit_delete_links', locals: { advocate: advocate }
+          - elsif advocate.active? && !advocate.enabled?
+            Inactive
+            = render partial: 'edit_delete_links', locals: { advocate: advocate }
           - else
             Inactive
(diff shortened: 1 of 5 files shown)


app/views/external_users/admin/external_users/index.html.haml
```diff
@@ -39,8 +39,7 @@
 
         = govuk_table_td('data-label': t('.actions')) do
           - if advocate.active? && advocate.enabled?
+            Inactive
```
jrmhaig, line 47: If I understand correctly, this will add 'Inactive' to all users, not just to those who are disabled or deleted. It looks as though there are now three cases;

* User is active and enabled - the edit and delete links should be available with no more information
* User is inactive but not deleted - the edit and delete links should be available but there could (should?) be a link to indicate that the user is inactive.
* User is deleted - the edit and delete links should not be shown and the 'inactive' note should be displayed

app/views/external_users/admin/external_users/index.html.haml
```diff
@@ -40,7 +40,8 @@
         = govuk_table_td('data-label': t('.actions')) do
           - if advocate.active? && advocate.enabled?
             .app-link-group
-              = govuk_link_to t('.edit_html', context: advocate.name), edit_external_users_admin_external_user_path(advocate)
-              = govuk_link_to t('.delete_html', context: advocate.name), external_users_admin_external_user_path(advocate), method: :delete, data: { confirm: t('.confirmation') }
-          - else
+            = render_edit_delete_links(advocate)
```
jrmhaig: This may be a matter of personal preference so please ask the opinion of others. I would find this easier to follow if `render_edit_delete_links` were replace by a partial instead. Also, this helper (or partial. if you decide to do that) already has a `.app-link-group` div so the previous line (and the similar line below) are not required.

mpw5 (approved):
This looks good to me. I've reviewed the code and tested the change locally.

My only question would be whether we've had design input on the change - I know this was something discussed previously. If not, we may want to check the current design with Matt before merging this PR, to ensure he is happy with the change before it goes live.

PR #7055, opened 2024-06-26
Group sentry gems

jrmhaig:
#### What

Configure Dependabot to group Sentry and Rails gems.

#### Ticket

N/A

#### Why

In each group the gems are usually all released at the same time to ensure that all version numbers remain the same. Therefore, when there is an update for Sentry there are 3 packages to upgrade and for an upgrade to Rails there are many more. In both cases the upgrades need to be done together.

#### How

```
    sentry:
      patterns:
        - "sentry-*"
    rails:
      patterns:
        - "rails"
        - "action*"
        - "active*"
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,6 +36,11 @@ updates:
     sentry:
       patterns:
         - "sentry-*"
+    rails:
+      patterns:
+        - "rails"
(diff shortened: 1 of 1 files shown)


mpw5 (approved):
👍

PR #7098, opened 2024-07-02
Create london_fees column/attribute in Claim model

AlexRyanShep:
Draft PR to allow for code review/testing; do not merge into master

#### What

Add new boolean attribute to the Claim model to allow indication of London or Non-London fees. Field is nullable to allow compatibility with existing claims without bulk applying a status as this could cause confusion by applying False to Claims which should qualify for London rates, or vice versa. 

#### Ticket

[CTSKF-845](https://dsdmoj.atlassian.net/browse/CTSKF-845)

[CTSKF-845]: https://dsdmoj.atlassian.net/browse/CTSKF-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/migrate/20240702094708_add_london_fees_attribute_to_claims.rb
+++ b/db/migrate/20240702094708_add_london_fees_attribute_to_claims.rb
@@ -1,0 +1,5 @@
+class AddLondonFeesAttributeToClaims < ActiveRecord::Migration[7.0]
+  def change
+    add_column :claims, :london_fees, :boolean, null: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


AlexRyanShep:
Note from me; schema.rb has been automatically modified during migration run to include `precision: nil` on a number of lines using datetime attributes. I did some research and it seems that this should have no actual effect and is maybe an aftereffect of running this migration now that we're on rails 7. According to the issue raised for this in the Rails repo, precision:nil was already implied on all those lines as they didn't specify anything else for it. https://github.com/rails/rails/issues/44571

mpw5:
This looks good to me. I think you're right about the schema changes, so I'm not concerned about it.

jrmhaig:
Looks good. 👍

AlexRyanShep:
Thanks both! I've merged this branch into my main feature branch, so closing this PR

PR #7105, opened 2024-07-03
Add consumes option to the Grape configuration

jrmhaig:
#### What

Add a global `consume` option for Grape.

#### Ticket

N/A

#### Why

See https://github.com/ruby-grape/grape-swagger/pull/872

The `consumes` option allows non-JSON formatted inputs to be accepted. With version 2.1.0 of `grape-swagger` this option is used to determine how the form in the Swagger documentation is displayed and `application/x-www-form-urlencoded` is required to ensure that each field is shown in a form rather than a single free-text field for JSON.

#### How

Add the `consumes` option to the two `root.rb` files.


AlexRyanShep (approved):
Lgtm

PR #7129, opened 2024-07-09
Pass class to serialize as keyword argument

jrmhaig:
#### What

Update `serialize` in three places to pass the class as a keyword argument.

#### Ticket

N/A (but related to [CCCD: Upgrade Rails to 7.1](https://dsdmoj.atlassian.net/browse/CTSKF-769))

#### Why

Passing the class as a positional argument is deprecated and will be removed in Rails 7.2.

#### How

Use the `type` keyword argument instead of a positional argument.


jrmhaig:
This update cannot be done prior to 7.1.

PR #7130, opened 2024-07-09
Move settings from secrets.yml to settings.yml

jrmhaig:
#### What

Move settings from `secrets.yml` into `settings.yml`.

#### Ticket

N/A (but related to [CCCD: Upgrade Rails to 7.1](https://dsdmoj.atlassian.net/browse/CTSKF-769))

#### Why

The `secrets.yml` file is deprecated and is removed in Rails 7.1. In preparation for the upgrade to Rails 7.1 the settings in that file need to be moved out. In CCCD settings are managed by the `config` gem with the `settings.yml` so this is where they can be moved.

#### How

Move the settings into `settings.yml` and update `Rails.application.secrets` to `Settings`, as appropriate.

The development and test version of the secret key base is moved into the appropriate environment specific `settings.yml` file.


mpw5 (approved):
👍

PR #7137, opened 2024-07-10
Update to Ruby 3.3.4

jrmhaig:
#### What

Update Ruby to version 3.3.4.

#### Ticket

N/A

#### Why

The CircleCI build was broken after the release of Ruby 3.3.4. In fact, the issue appears to be due to the docker image have Ruby 3.3.4 and the `Gemfile` and `.ruby-version` specifying 3.3.3.

Additionally, the docker image for 3.3.3-alpine20 has an [open critical security issue.](https://hub.docker.com/_/ruby/tags?page=&page_size=&ordering=&name=alpine3.20)

#### How

Update in various places to use Ruby 3.3.4.

Also use `ruby file: '.ruby-version'` in `Gemfile` so that this tracks `.ruby-version`.


mpw5 (approved):
👍

PR #7161, opened 2024-07-15
[CTSKF-766] Set GOVUKDesignSystemFormBuilder::FormBuilder as default

jrmhaig:
#### What

Set the default form builder to `GOVUKDesignSystemFormBuilder::FormBuilder`.

#### Ticket

[Set GOVUKDesignSystemFormBuilder as the default FormBuilder](https://dsdmoj.atlassian.net/browse/CTSKF-766)

#### Why

All forms have been converted to use the `govuk_design_system_formbuilder` gem and custom form elements have been removed. For consistency with the GovUK Design System all form should now be using `GOVUKDesignSystemFormBuilder::FormBuilder` as their form builder so for simplicity it can be set as the default instead of being explicitly set each time.

#### How

Add

```ruby
  config.action_view.default_form_builder = GOVUKDesignSystemFormBuilder::FormBuilder
```

to `config/application.rb`. The explicit `builder` option is then removed from all forms.

### DIFF
--- a/app/views/case_workers/admin/allocations/_allocation.html.haml
+++ b/app/views/case_workers/admin/allocations/_allocation.html.haml
@@ -5,7 +5,7 @@
   .govuk-grid-column-two-thirds
     = render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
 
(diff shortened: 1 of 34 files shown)


mpw5 (approved):

mpw5 (commented):
I approved this and then noticed some rspec tests are failing 🤦 

Should be good to go once those are fixed.

PR #7166, opened 2024-07-16
Ctskf 486 london fees front end

AlexRyanShep:
Draft PR for code review/testing. Do not merge to master

#### What
Adds new London Fee elements to the CCCD frontend (New radiobuttons in case details, new row in case summary table) and associated cucumber and spec tests. 

#### Ticket

[CTSKF-846](https://dsdmoj.atlassian.net/browse/CTSKF-846)

#### Why

[CTSKF-846]: https://dsdmoj.atlassian.net/browse/CTSKF-846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/external_users/claims/case_details/_summary.html.haml
+++ b/app/views/external_users/claims/case_details/_summary.html.haml
@@ -29,7 +29,8 @@
           = claim.unique_id
 
       - if claim.lgfs?
-        = govuk_summary_list_row_collection( t('external_users.claims.case_details.london_rates.london_rates') ) { t("external_users.claims.case_details.london_rates.london_rates_#{claim.london_rates_apply}") unless claim.london_rates_apply.nil? }
+        = govuk_summary_list_row_collection( t('external_users.claims.case_details.london_rates.london_rates') ) do
+          = t("external_users.claims.case_details.london_rates.london_rates_#{claim.london_rates_apply}")
 
       - if claim.transfer_court.present?
         = govuk_summary_list_row_collection( t('common.transfer_court') ) { claim.transfer_court.name if claim.transfer_court }
--- a/app/views/external_users/claims/case_details/_summary.html.haml
+++ b/app/views/external_users/claims/case_details/_summary.html.haml
@@ -29,7 +29,7 @@
           = claim.unique_id
 
       - if claim.lgfs?
-        = govuk_summary_list_row_collection( t('external_users.claims.case_details.london_fees.london_fee') ) { t("external_users.claims.case_details.london_fees.london_fee_#{claim.london_fees}") unless claim.london_fees.nil? }
+        = govuk_summary_list_row_collection( t('external_users.claims.case_details.london_rates.london_rates') ) { t("external_users.claims.case_details.london_rates.london_rates_#{claim.london_rates_apply}") unless claim.london_rates_apply.nil? }
 
       - if claim.transfer_court.present?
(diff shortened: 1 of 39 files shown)


db/migrate/20240702094708_add_london_fees_attribute_to_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddLondonFeesAttributeToClaims < ActiveRecord::Migration[7.0]
+  def change
+    add_column :claims, :london_fees, :boolean, null: true
```
jrmhaig: This might be a bit picky.

The convention in Active Record is that references are singular nouns and then a `has_many` relation is the plural. For this reason, `london_fees` suggest to me that there should be a model `LondonFee` with which the claim model has a one-to-many relation. Is it possible to have a different name? `london` might be enough, or the more wordy `firm_is_in_london`.

AlexRyanShep: @jrmhaig thanks for this! As discussed on slack I've refactored the attribute name to `london_rates_apply` and rolled-back/re-applied the migration locally. All is looking good and tests are passing

app/views/external_users/claims/case_details/_summary.html.haml
```diff
@@ -28,6 +28,9 @@
         .unique-id-small
           = claim.unique_id
 
+      - if claim.lgfs?
+        = govuk_summary_list_row_collection( t('external_users.claims.case_details.london_fees.london_fee') ) { t("external_users.claims.case_details.london_fees.london_fee_#{claim.london_fees}") unless claim.london_fees.nil? }
```
mpw5: This results in the `London fee rates` field being blank if neither London fees radio button is selected (which is possible as it isn't a mandatory field).

image

Is this intentional? I wonder if some holding text would be better, something like ('This claim may qualify for London fee rates. This will be assessed by a caseworker'). It's not a blocker for this ticket, but maybe worth getting some design input on.

Also, nitpicky, this is quite a long line - if there's an easy way to break it over multiple it lines, it might make it more readable.

mpw5 (approved):
LGTM. I've left a couple of comments about one line which I don't think are blockers but might be worth considering.

mpw5:
The other thing that might be worth thinking about is whether we want to add a feature flag to control the visibility of the radio buttons and summary field. It would potentially allow us to merge this PR without affecting production users, and avoid the issue of having multiple long-lived PRs open (also not something that needs to be done now but could be considered as an enhancement).

app/views/external_users/claims/case_details/_summary.html.haml
AlexRyanShep: This is something to talk about in our call maybe? I tried to break the line up but if I did it in any of the ways I'd usually do for Ruby code, it just broke the haml code here.

AlexRyanShep: I did a bit of digging and:

a) managed to figure out what I was doing wrong with the multi-line comments to make these more manageable
b) I've changed the logic slightly to have a case in the yml file for `london_rates_` which is what it searches for when the value is nil. For now I've used your holding text from this comment, but I will try and get some design input on the wording for that

Pushing the new commit for this now

PR #7224, opened 2024-08-02
CTSKF 847 London Fees Apply added to CCCD API

AlexRyanShep:
#### What
This ticket covers adding the new attribute to the API for incoming calls to add into their claims, along with api validation and step validation for LGFS claims. 

#### Ticket

[CTSKF-847](https://dsdmoj.atlassian.net/browse/CTSKF-847)

#### Why
Part of ongoing project to create a POC for introducing missing fees to fee calculator

[CTSKF-847]: https://dsdmoj.atlassian.net/browse/CTSKF-847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/interfaces/api/v1/claim_params_helper.rb
+++ b/app/interfaces/api/v1/claim_params_helper.rb
@@ -43,7 +43,8 @@ module API
         optional :transfer_case_number, type: String, desc: 'OPTIONAL: The case number or URN for the transfer court.'
         optional :london_rates_apply,
                  type: Boolean,
-                 desc: 'OPTIONAL: Whether the firm for the claim is based in a London Borough or not'
+                 desc: 'OPTIONAL: Whether the firm for the claim is based in a London Borough or not',
+                 optional_boolean_validation: true
       end
 
       params :legacy_agfs_params do
--- a/app/interfaces/api/v1/claim_params_helper.rb
+++ b/app/interfaces/api/v1/claim_params_helper.rb
@@ -43,7 +43,7 @@ module API
         optional :transfer_case_number, type: String, desc: 'OPTIONAL: The case number or URN for the transfer court.'
         optional :london_rates_apply,
                  type: Boolean,
-                 desc: 'OPTIONAL: Whether the firm for the claim is based in a London Borough or not',
+                 desc: 'OPTIONAL: Whether the firm for the claim is based in a London borough or not',
                  optional_boolean_validation: true
       end
 
--- a/app/interfaces/api/v1/claim_params_helper.rb
+++ b/app/interfaces/api/v1/claim_params_helper.rb
@@ -41,6 +41,9 @@ module API
         optional :supplier_number, type: String, desc: 'REQUIRED. The supplier number.'
         optional :transfer_court_id, type: Integer, desc: 'OPTIONAL: The unique identifier for the transfer court.'
         optional :transfer_case_number, type: String, desc: 'OPTIONAL: The case number or URN for the transfer court.'
+        optional :london_rates_apply,
+                 type: Boolean,
+                 desc: 'OPTIONAL: Whether the firm for the claim is based in a London Borough or not'
       end
 
       params :legacy_agfs_params do
--- a/spec/support/shared_examples_for_api.rb
+++ b/spec/support/shared_examples_for_api.rb
@@ -236,17 +236,18 @@ RSpec.shared_examples 'a claim validate endpoint' do |options|
       end
 
       it { expect_error_response('Enter a case number') }
+      it { expect(LogStuff).to have_received(:send).with(:error, hash_including(type: 'api-error')) }
     end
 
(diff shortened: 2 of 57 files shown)


app/interfaces/api/v1/claim_params_helper.rb
```diff
@@ -41,6 +41,9 @@ module ClaimParamsHelper
         optional :supplier_number, type: String, desc: 'REQUIRED. The supplier number.'
         optional :transfer_court_id, type: Integer, desc: 'OPTIONAL: The unique identifier for the transfer court.'
         optional :transfer_case_number, type: String, desc: 'OPTIONAL: The case number or URN for the transfer court.'
+        optional :london_rates_apply,
+                 type: Boolean,
+                 desc: 'OPTIONAL: Whether the firm for the claim is based in a London Borough or not'
```
mpw5: Really nitpicky - I don't think `Borough` should be capitalised in this instance. It's lower case in the front end and probably should match that here (in fact you could possibly reuse the complete text from the front end in the api): https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/2cee46d02281d843d9c30e4481cfb2dcc2dcbada/config/locales/en.yml#L1088

spec/support/shared_examples_for_api.rb
```diff
@@ -245,8 +245,15 @@
         post_to_validate_endpoint
       end
 
-      it { expect_error_response('london_rates_apply is invalid, london_rates_apply must be true, false or nil') }
-      it { expect(LogStuff).to have_received(:send).with(:error, hash_including(type: 'api-error')) }
+      it 'returns an error' do
+        skip('Only test for LGFS') if claim_user_type == 'Advocate'
+        expect_error_response('london_rates_apply is invalid, london_rates_apply must be true, false or nil')
+      end
```
mpw5: I think I understand why you've done it this way, but I'm not sure I'm completely happy with using `skip` to set the scope of these tests. If we're planning to also add London rates to AGFS claims at some point then _maybe_ it's ok to leave it as it is temporarily, but if this will only ever be for LGFS then I think we should look at structuring the tests differently to avoid the need to use `skip`.

For example you could create a new set of shared examples in this file (in fact two, one for the validate endpoint and one for the create endpoint) just to test the London rates:

```
RSpec.shared_examples 'validate London rates' do
  describe "POST #{ClaimApiEndpoints.for(options[:relative_endpoint]).validate}" do
    subject(:post_to_validate_endpoint) do
      post ClaimApiEndpoints.for(options[:relative_endpoint]).validate, valid_params, format: :json
    end

    before { allow(LogStuff).to receive(:send) }

    let(:claim_user_type) { 'Litigator' }

    <your contexts from these tests>
  end
end
```
and
```
RSpec.shared_examples 'create claim with London rates' do
  describe "POST #{ClaimApiEndpoints.for(options[:relative_endpoint]).create}" do
    subject(:post_to_create_endpoint) do
      post ClaimApiEndpoints.for(options[:relative_endpoint]).create, valid_params, format: :json
    end

    before { allow(LogStuff).to receive(:send) }

    let(:claim_user_type) { 'Litigator' }

    <your contexts from these tests>
  end
end
```
And then you can include them in just the LGFS API tests:
```
include_examples 'validate London rates'
include_examples 'create claim with London rates'
```

(I've not tested any of the above, so it might not actually work quite like that 🙂)

It adds some duplication of setup but more cleanly segregates the tests so that the tests for LGFS functionality only run against LGFS claims, rather than against all claims and we have to skip the ones that are out of scope. This sort of thing is one of the drawbacks of shared contexts. They can become overcomplicated surprisingly quickly.

mpw5 (commented):
This mostly looks good to me, but I have some concerns over the use of shared examples in testing. Happy to discuss if it would help.

app/interfaces/api/v1/claim_params_helper.rb
AlexRyanShep: Good catch! I'll change this

spec/support/shared_examples_for_api.rb
AlexRyanShep: Thanks Mark, I understand what you've said here. I'll check with Angela and Steffan whether we'd expect AGFS to eventually have this functionality and then we can chat about the best way to neaten this up? If it going to be LGFS only I like your solution above 👍

AlexRyanShep: I've had a talk with Angela and Steffan. There are no unimplemented fees in our backlog for AGFS which would make use of the london_rates_apply flag, and while we can't see the future, we've no particular reason to expect any will be added in the near future.  I'll have a go at structuring the tests into some new shared example sets like you've suggested above

PR #7300, opened 2024-08-20
[CTSKF-830] Rake task to create CSV of offence data

jrmhaig:
#### What

Create a rake task to dump all offence data into CSV files.

#### Ticket

[CCCD: Standardise offences across all environments (dev-lgfs)](https://dsdmoj.atlassian.net/browse/CTSKF-830)

#### Why

The offences data on the different environments differ and this can cause confusion with testing. As a first step towards bringing them into line this rake task will allow a copy of the production data to be generated as a CSV file that can then be added, with another rake task (to be written), into other environments.

#### How

The task `rails 'offences:extract[tmp/dumps]'` will create a new directory `tmp/dumps` and save four files from the offences, offence classes, offence bands and offence categories tables.

### DIFF
--- a/lib/tasks/offences.rake
+++ b/lib/tasks/offences.rake
@@ -1,5 +1,8 @@
 require_relative 'rake_helpers/repair_offences'
-require 'csv'
+require_relative 'rake_helpers/rake_utils'
+require 'fileutils'
+
+include Tasks::RakeHelpers::RakeUtils
 
 namespace :offences do
   desc 'Generate repair input file of claim id against offence id based on data from CCR'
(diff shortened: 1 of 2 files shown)


lib/tasks/offences.rake
```diff
@@ -18,4 +21,27 @@ namespace :offences do
     repair = Tasks::RakeHelpers::RepairOffences.new(file: args.filename, output: args.output, interactive:, dry_run:)
     repair.repair
   end
+
+  desc 'Extract details of offences and write to a CSV file'
+  task :extract, [:dir] => :environment do |_task, args|
+    dir = args.dir
+    if File.exist?(dir)
```
mpw5, line 28: I guess if I were being nitpicky, it might be worth specifying a default directory in case the user doesn't provide one themselves, otherwise this line will fail with an error. But as this is mostly just for our use in this specific case, it's probably overkill.

mpw5 (approved):
Nice 👍

PR #7384, opened 2024-09-05
Ctskf 852 cccd spacing between primary and secondary buttons

Katy600:
#### What
Adds spacing between primary and secondary buttons to the following pages:

- [x] Offence details
- [x] Manage provider
- [x] Confirmation
- [x] Certification
- [x] Supporting evidence
- [x] Travel expenses
- [x] Miscellaneous fees
- [x] Fixed fees
- [x] Defendant details
- [x] Case details
- [x] Choose your bill type
- [x] management_information

#### Ticket

[CCCD- Spacing between primary and secondary buttons](https://dsdmoj.atlassian.net/browse/CTSKF-852)

#### Why
To align with GDS default of 15px

#### How

It uses the `govuk-button-group` component from the gov design system that has a margin-right: -15px. 

On the management information page (caseworker) it displays two buttons `Download` and `Update report`. These either display on their own but also together. Therefore I had to change the sequence of the if statement to add the 'govuk-button-group'

--------

#### TODO (wip)
I could not find this one though. 

- [ ] Disbursements

### DIFF
--- a/app/views/case_workers/admin/management_information/_stats_report.html.haml
+++ b/app/views/case_workers/admin/management_information/_stats_report.html.haml
@@ -10,10 +10,11 @@
   %p
     = t('unavailable_report', scope: i18n_scope)
 
-- if report_rec&.started_at.present?
-  = govuk_link_button(t('download_html', scope: i18n_scope, report_name: report_name_html),
-                    case_workers_admin_management_information_download_url(report_type: report_object.name, format: :csv))
+.govuk-button-group
+  - if report_rec&.started_at.present?
+    = govuk_link_button(t('download_html', scope: i18n_scope, report_name: report_name_html),
+                      case_workers_admin_management_information_download_url(report_type: report_object.name, format: :csv))
 
-- if report_object.updatable?
-  = govuk_link_button_secondary(t('update_report_html', scope: i18n_scope, report_name: report_name_html),
-                              case_workers_admin_management_information_generate_url(report_type: report_object.name))
+  - if report_object.updatable?
+    = govuk_link_button_secondary(t('update_report_html', scope: i18n_scope, report_name: report_name_html),
+                                case_workers_admin_management_information_generate_url(report_type: report_object.name))
--- a/app/views/external_users/claims/buttons/_continue.html.haml
+++ b/app/views/external_users/claims/buttons/_continue.html.haml
@@ -1,6 +1,6 @@
 .form-section
   .form-group.form-buttons.button-holder
     - commit = claim.next_step? ? 'commit_continue' : 'commit_submit_claim'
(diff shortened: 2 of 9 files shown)


app/views/external_users/claims/confirmation.html.haml
```diff
@@ -17,5 +17,6 @@
         = render partial: 'shared/disk_evidence_info', locals: { f: claim }
 
       .button-holder
+      .govuk-button-group
```
jrmhaig: Should this be nested inside `button-holder`?

app/views/case_workers/admin/management_information/_stats_report.html.haml
```diff
@@ -10,10 +10,18 @@
   %p
     = t('unavailable_report', scope: i18n_scope)
 
-- if report_rec&.started_at.present?
+- if report_rec&.started_at.present? && report_object.updatable?
```
jrmhaig: How does this work if both the `if` are inside the `govuk-button-group`? This will mean that single buttons are inside a group in the case when one or the other isn't required but I don't think that is a problem.

app/views/external_users/claims/buttons/_continue.html.haml
```diff
@@ -1,6 +1,7 @@
 .form-section
   .form-group.form-buttons.button-holder
     - commit = claim.next_step? ? 'commit_continue' : 'commit_submit_claim'
-    = govuk_button(t('buttons.continue'), id: 'save_continue', name: commit)
+    .govuk-button-group
+      = govuk_button(t('buttons.continue'), id: 'save_continue', name: commit)
 
-    = govuk_button_secondary(t('buttons.save_a_draft'), name: 'commit_save_draft')
+      = govuk_button_secondary(t('buttons.save_a_draft'), name: 'commit_save_draft')
```
jrmhaig, line 6: This is a very minor thing. Now that the buttons are in a group then the blank line above can be removed. This will make it clearer that these are grouped. I think there may be a couple of other examples of this.

jrmhaig (commented):
Looks good. I have added some comments to think about.

app/views/external_users/claims/confirmation.html.haml
Katy600: Thank you @jrmhaig . You are right. I have nested it inside button-holder and it seems to format correctly.

app/views/case_workers/admin/management_information/_stats_report.html.haml
Katy600: @jrmhaig -I have a memory of attempting it like that and the formatting went a bit strange but now it seems to work! So yes that looks so much better. Thank you Joe!

jrmhaig (approved):
:+1:

PR #7443, opened 2024-09-16
Silence CSP alerts for chrome-extensions

mpw5:
#### What

Silence CSP alerts for chrome-extensions

#### Why

Since introducing a Content Security Policy we have seen a large volume of alerts for CSP violations caused by Chrome extensions (indicated by a `source-file` of `chrome-extension`). This results in a lot of noise in the `laa-cccd-alerts` Slack channel which distracts from other alerts.

#### How

Updates the `CspReportsController` so that Slack notifications are not created and sent when violations are caused by Chrome extensions.

### DIFF
--- a/app/controllers/csp_reports_controller.rb
+++ b/app/controllers/csp_reports_controller.rb
@@ -3,13 +3,15 @@ class CspReportsController < ApplicationController
   skip_forgery_protection
 
   def create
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
Seems reasonable

PR #7468, opened 2024-09-20
Format headings by changing heading style and and add spacing

Katy600:
#### What
This ticket implements a different heading size for a better hierarchy on the page and adds padding to space out the sections.
#### Ticket

[CCCD - Group section on a page to visually aid users when making an application](https://dsdmoj.atlassian.net/browse/CTSKF-491)

#### Why
To make the sections more distinguishable on the page - more readable
#### How

- It changes the heading size using form builder-  adding 'size: s' to the legend. 
- It uses the gov design to add spacing.

### DIFF
--- a/app/views/external_users/claims/additional_information/_fields.html.haml
+++ b/app/views/external_users/claims/additional_information/_fields.html.haml
@@ -1,4 +1,4 @@
-%h3.govuk-heading-m{ class: 'govuk-!-margin-top-6' }
+%h3.govuk-heading-m{ class: "govuk-!-padding-top-7" }
   = t('.additional_information')
 
 = f.govuk_text_area :additional_information,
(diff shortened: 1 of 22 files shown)


jrmhaig:
This looks good for the 'Case details' page. Is this PR to include changes on the Graduated fees page too?

Gemfile.lock
```diff
@@ -254,6 +254,7 @@ GEM
     faraday-net_http (3.3.0)
       net-http
     ffi (1.17.0-arm64-darwin)
+    ffi (1.17.0-x86_64-darwin)
```
mpw5: There are a few changes to this file which I wouldn't expect as part of this PR (we're not changing any gems so I'm not sure why the `Gemfile.lock` has changed).

We should probably remove the file from the PR.

mpw5 (approved):
The actual change look good to me. It looks like a unrelated change to the `Gemfile.lock` has crept in somehow - it would be good if you could remove that before merging.

jrmhaig (approved):

PR #7496, opened 2024-09-26
Create a helper drop down for miscellaneous fees 

Katy600:
#### What
Showing the AGFS miscellaneous fees available for the user to select.
Using the GDS details component[Details](https://design-system.service.gov.uk/components/details/)

#### Ticket

[CCCD - Reduce the cognitive complexity of selecting AGFS Miscellaneous fees](https://dsdmoj.atlassian.net/browse/CTSKF-495)

#### Why
The list of AGFS miscellaneous fee types is lengthy. There are also multiple versions of each fee in the list e.g. abuse of process hearings (half day), abuse of process hearings (half day uplift) etc. This makes it difficult for users to see all their choices in one glance, can create usability problems, and ultimately lead to users not claiming for eligible fees.

#### How

- Use an existing `govuk_detail` helper method that implements the GDS details component.
- Create a partial that renders the drop down on the miscellaneous fees page. 
- Create a method that uses an existing method `eligible_misc_fee_types` that generates the relevant fees for that user -  format the results to remove brackets and duplication.

For example:
 "Abuse of process hearings (half day)",
 "Abuse of process hearings (half day uplift)",
 "Abuse of process hearings (whole day)",
 "Abuse of process hearings (whole day uplift)"

 To this: 
"Abuse of process hearings"
- Refactor this into a helper method

### DIFF
--- a/app/views/external_users/claims/misc_fees/advocates/_misc_fee_fields.html.haml
+++ b/app/views/external_users/claims/misc_fees/advocates/_misc_fee_fields.html.haml
@@ -8,6 +8,8 @@
     = link_to_remove_association f, wrapper_class: 'misc-fee-group', class: 'govuk-link govuk-!-display-none fx-numberedList-action' do
       = t('common.remove_html', context: t('.misc_fee'))
 
+    = render partial: 'external_users/claims/misc_fees/advocates/miscellaneous_fees', locals: { claim: claim}
+
     = f.govuk_collection_select :fee_type_id,
       claim.eligible_misc_fee_type_options_for_select.map{ |ft| [ft.first, ft.second, { 'data-unique-code': ft.third[:data][:unique_code] }] },
       :second,
--- a/spec/helpers/case_workers/claims_helper_spec.rb
+++ b/spec/helpers/case_workers/claims_helper_spec.rb
@@ -112,4 +112,58 @@ describe CaseWorkers::ClaimsHelper do
       expect(helper.claim_count).to eq 3
     end
   end
+
+  describe '#format_miscellaneous_fee_names' do
+    subject { format_miscellaneous_fee_names(claim) }
+
+    let(:claim) { create(:claim) }
+
+    before do
+      allow(claim).to receive(:eligible_misc_fee_types).and_return(fee_type)
+    end
+
+    context 'When there are no eligible fee types' do
+      let(:fee_type) { [] }
+
+      it { is_expected.to be_empty }
+    end
+
+    context 'When there are brackets in the fee type description' do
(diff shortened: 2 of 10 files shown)


app/views/case_workers/claims/_miscellaneous_fees.html.haml
```diff
@@ -0,0 +1,4 @@
+= govuk_detail "Types of miscellaneous fees you can claim", id: "#misc-fees-list" do
```
mpw5: Could this text be moved into a translation file (probably `config/locales/en.yml`) and then referenced here?

mpw5: Also I'm not sure that you need `id: "#misc-fees-list"` any more, as we're not using to identify the component in the feature tests.

app/views/external_users/claims/misc_fees/advocates/_misc_fee_fields.html.haml
```diff
@@ -8,6 +8,8 @@
     = link_to_remove_association f, wrapper_class: 'misc-fee-group', class: 'govuk-link govuk-!-display-none fx-numberedList-action' do
       = t('common.remove_html', context: t('.misc_fee'))
 
+    = render partial: 'case_workers/claims/miscellaneous_fees', locals: { claim: claim}
```
mpw5: I think the new partial you've created is in the wrong place. You currently have it in `app/views/case_workers/claims/miscellaneous_fees` however it is being displayed to external users, not caseworkers. I think you probably need to move it to `app/views/external_users/claims/misc_fees/advocates/miscellaneous_fees` and then update the reference here.

mpw5 (changes requested):
I think there are a couple of small changes you could make - nothing major but they would be good to fix before releasing. I'm happy to work on it with you if you like.

spec/helpers/case_workers/claims_helper_spec.rb
```diff
@@ -112,4 +112,68 @@
...
+    end
+
+    context 'When there are brackets in the fee type description' do
+      let(:fee_type) do
+        [instance_double(Fee::MiscFeeType, description: 'Abuse of process hearings (half day)')]
+      end
+
+      before do
+        allow(claim).to receive(:eligible_misc_fee_types).and_return(fee_type)
+      end
```
jrmhaig: For a small cleanup, this can go outside of the contexts as they are identical in most of the contexts. If you decide to do this then for the 'no eligible fee types' context you will need to add `let(:fee_type) { [] }`.

mpw5 (approved):
👍

PR #7515, opened 2024-10-02
[CTSDK-940] Calculated determination correctly when vat_amount includes commas

jrmhaig:
#### What

Ensure that VAT added to the determination calculator for LGFS claims are correctly added to the total when submitted.

#### Ticket

[CCCD - VAT truncated in LGFS calculator](https://dsdmoj.atlassian.net/browse/CTSKF-940)

#### Why

Case workers may add figures in the determination calculator with commas for readability, or simply because they are cut-and-pasted from other applications. These should be dealt with in a consistent manner and in a way that does not inconvenience the user. The `vat_amount` value was getting truncated at the first comma but all other fields are interpreted more sensibly.

#### How

Add `vat_amount` to the list of `numeric_attributes` in the `Determination` model.

### DIFF
--- a/app/models/determination.rb
+++ b/app/models/determination.rb
@@ -16,7 +16,7 @@
 
 class Determination < ApplicationRecord
   include NumberCommaParser
-  numeric_attributes :fees, :expenses, :disbursements
+  numeric_attributes :fees, :expenses, :disbursements, :vat_amount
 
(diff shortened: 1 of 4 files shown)


AlexRyanShep (approved):
Looks good to me!

PR #7567, opened 2024-10-16
Ctskf 497 cccd   reduce the cognitive complexity of uploading files on the supporting evidence page

Katy600:
#### What
Update the upload evidence page for all advocate/litigators journeys to be more GDS compliant and easier to comprehend. Update the page to be more accessible by using up to date components and styles.

- [x] Use GDS tags for the status of the upload  [Tag](https://design-system.service.gov.uk/components/tag/) 
The existing ones have been edited but no new ones have been created - such as 'File to large'
- [x] Use GDS link style for ‘remove’ link [Links](https://design-system.service.gov.uk/styles/links/) 
- [x] Check spacing as per design [CLAIR Taskforce s.28 file](https://www.figma.com/design/ji0NwMxvla3H2x8XhXT4ZN/CLAIR-Taskforce-s.28-file?node-id=1894-1455&t=gGPN5c00Dz8UPsPO-1)

#### Ticket

[CCCD - Reduce the cognitive complexity of uploading files on the supporting evidence page](https://dsdmoj.atlassian.net/browse/CTSKF-497)

#### Why
The current situation takes features from both which is confusing for users. Visually it appears that users can only upload one file, when in fact they are able to drag and drop & upload multiple files at the same time.

#### TODO (wip)

 - [ ] Replace upload component with MOJ multiple file upload component [Upload files - MoJ Design System](https://design-patterns.service.justice.gov.uk/patterns/upload-files/)
 - [ ] Use  GDS error summary to highlight invalid uploads [Error summary](https://design-system.service.gov.uk/components/error-summary/)

### DIFF
--- a/app/views/external_users/claims/supporting_documents/_new.html.haml
+++ b/app/views/external_users/claims/supporting_documents/_new.html.haml
@@ -8,15 +8,19 @@
   - @claim.documents.includes(:document_blob, :converted_preview_document_attachment).each do |document|
     = f.hidden_field :document_ids, multiple: true, value: document.id, id: "claim_document_ids_#{document.id}"
 
-%h3.govuk-heading-m{ class: "govuk-!-padding-top-7" }
-  = t('.upload_supporting_evidence')
+%h3.govuk-heading-m{ class:'govuk-!-static-margin-bottom-2' }
+  = t('.supporting_evidence_docs')
+%span
+  = t('.files accepted')
+%p
+  = t('.maximum_file_size')
 
 = f.govuk_file_field :documents,
-  form_group: { class: 'dropzone' },
+  form_group: { class: 'dropzone govuk-!-static-margin-bottom-4' },
   label: { text: t('.upload_file') },
   multiple: true
 
-= govuk_table(id: 'dropzone-files', class: 'files hidden') do
+= govuk_table(id: 'dropzone-files', class: 'files hidden govuk-!-static-margin-bottom-7') do
   = govuk_table_caption do
     = t('.uploading_docs')
 
--- a/app/webpack/javascripts/modules/external_users/claims/Dropzone.js
+++ b/app/webpack/javascripts/modules/external_users/claims/Dropzone.js
@@ -168,13 +168,13 @@ moj.Modules.Dropzone = {
       error: function (xhr, status, error) {
         const fileName = file.name
         if (status === 'timeout') {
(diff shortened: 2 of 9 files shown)


app/webpack/javascripts/modules/external_users/claims/Dropzone.js
```diff
@@ -161,19 +161,23 @@ moj.Modules.Dropzone = {
...
       }.bind(this),
 
       error: function (xhr, status, error) {
         const fileName = file.name
+        const errorMessage = xhr.responseJSON ? xhr.responseJSON.error : xhr.responseText
 
         if (status === 'timeout') {
-          tableRow.replaceWith(this.notificationHTML(fileName, 'error', 'The server failed to process your file.'))
-          this.status.html('The server failed to process your file.')
+          tableRow.replaceWith(this.notificationHTML(fileName, 'govuk-tag govuk-tag--red word-wrap', 'Upload timed out'))
```
mpw5: One minor refactor you could make to remove some repetition could be to define `govuk-tag govuk-tag--red word-wrap` as a `const` at the start of the `uploadFile` function (eg `const govukError = 'govuk-tag govuk-tag--red word-wrap'`) 

You can then replace the three uses of it in this block of code, so 
`tableRow.replaceWith(this.notificationHTML(fileName, 'govuk-tag govuk-tag--red word-wrap', 'Upload timed out'))` 

becomes 

`tableRow.replaceWith(this.notificationHTML(fileName, govukError, 'Upload timed out'))` etc.

This means if we want to change that style in future we only need to do it one place.

app/views/external_users/claims/supporting_documents/_new.html.haml
```diff
@@ -1,5 +1,5 @@
 - unless @claim.agfs? && @claim.interim?
-  %h2.govuk-heading-l
+  %h2.govuk-heading-l{ class: 'govuk-!-static-margin-bottom-0'}
```
mpw5: Just checking that `0` is the correct margin here? (Not suggesting you need to change it, just a sense-check that it is what is actually intended 🙂)

mpw5 (approved):
LGTM. I've left one question and one suggestion but neither are blockers so feel free to ignore. Providing Kristina is happy with it, I'm happy for you to merge this.

app/webpack/javascripts/modules/external_users/claims/Dropzone.js
Katy600: Thanks Mark

Katy600:
Thanks @mpw5. I have made the changes you suggested.

PR #7598, opened 2024-10-22
[CTSKF-941] Update Google Analytics Tags

mpw5:
#### What

Updates the integration with Google Analytics to use Google Analytics tags instead of legacy Universal Analytics tags.

#### Ticket

[CTSKF-941](https://dsdmoj.atlassian.net/browse/CTSKF-941)

#### Why

We probably should have done it when we moved from UA to GA4 last year 😁 

Using the correct tags will prevent inconsistent data and the potential loss of functionality, and may potentially allow us to take advantage of new features.

#### How

* Updates the `GA_TRACKER_ID` value in the `app-config.yaml` file for all environments.
* Updates the javascript in `_analytics.js.haml` to bring it in line with Google Analytics 4. The tags used by GA4 can be used by both Analytics and Google Tag Manager, which means we can replace the existing javascript code for each of those services with a single snippet which functions for both, and remove redundant `GTM_TRACKER_ID`s. See [here](https://support.google.com/tagmanager/answer/12002338?sjid=1675847859540956513-EU#zippy=%2Cset-up-your-google-tag-in-google-tag-manager-instructions) for the Google guidance for this. The solution was heavily borrowed from the [VCD implementation](https://github.com/ministryofjustice/laa-court-data-ui/blob/1bc627ac2a34bc70d7878a3f2e6f6efb82bc5235/app/views/layouts/_google_analytics.haml).

[CTSKF-941]: https://dsdmoj.atlassian.net/browse/CTSKF-941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.k8s/live/api-sandbox/app-config.yaml
+++ b/.k8s/live/api-sandbox/app-config.yaml
@@ -10,7 +10,7 @@ data:
   RAILS_SERVE_STATIC_FILES: only-presence-required-to-enable
   GRAPE_SWAGGER_ROOT_URL: 'https://api-sandbox.claim-crown-court-defence.service.justice.gov.uk'
   CASEWORKER_API_URL: 'http://cccd-app-service'
-  GA_TRACKER_ID: 'UA-37377084-48'
+  GA_TRACKER_ID: 'G-GK279GHP21'
   MAINTENANCE_MODE: 'false'
   SETTINGS__SLACK__BOT_NAME: 'Injectotron'
   SETTINGS__SLACK__FAIL_ICON: ':katyperry:'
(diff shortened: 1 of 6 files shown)


VinceChiuMOJ (approved):

jrmhaig:
I wonder if we can change app/views/layouts/_analytics.js.haml into just a Javascript file in Webpack now. The only thing erb is being used for now, as far as I can tell, is to add the `nonce=true` to the script tag, which is only required because the code is inline.

***Update:*** OK, I see that it is also picking out environment variables.

PR #7611, opened 2024-10-24
Only allow patch level dependabot updates for Rails

jrmhaig:
#### What

Prevent Dependabot from upgrading Rails beyond 7.0.*

#### Ticket

N/A

#### Why

Major and minor level upgrades of Rails (according to semantic version) are major version upgrades and so require a more detailed process rather than simply installing new gems.

#### How

Add Dependabot configuration to ignore any versions of `rails`, `action*` or `active*` greater than or equal to 7.1.


H3ll3m4 (approved):
🚀

PR #7616, opened 2024-10-24
CTSKF-832_vinceTest1004

VinceChiuMOJ:
### Ticket:
https://dsdmoj.atlassian.net/browse/CTSKF-832

--------

### 241006 updates:
Changed the attachment variable name to attachments (plural for multiple). New attachments (one) can be added and download afterwards. The old attachment did not work anymore. Fixed by updating the name column in active_storage_attachments.

### 241025 updates:
Tried to modified rspec tests to pass. 
Accidentally fixed the issue for remove file link not showing for attachments.

Here are some dirty workaround to get the tests passed for now:
- spec/factories/messages.rb:
Could not make Rack::Test::UploadFile work with attachments, fix by direct attach file for now.

- spec/models/message_spec.rb:
There is no blob in have_many_attached, comment out the test case for now.

- app/webpack/stylesheets/application.scss:
no idea how to fix the error as it is related to the file name. Can it be muted?


jrmhaig:
Changing `attachment` to `attachments` means that the `name` field of related `ActiveStorage::Attachment` instances will need to be modified or users will lose access to previously created documents. Is there a plan for this?

package.json
```diff
@@ -27,6 +27,7 @@
     "@babel/plugin-proposal-private-methods": "^7.18.6",
     "@babel/plugin-transform-runtime": "^7.25.9",
     "@babel/preset-env": "^7.25.9",
+    "@ministryofjustice/frontend": "^2.2.4",
```
jrmhaig, line 30: Is this being used yet? If not I would suggest removing it to reduce the size of this PR.

PR #7665, opened 2024-11-01
Survey Monkey improvements

jrmhaig:
#### What

* Allow multiple Survey Monkey surveys.
* Attach pages in Survey Monkey to collector ids.

#### Ticket

N/A

#### Why

A new survey is going to be set up to get feedback on the court data view page.

When Survey Monkey was first set up there was a single collector id. This works when there is only a single survey, as is currently the case with the feedback form. However, a collector is linked to a specific survey so multiple surveys need multiple collector ids.

#### How

* Refactor the `SurveyMonkeySender` class to allow multiple sub-classes for different forms.
* Update the Survey Monkey library to add a collector to each page.


AlexRyanShep (approved):
👍

PR #7675, opened 2024-11-04
Fix flakey test

jrmhaig:
#### What

Fix one flaky test.

#### Ticket

N/A

#### Why

Some tests in CircleCI sporadically fail with false positives due to edge cases in the test setup. Specifically, one test fails if a randomly created defendant name matches the name used in a defendant created to search.

#### How

Use the `create_defendant_and_rep_order: false` option when creating claims from the factories.


mpw5 (approved):
❤️

PR #7676, opened 2024-11-04
Configure Survey Monkey on API Sandbox

jrmhaig:
#### What

Allow the feedback form to submit responses to Survey Monkey.

#### Ticket

N/A

#### Why

External users have access to API Sandbox and there is a feedback form that they may decide to use. At the moment this is not configured so attempting to submit this form results in an error.

#### How

Set `SURVEY_MONKEY_COLLECTOR_ID` in the `api-sandbox` environment to the same as in the `production` environment.


mpw5 (approved):
👍

PR #7722, opened 2024-11-12
[CTSKF-985] Add beta_tester role for case workers

jrmhaig:
#### What

Create a new 'beta tester' role for case workers.

#### Ticket

[CCCD - Add new case worker role for feature testing](https://dsdmoj.atlassian.net/browse/CTSKF-985)

#### Why

To allow new experimental features to be visible to only a selection of case workers.

#### How

Add `beta_tester` to the list of roles defined in the `CaseWorker` class.

Screenshot 2024-11-12 at 14 54 39

### DIFF
--- a/app/models/case_worker.rb
+++ b/app/models/case_worker.rb
@@ -14,7 +14,7 @@
 class CaseWorker < ApplicationRecord
   auto_strip_attributes squish: true, nullify: true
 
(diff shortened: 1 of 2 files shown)


Katy600 (approved):
It all makes sense to me! :)

mpw5 (approved):

PR #7755, opened 2024-11-19
Log request data in response

jrmhaig:
#### What

Add request data to API response log.

#### Ticket

N/A

#### Why

When requests are sent to the API using JSON in the body then the data is not available to be logged before the requests is processed. All parameters are added to api.request.input during the processing of the request and this is available for logging after. Therefore, to ensure that the correct data is being logged it is added to the log after the request as well as before.

#### How

Create a hash of data for logging both before and after. Use `env['api.request.input']` for getting request parameters for the log if it exists. Otherwise fall back to `env['rack.request.form_hash']` or `env['rack.request.query_hash']`.

### DIFF
--- a/app/interfaces/api/logger.rb
+++ b/app/interfaces/api/logger.rb
@@ -1,15 +1,8 @@
 module API
   class Logger < Grape::Middleware::Base
     def before
-      log_api(:info, 'api-request',
-              { request_id:,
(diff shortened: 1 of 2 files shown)


AlexRyanShep (approved):
Looks good to me!

PR #7766, opened 2024-11-21
Email sanitation

jrmhaig:
#### What

Improve the email address sanitization, including its testing.

#### Ticket

N/A

#### Why

Email addresses are sanitized before adding to logs to avoid unnecessary storage of PII. A bug in this sanitization can cause errors that prevent the service from operating correctly.

#### How

Extract the email sanitization to a separate service that can be tested in isolation from the logger.


mpw5 (approved):
Good work, GitHub Copilot 🤖

PR #7783, opened 2024-11-25
Transfer attachment to attachments on claim

jrmhaig:
#### What

Ensure that attachments that were added to messages are still available with the preparation for the multiple attachments feature.

#### Ticket

[CCCD - Allow for multiple documents on a message](https://dsdmoj.atlassian.net/browse/CTSKF-832)

#### Why

The implementation of multiple attachments will result in the attachments being detached from messages. This is because the name of the relation has changed from 'attachment' to `attachments` and this name is used in the database backing `ActiveStorage::Attachment` to enable [polymorphic associations.](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations)

#### How

`Message.attachment` is reintroduced with `has_one_attached` alongside the new `Message.attachments`.

A new `after_initialize` callback is created to check if `attachment` is attached and, if so, attach it to `attachments`.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -21,7 +21,7 @@ class Message < ApplicationRecord
 
   attr_accessor :claim_action, :written_reasons_submitted
 
-  # has_one_attached :attachment
+  has_one_attached :attachment
   has_many_attached :attachments
 
   validates :attachments,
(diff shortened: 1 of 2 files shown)


app/models/message.rb
VinceChiuMOJ, line 1: Do we need to delete the old attachment?

jrmhaig:
Following discussion, I'm closing this in favour of a different approach.

PR #7900, opened 2024-12-16
Allow the `advocate_email` field for supplementary and interim claims.

jrmhaig:
#### What

Allow the `advocate_email` field for supplementary and interim claims.

#### Ticket

N/A

#### Why

The `advocate_email` field for claims submitted via the API is deprecated in favour of `user_email` but at least one 3rd party client is still using it. It was never enabled for supplementary and interim claims and this causes problems for some providers.

#### How

Replace the `user_email` field with `legacy_agfs_params` in the configuration for these endpoints:

* /api/external_users/claims/advocates/supplementary
* /api/external_users/claims/advocates/supplementary/validate
* /api/external_users/claims/advocates/interim
* /api/external_users/claims/advocates/interim/validate

### DIFF
--- a/app/interfaces/api/logger.rb
+++ b/app/interfaces/api/logger.rb
@@ -24,6 +24,7 @@ module API
         path:,
         creator_email:,
         user_email:,
+        advocate_email:,
         input_parameters: request_data.keys
       }
     end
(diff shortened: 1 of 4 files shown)


mpw5 (approved):
👍

PR #4442, opened 2021-11-18
[CFP-331] Removal of dead codes

imtiazAhmd:
#### What
During the work commenced on the Final Table Migration [CFP-235](https://dsdmoj.atlassian.net/browse/CFP-235), we have identified some dead codes in our CCCD codebase. Dead code adds unnecessary complexity to the codebase. It makes the codebase more difficult to understand and maintain in the long run. It impacts performance and security.

#### Ticket
[CFP-331 Remove all identified potential dead codes](https://dsdmoj.atlassian.net/browse/CFP-331)

--------
#### Documentation (Decision Log)
https://dsdmoj.atlassian.net/l/c/uYjaPHiA

--------

#### TODO
- [x] app/views/external_users/claims/supporting_documents/_summary.html.haml
- [x] app/views/external_users/claims/supporting_evidence_checklist/_summary.html.haml
- [x] app/views/case_workers/admin/case_workers/_claim.html.haml
- [x] app/views/shared/_summary.html.haml
--------
 **Identify the use-cases of `user_message_statuses` and if found not in use, remove references from these files:** 
   - [x] app/controllers/user_message_statuses_controller.rb
   - [x] app/views/user_message_statuses/index.html.haml
   - [x] app/views/user_message_statuses/update.js.erb
   - [x] config/routes.rb
   - [x] spec/controllers/user_message_statuses_controller_spec.rb
--------
**Sanity check on `staging`
   - [x] as `advocate / advocateAdmin`
   - [x] as `litigator / litigatorAdmin`
   - [x] as `caseworker / caseworkerAdmin`

PR #7988, opened 2024-12-30
Rubocop fixes

jrmhaig:
#### What

Fix a few more Rubocop issues.

#### Ticket

N/A

#### Why

General housekeeping.

#### How

Fix `Metrics/MethodLength` in;

* `app/controllers/pages_controller.rb`
* `spec/models/remote/case_type_spec.rb`
* `spec/support/matchers/hash_matchers.rb`

Remove;

* `app/services/message_queue/message_template.rb`. This has not been used since #2231.
* `CaseWorkers::Admin::AllocationsHelper.owner_column_header`. This has not been used since #1901.
* `ClaimsHelper.claim_allocation_checkbox_helper`. This has not been used since #4442.

**Note:** Although these do not appear to be used anymore this should be monitored after merging to ensure that there are no side-effects.


mpw5 (approved):
👍

PR #7990, opened 2024-12-31
CTSKF-1002 - CCCD - Task to copy message attachment (singular) to attachments (plural)

VinceChiuMOJ:
#### What

1. Add a new has_many_attached :attachments to the Message model alongside the existing has_one_attached :attachment
2. Add a after_create callback in the Message model to attach an attachment, if it exists, to attachments.
3. Create and run a Rake task to attach attachments on existing messages to attachments.

#### Ticket
https://dsdmoj.atlassian.net/browse/CTSKF-1002

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -22,6 +22,7 @@ class Message < ApplicationRecord
   attr_accessor :claim_action, :written_reasons_submitted
 
   has_one_attached :attachment
+  has_many_attached :attachments
 
   validates :attachment,
             size: { less_than: 20.megabytes },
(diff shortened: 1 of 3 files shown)


app/models/message.rb
```diff
@@ -107,4 +109,11 @@ def process_written_reasons
   def claim_updater
     Claims::ExternalUserClaimUpdater.new(claim, current_user: sender)
   end
+
+  def duplicate_message_attachment
+    return unless attachment.attached?
+
+    attachment_blob = attachment.blob
+    attachments.attach(attachment_blob) unless attachments.find_by(blob: attachment_blob)
```
jrmhaig: I'm not entirely sure (need to check) but I think it is not possible to attach the same blob twice, in which case this can be reduced to:

```suggestion
    attachments.attach(attachment.blob)
```

However, this is only temporary so it isn't important if it turns out to be more complicated.

jrmhaig (commented):
Looks OK. In `spec/models/message_spec.rb` there is a test for `attachment`. This will need ultimately to be replaced by a similar one for `attachments`, and it could go in alongside it now.

```ruby
it { is_expected.to have_one_attached(:attachment) }
```

app/models/message.rb
VinceChiuMOJ: Yes, I just checked you are right.

AlexRyanShep (approved):
Looks good to me

jrmhaig (approved):

PR #8064, opened 2025-01-16
Ctskf 832 - CCCD - Allow for multiple documents on a message

VinceChiuMOJ:
## CCCD - Allow for multiple documents on a message
https://dsdmoj.atlassian.net/browse/CTSKF-832

--------

### 241006 updates:
Changed the attachment variable name to attachments (plural for multiple). New attachments (one) can be added and download afterwards. The old attachment did not work anymore. Fixed by updating the name column in active_storage_attachments.

### 241025 updates:
Tried to modified rspec tests to pass. 
Accidentally fixed the issue for remove file link not showing for attachments.

### 250113 updates:
Checked out a new branch to refine the changes did before on top of CTSKF-1002.

- spec/models/message_spec.rb:
No idea why there is no change in blob count, will further investigate

### 250122 updates:
Passed all the tests but there are some uncertainties.

- spec/models/message_spec.rb:
Commented out the tests for blob count change. This test may not be applicable when we move on to CTSKF-833 where we create the attachments as documents. Thus the blob will still exist after we remove the message.

- features/step_definitions/messaging_steps.rb:
Not sure if it is OK not using those hooks.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -24,7 +24,7 @@ class Message < ApplicationRecord
   has_one_attached :attachment
   has_many_attached :attachments
 
-  validates :attachment,
+  validates :attachments,
             size: { less_than: 20.megabytes },
             content_type: %w[
               application/pdf
@@ -51,7 +51,10 @@ class Message < ApplicationRecord
 
   after_create :generate_statuses, :process_claim_action, :process_written_reasons, :send_email_if_required,
                :duplicate_message_attachment
-  before_destroy -> { attachment.purge }
+  before_destroy do
+    attachment.purge
+    attachments.purge
+  end
 
   class << self
     def for(object)
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -21,7 +21,6 @@ class Message < ApplicationRecord
 
   attr_accessor :claim_action, :written_reasons_submitted
 
-  has_one_attached :attachment
   has_many_attached :attachments
 
   validates :attachments,
@@ -49,12 +48,8 @@ class Message < ApplicationRecord
 
   scope :most_recent_last, -> { includes(:user_message_statuses).order(created_at: :asc) }
 
-  after_create :generate_statuses, :process_claim_action, :process_written_reasons, :send_email_if_required,
-               :duplicate_message_attachment
-  before_destroy do
-    attachment.purge
-    attachments.purge
-  end
+  after_create :generate_statuses, :process_claim_action, :process_written_reasons, :send_email_if_required
+  before_destroy -> { attachments.purge }
 
   class << self
     def for(object)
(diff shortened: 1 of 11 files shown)


features/step_definitions/messaging_steps.rb
```diff
@@ -41,7 +42,9 @@
 When(/^I upload a file$/) do
   available_docs = Dir.glob "#{Rails.root}/spec/fixtures/files/*.pdf"
   @uploaded_file_path = available_docs.first
-  @external_user_claim_show_page.messages_panel.upload_file(@uploaded_file_path)
+  # @external_user_claim_show_page.messages_panel.upload_file(@uploaded_file_path)
```
mpw5: This can be deleted, assuming it's no longer needed.

features/step_definitions/messaging_steps.rb
```diff
@@ -1,4 +1,5 @@
 When(/^I send a message '(.*?)'$/) do |text|
+  sleep Capybara.default_max_wait_time
```
mpw5: `default_max_wait_time` is currently set to [10 seconds](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/f2aa1101049e2d6a4201609da258a259f1a36dec/features/support/capybara.rb) which seems like quite a long time to have to wait. Have you tested using a shorter wait time to see if that works? Maybe just one or two seconds?

spec/models/message_spec.rb
```diff
@@ -215,8 +215,8 @@
     context 'with an attachment' do
       let(:trait) { :with_attachment }
 
-      it { expect { destroy_message }.to change(ActiveStorage::Attachment, :count).by(-2) }
-      it { expect { destroy_message }.to change(ActiveStorage::Blob, :count).by(-1) }
+      it { expect { destroy_message }.to change(ActiveStorage::Attachment, :count).by(-(message.attachments.count + (message.attachment.attached? ? 1 : 0))) }
+      # it { expect { destroy_message }.to change(ActiveStorage::Blob, :count).by(-message.attachments.count) }
```
mpw5: Is this supposed to be commented out?

app/models/message.rb
```diff
@@ -24,7 +24,7 @@ class Message < ApplicationRecord
   has_one_attached :attachment
```
mpw5: Could this line be removed now? We have `has_many_attached :attachments` - do we also need to allow a message to `have_one_attached: attachment`?

mpw5 (commented):
I've left some comments and questions - happy to discuss if you like 🙂

features/step_definitions/messaging_steps.rb
VinceChiuMOJ: removed

spec/models/message_spec.rb
VinceChiuMOJ: yes

app/models/message.rb
VinceChiuMOJ: removed related codes

features/step_definitions/messaging_steps.rb
VinceChiuMOJ: I kept getting error on this line before, but when I tried to show it to Joe, I cannot replicate it. I will just remove it for now.

mpw5 (approved):
👍

jrmhaig (approved):

PR #8077, opened 2025-01-21
Some pre Rails 7.1 changes

jrmhaig:
#### What

Some small changes that can be made in advance of upgrading Rails 7.1.

#### Ticket

[CCCD: Upgrade Rails to 7.1](https://dsdmoj.atlassian.net/browse/CTSKF-769)

#### Why

Make the Rails 7.1 upgrade a little smoother.

#### How

* Explicitly convert `helper.govuk_warning_text_description` to a string in a test.
* Move a before action callback out of a concern as it is only relevant to some controllers.
* Cater for slight differences between the deprecated Rails specific `sum` method and the core Ruby version.
* Monkey patch `ActionView::Helpers::FormBuilder#as_json` to avoid fatal errors caused by `logstasher`.

### DIFF
--- a/app/controllers/concerns/provider_admin_concern.rb
+++ b/app/controllers/concerns/provider_admin_concern.rb
@@ -2,7 +2,7 @@ module ProviderAdminConcern
   extend ActiveSupport::Concern
 
   included do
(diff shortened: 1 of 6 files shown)


mpw5 (approved):
👍

PR #8128, opened 2025-02-03
[CTSKF-902] Refactor callback actions

jrmhaig:
#### What

Prevent exceptions due to missing callback actions in Rails 7.1.

#### Ticket

[CCCD - Fix missing callback actions](https://dsdmoj.atlassian.net/browse/CTSKF-902)

#### Why

With Rails 7.1 an exception will be raised if a callback in a controller is defined for an action that does not exist. This can happen when a helper is included, and callbacks are defined for actions that may or may not be present. This happens with `PasswordHelper` where callbacks are defined for the `create` action, which is not defined in `SuperAdmins::Admin::SuperAdminsController`.

#### How

The callbacks that are defined in `PasswordHelper` are used to set instance variables related to the parameters of the request. It isn't necessary to manage parameters with callbacks in this way and instead methods (with memoization) can be used instead.

#### ~Note~

~This change is forked off commits that start to install Rails 7.1. These commits should be removed before merging:~

* ~34e517c01f9e9e3caedb651e8eb6da3c571519f5~
* ~2a3298b8058f005d01d0cad261fff013fa43c38f~

~The test failure of `spec/lib/previous_version_of_claim_spec.rb:27` is a known issue with Rails 7.1. See https://dsdmoj.atlassian.net/browse/CTSKF-899~

### DIFF
--- a/app/controllers/concerns/password_helpers.rb
+++ b/app/controllers/concerns/password_helpers.rb
@@ -1,12 +1,6 @@
 module PasswordHelpers
   extend ActiveSupport::Concern
 
-  included do
-    before_action :set_resource_params, only: %i[create update_password]
-    before_action :set_temporary_password, only: :create
-    before_action :set_user_params, only: :update_password
-  end
-
   def update_password
     user = user_for_controller_action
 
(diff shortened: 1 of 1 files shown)


config/initializers/00_extensions.rb
```diff
@@ -42,10 +42,20 @@ class FalseClass
...
-    class FormBuilder
-      include Extensions::FormBuilderExtension
-    end
+# module ActionView
+#   module Helpers
+#     class FormBuilder
+#       include Extensions::FormBuilderExtension
+#     end
+#   end
+# end
```
mpw5: Cant this be removed?

mpw5 (approved):
LGTM. I've left a comment about a comment, but other than that and removing the two Rails 7.1 commits, I think it's good to go.

config/initializers/00_extensions.rb
jrmhaig: That commit it going to be removed before merging as it is only there for Rails 7.1.

mpw5: Oh, of course, I didn't spot that it was part of one of those commits 🤦

PR #8151, opened 2025-02-06
Ctskf 984 cccd   tidy up and standardise 'manage users' table

Katy600:
#### What
Implement GDS styles and styling tweaks to the ‘Manage Users’ page visible to Advocates and Litigators and also the Caseworkers ‘manage caseworkers’ page

- Use [GDS tags](https://design-system.service.gov.uk/components/tag/multiple-tags/) to style inactive/active to make it visually clear which is active or inactive
- Rename ‘Email notifications of messages?’ to ‘Email notifications’
- Rename ‘Manage details’ to ‘Actions' 
- Add govuk-spacing(6); (30px) bottom padding to the ‘create new user’ button as currently there is no space between it and the table

#### How
Adds a govuk_tag helper method to check if the user is active or inactive and then displays the corresponding gov_tag.
#### Ticket

[CCCD - Tidy up and standardise 'Manage Users' table](https://dsdmoj.atlassian.net/browse/CTSKF-984)

#### Why
To make the page more accessible and legible

### DIFF
--- a/app/views/external_users/admin/external_users/index.html.haml
+++ b/app/views/external_users/admin/external_users/index.html.haml
@@ -4,8 +4,9 @@
 = render partial: 'layouts/header', locals: { page_heading: t('.page_heading', provider_name: "#{current_user.provider.name}") }
 
 = content_for :search_html_block do
-  .form-group
-    = govuk_link_to t('.new_user'), new_external_users_admin_external_user_path
+  .div{ class: "govuk-!-margin-bottom-6" }
+    .form-group
+      = govuk_link_to t('.new_user'), new_external_users_admin_external_user_path
 
 = render partial: 'shared/search_form', locals: { search_path: external_users_admin_external_users_path(anchor: 'search-button'), hint_text: t('hint.search_users'), button_text: t('search.users') }
 
--- a/app/views/external_users/admin/external_users/index.html.haml
+++ b/app/views/external_users/admin/external_users/index.html.haml
@@ -4,9 +4,8 @@
 = render partial: 'layouts/header', locals: { page_heading: t('.page_heading', provider_name: "#{current_user.provider.name}") }
 
 = content_for :search_html_block do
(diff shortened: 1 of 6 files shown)


app/views/external_users/admin/external_users/index.html.haml
```diff
@@ -4,8 +4,9 @@
 = render partial: 'layouts/header', locals: { page_heading: t('.page_heading', provider_name: "#{current_user.provider.name}") }
 
 = content_for :search_html_block do
-  .form-group
-    = govuk_link_to t('.new_user'), new_external_users_admin_external_user_path
+  .div{ class: "govuk-!-margin-bottom-6" }
+    .form-group
```
jrmhaig: Does it work if the class is added to the `form-group` instead of wrapping in a new `div`?

lib/govuk_component/tag_helpers.rb
```diff
@@ -8,5 +8,12 @@ def govuk_tag(body = nil, color = nil, **tag_options)
 
       tag.strong(body, **tag_options)
     end
+
+    def govuk_tag_active_user?(user)
+      status = user.active? && user.enabled? ? 'Active' : 'Inactive'
+      tag_class = status == 'Active' ? 'govuk-tag--green' : 'govuk-tag--red'
+
+      content_tag(:strong, status, class: "govuk-tag #{tag_class}")
```
jrmhaig: This could use the `govuk_tag` method, above, such as;

```suggestion
      if user.active? && user.enabled? ? govuk_tag('Active', 'green') : govuk_tag('Inactive', 'red')
```

jrmhaig (commented):
A couple of suggestions to think about.

lib/govuk_component/tag_helpers.rb
Katy600: Oh yes! That is much better!

PR #8167, opened 2025-02-10
[CTSKF-1030] Add new AGFS refuse/reject reasons

mpw5:
#### What

Add new AGFS refuse/reject reasons

#### Ticket

[CTSKF-1030](https://dsdmoj.atlassian.net/browse/CTSKF-1030)

#### Why

Caseworkers require additional reasons to select when refusing or rejecting an AGFS claim, to reduce the use of the 'Other' option.

#### How

All refuse and reject reasons are held in `config/locales/claim_state_transition_reason.en.yml` with code in `/app/models/claim_state_transition_reason.rb` used to select the appropriate options. For refuse reasons, the current behaviour is to select a list of reasons based on the type claim, while for reject reasons the same list is displayed for all claims, regardless of fee scheme or claim type.

This PR:
* adds the required additional refuse reasons to `config/locales/claim_state_transition_reason.en.yml` for each of the following keys: `refused_advocate_claims`, `refused_advocate_supplementary_claims`, `refused_advocate_interim_claims`, `refused_advocate_hardship_claims`;
* adds the required additional reject reasons to `config/locales/claim_state_transition_reason.en.yml`, introducing the concept of different reasons for AGFS and LGFS claims by defining `rejected_advocate_claims` and `rejected_litigator_claims` keys. `rejected_litigator_claims` contains only the existing rejection reasons, `rejected_agfs_claims` is extended to include the new reasons;
* modifies the `reject_reasons_for` method in `/app/models/claim_state_transition_reason.rb` to select reject reasons based on the fee scheme of the claim;
* does some tidying up of the translation file to make it more readable and maintainable in future. All wordings now defined in `en.dictionary.claim_state_transition_reason`, with aliases to enable them to be referred to elsewhere. `en.claim_state_transition_reason` now defines all the relevant claim types and the reasons that should be displayed for both refusals and rejections, using the aliases defined in the dictionary.

(These changes result in some duplication of data in the translation file. It might be possible to DRY this out, possibly by defining sets of common reasons and adding to them as needed based on case type/fee scheme, however the current approach makes it easier to see and configure which reasons are displayed, and making changes may lead to changes in the order in which the reasons are displayed on screen, which may be disruptive to caseworkers. There are also additional changes still to be made for LGFS claims; it may be worth investigating improvements once those changes have been implemented.)

[CTSKF-1030]: https://dsdmoj.atlassian.net/browse/CTSKF-1030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/claim_state_transition_reason.rb
+++ b/app/models/claim_state_transition_reason.rb
@@ -26,7 +26,7 @@ class ClaimStateTransitionReason
     end
 
     def reject_reasons_for(claim)
-      reasons = reasons_for('rejected')
+      reasons = reasons_for("rejected_#{claim.agfs? ? 'advocate' : 'litigator'}_claims")
       reasons.insert(6, reasons_for(:disbursement)) if disbursement_only?(claim)
       reasons.flatten
     end
--- a/config/locales/claim_state_transition_reason.en.yml
+++ b/config/locales/claim_state_transition_reason.en.yml
@@ -56,6 +56,24 @@ en:
       duplicate_claim:
         short: *duplicate_claim_short
         long: *duplicate_claim_long
+      stayed_quashed_indictment:
+        short: &stayed_quashed_indictment_short Second fee for stayed/quashed indictment refused
+        long: &stayed_quashed_indictment_long Second fee not payable for stayed/quashed indictment. More information provided on previous note.
+      predate_rep_order:
+        short: &predate_rep_order_short Hearings not covered by legal aid
+        long: &predate_rep_order_long Hearings predate the rep order so are unable to be paid.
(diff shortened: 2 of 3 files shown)


jrmhaig (approved):
In terms of the code this looks OK. It might be worth at some point looking at how we are creating the messages. This image shows what I am seeing, which includes a lot of duplication:

Screenshot 2025-02-10 at 14 34 40

The duplication appears to come from displaying both the short and long form of the text.

PR #8174, opened 2025-02-11
Ctskf 984 cccd  tidy up and standardise manage users table

Katy600:
What
This ticket focuses on updating the CCCD case worker summary page only (other summary pages on the provider side are still waiting for designed) and implementing the gov design summary cards to display the claim information in a more readable way.

While doing this ticket I realised that some of the Caseworker summary page shares some files that display on the summary page with the external users summary page - both in the shared folder as well as in the external folder. This means I was unable to make the changes in isolation. I therefore temporarily duplicated these files to update the caseworker summary page with the new format only. There is a ticket that aims to consolidate these designs - so hopefully that will undo some of the strange file structure and duplication that I have introduced
[Summary cards](https://design-system.service.mod.gov.uk/components/summary-list/)

Ticket
[CCCD - Implement Summary Page accessibility improvements (caseworker) wrap in summary cards.](https://dsdmoj.atlassian.net/browse/CTSKF-993)

Why
The CCCD summary page displays essential information but suffers from poor usability. To improve user experience and accessibility, we will implement a new design pattern developed from other projects.

Additional thoughts
I followed the design in Figma very closely and sometimes it didn't always match up with the ticket description. I did get rid of the sortable arrows in the travel table - but I think I need to go and check this with Kristina.
There were no designs for when there was nothing to display in a section- so if feels like it doesn't match the rest of the design.

### DIFF
--- a/spec/lib/govuk_component/tag_helpers_spec.rb
+++ b/spec/lib/govuk_component/tag_helpers_spec.rb
@@ -34,4 +34,31 @@ RSpec.describe GovukComponent::TagHelpers, type: :helper do
       end
     end
   end
+
+  describe '#govuk_tag_active_user?' do
+    it 'responds to govuk_tag_active_user?' do
+      expect(helper).to respond_to(:govuk_tag_active_user?)
+    end
+
+    context 'when external user is inactive' do
+      let(:user) { create(:user, :active, :disabled) }
+
+      it 'displays a red inactive tag' do
+        tag_class = 'govuk-tag govuk-tag--red'
+        tag_text = 'Inactive'
+
+        expect(govuk_tag_active_user?(user)).to have_tag(:strong, with: { class: tag_class }, text: tag_text)
+      end
+    end
+
+    context 'when external user is active' do
+      let(:user) { create(:user, :active, :enabled) }
+
+      it 'displays a green inactive tag' do
+        tag_class = 'govuk-tag govuk-tag--green'
+        tag_text = 'Active'
+        expect(govuk_tag_active_user?(user)).to have_tag(:strong, with: { class: tag_class }, text: tag_text)
+      end
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


spec/lib/govuk_component/tag_helpers_spec.rb
```diff
@@ -34,4 +34,31 @@
...
+
+    context 'when external user is inactive' do
+      let(:user) { create(:user, :active, :disabled) }
+
+      it 'displays a red inactive tag' do
+        tag_class = 'govuk-tag govuk-tag--red'
+        tag_text = 'Inactive'
+
+        expect(govuk_tag_active_user?(user)).to have_tag(:strong, with: { class: tag_class }, text: tag_text)
+      end
```
jrmhaig: In helper specs `helper` is made available so that you can call methods on it, such as:

```ruby
helper.govuk_tag_active_user?(user)
```

This will mean that you do not need to `include SharedHelpers` in the helper module.

More broadly, it is good to separate out the test set up from the execution, so you can put `tag_class` and `tag_text` in `let`s and `helper.govuk_tag_active_user?(user)` in a `subject`. Then you can have;

```suggestion
      subject { helper.govuk_tag_active_user?(user) } # Further up, outside the contexts

      let(:tag_class) { 'govuk-tag govuk-tag--red' }
      let(:tag_text) { 'Inactive' }

      it { is_expected.to have_tag(:strong, with: { class: tag_class }, text: tag_text) }
```

Katy600: Thanks @jrmhaig I didn't realise that about the helper - thank you. That makes sense.

Gemfile.lock
```diff
@@ -268,6 +268,7 @@ GEM
     faraday-net_http (3.4.0)
       net-http (>= 0.5.0)
     ffi (1.17.1-arm64-darwin)
+    ffi (1.17.1-x86_64-darwin)
```
mpw5: You'll need to remove the commit which adds these changes, as it will break the pipeline again 😁

mpw5 (approved):
LGTM

PR #8193, opened 2025-02-14
[CTSKF-1029] Add new LGFS refuse/reject reasons

mpw5:
#### What

Add new LGFS refuse/reject reasons

#### Ticket

[CTSKF0-1029](https://dsdmoj.atlassian.net/browse/CTSKF-1029)

#### Why

Caseworkers require additional reasons to select when refusing or rejecting an LGFS claim, to reduce the use of the 'Other' option.

#### How

* Adds the new refuse and reject reasons to `config/locales/claim_state_transition_reason.en.yml`.
* Some complexity is added by the requirement to display one set of refusal reasons for new claims and another set for redeterminations. To handle this, extra functionality is added to the `refuse_reasons_for` method in the`ClaimStateTransitionReason` model.

### DIFF
--- a/app/models/claim_state_transition_reason.rb
+++ b/app/models/claim_state_transition_reason.rb
@@ -32,7 +32,16 @@ class ClaimStateTransitionReason
     end
 
     def refuse_reasons_for(claim)
-      reasons_for("refused_#{claim.class.to_s.demodulize.tableize}")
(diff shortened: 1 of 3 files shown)


jrmhaig (approved):
👍 

This looks good for now. I wonder if there is a better way of managing these lists.

PR #8202, opened 2025-02-17
[CTSKF 833] Allow for multi-file uploads for messages

jrmhaig:
#### What

Allow users to add multiple files to messages on claims.

#### Ticket

[CCCD - Allow multiple attachments to be added to messages](https://dsdmoj.atlassian.net/browse/CTSKF-833)

#### Why

A frequent request from providers is the ability to attach multiple document to messages sent to case workers. When a claim is submitted case workers sometimes request extra evidence documents and this can only be done with messages. If multiple documents are requested then the provider needs to create multiple messages and this is inefficient.

#### How

By design messages previously only permitted a single document. This is because, prior to Active Storage, the attachment attribute referenced a [paperclip](https://github.com/thoughtbot/paperclip) instance and this did not allow for collections. Active Storage, however, provides a link to either a single documents or [collection of documents.](https://guides.rubyonrails.org/active_storage_overview.html#has-many-attached) This has been facilitated by #7990 and this PR introduces a multi-file upload to the front-end from [the Ministry of Justice Frontend](https://design-patterns.service.justice.gov.uk/components/multi-file-upload/)

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -43,6 +43,24 @@ class DocumentsController < ApplicationController
     end
   end
 
+  def upload
+    @document = Document.new(creator_id: current_user.id, document: params[:documents])
+    if @document.save_and_verify
+      render_success_response
+    else
+      render_error_response
+    end
+  end
+
+  def delete
+    # TODO: This does not do any checking to see if the document is owned by the current user
+    @document = Document.find(params[:delete])
+
+    @document.destroy
+
+    render json: { file: { filename: params[:delete] } }
+  end
+
   private
 
   def document
@@ -55,5 +73,22 @@ class DocumentsController < ApplicationController
       :form_id,
       :creator_id
     )
+  end
+
+  def render_success_response
+    render json: {
+      file: { originalname: @document.document.filename, filename: @document.id },
+      success: {
+        messageHtml: "#{@document.document.filename} uploaded"
+      }
+    }, status: :created
+  end
+
+  def render_error_response
+    render json: {
+      error: {
+        message: "#{@document.document.filename} #{@document.errors[:document].join(', ')}"
+      }
+    }, status: :accepted
   end
 end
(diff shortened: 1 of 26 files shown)


app/controllers/documents_controller.rb
```diff
@@ -43,6 +43,22 @@ def destroy
...
+    @document = Document.new(creator_id: current_user.id, document: params[:documents])
+    if @document.save_and_verify
+      render_success_response
+    else
+      render_error_response
+    end
+  end
+
+  def delete
+    document = Document.find_by(id: params[:delete], creator: current_user, claim: nil)
```
mpw5, line 56: I think there might be an issue with the way we're setting `current_user` here. `current_user` is an instance of the `User` model, however the `creator` relation on the `Document` [model](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/jh-CTSKF-833-multi_upload/app/models/document.rb#L5) is to the `ExternalUser` model.

This could be changed to `creator: current_user.persona` here and elsewhere in this class, so that we're using the external user object relating to the user. I've done some brief testing and it doesn't seem to break the functionality, and I _think_ it will fix the failing tests we're currently seeing.

spec/requests/case_workers/claims/court_data_spec.rb
```diff
@@ -6,6 +6,8 @@
     allow(LAA::Cda::ProsecutionCase).to receive(:search).and_return([])
   end
 
+  # after { sign_out user.user }
```
mpw5: Can this be removed now? It's the only change in [this commit](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/8202/commits/617e8f5a9f3c4af07b3e38286a2403b76aa9d39f), so possibly the whole commit could be deleted?

mpw5 (approved):
I think this looks good now. I've left one comment about a rogue comment from a WIP commit, but nothing major.

mpw5 (approved):
Reapproving after changes.

PR #8215, opened 2025-02-21
(wip) Simplify messaging

jrmhaig:
When the claim is in certain states (authorised, part_authorised, refused or rejected when not in interim lgfs claim) an external user can apply for redetermination and add a message. The option to add a message should only be present if the redetermination is selected. Currently this is done by making a request to the server for a dynamically created javascript function to be executed on the page. This PR changes this to put the form on the page as a hidden div that is revealed when the radio button is selected.

#### Ticket

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


AnnaC-C:
Closed due to being redundant

PR #8247, opened 2025-02-28
[CTSKF-970] Use multi-file upload component for supporting evidence 

mpw5:
#### What

Use the MOJ Multi-file Upload component to upload supporting evidence

#### Ticket

[CTSKF-970](https://dsdmoj.atlassian.net/jira/software/c/projects/CTSKF/boards/987)

#### Why

To improve the user experience of uploading files, making it clearer that multiple files can be uploaded and that files can be dragged and dropped. Brings the supporting document page in line with the document upload functionality on the messaging page, and bring CCCD in line with other MOJ applications.

#### How

* replaces the existing `Dropzone` implementation on the supporting evidence screen with the MOJ Design System multi-file upload component
* overrides the component's uploadFile function in `Modules.MultiFileUpload.js`. This is necessary as the `form_id` needs to be passed to the `DocumentsController` via a `FormData` object in order for the document to be attached to the claim
* updates the `upload` controller method to use `document_params` to create the Document object, in order to persist the `form_id`. Note, `form_id` is not present on documents uploaded when sending messages, however this approach works for both types of document
* removes the `create` method from `DocumentsController` as all documents are now created via the `upload` method
* removes `Dropzone` as this is now no longer needed
* rearranges the layout on both the supporting evidence and case summary screens to display the list of uploaded files below the upload component, for consistency with existing functionality
* hides the `Files added` section when it is empty on both the supporting evidence and case summary screens

Other changes: 

* fixes a minor css error on the case summary screen
* reinstates hint text to the upload component on the case summary screen which was removed in a previous change (this may have been intentional - I'm happy to remove this commit if the wording is no longer required)

--------

#### TODO (wip)

 - [x] Finish Jasmine test

[CTSKF-970]: https://dsdmoj.atlassian.net/browse/CTSKF-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/webpack/javascripts/modules/Modules.MultiFileUpload.js
```diff
@@ -1,12 +1,65 @@
...
+
+    const uploadFile = function (file) {
+      this.params.uploadFileEntryHook(this, file)
+      const formData = new FormData()
+      formData.append('document[document]', file)
+
+      if (formId) {
+        formData.append('document[form_id]', formId)
+      }
+
```
jrmhaig: Ideally we do not really want to overwrite what comes from moj-framework as this could result in missing any updates or bug fixes. However, I appreciate that it does not behave as we need it to at the moment.

Perhaps it is worth having;

```suggestion
      this.params.uploadFileParamsHook(this, formData)
```

and then, further down at the definition of `upload`, adding: 

```javascript
      uploadFileParamsHook: function (_uploader, formData) {
        const formIdElement = document.querySelector('#claim_form_id')
        if (formIdElement) {
          formData.append('document[form_id]', formIdElement.value)
        }
      },
```

This then gives us the option of submitting a feature request and/or a pull request to have this hook added.

app/webpack/stylesheets/_multi_file_upload.scss
```diff
@@ -0,0 +1,3 @@
+.moj-multi-file__uploaded-files:has(.moj-multi-file-upload__list:empty) {
```
jrmhaig, line 1: Nice. Could this also be used allow `app/webpack/javascripts/modules/Modules.Messaging.js` and `app/webpack/javascripts/modules/Modules.MultiFileUpload.js` to be cleaned up, where they show and hide the error box in javascript?

Also (sorry for my lack of knowledge of the finer details of CSS) is there a way to inherit from the `govuk-visually-hidden` class?

jrmhaig (approved):
This looks OK to me with my limited experience with Javascript. I've made one suggestion but it is not a blocker.

app/webpack/javascripts/modules/Modules.MultiFileUpload.js
VinceChiuMOJ, line 1: Are using our own copy of uploadFIle() instead of directly getting it from the component js?

app/webpack/stylesheets/_multi_file_upload.scss
mpw5: Yes, I think we can reuse this to replace the javascript in those files. But I'm having issues getting that error box to show at all - I'm not sure if it's a problem with the `uploadFileErrorHook` or just me doing something wrong, but rather than complicate this ticket more I'll tackle that separately.

On your second point, we can do this
```
.moj-multi-file__uploaded-files:has(.moj-multi-file-upload__list:empty) {
  @extend .govuk-visually-hidden
}
```
to use the `govuk-visually-hidden` class, instead of explicitly setting the styling ourselves, which I prefer so I've pushed an update to that commit.

PR #8251, opened 2025-03-03
Add disbursements without VAT section

jrmhaig:
#### What

Put the disbursements without vat back on the the claim summary page.

#### Ticket

N/A

#### Why

As part of the claim summary page refresh the disbursements section was accidentally dropped.

#### How

The partial for display the disbursements has been adjusted so that it can be used for both with and without VAT.


mpw5 (approved):
👍

PR #8252, opened 2025-03-03
Update to show transfer details

jrmhaig:
#### What

Display details of transfer on claim summary.

#### Ticket

N/A

#### Why

The details of the transfer need to be visible to case workers assessing a claim. It appears that different details were being shown accidentally.

#### How

Fix the partial to show the correct details.


mpw5 (approved):
⭐

PR #8274, opened 2025-03-06
Fix logic to display LGFS redet refusal reasons

mpw5:
#### What

The ClaimStateTransitionReason class contains logic used to display specific refuasal reasons for LGFS claims where a redetermination or written reasons have been requested.

Incorrect methods were used to identify such claims, meaning not all expected refusal reasons are displayed. `claim.redetermination?` and `claim.awaiting_written_reasons?` correctly identify a claim based on its state _before_ it has been allocated to a caseworker, however do not work _after_ it has been allocated to a caseworker, which is when the refusal reasons need to be available.

#### How

Replace `claim.redetermination?` and `claim.awaiting_written_reasons?` with `claim.opened_for_redetermination?` and `claim.written_reasons_outstanding?` to fix this, and update tests so that they accurately reflect the state of the claims being tested.

### DIFF
--- a/app/models/claim_state_transition_reason.rb
+++ b/app/models/claim_state_transition_reason.rb
@@ -34,7 +34,7 @@ class ClaimStateTransitionReason
     def refuse_reasons_for(claim)
       reasons = reasons_for("refused_#{claim.class.to_s.demodulize.tableize}")
       if claim.lgfs?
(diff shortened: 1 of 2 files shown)


Katy600 (approved):
You will probably want someone else to take a look - but it makes sense to me!

PR #8281, opened 2025-03-07
Ctskf 994 cccd create a moj sub nav component allocated  attempt 3

Katy600:
#### What
This ticket focuses on creating a sub-nav component to split the assessment over 2 pages putting the messaging service on its own tab. 

- Split assessment summary and messaging onto their own page

- Implement MOJ sub-nav component and move ‘messaging and update status’ to new tab called ‘Update claim status’

- Create 2 tabs called claim status and claim information

#### Ticket
[Create a MOJ sub-nav](https://dsdmoj.atlassian.net/browse/CTSKF-994)

#### Why
To bring design inline with LAA standard design for ‘view a thing’

To make the page more accessible, particularly for dyslexic users

#### How

--------


jrmhaig:
This is what I see at the top of the page for the links to the two tabs:

Screenshot 2025-03-10 at 09 34 21

There should be an indicator as to which tab is being viewed. I believe that the selected option needs to have a thicker line under it.

app/controllers/case_workers/claims_controller.rb
```diff
@@ -26,6 +26,17 @@ def archived; end
 
     def show
       prepare_show_action
+      @active_tab = params[:tab] || 'information'
+    end
+
+    def status
+      prepare_show_action
+      render 'case_workers/claims/claim_status'
+    end
```
jrmhaig: Is this action (and the `information` action below) being used? It looks as though `show` is being used and the `tab` parameter determines which page to show. I think this is preferable to having two extra actions and corresponding endpoints.

app/views/shared/_claim_history.html.haml
```diff
@@ -1,21 +1,22 @@
-- if current_user.persona.is_a?(CaseWorker)
-  - open_messages_url = messages_case_workers_claim_path(claim)
-  - message_controls_url = show_message_controls_case_workers_claim_path(claim)
-- else
-  - open_messages_url = messages_external_users_claim_path(claim)
-  - message_controls_url = show_message_controls_external_users_claim_path(claim)
+#claim-accordion
```
jrmhaig: This partial is used elsewhere (in `app/views/shared/_claim_accordion.html.hml`, which is still used for external users) so this change will affect that. Does the `claim-accordion` id need to be here? Could it be used in the `_claim_status.html.haml`, where this partial is used?

lib/moj_component/sub_navigation_helpers.rb
```diff
@@ -0,0 +1,21 @@
+module MojComponent
+  module SubNavigationHelpers
+    def moj_subnav(tag_options: {}, items: {}, active: nil)
+      tag_options[:class] = 'moj-sub-navigation'
+      tag.nav(**tag_options) do
+        tag.ul(class: 'moj-sub-navigation__list') do
```
Katy600, line 6: Go away

app/views/case_workers/claims/show.html.haml
```diff
@@ -5,6 +5,11 @@
   = render 'errors/error_summary', ep: @claim, error_summary: t('.error_summary')
   = render 'errors/injection_error_summary', ep: claim
   = render 'warnings/injection_warnings_summary', claim: claim
-
   = render partial: 'shared/claim_header', locals: { claim: claim }
-  = render partial: 'shared/new_claim_accordion', locals: { claim: claim, status_disabled: false }
+  = render partial: 'shared/case_summary', locals: { claim: claim }
+  = render 'case_workers/claims/sub_navigation_tab', claim: claim
+
+  - if  @active_tab == :status
```
mpw5: Super nit-picky: you have two spaces between  `if` and  `@active_tab`
```suggestion
  - if @active_tab == :status
```

.gitignore
```diff
@@ -84,3 +84,12 @@ yarn-debug.log*
...
 Brewfile.lock.json
+pg_upgrade_dump_*.custom
+pg_upgrade_dump_*.log
+pg_upgrade_globals.sql
+pg_upgrade_internal.log
+pg_upgrade_server.log
+pg_upgrade_utility.log
+pg_upgrade_dump_globals.sql
+pg_upgrade_dump_globals.sql
+pg_upgrade_dump_globals.sql
```
mpw5: Do these need to be here? If so, can they be simplified to a single line?
```suggestion
pg_upgrade_*
```

spec/views/case_workers/claims/show_spec.rb
```diff
@@ -4,11 +4,22 @@
...
   let(:case_worker) { create(:case_worker) }
+  let(:claim) { create(:claim) }
 
   before do
     initialize_view_helpers(view)
     sign_in(case_worker.user, scope: :user)
     allow(view).to receive(:current_user_persona_is?).and_return(false)
+    params[:tab] = 'information'
+
+    # Set up the controller instance variables
```
mpw5: More nitpicking - I probably wouldn't bother with the comment here, I think it's clear what the subsequent lines are doing so it doesn't add much.

spec/views/case_workers/claims/show_spec.rb
```diff
@@ -322,6 +333,8 @@
     end
 
     context 'when the claim was created before messaging feature released' do
+      # The majority of this test has been moved to the case_workers/claims/_claim_status.html.haml
+      # as the show page has now been split across two tabs claim_status and information_status
```
mpw5: I'd also possibly consider removing this comment - it was helpful to understand the change when reviewing the PR but I'm not sure it needs to be in the code long-term. Maybe it would have been better in the commit message?

features/step_definitions/caseworker_claim_steps.rb
```diff
@@ -79,3 +79,12 @@
 Then(/^the claim I've just updated is no longer in the list$/) do
   expect(@case_worker_home_page).not_to have_content(@claim.case_number)
 end
+
+Then("caseworkers claim message {int} contains {string}") do |position, text|
+
```
mpw5: Nitpick: the empty line could be deleted.

mpw5 (commented):
This mostly looks good to me. I've left some comments about some minor things, but I think you also need to address Joe's comment on `app/views/shared/_claim_history.html.haml` before this can be released.

mpw5 (approved):

features/001_messaging.feature
```diff
@@ -1,37 +1,37 @@
...
-    And I sign out
-    And I sign in as the case worker
-    Then the claim should be visible with 1 new message
-
-    When I open the claim
-    Then the response and uploaded file should be visible
-    Then the page should be accessible
-
-    And I eject the VCR cassette
+Given a "case worker" user account exists
```
jrmhaig: This feature has lost the scenario and the indentation.

app/webpack/stylesheets/application.scss
```diff
@@ -10,6 +10,7 @@ $govuk-fonts-path: "~govuk-frontend/dist/govuk/assets/fonts/";
 @import "~govuk-frontend/dist/govuk/core/typography";
 
 @import "@ministryofjustice/frontend/moj/components/multi-file-upload/_multi-file-upload.scss";
+@import "@ministryofjustice/frontend/moj/components/sub-navigation/_sub-navigation.scss";
```
jrmhaig: For some reason I was having trouble getting the stylesheet for this component to load so the subnav was not appearing correctly. This is probably a problem with my setup. In the process of sorting this out I found that the `_` on the filename is not required so this line can be:

```suggestion
@import "@ministryofjustice/frontend/moj/components/sub-navigation/sub-navigation.scss";
```

Likewise with the previous line and, if these are both changed, then the `// stylelint-disable scss/load-no-partial-leading-underscore` on line 2 could be removed.

app/webpack/stylesheets/application.scss
```diff
@@ -9,8 +8,9 @@ $govuk-fonts-path: "~govuk-frontend/dist/govuk/assets/fonts/";
 @import "~govuk-frontend/dist/govuk/base";
 @import "~govuk-frontend/dist/govuk/core/typography";
 
-@import "@ministryofjustice/frontend/moj/components/multi-file-upload/_multi-file-upload.scss";
-
+@import "@ministryofjustice/frontend/moj/components/multi-file-upload/multi-file-upload.scss";
+@import "@ministryofjustice/frontend/moj/components/sub-navigation/sub-navigation.scss";
+å
```
mpw5: I think this needs to be removed - it's breaking the compilation of assets.

PR #8282, opened 2025-03-07
Implement skipped jasmine tests 

mpw5:
#### What

Four jasmine tests in `Modules.MangementInformation_spec.js` have never been implemented and are marked as TODO.

Additionally, one test in `Helpers.API.Establishments_spec.js` completes with the warning `SPEC HAS NO EXPECTATIONS`. Although it does not fail it is not accurately testing the functionality and needs rewriting.

#### Why

To improve the test coverage of our javascript code (and for me to practice writing more javascript).

#### How

* adds tests for the `blockDisabledLinkClick`, `dateInputEvent`,`disableDownloadButton` and `activateDownload` functions in `Modules.MangementInformation_spec.js`
* this highlights an issue with the `disableDownloadButton` function in the `ManagementInformation` module, which, when the disable button is clicked, adds sets `aria-disabled` to be `true` and then immediately sets it to be `false` again. This removes the instruction to set it to `false`, and tidies up the code to try and introduce some consistency between the `disableDownloadButton` and `enableDownloadButton` functions
* fixes a problem in `modules-amountAssessed_spec.js` where a test overrides the jQuery method `$.fn.addClass` without resetting it, which then pollutes the use of that method in other tests, leading to intermittent failures. This commit replaces that overriding with a spy
* rewrites the `Helpers.API.Establishments.js ...getAsOptions` spec to use `Promise.all` to ensure that all asynchronous calls have completed before the expectations are asserted

### DIFF
--- a/app/webpack/javascripts/modules/case_worker/admin/Modules.ManagementInformation.js
+++ b/app/webpack/javascripts/modules/case_worker/admin/Modules.ManagementInformation.js
@@ -1,7 +1,6 @@
 moj.Modules.ManagementInformation = {
   el: '.fx-dates-chooser',
   download: '#provisional_assessments_date_download',
-  disableDownload: true,
 
   init: function () {
     if ($(this.el).length) {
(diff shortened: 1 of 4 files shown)


jrmhaig (approved):
Thanks. I notice that the download button actually appears active now when it should!

I don't know if it is worth adding to this PR but I notice from the [GDS documentation](https://design-system.service.gov.uk/components/button/#disabled-buttons) that the `govuk-button` class will automatically change the appearance depending on the presence of the `disabled` attribute without needing an extra class.

mpw5:
> I don't know if it is worth adding to this PR but I notice from the [GDS documentation](https://design-system.service.gov.uk/components/button/#disabled-buttons) that the `govuk-button` class will automatically change the appearance depending on the presence of the `disabled` attribute without needing an extra class.

Thanks @jrmhaig. I've pushed an additional commit to remove the extra class.

PR #8300, opened 2025-03-12
Revert file utils change

Katy600:
#### What

Put 'include FileUtils` back into the setup code.

#### Why
It was removed accidentally during a Rails version upgrade to make future upgrades easier.  It is being returned because when running 'bin/setup' it resulted in an error as it was no longer available in the `copy_sample_file` method.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -13,7 +13,7 @@ def copy_sample_file(src, dst)
     puts "\n #{dst} already exists! Skipping..."
   else
     puts "\n Copying #{src} to #{dst}"
-    cp(src, dst)
+    FileUtils.cp(src, dst)
   end
 end
 


.gitignore
```diff
@@ -84,3 +84,14 @@ yarn-debug.log*
...
 # ignore Brewfile changes
 Brewfile.lock.json
+
+# Ignore PostgreSQL upgrade files
+pg_upgrade_dump_*.custom
+pg_upgrade_dump_*.log
+pg_upgrade_dump_globals.sql
+pg_upgrade_internal.log
+pg_upgrade_server.log
+pg_upgrade_utility.log
```
jrmhaig: Is it worth having this?

```suggestion
pg_upgrade_*
```

bin/setup
VinceChiuMOJ, line 1: Change line 16 to FileUtils.cp instead of loading FileUtils globally?

.gitignore
Katy600: Thanks Joe - I will remove it. I wasn't sure what to do with it.

bin/setup
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
 require 'fileutils'
+include FileUtils
```
jrmhaig: A better fix would be to update line 16 to `FileUtils.cp`.

jrmhaig (approved):

PR #8318, opened 2025-03-17
CTSKF-412 - CCCD: Make API smoke tests more resilient

VinceChiuMOJ:
#### What
CCCD: Make API smoke tests more resilient

#### Ticket
https://dsdmoj.atlassian.net/browse/CTSKF-412

#### Why
CCCD contains a set of automated smoke tests for the API. These run every day on a schedule and also as part of the deployment pipeline to production (although a failure does not block the pipeline).

A recent change (edit: recent when this ticket was written - not so recent now!) caused the tests to start failing. This was caused by a change to the order of a list of advocate categories which led to an invalid category for the fee scheme being chosen. A [patch](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5685) has been put in place but the failure highlighted that these tests are fragile and can be improved.

#### How
There are several improvements that can be made: 

- [x] Make the selection of advocate category random (so failures caused by the selection of an invalid category are not hidden) and ensure only valid advocate categories are used. This may also apply to other data items such as offence ids, where the same value is used for all tests.
- [x] The tests contain hard-coded dates in many places. These dates are all in or around May 2015. This will failures in May 2025 as CCCD only allows dates within the last 10 years to be used. These should be updated, potentially to use dynamic dates?
- [ ] The AGFS tests all seem to run using fee schemes 9/10 - consider updating them to use more recent fee schemes? (this may occur naturally as a result of changing dates in the point above)

#### Findings
There are less hardcoded dates failing the tests than expected. With trial and error, it occurs to me that only main_hearing_date is triggering the 10 years limit.

For randomising the advocate category selection, only code related to the [patch](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/5685)  is causing failure. A private fetch_value method is added to get the correct categories list.

There is another ticket CTSKF-1090 created for new tests around new fee schemes.


mpw5 (approved):
Looks good to me! 

One minor change you could make in a few places, is where you have (for example) `Time.zone.today - 90.days`, this could be replaced with `90.days.ago` which is just a little neater. But we use both approaches at the moment anyway, so it isn't a blocker to releasing this.

jrmhaig (approved):

VinceChiuMOJ:
I was trying to change the date to XX.days.ago. Then I remember the reason why I use Time.zone.today instead is that there were many rspec errors regarding to the time format. So I will just stick with it for now.

PR #8324, opened 2025-03-18
[CTSKF-592] Fix expense duplication

mpw5:
#### What

When duplicating an expense claim, the location and mileage rate are not selected on the new claim. This fix ensures that the location and mileage are correctly duplicated. 

NB: the date of the expense is not duplicated. It appears this is the intended functionality rather than a bug, possibly because it is likely that expense claims will usually be incurred on different dates, so no changes have been made to alter this behaviour.

#### Ticket

[CCCD: Court/Prison not copied when duplicating expense](https://dsdmoj.atlassian.net/browse/CTSKF-592)

#### Why

To improve the user experience, and save time.

#### How

* sets a slight delay (50ms) on the selection of the location to allow the drop down enough time to populate
* selects the mileage radio button by `value` instead of by `id`, which was consistently failing
* adds tests for the affected functions.

### DIFF
--- a/app/webpack/javascripts/modules/external_users/claims/DuplicateExpenseCtrl.js
+++ b/app/webpack/javascripts/modules/external_users/claims/DuplicateExpenseCtrl.js
@@ -69,12 +69,14 @@ moj.Modules.DuplicateExpenseCtrl = {
 
     this.setInputValue($el, '.fx-travel-location input', data.location)
 
-    // select the option by the data.location value
-    $el.find('.fx-establishment-select select option').filter(function (idx, el) { // eslint-disable-line
-      if ($(el).text() === data.location) {
-        $(el).prop('selected', true)
-      }
-    })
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
I wondered if it were possible to get a notification sent back from the location selection box to indicate it was ready. However, that would probably get overly complicated and a 50ms wait is probably good enough.

mpw5:
> I wondered if it were possible to get a notification sent back from the location selection box to indicate it was ready. However, that would probably get overly complicated and a 50ms wait is probably good enough.

Thanks @jrmhaig, I did start to look at sending some sort of notification to indicate that the list of locations was loaded, but I couldn't get it to work. Probably down to my javascript abilities as much as anything. The delay isn't ideal but is short enough that it shouldn't disrupt users, and it's an improvement overall, so I'll release this as it is.

PR #8358, opened 2025-03-25
Future proof tests by dynamically calculated dates so only dates with…

Katy600:
Dynamically calculated dates so only dates within the last 10 years are used.

#### What
CCCD contains suites of rspec and cucumber tests, which run as part of the deployment pipeline.

These tests contain hard-coded dates in many places, often using dates in 2015 or subsequent years. CCCD only allows dates within the last 10 years to be used.

#### What I have not done:

In CCCD their are lots of json files which contain 2015 dates
````Claim-for-Crown-Court-Defence/spec/examples/exported_claim.json````

That look a bit like this:

````
{
    "claim": {
      "creator_email": "advocateadmin@example.com",
      "advocate_email": "advocate@example.com",
      "case_number": "A20161234",
      "case_type_id": 1,
      "first_day_of_trial": "2015-06-01",
      "estimated_trial_length": 3,
      "actual_trial_length": 3,
      "trial_concluded_at": "2015-06-03",
      "advocate_category": "QC",
      "offence_id": 1,
      "court_id": 1,
      "cms_number": "12345678",
      "additional_information": "string",
      "apply_vat": true,
      "trial_fixed_notice_at": "2015-06-01",
      "trial_fixed_at": "2015-06-01",
      "trial_cracked_at": "2015-06-01",
      "defendants": [
      
````
I have left those. Should I be converting them too though? 

#### Ticket
[CCCD: Future-proof tests](https://dsdmoj.atlassian.net/browse/CTSKF-1090)

#### Why
 A failure in any of these tests blocks the pipeline, meaning no changes to CCCD can be deployed.

#### How

Set date dynamically to 1 year previously 
 `date = (Time.zone.today - 1.year).strftime('%d-%m-%Y')`

### DIFF
--- a/features/claims/litigator/scheme_nine/transfer_claim_draft_submit.feature
+++ b/features/claims/litigator/scheme_nine/transfer_claim_draft_submit.feature
@@ -15,7 +15,7 @@ Feature: Litigator partially fills out a draft transfer claim, then later edits 
     Then I choose the litigator type option 'New'
     And I choose the elected case option 'No'
     And I select the transfer stage 'Before trial transfer'
-    And I enter the transfer date '2015-05-21'
+    And I enter the transfer date 3 years ago
     And I select a case conclusion of 'Cracked'
 
     And I click "Continue" in the claim form
@@ -125,7 +125,7 @@ Feature: Litigator partially fills out a draft transfer claim, then later edits 
     Then I choose the litigator type option 'New'
     And I choose the elected case option 'Yes'
     And I select the transfer stage 'Before trial transfer'
-    And I enter the transfer date '2015-05-21'
+    And I enter the transfer date 3 years ago
 
     When I click "Continue" in the claim form
     Then I should see a page title "Enter case details for litigator transfer fees claim"
--- a/features/step_definitions/litigator_transfer_claim_steps.rb
+++ b/features/step_definitions/litigator_transfer_claim_steps.rb
@@ -23,6 +23,11 @@ And(/^I enter the transfer date '(.*)'$/) do |date_string|
   @litigator_transfer_claim_form_page.transfer_detail.transfer_date.set_date date_string
 end
 
+And('I enter the transfer date {int} years ago') do |years|
+  date = years.years.ago.strftime('%d-%m-%Y')
+  @litigator_transfer_claim_form_page.transfer_detail.transfer_date.set_date date
+end
+
 And(/^I select a case conclusion of '(.*)'$/) do |name|
   @litigator_transfer_claim_form_page.wait_until_case_conclusion_visible
   @litigator_transfer_claim_form_page.case_conclusion.choose_autocomplete_option(name)
--- a/spec/controllers/external_users/advocates/claims_controller_spec.rb
+++ b/spec/controllers/external_users/advocates/claims_controller_spec.rb
@@ -2,6 +2,8 @@ require 'rails_helper'
 
 RSpec.describe ExternalUsers::Advocates::ClaimsController do
   let!(:advocate) { create(:external_user, :advocate) }
+  let(:date) { Time.zone.today - 1.year }
+  let(:trial_concluded_date) { date + 2.days }
 
   before { sign_in advocate.user }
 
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: features/claims/litigator/scheme_nine/transfer_claim_draft_submit.feature [hunk 1]
```diff
@@ -15,7 +15,7 @@ Feature: Litigator partially fills out a draft transfer claim, then later edits
     Then I choose the litigator type option 'New'
     And I choose the elected case option 'No'
     And I select the transfer stage 'Before trial transfer'
-    And I enter the transfer date '2015-05-21'
+    And I enter the transfer date '#{date.year}-05-21'
```
jrmhaig: It may be better to have a step 'I enter the transfer date (.*) year(s) ago'. It is a bit confusing having `date` defined in the step definition and then used here.

spec/controllers/external_users/advocates/claims_controller_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe ExternalUsers::Advocates::ClaimsController do
   let!(:advocate) { create(:external_user, :advocate) }
+  let!(:date) { Time.zone.today - 1.year }
```
jrmhaig: This doesn't need a `!` on the `let`. `let` will only evaluate when the variable is first used while `let!` ensures that the value is always calculated.

spec/controllers/external_users/advocates/claims_controller_spec.rb
```diff
@@ -525,14 +525,14 @@ def valid_claim_fee_params
...
-      'first_day_of_trial(1i)' => '2015',
+      'first_day_of_trial(3i)' => date.day,
+      'first_day_of_trial(2i)' => date.month,
+      'first_day_of_trial(1i)' => date.year,
       'estimated_trial_length' => '2',
       'actual_trial_length' => '2',
-      'trial_concluded_at(3i)' => '15',
-      'trial_concluded_at(2i)' => '05',
-      'trial_concluded_at(1i)' => '2015',
+      'trial_concluded_at(3i)' => date.day + 2,
```
jrmhaig: This will cause problems when `date.day` is within 2 days of the end of the month. It is safer to use another memoized variable:

```
let(:trial_concluded_date) { date + 2.days }
```

features/claims/litigator/scheme_nine/transfer_claim_draft_submit.feature
Katy600: Yes - that is a much better idea. :)

features/step_definitions/litigator_transfer_claim_steps.rb
```diff
@@ -23,6 +23,11 @@
   @litigator_transfer_claim_form_page.transfer_detail.transfer_date.set_date date_string
 end
 
+And(/^I enter the transfer date (.*) years ago$/) do |years|
```
jrmhaig: To get `years` as an integer the step can be defined without a regular expression:

```suggestion
And('I enter the transfer date {int} years ago') do |years|
```

The documentation is [here.](https://cucumber.io/docs/cucumber/step-definitions/?lang=ruby)

You can then create a date as `years.years.ago` instead of relative to `Time.zone.today`.

features/claims/litigator/scheme_nine/transfer_claim_draft_submit.feature
```diff
@@ -15,7 +15,7 @@ Feature: Litigator partially fills out a draft transfer claim, then later edits
     Then I choose the litigator type option 'New'
     And I choose the elected case option 'No'
     And I select the transfer stage 'Before trial transfer'
-    And I enter the transfer date '2015-05-21'
+    And I enter the transfer date '3' years ago
```
jrmhaig: The quote marks are being picked up in the step definition so `years` is becomes `'3'` and this doesn't convert correctly to an integer.

```suggestion
    And I enter the transfer date 3 years ago
```

jrmhaig (commented):
I've written a couple of comments. Let me know if it isn't clear.

jrmhaig (approved):

PR #8380, opened 2025-03-28
Configure Rubocop for Ruby 3.4

jrmhaig:
#### What

Configure Rubocop for Ruby 3.4

#### Ticket

N/A

#### Why

Rubocop is configured to use new features of each Ruby version.

There are also a few fixes and an update to the to-do file.

#### How

Set `TargetRubyVersion`


mpw5 (approved):
👍

PR #8400, opened 2025-04-02
Revert pr 8281

Katy600:
#### What
This ticket reverts the implementation of the sub-nav component.

The previous ticket split the assessment over 2 pages putting the messaging service on its own tab - it needs to be reverted back.

#### Ticket
[Revert - CCCD - Implement Summary Page accessibility improvements (caseworker) create a MOJ sub-nav component - allocated](https://dsdmoj.atlassian.net/browse/CTSKF-1165)

#### Why
Caseworkers do not like the claim status and claim information in two tabs, and it is slowing down the assessment process


jrmhaig (approved):
Thanks.

I see that the very first commit is rolling back an update to an npm library. This isn't a major problem as that library will get updated again ultimately but you might like to remove the commit before merging.

PR #8415, opened 2025-04-04
Uncomment the rails active record commit_transaction_on_non_local_return

Katy600:
#### What
Set the Rails.application.config.active_record.commit_transaction_on_non_local_return setting as the default value for Rails 7.1.

#### Ticket

[CCCD - Set active_record.commit_transaction_on_non_local_return](https://dsdmoj.atlassian.net/browse/CTSKF-1147)

#### Why
To align with with Rails 7.1 update


jrmhaig (approved):
:+1:

PR #8443, opened 2025-04-15
Add Google Analytics 4 to CSP

mpw5:
#### What

Updates the content security policy to allow Google Analytics 4 domains following documentation here: https://developers.google.com/tag-platform/security/guides/csp#google_analytics_4_google_analytics

#### Why

An attempt to reduce some of the noise in the #laa-cccd-alerts slack channel.

Before, we frequently see these alerts:

image

This change means the alerts no longer occur. This won't resolve all CSP alerts in that channel, but will reduce the volume.


jrmhaig (approved):
:+1:

PR #8462, opened 2025-04-22
Modify the location column to be a status column

Katy600:
#### What
Implement GDS styles and styling tweaks to the ‘Manage Users’ page for Caseworkers

Use [GDS tags](https://design-system.service.gov.uk/components/tag/multiple-tags/) to style inactive/active to make it visually clear which is active or inactive

Rename ‘Manage details’ to ‘Actions' 

#### Why

To make the page more accessible and legible

#### Ticket

[CCCD - Tidy up and standardise 'Manage Users' table - Caseworkers page](https://dsdmoj.atlassian.net/browse/CTSKF-984)

#### How

Modify the location column to be a status column
Edit the text to display updated content
Add the tag helper method to display the status tag Remove the Inactive text from the Actions column when inactive Put the Edit and Delete links under one another

### DIFF
--- a/app/views/case_workers/admin/case_workers/index.html.haml
+++ b/app/views/case_workers/admin/case_workers/index.html.haml
@@ -18,7 +18,7 @@
 
       = govuk_table_thead_collection [t('.surname'),
       t('.name'),
-      t('.location'),
+      t('.status'),
       t('.actions')]
 
       = govuk_table_tbody do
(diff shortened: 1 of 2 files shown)


app/views/case_workers/admin/case_workers/index.html.haml
```diff
@@ -27,13 +27,10 @@
             = govuk_table_td('data-label': t('.surname')) { case_worker.user.last_name }
 
             = govuk_table_td('data-label': t('.name')) { case_worker.user.first_name }
-
-            = govuk_table_td('data-label': t('.location')) { case_worker.location.name }
+            = govuk_table_td('data-label': t('.status')) { govuk_tag_active_user?(case_worker) }
```
mpw5: I get an error when I run this locally. I think the problem is this line.

`govuk_tag_active_user?(case_worker)` results in `enabled?` being called on `case_worker`, which causes a failure. This is because `case_worker` is an instance of `Caseworker`, which does not define `enabled?`. That method actually expects an instance of `User` to be passed to it, which you can do like this:

```suggestion
            = govuk_table_td('data-label': t('.status')) { govuk_tag_active_user?(case_worker.user) }
```

mpw5 (changes requested):
I think there's a bug in this - I've added a suggestion which I think will resolve it, but let me know if you want to discuss it.

app/views/case_workers/admin/case_workers/index.html.haml
Katy600: Thanks @mpw5. That makes sense! :)

mpw5 (approved):
👍

PR #8467, opened 2025-04-24
[CTSKF-1061] Use govuk-components gem for tables

mpw5:
#### What

Replaces the current helper methods for creating tables with helpers from the [govuk-components](https://govuk-components.netlify.app/components/table/) gem.

#### Ticket

[CTSKf-1061](https://dsdmoj.atlassian.net/browse/CTSKF-1061)

#### Why

To standardise our approach to using GOVUK components and reduce the amount of custom code in CCCD.

#### How

* for every table in CCCD, every call to the current helper method is replaced with the equivalent helper method from the gem
* old helper methods and tests are removed


app/views/pages/api_release_notes.html.haml
```diff
@@ -300,12 +270,11 @@
...
-      %br/
       A list of valid Case Stages (to be used when creating a hardship claim) can be retrieved via:
-      = govuk_table do
-        = govuk_table_row do
-          = govuk_table_td('data-label': "Api") do
-            = 'GET /api/case_stages'
+
+      %ul.govuk-list.govuk-list--bullet
+        %li
+          GET /api/case_stages
```
Katy600, line 277: Hi @mpw5, just checking as it looks very different from everything else.

mpw5: Ah, good spot. I should have put something in the commit message to explain this.

I took a slightly different approach here (and a few other places in this file) where a table was being used just to show a single line. Rather than create a new table, which felt a bit redundant, I just used a bullet point instead.

This is what it looked like before:

image

and this is what it looks like now:

image

app/views/pages/api_release_notes.html.haml
```diff
@@ -1347,131 +1188,95 @@
...
+            = row.with_cell(html_attributes: { 'data-label': "Action" }, text: 'Validate advocate claim')
+            = row.with_cell(html_attributes: { 'data-label': "Endpoint URL" }, text: 'POST /api/external_users/claims/validate')
+          = body.with_row do |row|
+            = row.with_cell(html_attributes: { 'data-label': "Action" }, text: 'Create advocate claim')
+            = row.with_cell(html_attributes: { 'data-label': "Endpoint URL" }, text: 'POST /api/external_users/claims')
+          = body.with_row do |row|
+            = row.with_cell(html_attributes: { 'data-label': "Action" }, text: 'Validate litigator final claim')
+            = row.with_cell(html_attributes: { 'data-label': "Endpoint URL" }, text: 'POST /api/external_users/claims/final/validate')
+          = body.with_row do |row|
+            = row.with_cell(html_attributes: { 'data-label': "Action" }, text: 'Create litigator final claim')
```
Katy600, line 1207: Hi @mpw5 I am not sure if this is worth pointing out but I can't find `Create litigator final claim' in the old code? Just wanted to check it was added intentionally. :)

mpw5: Eagle eyes! That was me fixing a mistake I spotted as I went along. The original text was `Create litigator interim claim` but it was a link to `POST /api/external_users/claims/final` which is used to create final claims (not interim). I've just updated the text. (There's another entry a few rows down for `Create litigator interim claim` which has the correct link).

app/views/pages/api_release_notes.html.haml
Katy600: Oh that makes sense - thanks @mpw5 :)

app/views/pages/api_release_notes.html.haml
Katy600: Thanks Mark - that is a good spot! :)

Katy600 (approved):
I have approved it @mpw5 but it is a lot isn't it!! So you might want to get another set of eyes on it. :)

app/views/case_workers/admin/allocations/_allocation.html.haml
```diff
@@ -42,14 +42,15 @@
...
-      = t('.table_summary')
-
-    = govuk_table_thead_collection ['',
-      t('.case_number'),
-      t('.court'),
-      t('.defendents'),
-      t('.type'),
-      t('.submitted'),
-      t('.total')]
+  = govuk_table(classes: 'app-table--responsive app-jq-datatable', html_attributes: { style: 'width: 100%' }) do |table|
```
jrmhaig: Is it necessary now to explicitly set `style: 'width: 100%'`? I may be wrong but I think it is included in the `govuk-table` class.

app/views/shared/summary/expenses/index.html.haml
```diff
@@ -1,42 +1,33 @@
 %h3.govuk-heading-m
   = t(".header_#{vat ? 'with_vat' : 'without_vat'}")
 
-= govuk_table(class: current_user.case_worker? ? 'expenses-data-table' : nil, style: 'width:100%') do
-  = govuk_table_caption(class: 'govuk-visually-hidden') do
-    = t('.caption')
+= govuk_table(classes: current_user.case_worker? ? 'expenses-data-table app-table--responsive' : 'app-table--responsive', html_attributes: { style: 'width:100%' }) do |table|
```
jrmhaig: Similar to another comment. Is `style: 'width: 100%'` necessary here? I think it is included in the `govuk-table` class.

jrmhaig (approved):
I'm happy to approve this. I have added a comment regarding a small clean-up that may be possible.

app/views/case_workers/admin/allocations/_allocation.html.haml
mpw5: Good point. I took the approach of changing as little about the tables as possible, but you're right that the width is set to 100% by the table class, so I've pushed a commit to remove the two instances you flagged.

VinceChiuMOJ (approved):

PR #8527, opened 2025-05-07
Ctskf 1055 cccd migrate notification banner component

Katy600:
#### What
To update the customized notification banner to use the new govuk component. 

#### Ticket
[CCCD: Migrate notification banner component to use govuk-components gem](https://dsdmoj.atlassian.net/browse/CTSKF-1055)

#### Why
To improve maintainability of the application

#### How
Specify the parameters title_text and text. 

Add `govuk_notification_banner(title_text: "Important", text: "You have 7 days left to send your application.")`


app/views/case_workers/admin/allocations/_allocation.html.haml
```diff
@@ -1,5 +1,6 @@
 #allocation
-  = govuk_notification_banner('Allocation', nil, class: 'govuk-!-display-none') unless !flash.empty?
+  = govuk_notification_banner(title_text: 'Allocation', success: true) do |nb|
```
Katy600: Go away

app/views/layouts/_flashes.html.haml
```diff
@@ -1,3 +1,4 @@
 - flash.each do |key, value|
   - unless %w{hidden ga timedout}.include? key
-    = govuk_notification_banner(key.capitalize, value, class: "govuk-notification-banner--#{key}")
+    = govuk_notification_banner(title_text: key.capitalize, success: key == 'alert', classes: "auto-dismiss-banner") do |nb|
```
mpw5: I don't think this is working quite right - when I log in (or log out) I get the banner in blue, when previously it was green. 

image

I think this is because the `key` in these cases is `notice` not `alert`, so `success` is set to `false`. Changing `key == 'alert'` to `key == 'notice'` might fix this, but I'm not sure what affect it might have elsewhere...

mpw5 (commented):
Mostly looks good. I've left one comment about the colour of the banner that appears when you log in/out of the system, which I think needs fixing. I'll be happy to approve once that's addressed 👍

app/views/layouts/_flashes.html.haml
Katy600: Oh dear - thanks Mark. I thought I got those the right way around. I will switch it.  :)

mpw5 (approved):
LGTM!

PR #8544, opened 2025-05-13
Add FCMH to supplementary claim options

jrmhaig:
#### What

Allow 'further case management hearings' fees (FCMH) to be claimed on AGFS supplementary claims.

#### Ticket

[Add Further Case Management Hearings (FCMHs) to the list of Supplementary Claims in CCCD](https://dsdmoj.atlassian.net/browse/CTSKF-1186)

#### Why

Providers have requested that this fee should be added to the list of claims that can be added to supplementary claims. Currently they will add a fee of a different hearing type and add a note. This creates more work for caseworkers, increases the time required to assess a claim and adds the possibility of errors.

#### How

Add `MIFCM` to the `AGFS_SUPPLEMENTARY_SHARED_TYPES` constant in `Fee::MiscFeeType`.

Note that FCMH fees were introduced with AGFS fee scheme 10.


VinceChiuMOJ (approved):
Looks good.

PR #8547, opened 2025-05-14
update node version to 18.20.0

ziaxgit:
#### What
Update minimum required node version

#### Ticket
Not created one

#### Why
The previous version 18.12.0 was no longer compatible

#### How
Updated .nvmrc file to use to version 18.20.0 instead of 18.12.0


Katy600 (approved):

jrmhaig (approved):
Thanks.

PR #8571, opened 2025-05-20
Temporarily disable tests due to hardwired dates

jrmhaig:
#### What

Disable some failing tests.

#### Ticket

N/A

#### Why

Representation order dates more than 10 years old are invalid and this results in some tests failing. Rather than simply updating these dates it may be useful to refactor these tests so that they are no longer time dependent. This is a larger piece of work so this fix is to get the pipeline working again.

#### How

Set six tests to 'pending' with a note.


ziaxgit:
LGTM.

mpw5 (approved):

PR #8591, opened 2025-06-02
CTSKF 1223 password strength

ziaxgit:
#### What
This PR changes the character limit for passwords to 12 - 64 characters following NIST guidelines, and introduces some UI changes to provide advice to users when setting new passwords

#### Ticket

[CTSKF-1223](https://dsdmoj.atlassian.net/browse/CTSKF-1223)

#### Why

Increase security of CCCD password policy in preparation for rotation of all CCCD passwords

[CTSKF-1223]: https://dsdmoj.atlassian.net/browse/CTSKF-1223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/devise/passwords/edit.html.haml
+++ b/app/views/devise/passwords/edit.html.haml
@@ -10,6 +10,13 @@
 
       = f.hidden_field :reset_password_token
 
+      = govuk_inset_text do
+        %p
+          =t('.password_help')
+
+      %p
+        =t('.password_length')
+
       = f.govuk_password_field :password,
         label: { text: t('.new_password') },
         autofocus: true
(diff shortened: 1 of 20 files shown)


mpw5 (approved):
Looks good to me! 

You have one failing test still, in `spec/models/user_spec.rb` - I think you've changed the format of the error from `Password must be at least 8 characters` to `must be at least 8 characters` (to remove the issue with 'password' appearing twice in the message), so you'll need to update the test to match.

Once that's done you should be good to go though 👍

mpw5 (approved):
Re-approving after changes.

PR #8603, opened 2025-06-03
Fix version of rspec-mocks

jrmhaig:
#### What

Prevent rspec-mocks from upgrading.

#### Ticket

N/A

#### Why

Version 3.13.3 of the `rspec-mocks` fixes a bug that some of our tests depend on. A longer term solution is to refactor our tests but for the moment the gem is pinned.

See https://github.com/rspec/rspec-mocks/pull/1596 and #8594

#### How

Pin the version of the gem in `Gemfile`.


mpw5 (approved):
👍

PR #8610, opened 2025-06-06
Update footer to use the new component 

Katy600:
#### What
This PR updates the customised footer helper method with the govuk-components gem footer method

#### Ticket

[CCCD: Adopt govuk-components gem for footer](https://dsdmoj.atlassian.net/browse/CTSKF-1065)

#### Why

We can use the govuk-components gem to simplify the CCCD codebase and improve maintainability of the application.

#### How

It updates `app/views/layouts/_footer.html.haml` and `app/views/grape_swagger_rails/application/index.html.haml` separately.  For some reason in the index it uses hard coded urls instead of ruby generated ones. It also doesn't contain a research link. So unfortunately it can't be reused. 

It uses the [footer with meta links that have custom HTML attributes](https://govuk-components.netlify.app/components/footer/#footer-with-meta-links-that-have-custom-html-attributes) that allow you to add the href links to a meta_items 

#### Issues
In the original footnote the 'Build by' is not part of the link so not underlined.  
Screenshot 2025-06-06 at 14 04 38

However in the current updated version it is all underlined as a href link.  I couldn't find an option in the component to allow only parts of the text to be underlined. 

I tried removing the href link from the meta data table and having the link solely in transaction file but that caused an error. 

```service_owner_html: <span>Built by</span> <a href="https://mojdigital.blog.gov.uk/"><abbr title="Ministry of Justice">MOJ Digital Services</abbr></a>```

```- meta_items = [{ text: t('.cookies'), href: cookies_path, },
    { text: t('.contact'), href: contact_us_page_path, },
    { text: t('.accessibility'), href: accessibility_statement_path, },
    { text: t('.terms_and_conditions'), href: tandcs_page_path },
    { text: t('.feedback'), href:  new_feedback_path},
    { text: t('.research'), href: Settings.research_panel_url},
    { text: t('.service_owner_html'),  REMOVE LINK HERE},
    ]
```

```controller` can't be used during initialization```

So it expects to see this href link when the component is initialized. So I am not sure the best way to get around this.

### DIFF
--- a/app/views/grape_swagger_rails/application/index.html.haml
+++ b/app/views/grape_swagger_rails/application/index.html.haml
@@ -97,42 +97,14 @@
 
         = yield
 
-    -# = render partial: 'layouts/footer'
     -# renderd partial err: undefined local variable or method `contact_us_page_path'
-    %footer.govuk-footer
-      .govuk-width-container
-        .govuk-footer__meta
-          .govuk-footer__meta-item.govuk-footer__meta-item--grow
-            %h2.govuk-visually-hidden
-              = t('layouts.footer.support_link')
 
-            %ul.govuk-footer__inline-list
-              %li.govuk-footer__inline-list-item
-                = govuk_footer_link_to t('layouts.footer.cookies'), 'http://www.gov.uk/help/cookies'
+    - meta_items = [{ text:  t('layouts.footer.cookies'), href: 'http://www.gov.uk/help/cookies', },
+    { text: t('layouts.footer.contact'), href: '/contact_us', },
+    { text: t('layouts.footer.accessibility'), href: 'accessibility-statement', },
+    { text: t('layouts.footer.terms_and_conditions'), href: 'tandcs' },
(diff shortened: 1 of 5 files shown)


mpw5 (commented):
At the moment the `Built by MOJ Digital Services` footer link seems to be a bit broken. The `Built by` part links to the MOJ Blog but the `MOJ Digital Services` part is a broken link.

I think this is because of the way we currently have a link embedded in the translation text that is displayed. 

I think if you change the `service_owner_html` translation text  from `<span>Built by</span> <a class="govuk-footer__link" href="%{link}"><abbr title="Ministry of Justice">MOJ Digital Services</abbr></a>` to `Built by <abbr title="Ministry of Justice">MOJ Digital Services</abbr>` it should fix it.

(I don't think it really matters that the whole of that item is now a link and not just the `MOJ Digital Services` part).

mpw5 (approved):
👍

PR #8611, opened 2025-06-06
Ctskf 1059 cccd migrate phase banner component to use govuk components gem

Katy600:
#### What
Update the customised phase banner to use the new phase banner component. 

#### Ticket
[CCCD: Migrate phase banner component to use govuk-components gem](https://dsdmoj.atlassian.net/browse/CTSKF-1059)

#### Why

We can use the govuk-components gem to simplify the CCCD codebase and improve maintainability of the application.
This ticket covers the work needed to migrate the phase banner component to use the gem.

#### How

Delete old components and any reference to them:

`lib/govuk_component/phase banner_helpers.rb has been deleted`

`lib/govuk_component/railtie.rb has been updated to remove the govuk_component.phase banner_helpers initializer`

`spec/lib/govuk_component/phase banner_helpers_spec.rb has been deleted`

Use the phase banner with a block to allow links to be added. 

 https://govuk-components.netlify.app/components/phase-banner/#phase-banner-with-block
  
  It seemed to error if I don't include a tag. The current one doesn't have one.  
  
 In order for the phase banner block to work the translation files needed to be split up.


jrmhaig:
I don't think we should be using the phase banner here. I suspect that it was added when the app was first being developed and then never removed. This will be why it is not possible to use without a tag. We need to find the correct way for these links to be displayed.

mpw5:
Superseded by https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/9131

PR #8664, opened 2025-06-26
Fix deprecation warning

jrmhaig:
#### What

Remove some deprecation warnings after upgrading to Rails 7.1.

#### Ticket

N/A

#### Why

After upgrading to Rails 7.1 `ActiveRecord::Base#serialize` takes a keyword argument for the type (or coder) instead of a positional arguments. A positional argument results in a deprecation warning and from Rails 7.2 onwards this will not work.

#### How

Change the second positional argument with keyward arguments on `serialize` calls in three places.


mpw5 (approved):
👍

PR #8713, opened 2025-07-10
Update the laa-cda gem

jrmhaig:
#### What

Update the laa-cda gem.

#### Ticket

N/A

#### Why

Requests are sent to CDA to search Common Platform using GET requests with URL parameters and these appear in the CDA logs. A POST endpoint has now been added so that search parameters can be added to the message body, which will prevent possibly PII data being stored in logs.

#### How

See https://github.com/ministryofjustice/laa-cda/pull/1


VinceChiuMOJ (approved):
looks good.

PR #8777, opened 2025-08-08
[CTSKF-1184] Remove webpack

mpw5:
#### What

Remove `webpack` and move to using more modern alternatives for asset management, namely `propshaft`, `esbuild`, `jsbundling-rails`, and `cssbundling-rails`.

#### Why

To modernise CCCD and remove blockers caused by the use of `webpack`.

#### How

* Add `propshaft` and `cssbundling-rails` gems to `Gemfile`
* Add `esbuild` to `package.json` and create new `build:js` and `build:css` steps for bundling assets
* Remove now-unnecessary dependencies from `package.json`, including `babel`, and delete their configuration files
* Restructure the code appropriately:
    * Move the contents of `app/webpack/images` to `app/assets/images`
    * Move the contents of `app/webpack/stylesheets` to `app/assets/stylesheets`
    * Move the contents of `app/webpack/javascripts` to `app/javascript`
    * Move `app/webpack/packs/application.js` to `app/javascript/application.js`
    * Delete `app/webpack/packs/adp_swagger_application.js` and `govuk-frontend.js`, as they are no longer needed
    * Break the contents of `app/webpack/javascripts/application.js` into smaller modules. This file could not be moved directly to `app/javascript/application.js` as it would cause a naming conflict. It has always been poorly named, and refactoring it into smaller, better-named files makes it more understandable, although `Miscellaneous` still is not ideal. The new files are:
        * `app/javascript/modules/Modules.Miscellaneous.js`
        * `app/javascript/modules/Modules.StringInterpolation.js`
        * `app/javascript/modules/Modules.TinyPubSub.js`
        * `app/javascript/modules/Modules.TruncPolyfill.js`
    * Update `app/javascript/plugins/jquery.numbered.elements.js` by removing its wrapper and making it compatible with the `esbuild` and `jquery` setup
* Add shims for `jquery`, `stickyfill`, and `accessible-autocomplete` to make them compatible with `esbuild`
* Make minor changes to modules, including `AllocationDataTable` and `MultiFileUpload`, to work with the new asset setup
* Remove the `jquery-datatables-checkboxes` library, which is not compatible with `esbuild`, and replace its functionality with DataTables Select features to implement checkboxes on the allocations page. This requires changes to `app/javascript/modules/Modules.AllocationDataTable.js` and `app/javascript/modules/case_worker/Allocation.js`, as well as changes to the allocation table markup, page objects, and the way SitePrism clicks the checkbox in feature tests
* Update the development Procfile to run the JavaScript and CSS build watchers independently

#### Notes

SonarCloud is flagging some issues with the 'new' code. In reality, almost none of this is new code; it is mostly existing code being moved around. I do not propose fixing this here. It will need to be addressed separately later.


AnnaC-C:
Converted to draft as it is back to work in progress

app/javascript/modules/Modules.TinyPubSub.js
```diff
@@ -0,0 +1,19 @@
+/* jQuery Tiny Pub/Sub - v0.7 - 10/27/2011
+ * http://benalman.com/
+ * Copyright (c) 2011 "Cowboy" Ben Alman; Licensed MIT, GPL */
```
AnnaC-C, line 3: This is quite old - Why is this here? 
We should look into changing this work around, or pushing the removal of jQuery

app/javascript/modules/Modules.TruncPolyfill.js
AnnaC-C, line 1: Great to split our large files up into small functional files 🙌

app/javascript/application.js
AnnaC-C, line 1: Refactoring some of this would be a great use-case for AI workflows (Later on)

app/javascript/modules/Modules.AllocationDataTable.js
AnnaC-C, line 1: This file could use splitting up - we can refactor it in a different PR this PR is big enough already!

app/javascript/modules/Modules.AllocationDataTable.js
```diff
@@ -425,10 +470,12 @@ moj.Modules.AllocationDataTable = {
 
   // API: clear check boxes
   clearCheckboxes: function () {
-    this.dataTable
-      .column(0)
-      .checkboxes
-      .select(false)
+    if (!this.dataTable) return
```
AnnaC-C: Let's add in a log for when the `dataTable` isn't valid. (let's do this for line 496 as well)

```suggestion
    if (!this.dataTable) 
    {
      LOGF("Data table isn't valid");
      return
    }
```

docker/Dockerfile
```diff
@@ -85,9 +85,10 @@ ENV BUILD_TAG=${BUILD_TAG}
 ENV APP_BRANCH=${APP_BRANCH}
 ENV LIVE1_DB_TASK=${LIVE1_DB_TASK}
 
-USER 1000
-
 # Create tmp/pids/ needed for puma
-RUN mkdir -p $APP_HOME/tmp/pids
+RUN mkdir -p $APP_HOME/tmp/pids \
+&& chown -R appuser:appgroup $APP_HOME
```
AnnaC-C, line 90: Why are we doing this?

docker/Dockerfile
```diff
@@ -85,9 +85,10 @@ ENV BUILD_TAG=${BUILD_TAG}
...
+
+USER 1000
```
AnnaC-C, line 92: Why are we calling the USER here instead of line 88?

AnnaC-C (commented):
Just a few comments to answer some whys

app/javascript/modules/Modules.TinyPubSub.js
mpw5: It's used in a few places where events are published/subscribed to, eg  https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/bf1b1d37c5bbde6178623e30da7a15fc50cefee7/app/javascript/modules/Helpers.Autocomplete.js#L38 publishes `/onConfirm/<field-id>/`, and https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/bf1b1d37c5bbde6178623e30da7a15fc50cefee7/app/javascript/modules/external_users/claims/CaseTypeCtrl.js#L43 subscribes to it.. So it's basically a way of passing data between modules.

We could probably write our own event bus ather than including someone else's code but it's pretty small and static, and we're licensing it correctly, so it's probably fine to leave for now. If we refactor our javascript/remove jquery then we should definitely look at replacing it though.

app/javascript/application.js
mpw5: 🤖

app/javascript/modules/Modules.AllocationDataTable.js
mpw5: I think the best we can do at the moment is adding some console logging, something like 
```
  clearCheckboxes: function () {
    if (!this.dataTable) {
      console.warn('AllocationDataTable: cannot clear checkboxes because the DataTable is not initialised')
      return
    }
    this.dataTable.rows().deselect()

    const selectAllCheckbox = document.querySelector('#select-all-claim')
    if (selectAllCheckbox) selectAllCheckbox.checked = false
  },
```

and 
```
  clearFilter: function (e, data) {
    this.clearCheckboxes()
    this.clearSearchConfig()

    if (!this.dataTable) {
      console.warn('AllocationDataTable: cannot clear filters because the DataTable is not initialised')
      return
    }

    this.dataTable
      .search('')
      .columns()
      .search('')
      .draw()
  },
```

Which isn't ideal but we don't have anything set up to send client side logging anywhere else at the moment. We could probably set up something in Sentry for better long term visibility but it would need us to install more libraries ([@sentry/browser](https://www.npmjs.com/package/@sentry/browser) I think) which is out of scope for this change).

docker/Dockerfile
mpw5: As a result of the asset build changes, `$APP_HOME/tmp` is no longer writeable by UID 1000 - the docker image fails to build when it tries to create `tmp/pids` (which is needed by the puma web server).

Adding `chown -R appuser:appgroup $APP_HOME` ensures the non-root application user owns the application directory and can create the puma pid directory.

docker/Dockerfile
mpw5: `USER 1000` cannot write to `$APP_HOME/tmp`, and even if the directory already existed, `chown` would fail because only root can change file ownership. We need to move `USER 1000` adter the `mkdir`/`chown` step to ensure the creation of the `pids` directory is successful.

app/javascript/modules/Modules.TinyPubSub.js
AnnaC-C: Let's leave it for now, but note this for any future refactoring. And it would be good to write it ourselves as it really is so small. Perhaps another team has made an event bus we could use as an external library as well

app/javascript/modules/Modules.AllocationDataTable.js
AnnaC-C: Yeah let's not worry too much right now, I am happy with some console logging OR a comment `TODO` to add client side logging when it is setup as a reminder. It's not be-all-end-all, but we can incorporate client side logging at a later date

AnnaC-C: @mpw5 
It's your choice if you want to add the console logging, a `TODO` or leave it with this PR as documentation we need to add client side logging 😄

AnnaC-C (dismissed):
Only the comment about logging remains, so when you've decided your implementation (or lack thereof) choice, this is all good to go!

AnnaC-C (dismissed):

AnnaC-C (approved):

PR #8846, opened 2025-09-01
Change the test test password

jrmhaig:
#### What

Update test password.

#### Ticket

N/A

#### Why

Some cucumber tests fails when run locally because the browser brings up an insecure password browser pop-up that needs to be cleared before proceeding.

#### How

Global search-and-replace `password1234` with `PasswordForTest`.


mpw5 (approved):
👍

PR #8944, opened 2025-09-26
Update Rails from 7.1 to 7.2

Katy600:
#### What
CCCD runs on Rails 7.1 which goes out of support on 01/10/2025. We should upgrade to 7.2 before then, but need to understand and document the changes required to upgrade.
#### Ticket

[CCCD: Understand and ticket the changes required to Upgrade to rails 7.2](https://dsdmoj.atlassian.net/browse/CTSKF-1333)

#### Why
Because it will go out of support, stop getting patches and could mean other upgrades are not compatible with our framework. 

#### How
Update the Gem file to have Rails 7.2
Run ` bin/rails app:update ` to get the latest changes. 
For most of the changes I selected the option to keep the current state so that it didn't override the CCCD specific settings.


config/initializers/01_autosave_association_patch.rb
```diff
@@ -28,11 +28,11 @@ module AutosaveAssociation
     private
       alias old_association_valid? association_valid?
 
-      def association_valid?(reflection, record, index=nil)
+      def association_valid?(reflection, record)
```
jrmhaig: To avoid similar issues in the future with the change in arguments;

```suggestion
      def association_valid?(reflection, ...)
```

(and similar below)

config/initializers/01_autosave_association_patch.rb
```diff
@@ -28,11 +28,11 @@ module AutosaveAssociation
...
         if reflection.options[:validate].is_a?(Proc)
           return true unless reflection.options[:validate].call(self)
         end
-        old_association_valid?(reflection, record, index)
+        old_association_valid?(reflection, record)
```
jrmhaig: As above

```suggestion
        old_association_valid?(reflection, ...)
```

jrmhaig (commented):
Some of the commit message in the penultimate commit appears to have gone missing. Can you edit the commit message, please? While you do this you could merge in the last commit with the Rubocop fix.

My other comment, regarding the arguments of the `association_valid?` method, is just a suggestion so it is not essential to change.

jrmhaig (approved):
Looks good. It would be useful to deploy to an environment and test it a bit. I can help.

PR #8957, opened 2025-10-06
Set log level for logstasher to INFO in dev

jrmhaig:
#### What

Set the log level for logstasher to INFO in the development environment

#### Ticket

N/A

#### Why

There is an issue with how logstasher is using to_json to serialise data that results in 'stack level too deep' errors.

#### How

```ruby
config.logstasher.log_level = Logger::INFO
```


mpw5 (approved):
I agree, it's not ideal but I'm happy to approve to allow us to run CCCD locally

PR #9053, opened 2025-11-06
[CTSKF-1063] Migrate warning text component

ziaxgit:
#### What
All instances of the warning text component have been updated to use the helper methods from the gem: [Warning text](https://design-system.service.gov.uk/components/warning-text/)

#### Ticket

[CTKSF-1063](https://dsdmoj.atlassian.net/browse/CTSKF-1063?atlOrigin=eyJpIjoiN2I2NzI4OTg1MWY1NGY2YTk2ZWE5NWMxNGJmZTk2NTgiLCJwIjoiaiJ9)

#### Why
To simplify the CCCD codebase and improve maintainability of the application.

#### How
The existing helper module and related code has been removed from CCCD:
> lib/govuk_component/warning_text_helpers.rb has been deleted
> lib/govuk_component/railtie.rb has been updated to remove the govuk_component.warning_text_helpers initializer
> spec/lib/govuk_component/warning_text_helpers_spec.rb has been deleted


app/views/external_users/claims/_api_promo_banner.html.haml
```diff
@@ -1,9 +1,10 @@
 - if claim.new_record? && show_api_promo_to_user?
 
   .js-callout-banner{'data-setting': 'api_promo_seen'}
-    = govuk_warning_text t('callout_banner.api_promo_warning_text')
+    = govuk_warning_text do
+      = t('callout_banner.api_promo_warning_text')
```
jrmhaig, line 5: Is there a reason for not using this?

```suggestion
    = govuk_warning_text(text: t('callout_banner.api_promo_warning_text'))
```

Is the `div` on line 7, below, intended to be in the block? Because I think that it isn't.

jrmhaig (commented):
Just one comment. Otherwise it looks OK.

app/views/external_users/claims/_api_promo_banner.html.haml
ziaxgit: The `div` on line 7 is part of the block, yes. Similar to `_unclaimed_fees_notice.html.haml`. I can see why the extra space can cause confusion so I've removed it and indented properly.

jrmhaig: Thanks. The indent defines the block, in the same way as in Python, so following your latest change you have:

```ruby
  .js-callout-banner{'data-setting': 'api_promo_seen'}
    = govuk_warning_text do
      = t('callout_banner.api_promo_warning_text')

      %div{class: 'govuk-!-font-weight-regular govuk-!-margin-top-4'}
        = t('callout_banner.api_promo_information_html', href: api_landing_page_path)

    %ul.govuk-list
      %li.hide-link
```

The `div` is inside the `govuk_warning_text` block but the `ul` containing the hide link is not. All of this is inside the 'js-callout-banner` div.

This results in;

Screenshot 2025-11-12 at 09 08 03

Is this what is intended? Or should the link also be indented, which would happen if the link is within the `govuk_warning_text` too? There isn't an official design for this so either is OK.

config/locales/en.yml
```diff
@@ -3033,7 +3033,7 @@ en:
   callout_banner:
     api_promo_hide_banner: Do not show API information again
     api_promo_information_html:
-      You may be able to create claims without re-typing. <a class="govuk-link" href="%{href}">Read more on API information</a>.
+      You may be able to create claims without re-typing. <br> <a class="govuk-link" href="%{href}">Read more on API information</a>.
```
VinceChiuMOJ, line 3036: Why we need a line break now?

app/views/external_users/claims/_api_promo_banner.html.haml
ziaxgit: I think the reason I kept the `ul` indented outside is because other warnings like `_hardship_claims_banner.html.haml` and `_clair_contingency_banner.html.haml` do the same so just keeping it consistent. If I indent the `ul` for this, it would make sense to do the same for the rest. If there's not any official design, shall we keep it as it is?

config/locales/en.yml
ziaxgit: Had a chat with Mark and we agreed the link looks better if it's on a new line. Other warnings also have the link on a new line so this makes it consistent.

app/views/external_users/claims/_api_promo_banner.html.haml
VinceChiuMOJ: Seems that the "more information link" usually go with the "hide link" under the ul.govuk-list in other warning text, which I think is better to go within the warning text like this one.

VinceChiuMOJ (approved):
Looks good to me.

PR #9137, opened 2026-01-22
[CTSKF-1478] Allow up to 10 PTPH fees

jrmhaig:
There was previous a limit of 3 PCPH fees allows. This has been increased, although there is still a limit for the moment.

#### What

Permit advocates to add up to 10 plea and trial preparation hearings on a claim.

#### Ticket

[Increase the number of PTPH fees that are claimable in CCCD](https://dsdmoj.atlassian.net/browse/CTSKF-1487)

#### Why

Providers have requested that the limit of 3 plea and trial preparation hearings on a claim should be lifted and there is no reason why this limit is in place.

#### How

Update the validator to allow up to 10 instead of 3.

Note that plea and trial preparation hearings (PTPH) were originally called plea and case management hearings (PCM) and this is still used in various places in the code.

### DIFF
--- a/config/locales/en/models/claim.yml
+++ b/config/locales/en/models/claim.yml
@@ -235,7 +235,7 @@ en:
               npw_invalid: Enter a valid quantity for number of prosecution witnesses fees
               pcm_invalid: Enter a valid quantity for plea and trial preparation hearing
               pcm_not_applicable: Plea and case management hearing fee not applicable to case type
-              pcm_numericality: Enter a valid quantity (1 to 3) for plea and case management hearing fees
+              pcm_numericality: Enter a valid quantity (1 to 10) for plea and case management hearing fees
               ppe_invalid: Enter a valid quantity for pages of prosecution evidence fees
               saf_invalid: Enter a valid quantity for standard appearance fees
             rate:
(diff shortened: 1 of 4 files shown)


Katy600 (approved):
That makes sense :)

config/locales/en/models/claim.yml
```diff
@@ -364,7 +364,7 @@ en:
               miste_numericality: Enter a valid quantity (1) for section 28 hearing
               miumu_numericality: Enter a valid quantity (1) for unused material (up to 3 hours)
               pcm_invalid: Enter a valid quantity for plea and trial preparation hearing
-              pcm_numericality: Enter a valid quantity (1 to 3) for plea and case management hearing fees
+              pcm_numericality: Enter a valid quantity (1 to 10) for plea and case management hearing fees
```
mpw5: My only suggestion would be to change this error to refer to PTPH instead of PCMH.
```suggestion
              pcm_numericality: Enter a valid quantity (1 to 10) for plea and trial preparation hearing fees
```

config/locales/en/models/claim.yml
```diff
@@ -235,7 +235,7 @@ en:
               npw_invalid: Enter a valid quantity for number of prosecution witnesses fees
               pcm_invalid: Enter a valid quantity for plea and trial preparation hearing
               pcm_not_applicable: Plea and case management hearing fee not applicable to case type
-              pcm_numericality: Enter a valid quantity (1 to 3) for plea and case management hearing fees
+              pcm_numericality: Enter a valid quantity (1 to 10) for plea and case management hearing fees
```
mpw5: My only suggestion would be to change this error to refer to PTPH instead of PCMH.
```suggestion
              pcm_numericality: Enter a valid quantity (1 to 10) for plea and trial preparation hearing fees
```

mpw5 (approved):
I've left a couple of comments about the wording of error messages, as the user selects PTPH but the error refers to PCMH, so it might be a good chance to align them.

But I think it's an optional change - PTPH/PCMH seem to be used inconsistently throughout so changing it in just this one place isn't going to have a big impact.

PR #9196, opened 2026-02-13
Remove extensions to the hash object

jrmhaig:
#### What

Remove the monkey patching of the `Hash` object.

#### Ticket

N/A

#### Why

The `lib/extensions` contains a number of monkey-patches, which are generally discouraged. Specifically, the `hash_extensions.rb` contains methods added to the `Hash` object which only used in tests.

#### How

The `all_keys` method returns the list of all keys in a nested hash. In the two places it is used;
* In `spec/api/v2/mi/agfs_scheme_ten_claims_spec.rb` the hash is not nested and so `all_keys` returns the same as `keys`.
* In `spec/api/v2/search_spec.rb` the hash contains one key, `:filter` that holds another hash, so the single test can be replaced by two tests. Additionally, 5 additional keys are added to the list.

The `all_values_for` method finds all values in a nested hash. This is used in one place;
* In `spec/support/transfer_brain_helpers.rb` the values that are found in the hash are all at a consistent depth and so the way these values are found can be simplified using chained `flat_map`s.

The `key_paths` method finds all possible chains of keys within a nested hash. This is used in one place;
* In `spec/models/claim_state_transition_reason_spec.rb` this is used to find the values at the lowest level of a hash. This hash is known to be three layers deep and so this can be simplified using changed `flat_map`s.

After fixing these then `lib/extensions/hash_extensions.rb` can be removed.


mpw5 (approved):
Nice to see that code go at last!

PR #9204, opened 2026-02-17
fix(fee dates): add correct error message for date

roni-c-justice-dev:
fix(fee dates): add correct error message for date < 2yrs before reporder
---------

Date for final hearing should not be before the date of earliest rep order, as no representation would have been in place for the case whilst it was live. We would always expect representation to have been granted before a case concluded.

#### Ticket

(https://dsdmoj.atlassian.net/jira/software/c/projects/CTSKF/boards/987?selectedIssue=CTSKF-1582)

#### Why
Entering a date which was before earliest rep order date but within 2 years was producing the error message that it was more than 2 years before the earliest rep order.  This is incorrect and confusing for the user.

#### How
Amend the details in base_validator and claim.yml to produce a relevant error message for those dates which would be within 2 years but before earliest rep order date.  Removed the 2 year validations following extensive investigation with the business, the earliest rep order date is the important one here as we would always expect representation to have been granted before a case concluded.

Amend all associated unit and BDD tests appropriately


Gemfile.lock
```diff
@@ -766,6 +768,7 @@ GEM
 PLATFORMS
   arm64-darwin-23
   arm64-darwin-24
+  arm64-darwin-25
```
Katy600, line 771: My only thought is - does this relate specifically to your computer and therefore should be included in the main repo?? There are lots of others arm64-darmin  though- so perhaps not! I always get confused about what to include - perhaps worth asking Joe or Mark.

roni-c-justice-dev: Have checked with Mark and it shouldn't affect anything, will leave it in there at present and it might even save us some tech debt at a later point.

mpw5 (approved):
👍

PR #9225, opened 2026-02-27
Add conditions to the casworker document ability to restrict access t…

Katy600:
…o allocated caseworker.

#### What

Prevent unauthorised access to documents that are not associated with a claim that has been allocated to them. 

#### Ticket

[CCCD: Unauthorised access to documents](https://dsdmoj.atlassian.net/browse/CTSKF-1543)

#### Why
Ensured that caseworkers cannot access documents by claims assigned to other caseworkers, improving data security and privacy.

#### How

Update the Ability class so these conditions are added.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -64,7 +64,9 @@ class Ability
     end
     can_administer_any_provider if persona.roles.include?('provider_management')
     can %i[upload delete], Document
-    can %i[show download], Document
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
Looks good. This feels like the sort of thing we ought to make the case workers aware of before merging, and then monitoring to see if it has adverse effects.

PR #9247, opened 2026-03-05
Ctskf 1545 cccd nil dereference + validation bypass in unarchive

Katy600:
#### What
Improved the unarchive flow to avoid crashes and stop masking real errors.
I did not remove `(validate: false) ` as 

"it looks like there were some changes to the validation rules which meant that a claim could end up in a situation where:
- it was originally valid
- it was archived
- the validation rules changed
- if someone tried to unarchive the claim it would now be invalid and the the unarchiving would fail." (Mark)

#### Ticket

[CCCD: Nil Dereference + Validation Bypass in Unarchive](https://dsdmoj.atlassian.net/browse/CTSKF-1545)

#### Why

Some archived claims had no version to reify, which caused crashes and was previously hidden by a broad rescue.
These updates make unarchiving safer, more predictable, and easier to debug.

#### How

Added guard for missing PaperTrail versions before reifying.
Ensured archiving always creates a version using save_with_version.
Replaced broad rescue StandardError with targeted DB‑level rescue (RecordNotSaved, StatementInvalid).
Added error logging for easier debugging.

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -160,7 +160,8 @@ module ExternalUsers
     def unarchive
       claim_url = external_users_claim_url(@claim)
       return redirect_to claim_url, alert: t('.not_archived') unless unarchive_allowed?
-      @claim = @claim.paper_trail.previous_version
+      version = @claim.versions.last
+      @claim = version.reify
       @claim.zeroise_nil_totals!
       @claim.save!(validate: false)
       redirect_to external_users_claims_url, notice: t('.unarchived')
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -160,13 +160,17 @@ module ExternalUsers
     def unarchive
       claim_url = external_users_claim_url(@claim)
       return redirect_to claim_url, alert: t('.not_archived') unless unarchive_allowed?
+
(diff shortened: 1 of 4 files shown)


VinceChiuMOJ (commented):
LGTM but I am not sure how to test it.

app/controllers/external_users/claims_controller.rb
```diff
@@ -160,18 +160,11 @@ def destroy
     def unarchive
       claim_url = external_users_claim_url(@claim)
       return redirect_to claim_url, alert: t('.not_archived') unless unarchive_allowed?
-
-      version = @claim.versions.last
-      return redirect_to claim_url, alert: t('.cannot_unarchive_no_version') unless version
-
-      @claim = version.reify
+      claim_previous_version
```
jrmhaig: A `before_action` hook would work better for this. It could (potentially) also cater for the `unarchived_allowed?` guard on the previous line.

app/controllers/external_users/claims_controller.rb
```diff
@@ -192,6 +185,19 @@ def resource_klass
 
     private
 
+    def rescue_and_log_errors
```
jrmhaig: I'd prefer, if possible, for the rescue not to be extracted to a separate method. If it helps reduce the line count, the two errors that are listed can go on a single line.

PR #9248, opened 2026-03-05
CTSKF-1593: Replace Standard with ESLint and Neostandard

mpw5:
#### What

Resolves [a security vulnerability](https://security.snyk.io/vuln/SNYK-JS-AJV-15274295) in the `ajv` package by

* Replacing `standard` with `neostandard`
* Removing `file-loader` and `url-loader` `webpack` libraries

#### Ticket

[CTSKF-1593](https://dsdmoj.atlassian.net/browse/CTSKF-1593)

#### Why

* Through transitive dependencies, `standard` introduces a [vulnerable](https://security.snyk.io/vuln/SNYK-JS-AJV-15274295) version of the `ajv` package. `standard` no longer appears to be maintained, which means this vulnerability, and potentially others, will not be fixed. `neostandard` is an actively supported fork of `standard` and adopting it removes the vulnerability and gives us an improved tool.
* `file-loader` and `url-loader` also introduce compromised versions of `ajv`. These are deprecated `webpack` packages. `url-loader` is unused and be safely removed. `file-loader` can be replaced with the built-in `webpack` `asset/resource` function used for other asset types.

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -70,9 +70,9 @@ module.exports = {
       },
       {
         test: /\.(ico)$/,
-        loader: 'file-loader',
-        options: {
-          name: '[name].[ext]'
+        type: 'asset/resource',
+        generator: {
+          filename: '[name][ext]'
         }
       },
       {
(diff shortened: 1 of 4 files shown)


jrmhaig (approved):
🎉

PR #9261, opened 2026-03-10
Add a guard clause to prevent filtered_state_transitions.second retur…

Katy600:
…ning nil if there are fewer than 2 transitions.

#### What

 Issue: filtered_state_transitions.second returns nil if there are fewer than 2 transitions. Calling .event on nil raises NoMethodError. 

  
#### Ticket

[CCCD: Nil Reference in Message Callback](CTSKF-1544)

#### Why
To prevent the message creation silently failing and claim state transitions not happen.

#### How
A guard clause has been added along side a comment explaining why it probably isn't necessary. 

I investigated the code and tried to reproduce a situation when `request_written_reasons` are generated in the local browser. I was only able to do this the once but I found other examples in the code base that showed that this only happens after a claim has been authorised- therefore making it unlikely that `claim.filtered_state_transitions.second.event` would ever be Nil.  

Screenshot 2026-03-11 at 13 05 58

Screenshot 2026-03-11 at 13 06 24

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -102,6 +102,9 @@ class Message < ApplicationRecord
   def process_written_reasons
     return unless claim.written_reasons_outstanding?
     return unless written_reasons_submitted == '1'
+    return unless claim.filtered_state_transitions.second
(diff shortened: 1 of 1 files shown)


jrmhaig (dismissed):
:+1:

mpw5 (approved):
👍

PR #9296, opened 2026-03-13
Remove this line as the creator_id belongs to the User model not Exte…

Katy600:
…rnalUsers

#### What
The Document model has a field creator_id that is set to current_user.id when a document is uploaded and this links the document to an instance of User. However, the model has a reference defined as belongs_to :creator, class_name: 'ExternalUser'.

#### Ticket

[CCCD - Confused definition of Document#creator_id](https://dsdmoj.atlassian.net/browse/CTSKF-1014)

#### How
Replace this line 
`belongs_to :creator, class_name: 'ExternalUser'`

With 
`belongs_to :creator, class_name: 'User'`

So we no longer have to call creator_id inside the documents controller. creator now references the User model directly.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -34,7 +34,7 @@ class DocumentsController < ApplicationController
   end
 
   def upload
-    @document = Document.new(document_params.merge(creator_id: current_user.id))
+    @document = Document.new(document_params.merge(creator: current_user))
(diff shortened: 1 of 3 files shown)


jrmhaig:
This looks good. Can it be deployed on dev-lgfs so that we can test it a bit to make sure that uploading, downloading and viewing documents still works as expected.

PR #9323, opened 2026-03-17
Remove the customised tag_helpers code

Katy600:
#### What
Migrate tag component to use govuk-components gem
#### Ticket

[CCCD: Migrate tag component to use govuk-components gem](https://dsdmoj.atlassian.net/browse/CTSKF-1062)

#### Why
Improve maintainability of the application

#### How

- lib/govuk_component/tag_helpers.rb has been deleted 
- lib/govuk_component/railtie.rb has been updated to remove the govuk_component.tag_helpers initializer
- spec/lib/govuk_component/tag_helpers_spec.rb has been deleted
- remove old style sheets that were called dynamically by the code and replaces with a helper method


app/helpers/tag_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+  STATE_COLOURS =
+    {
+      'authorised' => 'green',
+      'accepted' => 'green',
+      'allocated' => 'purple',
+      'draft' => 'blue',
+      'rejected' => 'red',
+      'refused' => 'red',
+      'unverified' => 'red',
+      'archived-pending-delete' => 'dark-grey',
```
mpw5: I think this needs to be changed to `archived_pending_delete` (with underscores not dashes), to match the actual state of the claim. At the moment I suspect the colour is defaulting to `grey` as `archived-pending-delete` (with dashes not underscores) is not a recognised state.

However, this will cause a failure as `dark-grey` is not a colour supported by the component (see the error below). `grey` would probably be good enough for now but we might want to ask Jonathan to review the tag colours in future.

app/helpers/tag_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+      'redetermination' => 'black',
```
mpw5: I get an error when attempting to view a claim in a state of `redetermination`. The error is 
```
invalid tag colour black, supported colours are grey, green, teal, blue, light-blue, red, purple, magenta, orange, and yellow
```

`grey` would probably be good enough for now but we might want to ask Jonathan to review the tag colours in future.

mpw5 (changes requested):
Excellent work. Unfortunately there are a couple of issues in the `TagHelper` class which need resolving - I've left comments.

app/helpers/tag_helper.rb
Katy600: Oh good spot. Thanks Mark :)

mpw5 (approved):
LGTM

PR #9336, opened 2026-03-18
Ctskf 1014 cccd confused definition of document creator

Katy600:
#### What
The Document model has a field creator_id that is set to current_user.id when a document is uploaded and this links the document to an instance of User. However, the model has a reference defined as belongs_to :creator, class_name: 'ExternalUser'.

#### Ticket

[CCCD - Confused definition of Document#creator_id](https://dsdmoj.atlassian.net/browse/CTSKF-1014)

#### How
Replace this line 
`belongs_to :creator, class_name: 'ExternalUser'`

With 
`belongs_to :creator, class_name: 'User'`

So we no longer have to call creator_id inside the documents controller. creator now references the User model directly.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -34,7 +34,7 @@ class DocumentsController < ApplicationController
   end
 
   def upload
-    @document = Document.new(document_params.merge(creator_id: current_user.id))
+    @document = Document.new(document_params.merge(creator: current_user))
     if @document.save_and_verify
       render_success_response
     else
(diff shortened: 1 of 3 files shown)


mpw5 (approved):
LGTM.

Tested in `dev-lgfs`:
- [x] View a previously uploaded file as an advocate
- [x] Upload a file as an advocate
- [x] View a newly uploaded file as an advocate
- [x] View a previously uploaded file as a litigator
- [x] Upload a file as an litigator
- [x] View a newly uploaded file as a litigator
- [x] View a previously uploaded file as a caseworker
- [x] View a newly uploaded file as a caseworker

PR #9374, opened 2026-03-24
Ctskf 1545 cccd nil dereference validation bypass in unarchive

Katy600:
#### What

Improved the unarchive flow to avoid crashes and stop masking real errors.
I did not remove `(validate: false) ` as 

"it looks like there were some changes to the validation rules which meant that a claim could end up in a situation where:
- it was originally valid
- it was archived
- the validation rules changed
- if someone tried to unarchive the claim it would now be invalid and the the unarchiving would fail.

#### Ticket

[CCCD: Nil Dereference + Validation Bypass in Unarchive](https://dsdmoj.atlassian.net/browse/CTSKF-1545)

#### Why

Some archived claims had no version to reify, which caused crashes and was previously hidden by a broad rescue.
These updates make unarchiving safer, more predictable, and easier to debug.

#### How

Added guard for missing PaperTrail versions before reifying.
Ensured archiving always creates a version using save_with_version.

### DIFF
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -160,12 +160,11 @@ module ExternalUsers
     def unarchive
       claim_url = external_users_claim_url(@claim)
       return redirect_to claim_url, alert: t('.not_archived') unless unarchive_allowed?
-      @claim = @claim.paper_trail.previous_version
+      revert_to_version_before_archived
+      return redirect_to claim_url, alert: t('.unarchivable') if @claim.nil?
       @claim.zeroise_nil_totals!
       @claim.save!(validate: false)
       redirect_to external_users_claims_url, notice: t('.unarchived')
-    rescue StandardError
-      redirect_to claim_url, alert: t('.unarchivable')
     end
 
     class << self
@@ -185,6 +184,12 @@ module ExternalUsers
     end
 
     private
+
+    def revert_to_version_before_archived
+      @claim = @claim.versions.select do |v|
+        v.object_changes['archived_pending_review'] || v.object_changes['archived_pending_delete']
+      end.compact.last.reify
+    end
 
     def log(message, error: nil, level: :info)
       log_data = {
--- a/app/controllers/external_users/claims_controller.rb
+++ b/app/controllers/external_users/claims_controller.rb
@@ -186,9 +186,12 @@ module ExternalUsers
     private
 
     def revert_to_version_before_archived
-      @claim = @claim.versions.select do |v|
-        v.object_changes['archived_pending_review'] || v.object_changes['archived_pending_delete']
-      end.compact.last.reify
+      version = @claim.versions.reverse.detect do |v|
+        obj = v.reify
+        obj && !obj.state.to_s.start_with?('archived_')
+      end
+
+      @claim = version&.reify
     end
 
     def log(message, error: nil, level: :info)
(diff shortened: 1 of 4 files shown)


app/controllers/external_users/claims_controller.rb
```diff
@@ -160,12 +160,11 @@ def destroy
     def unarchive
       claim_url = external_users_claim_url(@claim)
       return redirect_to claim_url, alert: t('.not_archived') unless unarchive_allowed?
-      @claim = @claim.paper_trail.previous_version
+      revert_to_version_before_archived
+      return redirect_to claim_url, alert: t('.unarchivable') if @claim.nil?
```
roni-c-justice-dev: Should this not use the `cannot_unarchive_no_version`

Katy600: Well spotted Roni :)

spec/controllers/external_users/claims_controller_spec.rb
```diff
@@ -824,15 +831,26 @@ def advocate
...
+      end
     end
 
     context 'when the claim is archived with assessed values' do
+      # It looks like there were some changes to the validation rules which meant that a claim could end up in a situation where:
+      # Validation is set to false so that this test is testing that behaviour.
+      # - it was originally valid
+      # - it was archived
+      # - the validation rules changed
+      # - if someone tried to unarchive the claim it would now be invalid and the the unarchiving would fail.
```
mpw5: I'm not sure this comment makes much sense (and I think I wrote it originally, so that's worrying!) I kind of understand what it's saying - I think it's trying to explain the purpose of the line `@claim.save!(validate: false)` in the `ExternalUser::ClaimsController`, but by itself here I think it lacks context and is a bit confusing.

spec/controllers/external_users/claims_controller_spec.rb
```diff
@@ -824,15 +831,26 @@ def advocate
...
       subject(:claim) { create(:advocate_claim, external_user: advocate) }
 
       before do
         claim.submit!
         claim.allocate!
-        claim.assessment.update(fees: 123.00, expenses: 23.45)
+        claim.assessment.update(fees: 123.00, expenses: 23.45) # cant be zero for part auth but must be zero for refused state
+        # the line above was relying on @claim.save! to have been called with validate:false.
```
mpw5: As above, I think these comments are trying to add context but aren't really helping.

app/controllers/external_users/claims_controller.rb
```diff
@@ -186,6 +185,15 @@ def resource_klass
...
     private
 
+    def revert_to_version_before_archived
+      version = @claim.versions.reverse.detect do |v|
+        obj = v.reify
+        obj && !obj.state.to_s.start_with?('archived_')
+      end
+
+      @claim = version&.reify
+    end
```
mpw5, line 195: Just checking that we've deployed this to an environment and tested that a variety of archived claims can be unarchived correctly?

mpw5 (commented):
I've left some comments. Nothing major, my main concern is the comments in the spec, which I think could be a bit confusing. I think they probably made more sense when you were investigating the behaviour but I don't think add much in isolation now. I think they either need rewriting or possibly removing.

spec/controllers/external_users/claims_controller_spec.rb
roni-c-justice-dev: just need to remove this comment given that you decided to keep validate:false

app/controllers/external_users/claims_controller.rb
Katy600: I just had another check that the archived claims are being unarchived correctly @mpw5 and seems to be working as expected. :)

mpw5 (approved):

PR #9438, opened 2026-04-14
[CTSKF-1484] Allocating claims to null caseworker

ziaxgit:
#### What
Fixes case worker input field validation logic on the client side so that if an admin tries to allocate claims without selecting a case worker they are shown an error message.

#### Ticket
[CTSKF-1484](https://dsdmoj.atlassian.net/browse/CTSKF-1484)

#### Why
- So that if no case worker is selected the page doesn't get stuck with message **'Allocating.. please wait a moment..'**
- If a case worker is initially selected but then the field is cleared, they are not able to allocate without selecting again.

#### How
- Added a `subscribe` block so `$.publish('/allocation/error/'` action doesn't fail silently.
- After a successful allocation, the number of claims field and case worker field (accessible autocomplete component) is reset.
- Added an extra feature test.

#### Test
**Scenario 1**
- Log in to CCCD as a caseworker admin
- On the Allocation page, enter 1 in the ‘Number of claims’ field
- Leave the ‘Case worker’ field blank
- Click ‘Allocate’
**Before change**: The message “Allocating.. please wait a moment..” is displayed but no claims are allocated.
**After change**: Page shows message 'Select a caseworker'

**Scenario 2**
- Log in to CCCD as a caseworker admin
- On the Allocation page, enter 1 in the ‘Number of claims’ field
- Select a caseworker from the ‘Case worker’ list
- Click ‘Allocate’
- Confirmation is displayed that the claim has been allocated
- Clear the ‘Case worker’ field
- Click ‘Allocate’
**Before change**: The message “1 claims have been allocated to” is displayed. The claim is actually allocated to the caseworker selected in step 3.
**After change**: Page shows message 'Select a caseworker'

[CTSKF-1484]: https://dsdmoj.atlassian.net/browse/CTSKF-1484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/webpack/javascripts/modules/Modules.AllocationDataTable.js
+++ b/app/webpack/javascripts/modules/Modules.AllocationDataTable.js
@@ -337,22 +337,21 @@ moj.Modules.AllocationDataTable = {
 
     // EVENT: Allocate claims
     $('.allocation-submit').on('click', function (e) {
-      self.ui.$notificationMsg.removeClass('govuk-!-display-none govuk-notification-banner--error govuk-notification-banner--success')
-      self.ui.$notificationMsg.find('.govuk-notification-banner__heading').html('Allocating.. please wait a moment..')
-
       e.preventDefault()
-      self.ui.$submit.prop('disabled', true)
 
       const quantityToAllocate = $('#quantity-to-allocate-field').val() || false
-
       const allocationCaseWorkerId = $('#allocation-case-worker-id-field-select').val()
 
       if (!allocationCaseWorkerId) {
-        $.publish('/allocation/error/', {
-          msg: 'Select a case worker.'
-        })
+        self.ui.$notificationMsg.removeClass('govuk-!-display-none govuk-notification-banner--success')
+        self.ui.$notificationMsg.addClass('govuk-notification-banner--error')
+        self.ui.$notificationMsg.find('.govuk-notification-banner__heading').html('Select a case worker.')
         return
       }
+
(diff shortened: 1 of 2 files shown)


mpw5 (approved):
Nice work. Reviewed and tested ✅ 

⭐  for adding a test!

app/webpack/javascripts/modules/Modules.AllocationDataTable.js
```diff
@@ -335,25 +335,31 @@ moj.Modules.AllocationDataTable = {
...
         $.publish('/allocation/error/', {
           msg: 'Select a case worker.'
         })
         return
       }
 
+      self.ui.$notificationMsg.removeClass('govuk-!-display-none govuk-notification-banner--error govuk-notification-banner--success')
+      self.ui.$notificationMsg.find('.govuk-notification-banner__heading').html('Allocating.. please wait a moment..')
+      self.ui.$submit.prop('disabled', true)
+
```
Katy600, line 362: Hi Zia, 

All this code mostly makes sense to me (I am not very confident with javascript). Is this group of code just allocating? Would it make sense to group it in the same way you did with `subscribe('/allocation/error/', function (e, data)`. Or would that be overkill?

ziaxgit: If I group it with subscribe, I would then have to create a publish block that calls it. If you check the diff changes, these 3 lines were already present in this way. All I did was change the ordering so if there's an error, the code returns early.

PR #9439, opened 2026-04-14
[SPIKE] Upgrade to Rails 8

mpw5:
#### What

This is a spike into upgrading rails to 8.0.

### Steps followed:

* Updated `rails` to `8.0.5` in `Gemfile`

* Ran `bin/rails app:update`

  * Kept current versions of files where prompted

* Incrementally enabled the following settings in `config/initializers/new_framework_defaults_8_0.rb`

  * `Rails.application.config.active_support.to_time_preserves_timezone = :zone`

  * `Rails.application.config.action_dispatch.strict_freshness = true`

  * `Regexp.timeout = 1`

* Ran tests, linting and started the server locally

  * See below for issues encountered

* Built and deployed the application to `dev-lgfs` and staging environments for testing

### Issues encountered:
* The following error occurs when starting the server: 
```
../actionpack-8.0.5/lib/action_dispatch/routing/mapper.rb:1185:in 'ActionDispatch::Routing::Mapper::Resources::Resource#initialize': Route resources :injection_attempts - :only and :except must include only [:index, :create, :new, :show, :update, :destroy, :edit], but also included [:dismiss] (ArgumentError).
``` 
This occurs because the use of `only: [:dismiss]` in https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/main/config/routes.rb#L206 is no longer valid, as `:dismiss` is not a standard REST action. It can be resolved by changing to `only: []` which has the same function but meets Rails 8’s validation.

* Rubocop fails with `Rails/StrongParametersExpect` offences in 14 places. This is as a result of a move away from using `params.require(…).permit(…)` for strong parameters to `params.expect(…)`. Rubocop offers an autocomplete function which breaks functionality, so it will be necessary to manually update the syntax in all affected controllers. However the existing syntax is not deprecated and still functions correctly, so the approach taken in this spike is to disable the cop so that it can be addressed incrementally in future.

### Deprecation warnings:

One new deprecation warning is shown when running the rspec test suite: 
```
DEPRECATION WARNING: Using semicolon as a query string separator is deprecated and will not be supported in Rails 8.1 or Rack 3.0. Use `&` instead. (called from Rack::Injectable#call at /home/circleci/repo/lib/rack/injectable.rb:8). 
```

This is triggered by three tests however it is difficult to identify which are causing it. The simplest solution might be to wait until we upgrade to Rails 8.1 at which point those tests may start to fail and we can fix them then.

#### Ticket

[CTSKF-1668](https://dsdmoj.atlassian.net/browse/CTSKF-1668)

[CTSKF-1668]: https://dsdmoj.atlassian.net/browse/CTSKF-1668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/routes.rb
```diff
@@ -203,7 +203,7 @@
 
   end
 
-  resources :injection_attempts, only: [:dismiss] do
+  resources :injection_attempts, only: [] do
     patch 'dismiss', format: :js, on: :member
```
roni-c-justice-dev, line 207: If we're removing dismiss from the previous line, does it still need keeping here?  Or might this be something we can remove?

mpw5: We still need to keep the lines for `dismiss` action in the `do` block.

By default a resource in rails will have all the usual RESTful actions (`index`, `show`, `new` etc) defined for it. `only:` is used to limit that to the ones we actually need. In this example we don't actually want to create any of the regular actions, just the bespoke `dismiss`. That can be achieved by leaving the array empty, which suppresses the creation of the default routes, and then defining the `dismiss` actions in the block.

(I suspect that what we had previously, `only: [:dismiss]`, wasn't actually doing anything - the `:dismiss` was being silently ignored. Rails 8 is more opinionated on this than Rails 7 so it now errors.)

https://guides.rubyonrails.org/routing.html#restricting-the-routes-created has some more details.

PR #9453, opened 2026-04-16
[CTSKF-1723] Upgrade rack-livereload gem

jrmhaig:
#### What

Upgrade the rack-livereload gem

#### Ticket

[CCCD - Upgrade Rack from version 2 to 3](https://dsdmoj.atlassian.net/browse/CTSKF-1723)

#### Why

We missed updating rack-livereload as it was bundled in with the rails 7.2 dependabot PR, which we closed. See #7614. Once closed, dependabot ignores the gem until there is a new version and there hasn't been one since.

#### How

Update `Gemfile`:

```ruby
gem 'rack-livereload', '~> 0.6.1'
```

Rack version 3, which comes as a result of this update, updates the naming of HTTP status 422 from 'Unprocessable Entity' to 'Unprocessable Content' and deprecated the old name. There were some changes required to use the new helper. See rack/rack#2137

To deal with the change of the multi-part date fields correctly a scope is added to the submission form, so for the `start_at` dates in the management information page (for example) the fields are now `report[start_at(1i)]`, `report[start_at(2i)` and `report[start_at(3i)]`. The strong params are then handled in the `ManagementInformationController#report_params` using the nested `params.expect(report: %i[report_type start_at update])` instead of the un-nested `params.permit(:report_type, :start_at)`. To allow for this, all buttons on the page have been changed to buttons in forms instead of link styled as buttons, as they were before, and they now all route to a `POST` create endpoint instead of `GET`s to `download` or `generate`.

#### To do

The handling of parameters has been modified in Rack 3 so that square brackets, when they are not part of nested attributes, are included in the field name. This means that a field in a form with the name `[field]` will appear in `params` with the square brackets whereas in Rack 2 the brackets were removed. This occurs when a form is created with a `url` and no model and causes problems with multi field dates such where the date is constructed by three fields for day, month and year. See cd626025ddc02ae5bf6253931b6ba96066cc1d31 for example.

Places where this may need to be fixed:

* [x] `SuperAdmins::StatsController#parse_time` in `app/controllers/super_admins/stats_controller.rb`
* [x] `CaseWorkers::Admin::ManagementInformationController#validate_and_set_date` in `app/controllers/case_workers/admin/management_information_controller.rb`
* [ ] ~`MultiparameterAttributeCleaner#find_and_clean_date` in `app/controllers/concerns/multiparameter_attribute_cleaner.rb`~ This didn't need to be fixed as the form in this case includes a model

### DIFF
--- a/app/controllers/super_admins/stats_controller.rb
+++ b/app/controllers/super_admins/stats_controller.rb
@@ -26,13 +26,25 @@ module SuperAdmins
       @date_err = graph_data.date_err
     end
 
-    def parse_time(date)
-      Time.zone.parse("#{params["#{date}(3i)"]}-#{params["#{date}(2i)"]}-#{params["#{date}(1i)"]}")
+    def parse_date(date)
+      return if stats_params[:stats].nil?
+
+      Date.new(
+        stats_params[:stats]["#{date}(1i)"].to_i,
+        stats_params[:stats]["#{date}(2i)"].to_i,
+        stats_params[:stats]["#{date}(3i)"].to_i
+      )
+    rescue Date::Error
+      nil
     end
 
     def set_times
-      @from = parse_time('date_from') || Time.current.at_beginning_of_month
-      @to = parse_time('date_to')
+      @from = parse_date('date_from') || Time.current.at_beginning_of_month
+      @to = parse_date('date_to')
+    end
+
+    def stats_params
+      params.permit(stats: %i[date_from date_to])
     end
   end
 end
--- a/config/locales/en/views/management_information.yml
+++ b/config/locales/en/views/management_information.yml
@@ -7,6 +7,10 @@ en:
           invalid_report_type: &invalid_report_type The requested report type is not supported
         validate_report_type_is_updatable:
           report_cannot_be_updated: &report_cannot_be_updated The requested report type cannot be updated
+        validate_report_type_with_date:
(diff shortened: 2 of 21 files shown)


app/controllers/super_admins/stats_controller.rb
```diff
@@ -27,12 +27,24 @@ def generate_graph_data
     end
 
     def parse_time(date)
```
mpw5: Minor nitpick. This method now returns a`Date` object instead of `Time`. I don't think it causes any problems, but it means the method name is inaccurate.

config/locales/en/views/management_information.yml
```diff
@@ -7,13 +7,19 @@ en:
           invalid_report_type: &invalid_report_type The requested report type is not supported
         validate_report_type_is_updatable:
           report_cannot_be_updated: &report_cannot_be_updated The requested report type cannot be updated
+        validate_report_type_with_date:
```
mpw5: `validate_report_type_with_date` and `validate_report_type_is_updatable_with_date` don't correspond to any controller methods now - they were removed in 585f7f7c07f593a8f66ae68fa0dda3668ea745bf. I think you can remove the entries from this file.

mpw5 (commented):
Excellent work. Couple of minor comments.

mpw5 (approved):
👍

PR #9459, opened 2026-04-17
Update code to allow additional preparation fees to be available in t…

Katy600:
…he miscellaneious drop down

#### What
One of the potential changes to the AGFS fee schemes being discussed by policy is to amend the regulations to allow additional preparation fees on guilty pleas.
The objective of this spike is to identify where these failsafes are operating in CCCD that would prevent payment of an additional preparation fixed fee being added to a guilty plea.
Investigate CCCD to see if there are similar failsafes in place with regards additional fixed fees and guilty pleas.

This PR attempts to see how it would be done. 

#### Ticket
[Investigate CCCD failsafes ahead of potential AGFS fee scheme changes](https://dsdmoj.atlassian.net/browse/CTSKF-1635)

#### Why
To allow users under the AGFS Scheme 15+ to claim for additional preparation fees for a guilty‑plea‑specific 

#### How
Introduced explicit guilty‑plea‑only misc fee definitions

##### Added a GUILTY_ONLY_TYPES constant to Fee::MiscFeeType.
These codes define miscellaneous fees that are only applicable to guilty‑plea bill types.
The same codes are reused by CaseType#is_guilty_fee?

##### Added a guilty‑plea fee filter
Implemented a new private method, filter_guilty_only_types, in Claims::FetchEligibleMiscFeeTypes.
This mirrors the existing filter_trial_only_types approach, conditionally excluding guilty‑only fees when they are not applicable.

The filter is explicitly limited to AGFS Scheme 15+, preserving behaviour for earlier schemes.

##### Updated AGFS misc fee eligibility flow

Updated eligible_agfs_misc_fee_types to apply guilty‑plea filtering when the claim case type represents a guilty‑plea fee.
Added apply_guilty_fee_filter? to keep the intent of the filtering logic clear and readable.
Trial‑only and guilty‑only fee concerns remain separate, predictable, and easy to extend.

### DIFF
--- a/app/models/fee/misc_fee_type.rb
+++ b/app/models/fee/misc_fee_type.rb
@@ -5,12 +5,14 @@ module Fee
                                          MIDSE MIDSU MIPHC MISHR MISHU MIUMU MIUMO MISTE MIAPF MIFCM].freeze
     AGFS_SUPPLEMENTARY_TYPES = (AGFS_SUPPLEMENTARY_ONLY_TYPES + AGFS_SUPPLEMENTARY_SHARED_TYPES).freeze
     TRIAL_ONLY_TYPES = %w[MIUMU MIUMO MISTE MIAPF].freeze
+    GUILTY_ONLY_TYPES = %w[MIUAV1 MIUAV2].freeze
 
     default_scope { order(Arel.sql("regexp_replace(\"fee_types\".\"description\", '[()]', '', 'g')")) }
 
     scope :without_supplementary_only, -> { where.not(unique_code: AGFS_SUPPLEMENTARY_ONLY_TYPES) }
     scope :supplementary, -> { where(unique_code: AGFS_SUPPLEMENTARY_TYPES) }
     scope :without_trial_fee_only, -> { where.not(unique_code: TRIAL_ONLY_TYPES) }
+    scope :without_guilty_fee_only, -> { where.not(unique_code: GUILTY_ONLY_TYPES) }
 
     def fee_category_name
       'Miscellaneous Fees'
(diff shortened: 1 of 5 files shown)


app/models/fee/misc_fee_type.rb
```diff
@@ -5,12 +5,14 @@ class MiscFeeType < Fee::BaseFeeType
                                          MIDSE MIDSU MIPHC MISHR MISHU MIUMU MIUMO MISTE MIAPF MIFCM].freeze
     AGFS_SUPPLEMENTARY_TYPES = (AGFS_SUPPLEMENTARY_ONLY_TYPES + AGFS_SUPPLEMENTARY_SHARED_TYPES).freeze
     TRIAL_ONLY_TYPES = %w[MIUMU MIUMO MISTE MIAPF].freeze
+    GUILTY_ONLY_TYPES = %w[MIUAV1 MIUAV2].freeze
```
mpw5: I'm not sure the changes in this commit are quite right. These codes don't relate to additional preparation fees (they are `Unsuccessful application to vacate a guilty plea` (half/whole days). 

The fee code for additional preparation fees is `MIAPF` which is currently in the `TRIAL_ONLY_TYPES` array. I also don't think we need to create a new array for guilty-only types, as we want the additional preparation fee to be available for trial _and_ guilty case types. Possibly we need a new `TRIAL_AND_GUILTY_TYPES` variable which will just contain `MIAPF`.

Although, thinking about it a bit more, if we want to do this as part of a new fee scheme, we might want to leave `MIAPF` as it is (to allow additional preparation fees to be claimed on trial claims in fee schemes 15 and 16) and create a new code to allow them to claimed on trial and guilty claims on later fee schemes.

It's not a big problem as this is just a proof of concept, and it nicely highlights _where_ the changes need to be made, I'm just adding some notes for anyone who looks at this when working on the implementation of this in future.

Katy600: Thanks @mpw5 , I did feel uncertain about what codes I should add so that is good to know. That makes sense about incorporating it into TRIAL_AND_GUILTY - although thinking about it, won't it be used separately? A provider will want to filter Trial **or** Guilty not both? I shall have another look.

AnnaC-C:
Closing pull request due to fee scheme 17 now being merged in.

PR #9477, opened 2026-04-22
Fix caseworker access to documents on archived claims

mpw5:
#### What

A recent change (https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/9225) restricted caseworker access so that they can only view documents on claims that are allocated to them. This has caused some problems for caseworkers who have a legitimate need to view documents on claims that have already been assessed by another caseworker (for example where the claims is for a co-defendant on a case they are assessing a claim for).

This fix relaxes the restriction on caseworker access to documents so that they can view documents on claims that are in an archived state, even if they are not allocated to them.

Note: for caseworkers, claims appear in the archive as soon as they have been assessed. Claims may be in any of the following states: `authorised`, `part_authorised`, `rejected`, `refused`, `archived_pending_delete`, or `archived_pending_review`. This change uses the `Claims::StateMachine::CASEWORKER_DASHBOARD_ARCHIVED_STATES` constant to check for archived states, which includes all of the above.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -65,7 +65,7 @@ class Ability
     can_administer_any_provider if persona.roles.include?('provider_management')
     can %i[upload delete], Document
     can %i[show download], Document do |document|
-      document.claim.case_workers.include?(persona)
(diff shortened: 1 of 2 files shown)


ziaxgit (approved):
👍

PR #9504, opened 2026-04-30
[CTSKF-1670] Upgrade to rails 8.0

mpw5:
#### What

Upgrades rails to v8.0.5

#### Ticket

[CTSKF-1670](https://dsdmoj.atlassian.net/browse/CTSKF-1670)

#### Why

The current version of rails goes out of support in August 2026. This upgrade ensures we remain on a supported version of the framework.

#### How

* Updated `rails` to `8.0.5` in `Gemfile`
* Ran `bin/rails app:update`, keeping existing versions of files where prompted
* Incrementally enabled the following settings in `config/initializers/new_framework_defaults_8_0.rb`:
  * `Rails.application.config.active_support.to_time_preserves_timezone = :zone`
  * `Rails.application.config.action_dispatch.strict_freshness = true`
  * `Regexp.timeout = 1`
* Set config.load_defaults to 8.0 in config/application.rb and deleted config/initializers/new_framework_defaults_8_0.rb
* Updated `routes.rb` to address a definition which is now no longer valid due to stricter rules
* Updated controllers to use `params.expect(…)` over `params.require(…).permit(…)`. This also required updating three tests were where parameters were previously not being set correctly. Under the previous approach the tests passed however with the more strict `params.permit()` approach the tests fail. These are updated to ensure the tests pass with the new strong parameter handling.

[CTSKF-1670]: https://dsdmoj.atlassian.net/browse/CTSKF-1670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Gemfile.lock
```diff
@@ -90,22 +87,23 @@ GEM
...
       tzinfo (~> 2.0, >= 2.0.5)
+      uri (>= 0.13.1)
     addressable (2.9.0)
       public_suffix (>= 2.0.2, < 8.0)
     amoeba (3.3.0)
       activerecord (>= 5.2.0)
-    annotate (3.2.0)
-      activerecord (>= 3.2, < 8.0)
-      rake (>= 10.4, < 14.0)
+    annotate (2.6.5)
```
jrmhaig: I don't know if this is a major issue as this is a development dependency. The `annotate` gem doesn't officially support Rails 8 so it is being downgraded to the latest version without an upper limit. I found this issue, which proposes a fork: https://github.com/ctran/annotate_models/issues/1050

Alternatively, could this gem be removed?

Gemfile.lock
```diff
@@ -874,30 +870,30 @@ DEPENDENCIES
   zendesk_api (~> 3.1)
 
 CHECKSUMS
-  actioncable (7.2.3.1) sha256=d3bf40a3f4fc79a09709878f0e5c43a5e2d8e6607089f6b38f9472b8715eb33c
-  actionmailbox (7.2.3.1) sha256=a4e73480c97ab2fff5a416f92c54b065b1a6564ea4a807d42e0b83a94d4ec541
-  actionmailer (7.2.3.1) sha256=f578b6d5c5f81a20b6f6a796187698890c8348c041daa5e2e7cf7814ac520467
-  actionpack (7.2.3.1) sha256=b66afe7f937273270cb63f03bde7af7ba850017867766e8848d06d3e12e1e4ca
-  actiontext (7.2.3.1) sha256=5b1418f407ea347b98084a62b9b6caa1d3b1eb482d18dbbb69fad43f242843e3
-  actionview (7.2.3.1) sha256=de19b86843391762ac24a6287c30fbba11cd475fa4d4b664924d5fb7a2f1ff7c
+  actioncable (8.0.5)
```
jrmhaig: The shas are missing for all the upgraded packages. I'm not sure why.

config/routes.rb
```diff
@@ -203,7 +203,7 @@
 
   end
 
-  resources :injection_attempts, only: [:dismiss] do
+  resources :injection_attempts, only: [] do
```
jrmhaig, line 206: I presume that there isn't a tidier way of doing this without requiring the `only: []`.

spec/controllers/external_users/advocates/claims_controller_spec.rb
```diff
@@ -199,7 +199,7 @@ def expense_params
       end
 
       context 'submit to LAA with incomplete/invalid params' do
-        let(:invalid_claim_params) { { claim_class: 'Claim::AdvocateClaim' } }
+        let(:invalid_claim_params) { { advocate_category: 'QC' } }
```
roni-c-justice-dev, line 202: Is there a reason this has completely changed?  Everything else thus far has been a version of the earlier test data

config/routes.rb
mpw5: We could do `resources :injection_attempts do` but that would also generate other routes (`index`, `show`, `new` etc) which don't have corresponding controller actions. To just limit it to `dismiss` we need `only: []`.

Gemfile.lock
mpw5: Good spot. They _were_ populated when I initially spiked this 😕 I've found a way to populate them and pushed an additional commit.

spec/controllers/external_users/advocates/claims_controller_spec.rb
mpw5: `claim_class` isn't a valid parameter for this controller. With the old approach, unrecognised keys were silently dropped and an empty hash returned, which meant that the claim failed validation and the test passed.

With the new approach, if every submitted key is unrecognised Rails raises `ActionController::ParameterMissing` immediately and the test fails before reaching model validation.

I've amended the test so that it now contains a valid parameter but is missing other required params (court, case_number, etc.). This means `ActionController::ParameterMissing` is not thrown and allows the claim to be validated (and fails validation causing the test to pass).

(It's a good question and I probably should have gone into more detail about it in the commit message 😅 Let me know if it still doesn't make sense).

Gemfile.lock
mpw5: We're not using this gem so I've removed it.

roni-c-justice-dev (approved):

jrmhaig (approved):

PR #9549, opened 2026-05-12
[CTSKF-1054] Use govuk-component gem for links

mpw5:
#### What

Use the `govuk-component` gem to generate `govuk_link`s, replacing legacy helper methods.

#### Ticket

[CTSKF-1054](https://dsdmoj.atlassian.net/browse/CTSKF-1054)

#### Why

To move away from bespoke code to a standardised library.

#### How

Most links in the application are unaffected as the `govuk_link_to` method from the gem functions the same way as CCCD's legacy method. The only changes required are:
* adjusting `govuk_link_to` method calls where the third argument is a hash of options - this now needs to use a double splat
* replacing `govuk_back_link_to` with `govuk_back_link` and rearranging the order of arguments
* replacing `govuk_skip_link_to` with `govuk_skip_link` and rearranging the order of arguments
* replacing `govuk_mail_to` with `govuk_link_to` - the gem does not offer way of creating mail links, so just use a standard `govuk_link` annotated with `mailto:`
* removing redundant code

[CTSKF-1054]: https://dsdmoj.atlassian.net/browse/CTSKF-1054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrmhaig (approved):
:+1:

PR #9592, opened 2026-05-20
Ctskf 1067 cccd tidy up after migration to govuk components gem

Katy600:
#### What
We have now migrated all known components to use the govuk-components gem. This ticket tidies up some residual code.

#### Ticket

[CCCD: Tidy up after migration to govuk-components gem](https://dsdmoj.atlassian.net/browse/CTSKF-1067)

#### Why
We no longer need it. 

#### How
The lib/govuk-components directory and any remaining code within it is deleted

lib/govuk_component.rb is deleted

require_relative '../lib/govuk_component' is deleted from config/application.rb

Any code, CSS or Javascript which is now unused as a result of previous tickets has been deleted

.rubocop_todo.yml has been regenerated


config/environments/development.rb
```diff
@@ -12,7 +12,6 @@
   # Need to specifically set the logstasher loglevel since it will overwrite the one set earlier
   # Ideally this will be set to Logger::DEBUG but this is causing infinite loops using to_json
   # serialise data for logs. This could be fixed by;
-  #   * Further patching like lib/extensions/govuk_components_extension.rb
```
mpw5, line 15: I think we do need `lib/extensions/govuk_components_extension.rb` after all. I added it when introducing the govuk-components gem last year: https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/8467/changes/fe46d953e6323d209e809a86733cb473e0e09720. It relates to the gem not our previous helper methods.

So I think you need to reinstate this line, the changes to `config/initializers/00_extensions.rb`, and the `lib/extensions/govuk_components_extension.rb` file. Sorry!

jrmhaig (dismissed):
I ran the branch locally and everything seems to be OK.

Just a heads-up; The sha for `bundler` in `Gemfile.lock` seems to be coming and going and I am not sure why. I think this will result in a conflict for you so you may need to rebase just to get that back in line.

jrmhaig (approved):

PR #9596, opened 2026-05-21
Enable escape_json_responses

Katy600:
#### What
Skips escaping HTML entities and line separators. When set to `false`, the

#### Ticket

[CCCD: Enable `escape_json_responses` for Rails 8.1](https://dsdmoj.atlassian.net/browse/CTSKF-1706)

#### Why

JSON renderer no longer escapes these to improve performance.

#### How
Uncomment this line. 
Set the Rails.configuration.action_controller.escape_json_responses = false as the default value for Rails 8.1.

### DIFF
--- a/config/initializers/new_framework_defaults_8_1.rb
+++ b/config/initializers/new_framework_defaults_8_1.rb
@@ -25,7 +25,7 @@
 #
 # Applications that want to keep the escaping behavior can set the config to `true`.
 #++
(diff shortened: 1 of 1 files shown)


mpw5 (approved):
Providing you've deployed to an environment and done some testing (just making sure basic functionality works correctly), I think this is fine to be released.

PR #9644, opened 2026-06-01
fix(error msg): point existing validations at new error msg

roni-c-justice-dev:
#### Ticket

[[CCCD - Bug in Server Side Validation Error Handling](https://dsdmoj.atlassian.net/jira/software/c/projects/CTSKF/boards/987?selectedIssue=CTSKF-1582)]

#### Why
An incorrect error message was being returned, with the result that any date entered for Final Hearing Date on Graduated fees page before the rep order date is returning an error message relating to a date that should be within 2 years of the rep order date.  This occurs regardless of which fee scheme applies.

#### How
Add a new error message that simply states the date cannot be after the earliest representation order date
--------

#### Information
Intensive investigations could find no evidence for this 2 year rule, however, upon removing the validation rule entirely and pushing to production, other items such as allocation and authorisation were failing if there was data present in the DB that would not be compatible with the revised validation.  This required the original change to be reverted.  A decision has therefore been taken to implement the changes to error message only, to prevent this issue from re-occurring

### DIFF
--- a/config/locales/en/models/claim.yml
+++ b/config/locales/en/models/claim.yml
@@ -143,6 +143,7 @@ en:
               check_not_in_future: Enter a date that is not in the future
               not_before_earliest_permitted_date: Enter a date later than two years before the earliest representation order date
               too_long_before_earliest_reporder: The fee date cannot be more than two years before the earliest representation order date
+              not_before_earliest_reporder: The fee date cannot be before the earliest representation order
         disbursement:
           attributes:
             disbursement_type_id:
@@ -271,6 +272,8 @@ en:
               check_not_too_far_in_past: Fixed fee date cannot be too far in the past
               invalid_date: Enter a valid fixed fee date
               too_long_before_earliest_reporder: The fixed fee date cannot be no earlier than two years before the earliest representation order date
+              not_before_earliest_reporder: The fixed fee date cannot be before the earliest representation order
+
             fee_type:
               blank: 'Choose a type for the #{fixed_fee}'
             quantity:
@@ -303,6 +306,7 @@ en:
               check_not_too_far_in_past: Graduated fee date cannot be too far in the past
               invalid_date: Enter a valid graduated fee date
               too_long_before_earliest_reporder: The graduated fee date cannot be no earlier than two years before the earliest representation order date
+              not_before_earliest_reporder: The graduated fee date cannot be before the earliest representation order date
             quantity:
               numericality: Enter a valid quantity for the graduated fee
         fee/hardship_fee:
--- a/config/locales/en/models/claim.yml
+++ b/config/locales/en/models/claim.yml
@@ -143,7 +143,7 @@ en:
               check_not_in_future: Enter a date that is not in the future
               not_before_earliest_permitted_date: Enter a date later than two years before the earliest representation order date
               too_long_before_earliest_reporder: The fee date cannot be more than two years before the earliest representation order date
-              not_before_earliest_reporder: Cannot be before the earliest representation order
+              not_before_earliest_reporder: Cannot be before the earliest representation order date
         disbursement:
           attributes:
             disbursement_type_id:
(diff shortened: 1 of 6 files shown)


config/locales/en/models/claim.yml
```diff
@@ -303,6 +306,7 @@ en:
               check_not_too_far_in_past: Graduated fee date cannot be too far in the past
               invalid_date: Enter a valid graduated fee date
               too_long_before_earliest_reporder: The graduated fee date cannot be no earlier than two years before the earliest representation order date
+              not_before_earliest_reporder: Cannot be before the earliest representation order date
```
mpw5, line 309: This entry has `date` at the end of it, the other two are just `Cannot be before the earliest representation order`. I'm not sure which is correct but they should probably be consistent.

spec/support/shared_examples_for_fee_validators.rb
```diff
@@ -25,7 +25,17 @@
 
       fee.date = Time.zone.today - 3.days
       expect(fee).not_to be_valid
-      expect(fee.errors[:date]).to include(match(/.* date cannot be no earlier than two years before .*/))
+      expect(fee.errors[:date]).to include(match(/.*Cannot be before the earliest representation order*/))
```
mpw5: I'm not sure it's making a huge difference but the regex originally ended with `.*`. Your new version seems to have lost the `.`.

spec/validators/date_attended_validator_spec.rb
```diff
@@ -27,7 +27,7 @@
     it do
       should_error_if_before_specified_date(
         date_attended, :date, representation_order_date - 2.years - 1.day,
-        'The fee date cannot be before the earliest representation order'
+        'The fee date cannot be more than two years before the earliest representation order date'
       )
```
Katy600, line 31: Hi @roni-c-justice-dev,  just to clarify the overall purpose of this ticket - its objective is to replace the error messages in all cases with `Cannot be before the earliest representation order date` even though the logic behind it is checking that the date is prior to 2 years `representation_order_date - 2.years - 1.day`. Future tickets have been created to tackle this issue separately - front end. What is the reason for having this test though - that checks the error message displays `cannot be more than two years`. Is that error message still relevant anywhere?

spec/support/shared_examples_for_fee_validators.rb
```diff
@@ -25,7 +25,17 @@
...
+    end
+
+    it 'adds same error if more than 2 years before the first repo order date' do
+      # because the fee date is still before the earliest reporder the message should be the same
```
mpw5, line 32: I'm not sure the comment adds much but I'm not going to make you take it out :)

But it got me thinking about the scope of the change. If a claim where a date more than two years before the rep order date is entered now shows the new message, do we need the original `too_long_before_earliest_reporder` error in the system any more? The only place it is referenced in `DateAttendedValidator`. It might be possible to remove the line that executes that validation (although at that point we're possibly moving towards the sort of changes which might have unexpected impacts on users 😬).

mpw5 (commented):
I've left a few comments, happy to discuss if necessary.

spec/support/shared_examples_for_fee_validators.rb
roni-c-justice-dev: Upon discussion and due to previous unexpected failure in production when the validation code was amended, will leave this in at present, to address under any future validation work more formally.

spec/validators/date_attended_validator_spec.rb
roni-c-justice-dev: Because we've not removed the 2 year logic entirely, but focused purely on the 3 fee scenarios that triggered the incorrect error, there are still potential situations in the codebase that can trigger this.  I've therefore chosen to leave in a test appropriate to that situation.

config/locales/en/models/claim.yml
roni-c-justice-dev: sorted thanks

mpw5 (approved):
LGTM 🎉

PR #9670, opened 2026-06-08
chore(alpine): update alpine and ruby

roni-c-justice-dev:
#### What
Update alpine version to 3.23, ruby to 3.4.8, cimg/ruby to cimg/ruby:3.4.8-browsers
Fix rubocop errors in multiple files
Fix attach_evidence and page_title tests to prevent flakiness in pipeline runs.

#### Ticket

(https://dsdmoj.atlassian.net/jira/software/c/projects/CTSKF/boards/987?selectedIssue=CTSKF-1614)
#### Why
Ruby version needed to be uplifted to allow the alpine version uplift which was showing a CVE at the time.
#### How
Update alpine version to 3.23
Update ruby to 3.4.8
Update cimg/ruby to cimg/ruby:3.4.8-browsers
Fix rubocop errors in multiple files
Fix attach_evidence and page_title tests to prevent flakiness in pipeline runs.
------


.circleci/config.yml
```diff
@@ -405,7 +405,7 @@ jobs:
 
   cucumber-tests:
     executor: test-executor
-    resource_class: small
+    resource_class: large
```
mpw5, line 408: Really interesting how much of a difference this has made in the time taken to execute those tests 🤯 

I looked at it a few years ago and found that increasing the `resource_class` didn't make much difference, so I increased the `parallelism` instead. I wonder if something has changed at CircleCI (different server architecture or something) that means the server size has more of an impact now. 

(Might be worth trying the rspec tests on X-large at some point, although it starts getting expensive 😬)

features/step_definitions/common_steps.rb
```diff
@@ -135,6 +135,7 @@
   expect(page).to have_title(page_title)
 end
 
+
```
mpw5: The nitpickiest of nitpicks: we don't really need the extra whitespace here.

Katy600 (dismissed):

Gemfile.lock
```diff
@@ -872,295 +897,8 @@ DEPENDENCIES
   webmock
   zendesk_api (~> 3.1)
 
-CHECKSUMS
```
mpw5, line 875: Something seems to have gone wrong with `bundle` and all the checksums have disappeared. The same thing happened to me a little while ago.

I think what I did was checkout the lockfile (`git checkout origin -- Gemfile.lock`) and then run `bundle install` again. Hopefully that should cause the checksums to reappear and you can recommit the file 🤞

mpw5 (commented):
Looks mostly good apart from a problem with the `Gemfile.lock` file which I don't think is your fault. I've suggested a way to fix it but let me know if it doesn't work and I'm happy to work with you on it.

Gemfile.lock
roni-c-justice-dev: I suspect I might have accidentally deleted them when dealing with a merge conflict in the lockfile. Will give that a shot

roni-c-justice-dev: should be resolved now

mpw5 (approved):
🥇

PR #9689, opened 2026-06-10
Revert "Remove images from the file path as images is now the default…

Katy600:
… place for favicon (#9680)"

This reverts commit 5a0260b6d147bec510be9554dc91b0d76196401f.

Revert changes to file path

#### What
This needs to be reverted as it seems to be a localised problem on my computer.


mpw5 (approved):
👍

PR #9714, opened 2026-06-16
Ctskf 880 cccd update h1 to reflect page content updated

Katy600:
#### What

Change the h1 on each step of the advocate/litigator journey to reflect the page content.
Add the bill type as a caption above the h1

#### Ticket

[CCCD - Update H1 to reflect page content](https://dsdmoj.atlassian.net/browse/CTSKF-880)

#### Why

The h1 on each page does not reflect what the content of the page is about. Every h1 in the advocate/litigator journey is ‘claim for X’. This is causing some problems with hierarchy and making it more difficult for users to understand what they need to do on each page

#### How

Use GOV design system caption to show bill type. So each heading will consist of:

`<span class="govuk-caption-xl">[bill type]</span>`
`<h1 class="govuk-heading-xl">[H1 as per below]</h1>`


app/views/layouts/_main_header.html.haml
```diff
@@ -0,0 +1,5 @@
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    - title(page_heading) unless content_for?(:page_title)
+    %h1#page-h1.govuk-heading-xl
+      = page_heading
```
mpw5, line 5: I'm not sure that `main_header` is the best name for this. I think I understand why we now need two header partials (because there are different types of page which need different header structures?) but I'm not sure the names of them make their purposes clear.

I wonder if we can rename one or both of them so the purpose is more obvious?

To me it looks like the `_header` partial now generates a caption, not a heading, so could this now be `_page_caption`?

And `_main_header` generates the heading for each page, so maybe something like `_page_header` (or `_page_heading`)? 

Just suggestions, I will leave the actual names up to you 🙂

app/views/layouts/_header.html.haml
```diff
@@ -1,5 +1,4 @@
 .govuk-grid-row
   .govuk-grid-column-two-thirds
     - title(page_heading) unless content_for?(:page_title)
-    %h1#page-h1.govuk-heading-xl
-      = page_heading
+    %span.govuk-caption-xl=page_heading
```
mpw5: I think this could be changed to
```suggestion
    %span.govuk-caption-xl
      =page_heading
```

This would be more inline with the previous code and the main_header partial, and is probably more readable.

mpw5 (commented):
I think this mostly looks really good. I've left a comment about the naming of partials and some suggested alternatives, and a minor code change.

Also, minor thing, but I'm not sure that a commit titled `Does this fix anything?` is ideal (although I completely get why you probably called it that 😁). Something more descriptive would be  helpful, or maybe just squash it into the previous commit as it looks like they're related?

Edit to add: I think it's also worth reviewing any places where `layouts/header` is currently used, now that you've changed that from creating a header to a caption. There are a couple of places I've spotted where I'm not sure it's correct, eg on `app/views/cookies/cookie_details.html.haml`

Before: 
image

After:
image

app/views/devise/confirmations/new.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = 'Resend confirmation instructions'
 
-= render partial: 'layouts/header', locals: {page_heading: 'Resend confirmation instructions'}
+= render partial: 'layouts/page_caption', locals: {page_heading: 'Resend confirmation instructions'}
```
mpw5: This was a header but is now a caption. Is that correct? Not necessarily a problem, if it's intentional and Jonathan is ok with it is fine, but I think we need to be sure it's right.

Before and after screenshots for example:

Image 

Image

app/views/devise/passwords/edit.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
+= render partial: 'layouts/page_caption', locals: {page_heading: t('.page_heading')}
```
mpw5: As above, this was a header but is now a caption, is that right?

app/views/devise/passwords/new.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
+= render partial: 'layouts/page_heading', locals: {page_heading: t('.page_heading')}
```
mpw5, line 4: As above.

app/views/devise/registrations/new.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
+= render partial: 'layouts/page_heading', locals: { page_heading: t('.page_heading') }
```
mpw5, line 4: As above.

app/views/devise/sessions/new.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
+= render partial: 'layouts/page_heading', locals: {page_heading: t('.page_heading')}
```
mpw5, line 4: As above.

app/views/devise/unlocks/new.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
+= render partial: 'layouts/page_heading', locals: {page_heading: t('.page_heading')}
```
mpw5, line 4: As above.

app/views/errors/internal_server_error.html.haml
```diff
@@ -1 +1 @@
-= render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
+= render partial: 'layouts/page_caption', locals: { page_heading: t('.page_heading') }
```
mpw5: As above.

app/views/errors/not_found.html.haml
```diff
@@ -1 +1 @@
-= render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
+= render partial: 'layouts/page_caption', locals: { page_heading: t('.page_heading') }
```
mpw5: As above.

app/views/external_users/admin/external_users/change_password.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
+= render partial: 'layouts/page_caption', locals: { page_heading: t('.page_heading') }
```
mpw5: As above.

app/views/external_users/claims/outstanding.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title')
 
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
+= render partial: 'layouts/page_caption', locals: {page_heading: t('.page_heading')}
```
mpw5: As above.

app/views/pages/hardship_claims.html.haml
```diff
@@ -1,4 +1,4 @@
-= render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
+= render partial: 'layouts/page_caption', locals: { page_heading: t('.page_heading') }
```
mpw5: As above.

app/views/pages/servicedown.html.haml
```diff
@@ -1,5 +1,5 @@
 %section
-  = render partial: 'layouts/header', locals: { page_heading: t('.page_heading') }
+  = render partial: 'layouts/page_caption', locals: { page_heading: t('.page_heading') }
```
mpw5: As above.

app/views/super_admins/admin/super_admins/change_password.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title', super_admin: @super_admin.name)
 
-= render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
+= render partial: 'layouts/page_caption', locals: {page_heading: t('.page_heading')}
```
mpw5: As above.

app/views/super_admins/admin/super_admins/show.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = t('.page_title', super_admin: @super_admin.name)
 
-= render partial: 'layouts/header', locals: {page_heading: 'Your account'}
+= render partial: 'layouts/page_caption', locals: {page_heading: 'Your account'}
```
mpw5: As above.

app/views/super_admins/offences/index.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for :page_title, flush: true do
   = 'Offences'
 
-= render partial: 'layouts/header', locals: { page_heading: 'Offences' }
+= render partial: 'layouts/page_caption', locals: { page_heading: 'Offences' }
```
mpw5: As above.

app/views/users/index.html.haml
```diff
@@ -1,21 +1,21 @@
 = content_for :page_title, flush: true do
   = t('.heading')
 
-= render partial: 'layouts/header', locals: { page_heading: t('.heading') }
+= render partial: 'layouts/page_caption', locals: { page_heading: t('.heading') }
```
mpw5: As above.

mpw5 (commented):
Nearly there, I've found a few examples that previously had headers but now have captions. I think they should probably be still be headers, unless there's a reason for changing them. I've left a comment on each one, happy to go over them with you.

app/views/devise/confirmations/new.html.haml
Katy600: Thank for your follow up feedback @mpw5 . 

I meant to ask you about a few file paths that I was struggling to reproduce and then I must have forgotten. I apologise about that. Parts of the devise file paths were areas that I wasn't sure how to check. So it should all look the same as before. However, the 'heading' partial no longer exists as I renamed that to 'page_caption' and then the newly formatted one was named 'page_heading'. So the other pages with 'page_heading' should be correct. 

I will go ahead and change the ones called page_caption to be page_headingg unless you think the naming of it is now confusing.

mpw5 (approved):
🎉

app/views/devise/unlocks/new.html.haml
mpw5: This was a mistake on my part, the heading here is actually correct, sorry 🤦

app/views/devise/sessions/new.html.haml
mpw5: This was a mistake on my part, the heading here is actually correct, sorry 🤦

app/views/devise/registrations/new.html.haml
mpw5: This was a mistake on my part, the heading here is actually correct, sorry 🤦

app/views/devise/passwords/new.html.haml
mpw5: This was a mistake on my part, the heading here is actually correct, sorry 🤦

PR #9724, opened 2026-06-18
chore(tests): ensure upload tests wait appropriately

roni-c-justice-dev:
#### What
Tests relying on upload and some tests relying on other components being available, have been failing intermittently in the pipelines.

#### Ticket
no direct ticket, picked up while working on CTSKF-1614 but separating issues into individual PRs for readability

#### Fix
edit the affected implementation along with some tests to ensure that the implemented behaviour waits fully, but that also the tests wait further where needed

### DIFF
--- a/features/page_objects/claim_form_page.rb
+++ b/features/page_objects/claim_form_page.rb
@@ -122,13 +122,28 @@ class ClaimFormPage < BasePage
   end
 
   def attach_evidence(count: 1, document: '*')
-    count ||= 1
-    available_docs = Dir.glob "#{Rails.root}/spec/fixtures/files/#{document.gsub('.pdf','')}.pdf"
-    available_docs[0...count].each do |path|
-      # element needs to be visible in order to attach_file
-      page.execute_script("$('.moj-multi-file-upload__input').css('position','unset')")
-      input_field = page.find("input[name='attachments']")
-      input_field.attach_file(path)
+    pattern = document.to_s.gsub('.pdf', '')
+    paths = Dir.glob(Rails.root.join("spec/fixtures/files/#{pattern}.pdf"))
+
+    raise "No files found for pattern: #{document}" if paths.empty?
+
+    paths.first(count).each do |path|
+      filename = File.basename(path)
+
+      # ✅ Re-find input every time (handles DOM replacement)
+      input_field = find("input[name='attachments']", visible: :all)
+
+      # ✅ Use Capybara-supported visibility handling
+      input_field.attach_file(path, visible: :all)
+
+      # ✅ Wait for upload confirmation
+      unless page.has_selector?(
+        '.moj-multi-file-upload__success',
+        text: filename,
+        wait: Capybara.default_max_wait_time
+      )
+        raise "Upload failed: #{filename} not displayed on #{page.current_path}"
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


features/page_objects/claim_form_page.rb
```diff
@@ -122,13 +122,28 @@ def add_govuk_misc_fee_if_required
...
-    available_docs = Dir.glob "#{Rails.root}/spec/fixtures/files/#{document.gsub('.pdf','')}.pdf"
-    available_docs[0...count].each do |path|
-      # element needs to be visible in order to attach_file
-      page.execute_script("$('.moj-multi-file-upload__input').css('position','unset')")
-      input_field = page.find("input[name='attachments']")
-      input_field.attach_file(path)
+    pattern = document.to_s.gsub('.pdf', '')
+    paths = Dir.glob(Rails.root.join("spec/fixtures/files/#{pattern}.pdf"))
+
+    raise "No files found for pattern: #{document}" if paths.empty?
```
mpw5: SonarCloud has flagged this with "Use a specific exception class instead of raising a string literal.". I think it's worth considering adding an exception class to it, as `raise` should always return an exception class, not a bare string.

`raise RuntimeError, "No files found for pattern: #{document}" if paths.empty?` or `raise ArgumentError, "No files found for pattern: #{document}" if paths.empty?` would probably both be appropriate.

features/page_objects/claim_form_page.rb
```diff
@@ -122,13 +122,28 @@ def add_govuk_misc_fee_if_required
...
+    paths.first(count).each do |path|
+      filename = File.basename(path)
+
+      # ✅ Re-find input every time (handles DOM replacement)
+      input_field = find("input[name='attachments']", visible: :all)
+
+      # ✅ Use Capybara-supported visibility handling
+      input_field.attach_file(path, visible: :all)
+
+      # ✅ Wait for upload confirmation
```
mpw5: Nitpick: I'm not sure the comments add much, I'd probably remove them. Not a blocker if you want to keep them though 🙂

features/step_definitions/common_claim_steps.rb
```diff
@@ -16,7 +16,10 @@
 end
 
 When(/^I select a case type of '(.*?)'$/) do |case_type|
+  patiently do
   @claim_form_page.auto_case_type.choose_autocomplete_option(case_type)
```
mpw5: Nitpick: this should be indented within the do/end block.

features/page_objects/claim_form_page.rb
```diff
@@ -122,13 +122,28 @@ def add_govuk_misc_fee_if_required
...
+      unless page.has_selector?(
+        '.moj-multi-file-upload__success',
+        text: filename,
+        wait: Capybara.default_max_wait_time
+      )
+        raise "Upload failed: #{filename} not displayed on #{page.current_path}"
```
mpw5: As above.

mpw5 (commented):
I've left a couple of comments, nothing major

features/page_objects/claim_form_page.rb
roni-c-justice-dev: Ironically it was flagging for raising a RuntimeError and telling me there was no need to do so.  Copilot suggested the string literal.  I'll try it with ArgumentError to see if it likes that better.

features/page_objects/claim_form_page.rb
roni-c-justice-dev: Have reduced the comments down to one that I think should stay

features/step_definitions/common_claim_steps.rb
roni-c-justice-dev: changed in next push

mpw5 (approved):

PR #9725, opened 2026-06-18
[CTSKF-867] Sortable table consistency

mpw5:
#### What

Addresses a number of issues with the following sortable tables:

**Advocate/Litigator**: Your claims, Your claim archive

**Caseworker**: Your claims, Your archive, Allocation

#### Ticket

[CTSKF-867)](https://dsdmoj.atlassian.net/browse/CTSKF-867)

#### Why

To improve accessibility and align with design standards

#### How

This PR combines the following fixes:

1) [fefb7abcadfac04e9b16fec8d79369660bcc3edf] prevents upper-casing of labels in the mobile versions of tables by removing the `text-transform: uppercase` styling.

image
image

------

2) [ead2ff3351b6609e60e521911d8371de0989cb15] changes the headers of the mobile versions of tables to a more accessible shade of grey by updating `background-color` to `govuk-colour("light-grey")`.

image
 image

------

3) [8ccc47610db1b62ca2fdf39ac2796df25c823f32] changes the row separators of mobile versions of tables to a more accessible shade of grey by setting `border-bottom` to `$govuk-border-colour`

image image

------

4) [7ca7f466ded2e43a4b05f1d1c4338d6c450db5e0] adds MOJ design system [sortable table](https://design-patterns.service.justice.gov.uk/components/sortable-table/) styling to the 'Your claims' and 'Your archive' tables for providers and caseworkers. Note that this does not use the actual sortable table component as that is only compatible with client-side tables whereas all our tables rely on server-side sorting and pagination. Instead the styling is copied to our own `_tables.scss` file. Additionally, it is not applied to the allocation and reallocation tables as the use of the DataTables gem makes adding bespoke styling to those tables challenging.

image
image

------

5) [a651f59936bc3588f5fd9c7192f4cdbdaf3dc228] fixes the width of the mobile version of the allocations table, which was previously too narrow, by setting the DataTables `autoWidth` variable to `false`.

image

image

------

6) [1fa38254af7469e3d2344d8e63ffedf8bbc36949] fixes an issue on mobile tables where the Case Type field, which is a combination of the case type and the claim state separated by a line break, is being displayed without the line break, as concatenated text on a single line, which makes it hard to read, buy adding `display: block;` styling.

image
 
image

------

7) [0ec10b8668514f911b0543766575d69e88de2bd0] fixes the alignment of checkboxes on the mobile version of the allocations table by adding some additional styling.

image
image

### DIFF
--- a/app/webpack/stylesheets/components/_table.scss
+++ b/app/webpack/stylesheets/components/_table.scss
@@ -81,7 +81,7 @@ tr {
 
     tbody tr:nth-child(odd),
     tbody tr:nth-child(even) {
-      background-color: none;
+      background-color: transparent;
     }
 
     th::before,
--- a/app/webpack/stylesheets/components/_table.scss
+++ b/app/webpack/stylesheets/components/_table.scss
@@ -181,3 +181,94 @@ td.active {
   --dt-row-selected-text: 0,0,0;
   --dt-row-selected-link: 26, 101, 166;
 }
+
+// Style sortable table column headers using MOJ Sortable Table styling
+$sortable-indicator-size: 22px;
+$sortable-indicator-mask-unsorted: "data:image/svg+xml,%3Csvg width='22' height='22' viewBox='0 0 22 22' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M8.1875 9.5L10.9609 3.95703L13.7344 9.5H8.1875Z' fill='black'/%3E%3Cpath d='M13.7344 12.0781L10.9609 17.6211L8.1875 12.0781H13.7344Z' fill='black'/%3E%3C/svg%3E";
+$sortable-indicator-mask-asc: "data:image/svg+xml,%3Csvg width='22' height='22' viewBox='0 0 22 22' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M6.5625 15.5L11 6.63125L15.4375 15.5H6.5625Z' fill='black'/%3E%3C/svg%3E";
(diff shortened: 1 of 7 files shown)


app/webpack/stylesheets/components/_table.scss
```diff
@@ -67,7 +67,7 @@ tr {
     }
 
     tbody tr td:first-child {
```
roni-c-justice-dev, line 69: Do we need a th in here if this is to apply to the table header as well?

app/webpack/stylesheets/components/_table.scss
```diff
@@ -89,7 +89,6 @@ tr {
       content: attr(data-label);
       float: left;
       font-weight: 700;
-      text-transform: uppercase;
```
roni-c-justice-dev, line 92: are we entirely certain that nothing further downstream in the backend is expecting an upper-case response?

app/webpack/stylesheets/components/_table.scss
mpw5: I've probably done a bad job of describing the change here. It's not actually the table header that's affected. On the mobile version of the table, the first item of each record is shaded grey, to break up the data. It's not an actual header but in my mind it kind of acts like one. `tbody tr td:first-child` is the correct selector for it.

image

app/webpack/stylesheets/components/_table.scss
mpw5: This only changes the way the text is displayed on the screen, not in the backend, and only applies to key names in the table, not the actual values, so no application data is affected

image

ziaxgit (approved):

app/webpack/stylesheets/components/_table.scss
roni-c-justice-dev: Makes sense, unless we have any Cucumber tests directly linked to them then happy to approve

roni-c-justice-dev (approved):

PR #9734, opened 2026-06-19
[Do not merge] CCCD: Ctskf-1741 integrate entra endpoints

VinceChiuMOJ:
#### What
Silas: Integrate real Entra ID endpoints into OmniAuth flow (dev)

#### Ticket
[CTSKF-1741](https://dsdmoj.atlassian.net/browse/CTSKF-1741)

[CTSKF-1741]: https://dsdmoj.atlassian.net/browse/CTSKF-1741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/initializers/zeitwerk_inflections.rb
AnnaC-C, line 1: Can move to `config/initializers/inflections`

app/controllers/omniauth_callbacks_controller.rb
AnnaC-C, line 1: For when it goes into main - splitting the functions, and adding some unit/integration tests

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,519 @@
...
+
+    auth_hash = auth.to_h
+    credentials = auth_hash['credentials'] || {}
+    credentials = credentials.except('token', 'id_token', 'refresh_token', :token, :id_token, :refresh_token)
+    auth_hash['credentials'] = credentials
+
+    payload = {
+      event: 'omniauth_callback_full',
+      auth: auth_hash
+    }
```
AnnaC-C, line 91: Can we add in some error messages for the authentication where it may be wrong. 

I'm not entirely sure what error message in practise, (invalid email? Role?) but it would be good to prove them out.

AnnaC-C: I've noticed some errors lower down so might not need to add them

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,519 @@
...
+
+    SupplierNumber.find_by!([SUPPLIER_NUMBER_CASE_INSENSITIVE_WHERE, matching_supplier_number]).provider
+  end
+
+  def matched_provider_from_firm_name(firm_name)
+    normalized_firm_name = normalize_firm_name(firm_name)
+
+    Provider.find_each do |provider|
+      return provider if normalized_firm_name == normalize_firm_name(provider.name)
+    end
```
mpw5, line 417: It feels like calling `normalize_firm_name` on every provider could be expensive and slow. It's probably fine for a proof of concept but something to be aware of when we go to production. We should test its performance at least.

Thinking it through a bit more, if this loop was very slow to complete, is there a risk that two users from the same firm could log in at the same time and, while the loop is completing, both cause a new provider record to be created, leading to a duplicate firm? We might want to add some sort of defensive constraint to prevent dupes.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,519 @@
...
+    user = nil
+    User.transaction { user = yield }
+
+    [user, nil]
+  rescue StandardError => e
+    [nil, provision_failed_message(email, e)]
+  end
+
+  def auto_create_providers_from_silas?
+    ActiveModel::Type::Boolean.new.cast(ENV.fetch('AUTO_CREATE_PROVIDERS_FROM_SILAS', 'false'))
```
mpw5, line 306: If I'm understanding this correctly it looks like this acts as a feature flag to control whether providers are auto-created, but there doesn't seem to be a similar flag for the auto-creation of external, caseworker or super admin users. Should there be? If this isn't being deployed to production in its current state then this probably isn't a problem but maybe extra caution would be useful anyway.

mpw5 (commented):
I've left a couple of extra comments in addition to what Anna has already mentioned. I think it's fine as a POC to enable further development work but there are probably a few things to address before it's ready for production.

config/initializers/zeitwerk_inflections.rb
VinceChiuMOJ: I tried to move it. It doesn't work and cause a series of errors on how devise recognise omniauth files

app/controllers/omniauth_callbacks_controller.rb
VinceChiuMOJ: Good point, we definitely need a better approach on this. I thought we may have something more unique like a firm_code when we go live thats why I just leave it this way in this POC. It does not seem to be the case now so we may need to revisit this provider name matching later.

app/controllers/omniauth_callbacks_controller.rb
VinceChiuMOJ: Yes, you get it right. I thought eventually we would be able to create everything from what we got from Silas and provider is the only blockers, so I just make a flag on providers. 
For super admin, I don't think it would be included in SiLAS work flow from now. My thought is to handle it when we implement the solution for multi-firm users.

PR #9757, opened 2026-06-24
Remove json_spec gem

mpw5:
The `json_spec` library is used in some rspec tests to provide JSON matchers. The gem has been unmaintained since 2017 and we probably shouldn't be using it any more. Alternatives exist, such as [rspec-json_expectations](https://github.com/waterlink/rspec-json_expectations) and [rspec-oj](https://github.com/mhenrixon/rspec-oj) however these also appear to have been abandoned.

This PR removes the library and replaces its usage with standard rspec equivalents using `JSON.parse` and `dig`.

This mostly affects test which check the content of JSON arrays. For example, tests in the form

```
is_expected.to have_json_size 1
```
have been replaced with 
```
expect(parsed.size).to eq(1)
```

and

```
is_expected.to be_json_eql('true').at_path('defendants/0/main_defendant')
```

have been replaced with
```
expect(parsed.dig('defendants', 0, 'main_defendant')).to eq(true)
```

which has the same affect and is not (in my opinion) particularly less readable.


spec/requests/pages/servicedown_spec.rb
```diff
@@ -19,14 +19,14 @@
...
-    it { expect(response.body).to include_json({ error: 'Service temporarily unavailable' }.to_json) }
+    it { expect(JSON.parse(response.body)).to include('error' => 'Service temporarily unavailable') }
   end
 
   context '/ping' do
     before { get '/ping' }
 
     it { expect(response).to be_ok }
-    it { expect(response.body).to be_json }
+    it { expect(response.media_type).to eq('application/json') }
```
AnnaC-C: Why are we changing from `body` to `media_type` ?

spec/requests/pages/servicedown_spec.rb
```diff
@@ -19,14 +19,14 @@
   shared_examples 'maintenance json' do
     it { expect(response).to have_http_status :service_unavailable }
     it { expect(response.body).to be_json }
```
AnnaC-C, line 21: Does this also need to be changed to 
`  it { expect(response.media_type).to eq('application/json') }`
?

AnnaC-C (commented):
A couple of questions. Looking a lot neater without the json_spec gem

spec/requests/pages/servicedown_spec.rb
mpw5: I... can't remember 😅 I think it might have been something I was experimenting with and left in. It's not necessary so I've reverted it to the previous code.

spec/requests/pages/servicedown_spec.rb
mpw5: No, actually the reverse. This is still correct, I've reverted the other changes 😁

AnnaC-C (approved):
LGTM 👍

PR #9777, opened 2026-06-30
Remove awesome_print from the Gemfile

Katy600:
#### What
The [awesome_print](https://rubygems.org/gems/awesome_print) gem appears to have been abandoned, having not had a release since 2019 in spite of multiple issues and pull requests.

The awesome_print gem allows strings to be coloured using (for example) ”hello, world”.green and this is used in various places such as cucumber tests and Rake tasks. Other gems provide the same functionality;

[amazing_print](https://rubygems.org/gems/amazing_print) is a fork of awesome_print that is still in active development.

[rainbow](https://rubygems.org/gems/rainbow) is already included as dependencies of other gems (rubocop and i18n-tasks). This would reduce the number of dependencies but may require various code_changes.

#### Ticket

[CCCD - Replace awesome_print](https://dsdmoj.atlassian.net/browse/CTSKF-1550)

#### Why
The awesome_print is no longer included as a dependency of CCCD

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,6 @@ gem 'auto_strip_attributes',  '~> 2.6.0'
 gem 'aws-sdk-s3',             '~> 1'
 gem 'aws-sdk-sns',            '~> 1'
 gem 'aws-sdk-sqs',            '~> 1'
-gem 'awesome_print'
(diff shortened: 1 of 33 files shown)


mpw5 (approved):
LGTM

PR #9778, opened 2026-06-30
Attempt to fix flaky cucumber test

mpw5:
#### What

One specific cucumber test has started failing intermittently: `features/super_admins/stats.feature`. This PR attempts to fix it by adding a short delay to the affected step, which will hopefully give the page enough time to load before the assertion is run.

#### Why

To reduce failed tests and speed up successful completion of the test suite, reducing costs and preventing developers wasting their time.


Katy600 (approved):

AnnaC-C (approved):
LGTM 👍  - Code question on behaviour

PR #9791, opened 2026-07-03
[CTSKF-1690] AGFS scheme 17: Additional preparation fee claim injection

mpw5:
#### What

Additional preparation fees are currently excluded from injection into CCR. From AGFS scheme 17, this fee can be injected. Changes are required to enable those fees to be injected and to ensure that warnings displayed to caseworkers on claims where additional preparation fees have _not_ been injected, are not displayed on scheme 17 claims.

#### Ticket

[CTSKF-1690](https://dsdmoj.atlassian.net/browse/CTSKF-1690)

#### Why

To make caseworkers lives easier by automatically creating additional preparation fees in CCR rather than make them manually create them.

#### How

1. To enable the injection of additional preparation fees on scheme 17 claims, this updates `API::Entities::CCR::BaseClaim` to  pass a new `agfs_scheme_17_or_later` flag from to the `CCR::Fee::MiscFeeAdapter`, which is used to determine whether additional preparation fees should be excluded from injection or not.

2. There are two places where warnings about additional preparation fee injection are displayed to caseworkers. To suppress these in both places this:
    - adds extra logic to `app/views/warnings/_injection_warnings_summary.html.haml` to check for scheme 17+ eligibility
    - adds a new method to `API::Helpers::SearchResultHelpers` to determine scheme 17+ eligibility, via the earliest representation order date from the claim's defendants, which is added to the unallocated claims SQL query in `API::V2::QueryHelper`.

[CTSKF-1690]: https://dsdmoj.atlassian.net/browse/CTSKF-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/ccr/fee/misc_fee_adapter_spec.rb
+++ b/spec/services/ccr/fee/misc_fee_adapter_spec.rb
@@ -12,7 +12,8 @@ RSpec.describe CCR::Fee::MiscFeeAdapter, type: :adapter do
 
   it_behaves_like 'a mapping fee adapter'
 
-  EXCLUSIONS = %i[BACAV MIPHC MIUMU MIUMO MIAPF].freeze
+  EXCLUSIONS_PRE_SCHEME_17 = %i[BACAV MIPHC MIUMU MIUMO MIAPF].freeze
+  EXCLUSIONS = %i[BACAV MIPHC MIUMU MIUMO].freeze
 
   MAPPINGS = {
     BACAV: %w[AGFS_MISC_FEES AGFS_CONFERENCE], # Conferences and views (basic fee)
@@ -60,7 +61,7 @@ RSpec.describe CCR::Fee::MiscFeeAdapter, type: :adapter do
   }.freeze
 
   def self.mappings(exclusions: true)
-    exclusions ? MAPPINGS.except(*EXCLUSIONS) : MAPPINGS
+    exclusions ? MAPPINGS.except(*EXCLUSIONS_PRE_SCHEME_17) : MAPPINGS
   end
 
   describe '#bill_type' do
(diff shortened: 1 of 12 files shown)


spec/services/ccr/fee/misc_fee_adapter_spec.rb
```diff
@@ -60,7 +58,7 @@
   }.freeze
 
   def self.mappings(exclusions: true)
-    exclusions ? MAPPINGS.except(*EXCLUSIONS) : MAPPINGS
+    exclusions ? MAPPINGS.except(*CCR::Fee::MiscFeeAdapter::MISC_FEE_BILL_MAPPING_EXCLUSIONS_PRE_SCHEME_17) : MAPPINGS
```
AnnaC-C:
```suggestion
    exclusions ? MAPPINGS.except(*described_class::MISC_FEE_BILL_MAPPING_EXCLUSIONS_PRE_SCHEME_17) : MAPPINGS
```

Let's be consistent with declarations
Or lower down change to the same variable here

AnnaC-C (changes requested):
Just one comment

spec/services/ccr/fee/misc_fee_adapter_spec.rb
mpw5: Instead I've stayed closer to what the test was doing originally and defined two sets of exclusions
```
  EXCLUSIONS_PRE_SCHEME_17 = %i[BACAV MIPHC MIUMU MIUMO MIAPF].freeze
  EXCLUSIONS = %i[BACAV MIPHC MIUMU MIUMO].freeze
```
which can be used in the tests, rather than referencing the exclusions in the code under test. Feels a bit neater and explicit about what is happening.

mpw5:
There are issues in CCR with enabling this injection, due to the possibility of discontinuances with prosecution evidence (which are eligible for APFs) being assessed down to discontinuances without prosecution evidence (which are not eligible for APFs). Solutions to this have been investigated and it has been agreed that for now we will not enable this injection: https://dsdmoj.atlassian.net/wiki/spaces/CT/pages/6229689614/CTSKF-1804+What+do+we+need+to+do+to+enable+auto+injection+to+CCR+of+claims+with+additional+prep

PR #9807, opened 2026-07-08
Add Rainbow call to RakeUtils method

mpw5:
#### What

Removing AwesomePrint means that calling a colour method on a string will no longer work. Most such calls have been updated to use Rainbow, but the RakeUtils method `continue?` was missed. This prevents rake tasks which use that method, such as fee scheme seeders, to fail.

This updates the `continue?` method to use Rainbow for colouring.


AnnaC-C (approved):
LGTM 👍

PR #9820, opened 2026-07-10
Update text colouring to use Rainbow

mpw5:
#### What

There are a two rake tasks where we previously used awesomeprint to colour text output to the terminal. Since moving to the Rainbow library, the previous syntax no longer works. This causes failures when the tasks are run, including in the daily smoke tests.

#### How

Updates both tasks to use the correct syntax: eg `'text'.colour` -> `Rainbow('text').colour`


AnnaC-C (approved):
LGTM 👍

PR #9821, opened 2026-07-10
Fix flaky test

Katy600:
#### What
**Flaky test 1**: In the cucumber tests `Given I am later on the Your claims page` the test fails as it is not hitting that page. Adds a patiently block to ensure that page is loaded correctly. 

**Flaky test 2** Fix a parsing issue. Enforcing a consistent, zero-padded format ensures the date is always interpreted correctly by the application, eliminating a source of intermittent test failures.

**Flaky test 3:** refactor and simplify the test to use the 'patiently' block. Remove the explicit 'sleep' and the logic for a second button click. The 'patiently' block will retry the click and the title expectation together, as a way to handle slow and transitions and asynchronous behaviour.

**Flaky test 4** 
This test fails because it doesn't reach the expected page. The fix uses Capybara's waiting mechanism to ensure the page navigation has completed before proceeding to the next step. This helps prevent race conditions.

**Flaky test 5**
Changing build to create  
By changing build(:case_type, :trial) to create(:case_type, :trial), you ensure the CaseType record is persisted to the database. This allows the service's queries to find the associated claims correctly, leading to consistent and accurate counts, thus fixing the flaky test.

**Flaky test 6**
Add URL matcher to prevent flaky tests
Add a set_url_matcher to the ExternalUserHomePage to allow for flexible URL matching. This will prevent flaky tests by ensuring the page is recognized at either /external_users or /external_users/claims

### DIFF
--- a/features/super_admins/stats.feature
+++ b/features/super_admins/stats.feature
@@ -5,7 +5,7 @@ Feature: Stats
     Given I am a signed in super admin
     Given I have created test claims for "10/08/2022"
     When I visit "/super_admins/stats"
-    And I enter the From date "01/8/2022"
-    And I enter the To date "31/8/2022"
+    And I enter the From date "01/08/2022"
+    And I enter the To date "31/08/2022"
     And I click Update
     Then the Javascript variable createChart should change
--- a/features/page_objects/external_user_home_page.rb
+++ b/features/page_objects/external_user_home_page.rb
@@ -2,4 +2,5 @@ require_relative "user_home_page"
 
 class ExternalUserHomePage < UserHomePage
   set_url "/external_users"
+  set_url_matcher %r{/external_users(/claims)?$}
 end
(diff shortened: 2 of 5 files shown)


features/super_admins/stats.feature
```diff
@@ -5,7 +5,7 @@ Feature: Stats
     Given I am a signed in super admin
     Given I have created test claims for "10/08/2022"
     When I visit "/super_admins/stats"
-    And I enter the From date "01/8/2022"
-    And I enter the To date "31/8/2022"
+    And I enter the From date "01/08/2022"
+    And I enter the To date "31/08/2022"
```
mpw5, line 9: I'm not sure that this will actually resolve anything (I think a change I made last week might have fixed this failing test instead) but this won't cause any problem and it looks neater like this, so I'm happy for it to be left in 🙂

mpw5 (approved):
LGTM. Hopefully this works, it will be great if we can reduce the number of failures we see 🤞

features/super_admins/stats.feature
Katy600: Yes, I think you are right. I am suspicious of a few of my fixes as it is hard to tell with flaky tests.

PR #9822, opened 2026-07-10
Fixing Software Composition Analysis 

AnnaC-C:
#### What

Fixing the build error for Software Composition Analysis build 

#### Ticket

- No ticket, just build error

#### Why

To remove the build error for the Software Composition Analysis build as it started running errors on the private Git repo now it isn't public.

#### How

Changed 
```
security-events: read
```

to 

```
security-events: write
actions: read
```

within the devsecops_sca_scan.yml file.

--------

### DIFF
--- a/.github/workflows/devsecops_sca_scan.yml
+++ b/.github/workflows/devsecops_sca_scan.yml
@@ -19,7 +19,7 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
-      security-events: read
+      security-events: write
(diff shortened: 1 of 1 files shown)


AnnaC-C:
Currently we have issues with GitHub Advanced Security from changing this repo to private (from public). 

https://mojdt.slack.com/archives/C08SV6MR2P7/p1783675740103109

So this PR may end up being redundant

PR #9864, opened 2026-07-21
Update SimpleCov config

mpw5:
#### What

SimpleCov is a gem used to calculate code coverage of tests. It was recently upgraded from v0.22 to v1.0. This has introduced some deprecation warnings when the rspec test suite runs.

#### How

This PR fixes those warnings by replacing calls to the `SimpleCov.add_filter` method with `.skip` and calls to `SimpleCov.add_group` with `.group`.

It also removes from config some skipped file locations which no longer exist.


roni-c-justice-dev (approved):
Looks good to me

PR #9977, opened 2026-08-20
feat: add workflow to rerun failing tests

roni-c-justice-dev:
#### What

Add a workflow which will rerun failing tests from rspec-tests, cucumber-tests or other-tests (in the to-be GHA pipeline)

#### Why
Flaky tests in rspec and cucumber tests are requiring manual operation to re-run failed tests.  These tests will be addressed on a future ticket, however this can help us in the meantime

Because this runs on workflow_run trigger it needs to be merged into main to run against any other branch.


mpw5 (approved):
Approving for now. We might need to think about how this works in the long term - I like the idea of retrying flaky tests automatically to cut down on manual intervention, but I think this will also retry any genuinely failing tests, which might be a bit of a waste of resources.

PR #10070, opened 2026-09-15
Stabilise flaky Cucumber claim steps

mpw5:
#### What

Attempts to stabilise and speed up the Cucumber test suite by improving synchronisation around claim journeys, evidence checklists, offence searches and fee calculations.

#### Why

Several Cucumber scenarios were intermittently failing because they relied on fixed sleeps, substring-based checklist matching, or retries that did not verify the desired browser state. The suite had also become slower due to unnecessary waits, including a full timeout before offence search results and evidence-checklist detection on pages where the checklist was absent.

#### How

* Make evidence checklist selection exact, case-insensitive and idempotent.
* Verify that requested evidence checkboxes are actually checked.
* Guard Continue-button retries to avoid submitting the claim twice after navigation has started.
* Replace fixed sleeps and per-character AJAX waits with condition-based waits and debounce-aware input handling.
* Wait for offence search results to appear instead of sleeping for the maximum timeout.
* Reduce selected page-transition wait times where appropriate.
* Add headless Chrome flags to reduce browser overhead in the test environment.


Katy600 (approved):
It looks very thorough.

PR #10071, opened 2026-09-15
Poc change pipeline

roni-c-justice-dev:
Experimental POC to test out new pipeline structure


.github/workflows/pull_request_deploy.yml
```diff
@@ -0,0 +1,176 @@
+name: Pull request deploy
+
+on:
+  pull_request:
+    branches:
+      - main
+    types: [opened, reopened, synchronize, ready_for_review]
```
mpw5, line 7: I wonder if we can remove `ready_for_review` from this? The pipelines will have already have run when the PR is opened and any additional commits pushed to it, so maybe this is redundant (although probably no harm leaving it in either)

.github/workflows/pull_request_tests.yml
```diff
@@ -0,0 +1,269 @@
+name: Pull request tests
+
+on:
+  pull_request:
+    branches:
+      - main
+    types: [opened, reopened, synchronize, ready_for_review]
```
mpw5, line 7: I wonder if we can remove `ready_for_review` from this? The pipelines will have already have run when the PR is opened and any additional commits pushed to it, so maybe this is redundant (although probably no harm leaving it in either)

.github/workflows/pull_request_deploy.yml
```diff
@@ -0,0 +1,176 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build-and-push:
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
```
mpw5, line 18: I think this has changed from the CircleCI approach a little. In Circle we needed to approve the build-and-push step before manually triggering deployments to environments. 

In GitHub Actions it looks like the build step will automatically run on any open PR, which might be a security risk - for example if someone opened a malicious PR which amended the build script to do something bad, it would run without any of us knowing about it.

I'd like to reinstate the approval of the build-and-push step. I think it just needs to be something like 

```
  build-and-push:
    if: github.event.pull_request.head.repo.full_name == github.repository
    runs-on: ubuntu-latest
    environment: dev
    permissions:
      id-token: write
      contents: read
```

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,684 @@
+name: CI CD
+
+on:
+  push:
+    branches:
+      - main
+  
+  workflow_dispatch:
+
+  schedule:
```
mpw5, line 10: It looks to me like this will run the whole pipeline on this schedule, rather than just the smoke tests, which is what CircleCI does. I think you've mentioned before why we need this approach but I can't remember the details. 

I can see we're guarding the smoke test steps with `github.event_name == 'schedule'` but I wonder if we also need to exclude the other steps (particularly deployments) with something like `github.event_name != 'schedule'`

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,684 @@
+name: CI CD
+
+on:
+  push:
+    branches:
+      - main
+  
+  workflow_dispatch:
```
mpw5, line 8: I think this command means that this workflow can be triggered manually for any branch, which effectively means we can deploy any open PR to production. I think if we remove it then it will only run on pushes to main and the schedule, which is what we actually want. Is there any reason why we need to have `workflow_dispatch` here?

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,684 @@
...
+          kubectl set image -f .k8s/live/staging/deployment.yaml cccd-app=$IMAGE --local -o yaml | kubectl apply -f -
+          kubectl apply \
+            -f .k8s/live/staging/service.yaml \
+            -f .k8s/live/staging/ingress.yaml
+          kubectl annotate deployment/claim-for-crown-court-defence kubernetes.io/change-cause="$(date +%Y-%m-%dT%H:%M:%S%z) - deploying: $IMAGE via GitHub Actions" --overwrite
+          kubectl rollout status deployment/claim-for-crown-court-defence
+
+  deploy-production:
+    runs-on: ubuntu-latest
+    needs: [rspec-tests, cucumber-tests, other-tests, build-and-push, smoke-test-main, ui-smoke-test-dev, deploy-staging]
```
mpw5, line 475: There are times when it's useful to deploy to production without having to deploy to staging first (we sometimes use staging for testing with caseworkers). Could we change this to remove the dependency on the deployment to staging? 

(This is possibly a bit of a smell for our overall process as I think you're right that ideally we should keep staging in line with production, so maybe it's something we need to think about more generally).

mpw5, line 475: I also think that the ui-smoke-test-dev step might not make sense here. 

Previously we 

* automatically deployed to dev
* ran the smoke-test-main and ui-smoke-test-dev jobs (so the smoke test ran against the new version of dev
* if they passed, allowed deployments to other environments

Now we run the smoke-test-main and ui-smoke-test-dev jobs on a push - this means the ui smoke test is running against the old version of dev, so there's no benefit to having it as gate for deploying to production.

Maybe we need to rethink the overall process - although auto-deploying to dev was annoying, it meant we could guarantee that the deploy step was working for at least one environment, and that environment was still useable after the deployment, before going to production. Now we don't seem to have that assurance 😬

mpw5 (changes requested):
I've left some comments. Very happy to talk them through - some of them might just be my limited understanding, but I think there are few things that will need changing.

AI also seems to be flagging some concerns. I haven't been through them all yet but we should definitely review them properly.

.github/workflows/pull_request_deploy.yml
roni-c-justice-dev: To have the same kind of approval gate in Github Actions, the advice was to create an environment for build-and-push and attach approval gates to that.  Addressed using that approach and code changed accordingly.  We will next need to re-check that deployments continue to work correctly following this change.

.github/workflows/ci_cd.yml
roni-c-justice-dev: I'm honestly not sure on this one, I think it was a leftover from the more generic approach.  Removed

.github/workflows/ci_cd.yml
roni-c-justice-dev: removed

roni-c-justice-dev: Have added in the deploy-dev step as a requirement for this job.  It keeps us linked to a manual dev deployment but makes sure that deployment also happened before the ui smoke test runs so it's testing the right version of dev.

.github/workflows/ci_cd.yml
roni-c-justice-dev: Amended to add a completely separate stage for scheduled smoke test as it needs the latest rather than the current SHA in a commit anyway.

.github/workflows/pull_request_deploy.yml
```diff
@@ -0,0 +1,176 @@
...
+        uses: aws-actions/amazon-ecr-login@d539f0932e70871a027e9d5a9d8fc38589180a64
+
+      - name: Build and push image
+        run: .github/build.sh
+
+  deploy-dev:
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    needs: build-and-push
+    runs-on: ubuntu-latest
+    environment: dev
```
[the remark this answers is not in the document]
roni-c-justice-dev: This is deliberate, we need the opportunity to quickly deploy to lower environments without waiting 13 minutes for a full test run

PR #10085, opened 2026-09-17
Speed up fee calculator response specs

mpw5:
#### What

Reduces the time taken to run shared fee calculator response specs

#### Why

The fee calculator response shared examples in `spec/support/shared_examples/services/shared_examples_for_fee_calculators.rb` were repeating expensive setup work, which slowed down the suite unnecessarily. These examples are shared between many other spec files and so delays were multiplied many times.

Running the tests in `spec/services/claims/fee_calculator/` previously took ~8 minutes. With this change in place the same tests run in ~90 seconds.

#### How

Use `:aggregate_failures` to group expectations. This means that instead of making a mocked fee calculator request for every `it` block in the tests, a single request is made and assertions run against the resulting response. This reduces overheads and speeds up the tests while maintaining test coverage.


Katy600 (approved):
That makes sense! And so much faster!!

PR #10089, opened 2026-09-18
Speed up slow-running rspec examples by removing duplicate setup

mpw5:
#### What

Part of an ongoing attempt to make the rspec test suite run a bit quicker.

#### Why

Because it's really quite slow at the moment.

#### How

Several spec files had `it` blocks that shared the same expensive `before`/`let!` setup and only differed in which assertion they checked. Each sibling example re-ran that setup from scratch, multiplying the slowest parts of the suite for no extra coverage.

This merges these siblings into single examples using `aggregate_failures`, so the shared setup runs once per scenario instead of once per assertion. All original expectations are preserved and still individually reported on failure; only the setup work is de-duplicated.

Affected specs:
`- spec/requests/super_admins/stats_request_spec.rb`
`- spec/models/claims/search_spec.rb`
`- spec/controllers/external_users/claims_controller_spec.rb`
`- spec/api/v2/mi/injection_errors_spec.rb`
`- spec/api/v2/mi/provisional_assessments_spec.rb`
`- spec/models/claims/state_machine_scopes_spec.rb`
`- spec/models/claims/financial_summary_spec.rb`
`- spec/models/claim/advocate_claim_spec.rb`


Katy600 (approved):
It is going to be so much faster! 🥇

PR #10099, opened 2026-09-21
[CTSKF-1752] Fix allocation navigation active state

mpw5:
#### What

When a user selects either the Allocation or Re-allocation primary navigation items both become active.

This started occurring on approximately 04/06/2026. There had been no specific change to the functionality - it is possible that was caused by changes to Rails (particularly the introduction of Rack v3) or the GOVUK or MOJ design systems. 

It seems that `current_page?` method in the `ApplicationHelper` module, which is used by the `_primary_navigation.html.haml` isn’t working as expected any more for pages where a param is part of the url.

#### Ticket

#### Why

To ensure correct navigation for users and reduce confusion.

#### How

* Adds a helper method `active_allocation_navigation_tab?` to `app/helpers/application_helper.rb` to replaces the use of `current_page?` in order to highlight Allocation or Re-allocation based on the value in the params, defaulting to the Allocation tab when no tab param is present.

* Also moves super admin primary navigation coverage tests to a general layout view spec and adds additional tests to cover the allocation/re-allocation menu functionality.

#### Screenshots

Before:

image

After:

image

image


AnnaC-C (approved):
👍  LGTM

